PR #3, opened 2019-05-17
Setup Apollo GraphQL client with cache persisting

donni106:
- setup Apollo GraphQL client: https://www.apollographql.com/docs/react/
  - a local development address is given for the GraphQL server uri
- implemented cache persisting with https://github.com/apollographql/apollo-cache-persist
  to AsyncStorage
- made first queries
  - on HomeScreen querying data from locally connected main server
  - on IndexScreen querying data from the client store
  - for `fetchPolicy`s see https://www.apollographql.com/docs/react/api/react-apollo#graphql-config-options-fetchPolicy
- increased version to 0.2.0

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -1,5 +1,88 @@
-import React from 'react';
+import React, { useEffect, useState } from 'react';
+import { ActivityIndicator, AsyncStorage, StyleSheet, View } from 'react-native';
+import { ApolloProvider } from 'react-apollo';
+import { ApolloClient } from 'apollo-client';
+import { HttpLink } from 'apollo-link-http';
+import { InMemoryCache } from 'apollo-cache-inmemory';
+import { persistCache } from 'apollo-cache-persist';
 
 import AppContainer from './navigation/AppContainer';
 
-export const App = () => <AppContainer />;
+export const App = () => {
+  const [client, setClient] = useState(null);
+  const [loaded, setLoaded] = useState(false);
+
+  const setupApolloClient = async () => {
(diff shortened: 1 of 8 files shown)


src/index.js
```diff
@@ -1,5 +1,86 @@
...
+    // Setup your Apollo Link, and any other Apollo packages here.
+    const client = new ApolloClient({
+      link,
+      cache,
+
+      // If you want to use Apollo Client's @client support to query the cache without using
+      // local resolvers, you must pass an empty object into the ApolloClient constructor
+      // resolvers option. Without this Apollo Client will not enable its integrated @client
+      // support, which means your @client based queries will be passed to the Apollo Client
+      // link chain. You can find more details about why this is necessary here.
```
SoleCincis: link is missing

src/screens/IndexScreen.js
```diff
@@ -49,19 +32,55 @@ export default class IndexScreen extends React.Component {
...
+        {({ data, client }) => {
+          const { cacheItems } = data;
+
+          return (
+            <View style={styles.container}>
+              <Text>Index Screen</Text>
+              {!!cacheItems &&
+                cacheItems.map((item) => (
+                  <Button
+                    key={`bla${item.itemId}`}
```
SoleCincis: To do : bla needs to be changed with a appropriate word

src/index.js
donni106: thx, added it

src/screens/IndexScreen.js
donni106: thx, changed it to `key`.

PR #4, opened 2019-05-17
Feature/styled components and react native elements

SoleCincis:
Setup of styled-components and react-native-elements

### DIFF
--- a/package.json
+++ b/package.json
@@ -19,7 +19,6 @@
   },
   "dependencies": {
     "expo": "^32.0.0",
-    "node": "^12.2.0",
     "react": "16.8.6",
     "react-native": "https://github.com/expo/react-native/archive/sdk-32.0.0.tar.gz",
     "react-native-elements": "^1.1.0",
--- a/package.json
+++ b/package.json
@@ -19,9 +19,12 @@
   },
   "dependencies": {
     "expo": "^32.0.0",
+    "node": "^12.2.0",
     "react": "16.8.6",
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -19,9 +19,12 @@
   },
   "dependencies": {
     "expo": "^32.0.0",
+    "node": "^12.2.0",
```
SoleCincis: this is creating a conflict 
Screenshot 2019-05-17 at 16 32 51

donni106: Ok. Please remove the node package from the projects package.json. Node should be used from your global system packages.

donni106: Ok. Please remove the node package from the projects package.json. Node should be used from your global system packages.

src/config/texts.js
```diff
@@ -14,3 +14,12 @@ export const texts = {
     policy: 'Datenschutz'
   }
 };
+
+import styled, { css } from 'styled-components/native';
+import { colors } from './colors';
+
+export const ScreenTitle = styled.Text`
```
donni106: This has to be removed here and put somewhere in /components/. This is a styled `Text` component and not part of text configs.

package.json
donni106: Resolved with cherry picking the [commit](https://github.com/ikuseiGmbH/smart-village-app-app/commit/cc2880c3fe49f4f1a374df8c3aaebd144d65b4fa) from branch: **feauture/working-on-components**

src/config/texts.js
SoleCincis: did it in new branch 'workin-on-components'

PR #5, opened 2019-05-21
Create a custom drawer navigator

donni106:
- created custom drawer and custom drawer items components
  based on the original ones from React Navigation
  - needed to set a gradient background
  - needed to change the navigation behaviour with popToTop() in stacks
  - needed to add an close icon on the top right
  - conditionally use drawerType depending on platform
    - on iOS `slide` is more common
  - changed fontWeight depending on `focused`
    - `bold` when active, otherwise `normal`
- created a DiagonalGradient component to use everywhere needed
- created device config holding different values needed in some screens,
  like platform or dimensions
- added StatusBar to set the bar style to `light-content`, as we want
  to have white text on our gradient header
- updated colors to new shades of green and blue

---

Drawer


src/navigation/DrawerNavigatorItems.js
```diff
@@ -0,0 +1,146 @@
...
+  activeLabelStyle,
+  inactiveLabelStyle,
+  drawerPosition,
+  navigation
+}) => {
+  const Touchable = device.platform === 'ios' ? TouchableOpacity : TouchableNativeFeedback;
+
+  /**
+   * based on:
+   *   https://github.com/react-navigation/drawer/blob/c5954d744f463e7f1c67941b8eb6914c0101e56c/src/views/DrawerSidebar.tsx#L67
```
SoleCincis: this link doesn't lead to a specific page , but the the google search react-navigation/drawer.

donni106: for me this link is working fine: https://github.com/react-navigation/drawer/blob/c5954d744f463e7f1c67941b8eb6914c0101e56c/src/views/DrawerSidebar.tsx#L67

PR #6, opened 2019-05-21
Feature/working on components

SoleCincis:
Create 10 new components
-  which are making up the layout of the main screens like `Home`, `Index` and `Detail `
- `Gradient`, `StandardButton`, `SvgShape` components are not currently in use
  - `SvgShape` will probably be erased from the project 
- `config` now contains not just colors and texts
  -  also object of html tags in order to style the text in `TextContent`


src/components/Title.js
```diff
@@ -1,8 +1,17 @@
...
-  color: ${colors.lightestText};
+export const Title = styled.Text`
+  color: ${colors.primary};
   font-weight: bold;
-  text-align: center;
+  font-size: 20;
+  padding: 10px;
+  text-transform: uppercase;
 `;
+export const TitleContainer = styled.View`
```
donni106: please separate different `const`s with a line break for better readability.

src/screens/HomeScreen.js
```diff
@@ -22,30 +22,43 @@ const items = [
...
+        <TitleContainer>
+          <Title>nachrichten</Title>
+        </TitleContainer>
+        <TextList navigation={navigation} data={items} />
+        <Title>orte & routen</Title>
+        <CardList data={cards} />
+        <Title>veranstaltungen</Title>
+        <TextList navigation={navigation} data={items} second={true} />
+        <Title>service</Title>
+        <TextList navigation={navigation} data={items} listService={true} />
```
donni106: `listService` better `subtitle`

src/screens/HomeScreen.js
```diff
@@ -22,30 +22,43 @@ const items = [
...
+      <ScrollView>
+        <TopVisual />
+        <TitleContainer>
+          <Title>nachrichten</Title>
+        </TitleContainer>
+        <TextList navigation={navigation} data={items} />
+        <Title>orte & routen</Title>
+        <CardList data={cards} />
+        <Title>veranstaltungen</Title>
+        <TextList navigation={navigation} data={items} second={true} />
```
donni106: instead of `second` we can think of some more clear naming like `withBackground` or `alternativeLayout`?
`second` will not make clear, what exactly will be conditionally changed.

src/components/Title.js
```diff
@@ -8,10 +8,10 @@ export const Title = styled.Text`
...
 `;
-export const TitleContainer = styled.View`
-  box-shadow: 0 6px 6px black;
-`;
 
 //experimentig with the shadow ,
 //it seams that titles on homepage are inside of a box witch has a shadow
 // I need to recrate the effect of having a shadow from the line
+export const TitleContainer = styled.View`
+  box-shadow: 0 6px 6px black;
```
donni106: for shadows maybe we can try to have a single <Shadow /> styled component, that we add after containers which needs a shadow.

example:
```
<TitleContainer>
  <Title>asdasd<Title>
</TitleContainer>
<Shadow />
<...>
```

SoleCincis: Shadow is still failing, but I created a fake shadow effect for now with border.

src/screens/HomeScreen.js
SoleCincis: We want to have a subtitle as default. So now we can have a prop `noSubtitle` which can be true, to render no subtitle on the `ListItem`.

PR #9, opened 2019-05-27
Feature/update header

SoleCincis:
Icons and Gradient background on header:
- Icon component
  -  is transforming and rendering my svg files with series of style props 
  - is wrapped in to a` TouchableOpacity` in order to be used for navigating
- LinearGradien component 
  - is working in a similar way , has a `Defaultprops` with color and direction of the gradient
  - but is can also accept new props and children optional not required 
  - in order to make the linear background work on my header , it must be used within  
    `headerBackground` props , positioned in my` defaultStackNavigatorConfig`


src/components/DiagonalGradient.js
```diff
@@ -4,16 +4,22 @@ import { LinearGradient } from 'expo';
...
-export const DiagonalGradient = ({ children }) => (
-  <LinearGradient
-    colors={[colors.primary, colors.secondary]}
-    start={{ x: 0, y: 0 }}
-    end={{ x: 1, y: 1 }}
-  >
-    {children}
-  </LinearGradient>
-);
+const defaultProps = {
```
donni106: maybe we can use the components default props, which would be the more typical way. I will try that out:

```
DiagonalGradient.defaultProps = {
  colors={[colors.primary, colors.secondary]}
  start={{ x: 0, y: 0 }}
  end={{ x: 1, y: 1 }}	  
  style: { flex: 1 }
};
```

than we do not have do set it explicitly here, because the PropTypes logic knows what to do:

```
<LinearGradient {...props}>
```

donni106:
Thx. I will fix the conflicts and merge that.

PR #22, opened 2019-06-13
Feature/poi detail screen

SoleCincis:
Completed priceCard and InfoCard:
**InfoCard**
- added a lot of checks for presence before rendering
  - we only want to render things if there is data
- refactored Text components to use them more globally
- transformed each info into a link using onPress navigates to correlated location
  - using mapHelper to link to maps
  - using tel from openLink/helpers to make a call phone
- passing a prop linkText that changes the color of taxt from black to green color of links

**priceCard**
- background color of each single priceBox is now with opacity (0.6)
- created and using a price helper able to format any number in to a price
 - as a guide : https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/toLocaleString

### DIFF
--- a/src/components/Text.js
+++ b/src/components/Text.js
@@ -7,6 +7,11 @@ export const RegularText = styled.Text`
   flex: 1;
   font-family: titillium-web-regular;
   font-size: ${normalize(16)};
+  ${(props) =>
+    props.linkText &&
+    css`
+      color: ${colors.primary};
+    `};
 `;
 
 export const BoldText = styled(RegularText)`
--- a/src/helpers/mapHelper.js
+++ b/src/helpers/mapHelper.js
@@ -1,0 +1,19 @@
+import { Platform } from 'react-native';
+
+export function locationString(address) {
+  if (!address) return '';
+
+  return encodeURIComponent(address);
+}
+
+// a maps link is different between the platforms
+export function locationLink(mapsString) {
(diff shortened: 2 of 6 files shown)


src/components/Text.js
```diff
@@ -7,6 +7,11 @@ export const RegularText = styled.Text`
   flex: 1;
   font-family: titillium-web-regular;
   font-size: ${normalize(16)};
+  ${(props) =>
+    props.linkText &&
```
donni106: we are inside of a text element, to we can simply use a prop `link` here

src/components/Text.js
```diff
@@ -7,6 +7,11 @@ export const RegularText = styled.Text`
   flex: 1;
   font-family: titillium-web-regular;
   font-size: ${normalize(16)};
+  ${(props) =>
```
donni106: nice to have empty line before props stuff

src/helpers/mapHelper.js
```diff
@@ -0,0 +1,19 @@
...
+
+export function locationString(address) {
+  if (!address) return '';
+
+  return encodeURIComponent(address);
+}
+
+// a maps link is different between the platforms
+export function locationLink(mapsString) {
+  switch (Platform.OS) {
```
donni106: we have a Platform check in configs device, we use everywhere. we can also use that here.

PR #23, opened 2019-06-17
Feature/update layout list

SoleCincis:
Update on button:
- our Button has now  inside  the diagonal gradient
 - added  `propType` method  , `oneOfType` in order to have more the one possible prop types
 - bottomDivider has a good logic for the purpose of not showing the divider after the last list element

No moro margin between subtitle and title.

Update on Title and Subtitle texts :
- Text component is now bundling all the possible text combinations
 -  ListTitle is now BoldText and `ListSubtitle` is `RegularText`
- all style variations are gone (like `AlternativeLayout` ) 
  - because `TextList` is now looking always the same
 - except for `noSubititle` which is still visible in `publicJsonFileContent`

### DIFF
--- a/src/components/Text.js
+++ b/src/components/Text.js
@@ -22,20 +22,17 @@ export const RegularText = styled.Text`
 
 export const BoldText = styled(RegularText)`
   font-family: titillium-web-bold;
+
+  ${(props) =>
+    props.light &&
+    css`
+      color: ${colors.lightestText};
+      font-size: ${normalize(14)};
+    `};
 `;
 
 export const LightestText = styled(RegularText)`
   color: ${colors.lightestText};
-
-  ${(props) =>
-    props.bold &&
-    css`
-      font-family: titillium-web-bold;
-    `};
-`;
-
-export const PriceText = styled(LightestText)`
-  font-size: ${normalize(14)};
 
   ${(props) =>
     props.bold &&
--- a/src/screens/HomeScreen.js
+++ b/src/screens/HomeScreen.js
@@ -15,7 +15,7 @@ import {
   CardList,
   DiagonalGradient,
   Image,
-  LightestText,
+  BoldText,
   ServiceBox,
   TextList,
   Title,
@@ -250,7 +250,7 @@ export const HomeScreen = ({ navigation }) => (
               bottomDivider: false
             });
 
-          return <TextList navigation={navigation} data={eventRecords} alternativeLayout />;
+          return <TextList navigation={navigation} data={eventRecords} />;
         }}
       </Query>
       <TitleContainer>
@@ -295,9 +295,7 @@ export const HomeScreen = ({ navigation }) => (
                               style={styles.serviceImage}
                               PlaceholderContent={null}
                             />
-                            <LightestText bold small>
-                              {item.title}
-                            </LightestText>
+                            <BoldText light>{item.title}</BoldText>
                           </View>
                         </TouchableOpacity>
                       </ServiceBox>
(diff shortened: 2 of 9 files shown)


src/components/Text.js
```diff
@@ -22,20 +22,17 @@ export const RegularText = styled.Text`
...
 export const BoldText = styled(RegularText)`
   font-family: titillium-web-bold;
-`;
-
-export const LightestText = styled(RegularText)`
-  color: ${colors.lightestText};
 
   ${(props) =>
-    props.bold &&
+    props.light &&
```
donni106: Why do you have a `light` prop now for `BoldText`? Cannot read about it in the commit message.
It somehow makes no sense for me, because "light" text for bold is equal to `RegularText`, isn't it?

src/screens/HomeScreen.js
```diff
@@ -230,27 +243,36 @@ export const HomeScreen = ({ navigation }) => (
...
-                rootRouteName: 'EventRecords'
-              },
-              bottomDivider: false
-            });
+          return (
+            <View>
+              <TextList navigation={navigation} data={eventRecords} />
 
-          return <TextList navigation={navigation} data={eventRecords} alternativeLayout />;
+              {!!eventRecords && (
```
donni106: since we are checking for the `length` of the `eventRecords` object in line 250, we do not need this check here (again). it is safe to render `<Wrapper><Button...` directly.

src/screens/HomeScreen.js
```diff
@@ -77,29 +79,39 @@ export const HomeScreen = ({ navigation }) => (
...
-                rootRouteName: 'NewsItems'
-              },
-              bottomDivider: false
-            });
+          return (
+            <View>
+              <TextList navigation={navigation} data={newsItems} />
 
-          return <TextList navigation={navigation} data={newsItems} />;
+              {!!newsItems && (
```
donni106: since we are checking for the `length` of the `newsItems ` object in line 86, we do not need this check here (again). it is safe to render `<Wrapper><Button...` directly.

src/screens/HomeScreen.js
```diff
@@ -158,24 +170,25 @@ export const HomeScreen = ({ navigation }) => (
...
-                      queryVariables: {},
-                      rootRouteName: 'PointsOfInterest'
-                    },
-                    bottomDivider: false,
-                    topDivider: true
-                  }
-                ]}
-              />
+
+              {!!pointsOfInterest && (
```
donni106: we can get rid of the check for `!!pointsOfInterest` here when checking before with an early `return null`, like for news items and event records. than it would be is safe to render `<Wrapper><Button...` directly.

src/screens/HomeScreen.js
```diff
@@ -77,29 +79,39 @@ export const HomeScreen = ({ navigation }) => (
                   message: shareMessage(newsItem, 'newsItem')
                 }
               },
+              bottomDivider: index === data.newsItems.length - 1 ? false : true,
```
donni106: We can upgrade the condition to only `index === data.newsItems.length - 1`, because if this is false, it returns false and if it is true it returns true. We do not have to write it explicitly.

src/components/Text.js
SoleCincis: it's not a good prop name , it's used on PriceCard for prices  a small(normalize(14)) bold and white  text .  I can maybe call it white ?

donni106: What about `<LightestText small...` for that cases?

SoleCincis: TODO: restructure RegularText will be jest Text with 4 different props : bold , small, link and white

donni106: consider `lightest` instead of `white`. and we can change `link` to `primary` also, which would be more general.

PR #24, opened 2019-06-17
Update on header

SoleCincis:
Created a condition so that : 
  - iOS has bold text
  - android regular text (bold text is was not good looking on android header)
- fontWeight is required form default stack navigation

### DIFF
--- a/src/navigation/defaultStackNavigatorConfig.js
+++ b/src/navigation/defaultStackNavigatorConfig.js
@@ -2,7 +2,7 @@ import React from 'react';
 import { StyleSheet, TouchableOpacity } from 'react-native';
 import CardStackStyleInterpolator from 'react-navigation-stack/src/views/StackView/StackViewStyleInterpolator';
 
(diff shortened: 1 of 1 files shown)


donni106:
what do you mean with

> * fontWeight is required form default stack navigation

src/navigation/defaultStackNavigatorConfig.js
```diff
@@ -15,8 +15,8 @@ export const defaultStackNavigatorConfig = (initialRouteName) => {
       headerBackground: <DiagonalGradient />,
       headerTitleStyle: {
         color: colors.lightestText,
-        fontFamily: 'titillium-web-regular',
-        fontWeight: '400'
+        fontWeight: device.platform === 'ios' ? '600' : '400',
+        fontFamily: device.platform === 'ios' ? 'titillium-web-bold' : 'titillium-web-regular'
```
donni106: why do you have changed the order of `fontFamily` and `fontWeight` :smirk:?

PR #29, opened 2019-06-27
Performance optimizations and fixes

donni106:
**Update packages and add react-native-expo-image-cache for image caching**

- https://github.com/wcandillon/react-native-expo-image-cache
- it seems, that a better solution would be https://github.com/DylanVann/react-native-fast-image,
  but this package is not integrated in Expo yet
  - https://expo.canny.io/feature-requests/p/add-react-native-fast-image

**Secure access of `mediaContents` of a point of interest**

**Remove unnecessary `loaded` state and check `client` presence**

- having `loaded` was redundant because it was only set
  after `client` is set, so we can rely on `client` for further rendering
  in the same way and get rid of `loaded`

**Remove logic with useEffect (componentDidMount) for PointOfInterest**

- PointOfInterest is only rendered from DetailScreen, when data
  are fetched, otherwise a loading indicator is shown
- so when calling PointOfInterest we already have data to use for
  the screen, we do not need to build variables after mount and
  render an empty component until then

**Add ListFooterComponent to CardList and TextList**

- added a ActivityIndicator logic to CardList and TextList
  to indicate loading/processing while scrolling
- refactored keyExtractor methods to return a readable key

**Add removeClippedSubviews prop to FlatLists**

If true, views that are outside of the viewport are detached from the native view hierarchy.

Pros: This reduces time spent on the main thread, and thus reduces the risk of dropped frames, by excluding views outside of the viewport from the native rendering and drawing traversals.

Cons: Be aware that this implementation can have bugs, such as missing content (mainly observed on iOS), especially if you are doing complex things with transforms and/or absolute positioning. Also note this does not save significant memory because the views are not deallocated, just detached.

https://facebook.github.io/react-native/docs/next/optimizing-flatlist-configuration#removeclippedsubviews

**Update queries and pass data already from IndexScreen to DetailScreen**

- added attributes to index queries to fetch more data on IndexScreen
- passed down `details` in params to DetailScreen for better offline
  experience
  - so if the network gets loss after IndexScreen we can present
    data on the DetailScreen anyway, because they are fetched already
- secured access of `mediaContents` of a points of interest
  from IndexScreen
- refactored DetailScreen for better overview of the render method
- map through all `contentBlocks` for presenting `body` in newsItem
  DetailScreen

**Edit TODO comments**

- removed were possible
- changed to NOTE: where there are a lot presence checks

**Remove unused StandardButton**

**Add `useForeground` to Button because of DiagonalGradient**

Set to true to add the ripple effect to the foreground of the view, instead of the background. This is useful if one of your child views has a background of its own, or you're e.g. displaying images, and you don't want the ripple to be covered by them.

http://facebook.github.io/react-native/docs/touchablenativefeedback.html#useforeground

**Refactor image size calculation to helper**

- created imageHeight() imageWidth() methods in a imageHelper
  and use them everywhere to avoid duplications
- adjusted imagesMaxWidth in HtmlView with substraction
  of wrapper padding
- align HtmlView  centered
  - currently we get only very small images through RSS API,
    which we cannot scale, so centered looks better

**Set line height and font size everywhere concordantly**

- removed font-size from BoldText because we can use `small`
  prop instead

**Adjust headerTitleStyle**

**Update graphql package from 14.3.1 to 14.4.0**

https://github.com/graphql/graphql-js/releases/tag/v14.4.0

**Implement dynamic fetch policy for graphql queries**

- created a helper for returning the fetch policy depending on the network state
- created a helper for returning boolean isConnected on the basis of NetInfo.getConnectionInfo()
- implemented a NetworkProvider with React.Context for holding the connection state
  and passing it down to necessary child components
  - thx to https://www.freecodecamp.org/news/easily-manage-connection-status-updates-in-react-native-28c9b4b0647f/
  - React Context API: https://reactjs.org/docs/context.html
- updated queries to use the new fetch policy method
- updated call of `auth` conditionally only when online
- refactored IndexScreen for better overview of the render method

**Update graphql package from 14.4.0 to 14.4.1**

https://github.com/graphql/graphql-js/releases/tag/v14.4.1

**Fix typo in shareHelper for point of interest**

- sharing a point of interest resulted in undefined body
  - fixed typo in case statement
- added description to sharing content
- removed pre-text from sharing content because the source
  is added already to the end of the sharing content

**Adjust version of react-native-webview**

Warning after upgrading expo-cli:

"Some of your project's dependencies are not compatible with currently installed expo package version:
 - react-native-webview - expected version range: ~5.8.1 - actual version installed: ^5.12.0
Your project may not work correctly until you install the correct versions of the packages.
To install the correct versions of these packages, please run: expo install [package-name ...]"

Ran `expo install react-native-webview`.

**Fix returning nothing with early return of null for no data**


donni106:
Work in progress...there are still some TODOs (error handling, polling on home screen, ...)

donni106:
> ℹ️ **Work in progress...there are still some TODOs (error handling, polling on home screen, ...)**

With checking for online/offline and reacting to changes, handling errors should be better now.

Polling is not working. First followed the instructions here: https://www.apollographql.com/docs/react/essentials/queries/#polling-and-refetching
After nothing happened, I tried different ways. Nothing worked.

PR #30, opened 2019-07-02
Feature/event record detail screen

SoleCincis:
- refactored event details from DetailScreen to an own component
  - added detail renderings according to the data
  - similar to PointOfInterest

 - added static texts to texts
 - added eventDate helper method for displaying the event date
 - on query used gql method to rename object and keys

- temporary logic in order to show priceCard just when description is first index
 - find a solution in back end to have description null when no data available
- in Operating Company we still have fax icon always visible because his value from data is boolean


src/components/screens/EventRecord.js
```diff
@@ -0,0 +1,147 @@
...
+      description,
+      mediaContents,
+      prices,
+      title,
+      webUrls,
+      operatingCompany,
+      openingHours
+    } = data;
+
+    setPage({
```
donni106: I will remove this `setPage...` in resolving merge conflict while merging to master. So we get rid of this extra step like already done in `PointOfInterest`.

src/queries/eventRecords.js
```diff
@@ -31,12 +40,58 @@ export const GET_EVENT_RECORD = gql`
           url
         }
       }
+      addresses {
+        city
+        street
+        zip
+      }
+      contact: contacts {
```
donni106: We have to be careful here. One `contact` in `PointOfInterest` cannot be handled the same way like multiple `contacts` for `EventRecord`. One is an Object and the other an Array of Objects.

src/config/texts.js
```diff
@@ -1,6 +1,12 @@
 import appJson from '../../app.json';
 
 export const texts = {
+  eventRecord: {
+    prices: 'Preise',
+    description: 'Beschreibung',
```
donni106: alphabetical order is preferred

PR #33, opened 2019-07-09
Feature/tour detail screen

SoleCincis:

**New screen Tour :** 
 - operetingCompany is empty but still displayed cause there should be data coming eventually
- in Tour we have no openigTime or prices
- new and definitive Icon for lengthKm
- tour is also correctly placed in detailScreen

**Nested queries&Mix of Tours and POIs:** 
- created a new query containing both (poi and tours)
- in index screen created new case able to display a long list of poi and tours
  -  created  logic  to display poi and tours in alphabetical order 
   - as a guide line used https://developer.mozilla.org/en-
- home screen random mix:
     - use a shuffle method from external package lodash https://lodash.com/docs/4.17.11#shuffle
     - use order method from server with key RAND to always get different pois and tours to display on preview home screen 
    - it's necessary to have 2 different param orderTour and orderPoi
    - passing the order RAND (random) to both , but it would be possible to have a different order in each of them


src/components/screens/InfoCard.js
```diff
@@ -73,6 +74,18 @@ export const InfoCard = ({ addresses, category, contact, contacts, webUrls }) =>
         <RegularText>{category.name}</RegularText>
       </InfoBox>
     )}
+    {!!lengthKm && (
```
donni106: empty line before the new block would be nice but is not necessary

src/components/screens/InfoCard.js
```diff
@@ -73,6 +74,18 @@ export const InfoCard = ({ addresses, category, contact, contacts, webUrls }) =>
...
+      <InfoBox>
+        <RNEIcon
+          name="map"
+          type="material"
+          color={colors.primary}
+          iconStyle={{ marginRight: normalize(10) }}
```
donni106: now as we have this iconStyle multiple times in this file, it would be nicer to have a const styles = ... for that, which we can use everywhere here for icons.
it helps for the future if we want to change the margin for example from 10 to something else. now we would need to change it on multiple places. if we would have a const for that particular style, we only need to change it on one place.

src/queries/index.js
```diff
@@ -14,13 +18,22 @@ export const getQuery = (query) => {
     return GET_NEWS_ITEMS;
   case 'newsItem':
     return GET_NEWS_ITEM;
+
```
donni106: spaces between case statements are not necessary

src/queries/pointsOfInterestAndTours.js
```diff
@@ -0,0 +1,189 @@
...
+        logo {
+          url
+        }
+        name
+      }
+    }
+  }
+`;
+
+export const GET_POINT_OF_INTEREST_AND_TOUR = gql`
```
donni106: this makes no sense. reading it out: "let get us one point of interest and tour". there cannot be 1 poi and tour logically. there are at least 2 (1 poi and 1 tour), and therefore we would use the upper query.

need to remove this query.

src/screens/IndexScreen.js
```diff
@@ -113,9 +114,104 @@ export class IndexScreen extends React.PureComponent {
...
+              query: 'tour',
+              queryVariables: { id: `${tour.id}` },
+              rootRouteName: 'Tours',
+              shareContent: {
+                message: shareMessage(tour, 'tour')
+              },
+              details: tour
+            }
+          }));
+      const mix = [...(pointsOfInterest || []), ...(tours || [])];
```
donni106: `mix` can be more exact...something like `pointsOfInterestAndTours`

PR #42, opened 2019-08-22
Updates in events detail screen 

SoleCincis:
Changes according to CMS 
- rendering first and last name (InfoCard + OperatingCompanyInfo)
- rendering description instead of url when there is one  (InfoCard + Link)
- rendering addition in address (OperatingCompanyInfo)

### DIFF
--- a/src/components/Link.js
+++ b/src/components/Link.js
@@ -9,11 +9,11 @@ import { Icon } from './Icon';
 import { RegularText } from './Text';
 import { WrapperRow } from './Wrapper';
 
-export const Link = ({ url, title }) => (
+export const Link = ({ url, descrption }) => (
   <TouchableOpacity onPress={() => openLink(url)}>
     <WrapperRow>
       <Icon icon={link(colors.secondary)} style={styles.icon} />
-      <RegularText link>{title}</RegularText>
+      <RegularText link>{descrption}</RegularText>
     </WrapperRow>
   </TouchableOpacity>
 );
@@ -27,5 +27,5 @@ const styles = StyleSheet.create({
 
 Link.propTypes = {
   url: PropTypes.string.isRequired,
-  title: PropTypes.string.isRequired
+  descrption: PropTypes.string.isRequired
 };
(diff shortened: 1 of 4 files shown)


src/components/Link.js
```diff
@@ -9,11 +9,11 @@ import { Icon } from './Icon';
 import { RegularText } from './Text';
 import { WrapperRow } from './Wrapper';
 
-export const Link = ({ url, title }) => (
+export const Link = ({ url, descrption }) => (
```
donni106: [typo] `descrption` => `description`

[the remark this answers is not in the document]
donni106: Have you checked, where we use the Link component, if it is possible to change the key here from `title` to `description`? This means, the props given needs to have a `description` key. If there were data with `title` before, this would not work anymore with the changed name. Be careful!
I have checked the code and we do not use this Link component currently.
So I will only fix the typo.

src/components/screens/InfoCard.js
```diff
@@ -113,15 +124,18 @@ export const InfoCard = ({ addresses, category, contact, contacts, webUrls }) =>
...
+          return null;
+        }
 
         return (
           <InfoBox key={index}>
             <Icon icon={urlIcon(colors.primary)} style={styles.margin} />
             <TouchableOpacity onPress={() => openLink(url)}>
-              <RegularText link>{url}</RegularText>
+              {!description && <RegularText link>{url}</RegularText>}
+              {!!description && <RegularText link>{description}</RegularText>}
```
donni106: Well done, we will take this!

But :)
Now there is a duplication of the `<RegularText link>...</RegularText>`. Only the `...` data inside of the RegularText is different. Though, we could think of having only one and making the data conditionally.

This could look something like this:

`<RegularText link>{!!description ? description : url}</RegularText>` what we read: if there is a `description`, use the `description`, otherwise use the `url`.

Another option would be to declare a `const linkText` conditionally before the `return` and use that inside `<RegularText link>{linkText}</RegularText>`, where `linkText` is set to `description` or `url` depending on the data.

PR #43, opened 2019-09-12
Update news item rendering to consider stories

donni106:
- data for news item can have multiple content blocks forming a longer story
- each content block is now a different section in a whole story
- a news item renders a main image(carousel) and a main title from the first content block
  - that is why that data is ignored inside of the story sections for the first index in the iteration
- the order of elements inside of a sections in the story is:
  - title
  - intro
  - image(carousel)
  - text
  - audio/video
- added a new wrapper with only horizontal padding to separate each section from the other without vertical space
- removed bottom margin from the logo as the body text now comes inside of the story sections
- added necessary keys in the news items queries

Render audio and video as html contents (iframe) in a WebView

- there were problems playing sound of SoundCloud iframes in the HTML renderer
- with WebView the sound plays correctly
- for correct display with zoomlevel, a hacky javascript include is necessary

Render audio and video also in EventRecord

### DIFF
--- a/src/components/screens/NewsItem.js
+++ b/src/components/screens/NewsItem.js
@@ -9,46 +9,123 @@ import { Image } from '../Image';
 import { Logo } from '../Logo';
 import { Title, TitleContainer, TitleShadow } from '../Title';
 import { Touchable } from '../Touchable';
-import { Wrapper } from '../Wrapper';
+import { Wrapper, WrapperHorizontal } from '../Wrapper';
 import { ImagesCarousel } from '../ImagesCarousel';
 import { momentFormat, openLink, trimNewLines } from '../../helpers';
-import { RegularText } from '../Text';
+import { BoldText, RegularText } from '../Text';
 
 /* eslint-disable complexity */
 /* NOTE: we need to check a lot for presence, so this is that complex */
 export const NewsItem = ({ data }) => {
-  const { dataProvider, contentBlocks, publishedAt, sourceUrl } = data;
+  const { dataProvider, mainTitle, contentBlocks, publishedAt, sourceUrl } = data;
 
-  const subtitle = `${momentFormat(publishedAt)} | ${dataProvider && dataProvider.name}`;
-  const title = !!contentBlocks && !!contentBlocks.length && contentBlocks[0].title;
   const logo = dataProvider && dataProvider.logo && dataProvider.logo.url;
-  const body =
-    !!contentBlocks &&
-    !!contentBlocks.length &&
-    contentBlocks.map((contentBlock) => contentBlock.body).join('');
   const link = sourceUrl && sourceUrl.url;
-  let images = [];
+  const subtitle = `${momentFormat(publishedAt)} | ${dataProvider && dataProvider.name}`;
+  // the title of a news item is either a given main title or the title from the first content block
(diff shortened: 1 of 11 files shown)


src/components/screens/NewsItem.js
```diff
@@ -113,7 +124,19 @@ export const NewsItem = ({ data }) => {
             !!mediaContent.sourceUrl.url &&
             section.push(
               <WrapperHorizontal key={`${index}-${contentBlock.id}-mediaContent${mediaContent.id}`}>
-                <HtmlView html={trimNewLines(mediaContent.sourceUrl.url)} />
+                <WebView
+                  source={{ html: trimNewLines(mediaContent.sourceUrl.url) }}
+                  style={styles.iframeWebView}
+                  scrollEnabled={false}
+                  bounces={false}
+                  injectedJavaScript={INJECTED_JAVASCRIPT_FOR_IFRAME_WEBVIEW}
```
SoleCincis: ''for correct display with zoomlevel, a hacky javascript include is necessary'' is this the hacky js that you mentioned ? what is it ?

src/components/screens/NewsItem.js
```diff
@@ -14,6 +15,16 @@ import { ImagesCarousel } from '../ImagesCarousel';
 import { momentFormat, openLink, trimNewLines } from '../../helpers';
 import { BoldText, RegularText } from '../Text';
 
+// necessary hacky way of implementing iframe in webview with correct zoom level
+// thx to: https://stackoverflow.com/a/55780430
+const INJECTED_JAVASCRIPT_FOR_IFRAME_WEBVIEW = `
```
SoleCincis: Is this a sort of importing ?

src/components/screens/NewsItem.js
```diff
@@ -113,7 +124,19 @@ export const NewsItem = ({ data }) => {
             !!mediaContent.sourceUrl.url &&
             section.push(
               <WrapperHorizontal key={`${index}-${contentBlock.id}-mediaContent${mediaContent.id}`}>
-                <HtmlView html={trimNewLines(mediaContent.sourceUrl.url)} />
+                <WebView
+                  source={{ html: trimNewLines(mediaContent.sourceUrl.url) }}
```
SoleCincis: what is this helper doing? I can't find the commit for it.

src/components/screens/NewsItem.js
donni106: yes, that constant `INJECTED_JAVASCRIPT_FOR_IFRAME_WEBVIEW` contains the "hacky javascript"

src/components/screens/NewsItem.js
donni106: no, it is a constant definition for using it as the prop `injectedJavaScript` inside of the WebView

src/components/screens/NewsItem.js
donni106: you mean the helper `trimNewLines`?  it is defined in [/src/helpers/htmlViewHelper.js](https://github.com/ikuseiGmbH/smart-village-app-app/blob/master/src/helpers/htmlViewHelper.js). it is used for removing unnecessary and unwanted new lines in a text, that we want to render as html.

PR #71, opened 2020-07-07
Implement global settings for optional navigator

### LINKED ISSUE #70: Ability to choose between drawer and tab navigation
The app should have the options to run with drawer navigator or with tab navigator.
Therefore a json on the server with global settings is necessary.
Default and fallback should be the drawer navigation, as this was initially implemented.

donni106:
Resolves #70


donni106:
Code climate issues will be attacked on a later point.

Issue #49, opened 2020-05-13
Add information text to empty screens


donni106:
Currently screens are empty and white, if there is no data.
I would be nice to have a short text on screens without data. An empty screen is always bad user experience.

|example from GitHub app|example Google Play|example Expo|
|---|---|---|
|![Screenshot_20200513-104755](https://user-images.githubusercontent.com/1942953/81791817-8496b380-9507-11ea-8b42-ea52098022e8.png)|![Screenshot_20220119-181156](https://user-images.githubusercontent.com/1942953/150180617-ea1e1f44-bcfe-4708-a2ce-494abc8c3207.png)|![Screenshot_20220119-204559](https://user-images.githubusercontent.com/1942953/150202982-78846b5a-65b6-4f56-ab91-db43434429fe.png)|

SoleCincis:
I would be happy to work on this feature if requested

donni106:
there are a few examples where this could happen. two of them are pointed out in the following table with screenshots.

|home screen events widget|no events|home screen about item|no content|
|---|---|---|---|
|![Simulator Screen Shot - iPhone 8 Plus - 2022-01-19 at 18 01 31](https://user-images.githubusercontent.com/1942953/150178917-29fbf1f5-2b35-4504-9eb3-0d772d9a21b7.png)|![Simulator Screen Shot - iPhone 8 Plus - 2022-01-19 at 18 01 35](https://user-images.githubusercontent.com/1942953/150178931-bf50527e-521e-428d-85aa-fbaaa2fdfd97.png)|![Simulator Screen Shot - iPhone 8 Plus - 2022-01-19 at 18 01 49](https://user-images.githubusercontent.com/1942953/150178965-cf0f9af1-a58d-4f2c-adb4-a54762fb4e67.png)|![Simulator Screen Shot - iPhone 8 Plus - 2022-01-19 at 18 01 51](https://user-images.githubusercontent.com/1942953/150178973-df22e4d6-a1fb-4cff-b9d9-1f2212fd0354.png)|

it would be nice to have some default message saying that there is nothing to see. maybe some sort of fallback component for empty things.

PR #80, opened 2020-07-28
Update on title header in webView news

SoleCincis:
- just 'Nachrichte' will be visible now on header of webView news article
  - hardcoded in title params instead of previeus condition
-const title is in use in this file in several places , should not be deleted
- this change have been not tested because of network issues

### DIFF
--- a/src/components/screens/NewsItem.js
+++ b/src/components/screens/NewsItem.js
@@ -30,19 +30,17 @@ const INJECTED_JAVASCRIPT_FOR_IFRAME_WEBVIEW = `
 /* eslint-disable complexity */
 /* NOTE: we need to check a lot for presence, so this is that complex */
 export const NewsItem = ({ data, navigation }) => {
-  const { dataProvider, mainTitle, contentBlocks, publishedAt, sourceUrl, settings } = data;
+  const { dataProvider, contentBlocks, publishedAt, sourceUrl, settings } = data;
 
   const logo = dataProvider && dataProvider.logo && dataProvider.logo.url;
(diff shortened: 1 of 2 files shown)


SoleCincis:
Working on issues !

src/components/screens/NewsItem.js
```diff
@@ -42,7 +42,7 @@ export const NewsItem = ({ data, navigation }) => {
     navigation.navigate({
       routeName: 'Web',
       params: {
-        title,
+        title: 'Nachrichte',
```
donni106, line 45: 'Nachricht' ⚠️

SoleCincis: "Please remove the variable, that is not used anymore" , do you mean const title ? I was getting a issues on when I removed it . 
"And please check the other detail screens, if they are all singular", I did they are all singular and Touren is always plural , right ?

donni106: But then a single Tour needs to have "Tour" doesn't it?

SoleCincis: They are all singular now !

donni106:
In the first commit the issue #49 is referencedm which is not correct! There is no issue for this.
I will rebase your commits @SoleCincis and edit the commit message.

SoleCincis:
> In the first commit the issue #49 is referenced, which is not correct! There is no issue for this.
> I will rebase your commits @SoleCincis and edit the commit message before merge.

yess that was out of place! I was looking at this issue that day because I'd like to work on it as soon as possible .

PR #84, opened 2020-08-18
feat(external links): open all links in internal web view

donni106:
- added `openWebScreen` navigation method in components
  where necessary to open links
- if a link is determined, do not open the url in an external browser,
  but navigate to the web view with the web url to open

### DIFF
--- a/src/components/HtmlView.js
+++ b/src/components/HtmlView.js
@@ -53,12 +53,12 @@ const htmlConfig = {
   ignoredTags: IGNORED_TAGS
 };
 
-export const HtmlView = (props) => (
+export const HtmlView = ({ html, tagsStyles, openWebScreen }) => (
   <HTML
-    {...props}
+    html={html}
     {...htmlConfig}
-    onLinkPress={(evt, href) => openLink(href)}
-    tagsStyles={{ ...styles.html, ...props.tagsStyles }}
+    onLinkPress={(evt, href) => openLink(href, openWebScreen)}
+    tagsStyles={{ ...styles.html, ...tagsStyles }}
     emSize={normalize(16)}
     baseFontStyle={styles.baseFontStyle}
     ignoredStyles={['width', 'height']}
@@ -76,7 +76,9 @@ export const HtmlView = (props) => (
 );
 
 HtmlView.propTypes = {
-  tagsStyles: PropTypes.object
+  html: PropTypes.string,
+  tagsStyles: PropTypes.object,
+  openWebScreen: PropTypes.func
 };
 
 HtmlView.defaultProps = {
--- a/src/components/screens/EventRecord.js
+++ b/src/components/screens/EventRecord.js
@@ -4,7 +4,7 @@ import { ActivityIndicator, StyleSheet, View } from 'react-native';
 import { WebView } from 'react-native-webview';
 import _filter from 'lodash/filter';
 
-import { device, normalize, texts } from '../../config';
+import { colors, device, normalize, texts } from '../../config';
 import { HtmlView } from '../HtmlView';
 import { Image } from '../Image';
 import { LoadingContainer } from '../LoadingContainer';
--- a/src/components/screens/EventRecord.js
+++ b/src/components/screens/EventRecord.js
@@ -10,6 +10,7 @@ import { Image } from '../Image';
 import { LoadingContainer } from '../LoadingContainer';
 import { Logo } from '../Logo';
 import { Title, TitleContainer, TitleShadow } from '../Title';
+import { Touchable } from '../Touchable';
 import { Wrapper, WrapperHorizontal } from '../Wrapper';
 import { PriceCard } from './PriceCard';
 import { InfoCard } from './InfoCard';
@@ -30,7 +31,7 @@ const INJECTED_JAVASCRIPT_FOR_IFRAME_WEBVIEW = `
 
 /* eslint-disable complexity */
 /* NOTE: we need to check a lot for presence, so this is that complex */
-export const EventRecord = ({ data }) => {
+export const EventRecord = ({ data, navigation }) => {
   const {
     addresses,
     category,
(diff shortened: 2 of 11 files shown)


src/components/HtmlView.js
```diff
@@ -53,12 +53,12 @@ const htmlConfig = {
...
-export const HtmlView = (props) => (
+export const HtmlView = ({ html, tagsStyles, openWebScreen }) => (
   <HTML
-    {...props}
+    html={html}
     {...htmlConfig}
-    onLinkPress={(evt, href) => openLink(href)}
-    tagsStyles={{ ...styles.html, ...props.tagsStyles }}
+    onLinkPress={(evt, href) => openLink(href, openWebScreen)}
+    tagsStyles={{ ...styles.html, ...tagsStyles }}
```
SoleCincis, line 61: I can see that openWebScreen method is applied everywhere is needed but where did you define what this method/function is doing ?  why is it not part the commits ?

src/components/screens/EventRecord.js
```diff
@@ -44,6 +45,18 @@ export const EventRecord = ({ data }) => {
     title,
     webUrls
   } = data;
+  const link = webUrls && webUrls.length && webUrls[0].url;
+  const rootRouteName = navigation.getParam('rootRouteName', '');
+  // action to open source urls
+  const openWebScreen = (webUrl) =>
```
donni106, line 51: every component has it's special `openWebScreen` method, because there are different params. here for example for `EventRecord` @SoleCincis

src/components/screens/Tour.js
```diff
@@ -30,6 +30,17 @@ export const Tour = ({ data }) => {
...
+  const rootRouteName = navigation.getParam('rootRouteName', '');
+  // action to open source urls
+  const openWebScreen = (webUrl) =>
+    navigation.navigate({
+      routeName: 'Web',
+      params: {
+        title: 'Tour',
+        webUrl,
+        rootRouteName
+      }
```
SoleCincis, line 42: why link it's not needed here?

src/helpers/linkHelper.js
```diff
@@ -23,15 +23,30 @@ function ensureProtocol(link) {
...
+ *
+ * @param {string} linkWithProtocol
+ *
+ * @return true if linkWithProtocol is a web page
+ */
+function isWeb(linkWithProtocol) {
+  return linkWithProtocol.match(/^https?:/);
+}
+
 // https://facebook.github.io/react-native/docs/linking.html#opening-external-links
```
SoleCincis, line 37: I can check what ensureProtocol does on this link

src/components/screens/EventRecord.js
```diff
@@ -44,6 +45,18 @@ export const EventRecord = ({ data }) => {
...
+    navigation.navigate({
+      routeName: 'Web',
+      params: {
+        title: 'Veranstaltung',
+        webUrl: !!webUrl && typeof webUrl === 'string' ? webUrl : link,
+        rootRouteName
+      }
```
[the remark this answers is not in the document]
donni106: the new logic implemented let's every link open inside the web view in the app. there is no more browser opened.
this condition decides, whether to use a given `webUrl` or the `link`. both are strings representing an url.

src/screens/HtmlScreen.js
```diff
@@ -36,14 +36,25 @@ export class HtmlScreen extends React.PureComponent {
...
+    const subQuery = navigation.getParam('subQuery', '');
     const isConnected = this.context.isConnected;
     const fetchPolicy = graphqlFetchPolicy(isConnected);
+    // action to open source urls
+    const openWebScreen = (webUrl) => {
+      return navigation.navigate({
+        routeName: 'Web',
+        params: {
+          title,
+          webUrl: !!webUrl && typeof webUrl === 'string' ? webUrl : subQuery.webUrl,
```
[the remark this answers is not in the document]
donni106: the main data for this component comes from the `subQuery` object.
it is the same as before:

Bildschirmfoto 2020-08-19 um 19 29 24

so we kepp that value as fallback here.

src/components/screens/PointOfInterest.js
```diff
@@ -29,6 +29,17 @@ export const PointOfInterest = ({ data }) => {
...
+  const rootRouteName = navigation.getParam('rootRouteName', '');
+  // action to open source urls
+  const openWebScreen = (webUrl) =>
+    navigation.navigate({
+      routeName: 'Web',
+      params: {
+        title: 'Ort',
+        webUrl,
+        rootRouteName
+      }
```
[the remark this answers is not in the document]
donni106: it is needed. the `webUrl` is the link (string with url of some web page).

src/components/screens/Tour.js
donni106: it is needed. the `webUrl` is the link (string with url of some web page).

src/helpers/linkHelper.js
donni106: the whole `Linking` stuff is documented with different cases.
i can see, that the documentation has updated to `async/await`. we are still using `.then(...)`, which achieves the same.

src/components/HtmlView.js
donni106: commented here: https://github.com/ikuseiGmbH/smart-village-app-app/pull/84#discussion_r473092530

src/components/screens/EventRecord.js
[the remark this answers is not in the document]
SoleCincis: Ok, so there is no more difference between Link and webUrl, if not, could we call them all webUrl or link?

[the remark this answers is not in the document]
donni106: This method gets called with a `webUrl` or is using `link` from `const link = webUrls && webUrls.length && webUrls[0].url;`. We cannot give identical variable names to both different variables.
But I agree, that the namings can confuse. I have no simpler or more clear idea right now.

SoleCincis (commented):
approved

src/screens/HtmlScreen.js
```diff
@@ -36,14 +36,25 @@ export class HtmlScreen extends React.PureComponent {
...
 
+    const title = navigation.getParam('title', '');
+    const rootRouteName = navigation.getParam('rootRouteName', '');
+    const subQuery = navigation.getParam('subQuery', '');
     const isConnected = this.context.isConnected;
     const fetchPolicy = graphqlFetchPolicy(isConnected);
+    // action to open source urls
+    const openWebScreen = (webUrl) => {
+      return navigation.navigate({
+        routeName: 'Web',
```
donni106, line 50: Here is a mistake. A condition is necessary to decide, whether to navigate to `Web` or a given `routeName` from `subQuery`. I fixed it with https://github.com/ikuseiGmbH/smart-village-app-app/commit/1baffbb3baa72c783b1bbc5f5bef32cb41859b56

donni106:
Passing `openWebScreen` prop was missing in `Tour` and `PointOfInterest`. I fixed it with https://github.com/ikuseiGmbH/smart-village-app-app/commit/f9132f7c9863882285fff9782d76c4e3df5fe343

PR #85, opened 2020-08-19
Pagination for text lists

donni106:
This pull request introduces pagination for text lists. This will improve performance and usability significantly, as only a bunch of data is requested at a time. If there are plenty of data for a list, it takes too much time to query and render everything at one time.

Please see the commits for more details.

### DIFF
--- a/src/components/TextList.js
+++ b/src/components/TextList.js
@@ -2,7 +2,7 @@ import PropTypes from 'prop-types';
 import React from 'react';
 import { ActivityIndicator, FlatList } from 'react-native';
 
-import { normalize } from '../config';
+import { colors, normalize } from '../config';
 import { TextListItem } from './TextListItem';
 
 export class TextList extends React.PureComponent {
@@ -25,7 +25,9 @@ export class TextList extends React.PureComponent {
         )}
         ListFooterComponent={
           data.length > 10 &&
-          !listEndReached && <ActivityIndicator style={{ margin: normalize(14) }} />
+          !listEndReached && (
+            <ActivityIndicator color={colors.accent} style={{ margin: normalize(14) }} />
+          )
         }
         onEndReached={() => this.setState({ listEndReached: true })}
       />
--- a/src/components/TextList.js
+++ b/src/components/TextList.js
@@ -12,6 +12,20 @@ export class TextList extends React.PureComponent {
 
   keyExtractor = (item, index) => `index${index}-id${item.id}`;
 
+  onEndReached = async () => {
+    const { query, fetchMoreData } = this.props;
+
+    if (fetchMoreData) {
+      // if there is a pagination, the end of the list is reached, when no more data is returned
+      // from partially fetching, so we need to check the the data to determine the lists end
+      const { data: moreData } = await fetchMoreData();
+
+      this.setState({ listEndReached: !moreData[query].length });
+    } else {
+      this.setState({ listEndReached: true });
+    }
+  };
+
   render() {
     const { listEndReached } = this.state;
     const { data, navigation, noSubtitle } = this.props;
@@ -29,7 +43,8 @@ export class TextList extends React.PureComponent {
             <ActivityIndicator color={colors.accent} style={{ margin: normalize(14) }} />
           )
         }
-        onEndReached={() => this.setState({ listEndReached: true })}
+        onEndReachedThreshold={0.5}
+        onEndReached={this.onEndReached}
       />
     );
   }
@@ -38,7 +53,9 @@ export class TextList extends React.PureComponent {
 TextList.propTypes = {
   navigation: PropTypes.object.isRequired,
   data: PropTypes.array.isRequired,
-  noSubtitle: PropTypes.bool
+  noSubtitle: PropTypes.bool,
+  query: PropTypes.string,
+  fetchMoreData: PropTypes.func
 };
 
 TextList.defaultProps = {
--- a/src/queries/eventRecords.js
+++ b/src/queries/eventRecords.js
@@ -1,8 +1,8 @@
 import gql from 'graphql-tag';
 
 export const GET_EVENT_RECORDS = gql`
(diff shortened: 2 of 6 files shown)


src/components/TextList.js
```diff
@@ -29,7 +43,8 @@ export class TextList extends React.PureComponent {
             <ActivityIndicator color={colors.accent} style={{ margin: normalize(14) }} />
           )
         }
-        onEndReached={() => this.setState({ listEndReached: true })}
+        onEndReachedThreshold={0.5}
```
SoleCincis, line 46: what is this 0.5 ?

src/components/TextList.js
```diff
@@ -12,6 +12,20 @@ export class TextList extends React.PureComponent {
...
+  onEndReached = async () => {
+    const { query, fetchMoreData } = this.props;
+
+    if (fetchMoreData) {
+      // if there is a pagination, the end of the list is reached, when no more data is returned
+      // from partially fetching, so we need to check the the data to determine the lists end
+      const { data: moreData } = await fetchMoreData();
+
+      this.setState({ listEndReached: !moreData[query].length });
+    } else {
```
SoleCincis, line 24: this determine the end of lists ?

src/screens/IndexScreen.js
```diff
@@ -209,10 +218,26 @@ export class IndexScreen extends React.PureComponent {
...
+            fetchMore({
+              variables: {
+                offset: listItems.length
+              },
+              updateQuery: (prevResult, { fetchMoreResult }) => {
+                if (!fetchMoreResult || !fetchMoreResult[query].length) return prevResult;
+
+                return { [query]: [...prevResult[query], ...fetchMoreResult[query]] };
+              }
+            });
```
SoleCincis, line 231: please here I need an explanation line by line ,   [Offset-based](https://www.apollographql.com/docs/react/v2.4/features/pagination/#offset-based) didn't help me

src/screens/HomeScreen.js
```diff
@@ -323,7 +323,7 @@ export const HomeScreen = ({ navigation }) => {
                             params: {
                               title: 'Veranstaltungen',
                               query: 'eventRecords',
-                              queryVariables: { order: 'listDate_ASC' },
+                              queryVariables: { limit: 15, order: 'listDate_ASC' },
                               rootRouteName: 'EventRecords'
                             }
                           })
```
SoleCincis, line 329: limit of 15 queryVariables per time

src/queries/eventRecords.js
```diff
@@ -1,8 +1,8 @@
 import gql from 'graphql-tag';
 
 export const GET_EVENT_RECORDS = gql`
-  query EventRecords($limit: Int, $order: EventRecordsOrder) {
-    eventRecords(limit: $limit, order: $order) {
+  query EventRecords($limit: Int, $offset: Int, $order: EventRecordsOrder) {
+    eventRecords(limit: $limit, skip: $offset, order: $order) {
```
SoleCincis, line 5: offset = asking for the next two in the list?

src/screens/IndexScreen.js
SoleCincis: I expected offset to be an integer..

src/components/TextList.js
donni106: Are you asking for `onEndReachedThreshold` or what the value 0.5 achieves?
Have you checked the official React Native documentation? Both is described there: https://reactnative.dev/docs/flatlist#onendreachedthreshold
Maybe in this StackOverflow it is more detailed: https://stackoverflow.com/questions/39366356/what-onendreachedthreshold-really-means-in-react-native

src/components/TextList.js
donni106: `onEndReached` of a FlatList in React Native is triggered, when a user scrolled at the end: https://reactnative.dev/docs/flatlist#onendreached

We pass this method to the prop to call its logic.
With pagination, the end of a list is reached multiple times, because data gets loaded in partially fetched batches. We have a state variable to save the real end.
If there is a pagination, the real end of the list is reached, when no more data is returned from partially fetching, so we need to check the the data to determine the lists end.
Otherwise, without pagination, the end is reached when the lists end is reached. We do not need further logics and can set the variable.

src/screens/IndexScreen.js
donni106: correct, `offset` needs to be an integer/a number.

`fetchMore` is the method from Apollos query component to fetch data in batches. So with pagination we want exactly that. Maybe you can search for general articles about pagination, when the specific one is too detailed?
We call that method everytime a user scrolls at the end of a list to fetch more data to append to the list. The offset increases with every call, because the list gets more items with every call.
`updateQuery` is the callback method from Apollo, which needs to be used to add newly fetched data (`fetchMoreResult `) to the existing list (`prevResult`).
As you can see in Insomnia for example, our returned data from the server is always in a schema like
```
{
  "data": {
    "eventRecords": [
      ...
    ]
  }
}
```

`query` is the type, for example `eventRecords`
So the return of `updateQuery` in `fetchMore` is `{ "eventRecords": [...]  }`, as we need it to work with the data passed to the list. It must be the same structure as the initial call.

src/screens/HomeScreen.js
donni106: no. `limit` is a query variable with the value 15.

src/queries/eventRecords.js
donni106: with `offset`, the query returns not from the first element, but from the element index given per `offset` query variable.

src/screens/HomeScreen.js
SoleCincis: [queries-limit](https://www.postgresql.org/docs/8.2/queries-limit.html ) not helping me ... 
why do you set 15 as a value for the query variable limit ?

donni106: From your link:

> If a limit count is given, no more than that many rows will be returned

We want to request max. 15 items with every request.

PR #87, opened 2020-08-24
feat(network error): retry authentication with forcing new token

donni106:
- added logic on network error while `auth()` on app
  start to retry the authentication with a forced fresh
  token request
  - if the app fails once with the default logic, the
    app tries the request again with requesting a new
    auth token
  - this can only be done once, otherwise we could
    crash in an endless loop of retries

---

This feature makes it possible to switch secrets and change the main server without problems. This was necessary for smooth development.

### DIFF
--- a/src/auth.js
+++ b/src/auth.js
@@ -19,9 +19,15 @@ const isTokenValid = async () => {
   return now < expires;
 };
 
-export const auth = async (callback) => {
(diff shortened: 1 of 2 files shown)


donni106:
This feature makes it possible to switch secrets and change the main server without problems.

PR #90, opened 2020-09-17
Add filtering of news items based on data providers

donni106:
In order to filter news items, there is a new filter dropdown of data providers in the header of the list on the index screen of news items. A selection re-queries the server with the selected data provider as query variable, to show a filtered list.

1 | 2
------------ | -------------
Simulator Screen Shot - iPhone 8 Plus - 2020-09-17 at 12 21 36 | Simulator Screen Shot - iPhone 8 Plus - 2020-09-17 at 12 21 37


donni106:
We still have the `IndexFilter`component in here, which we do not use. This needs to be removed.
Please add a removal commit @SoleCincis

src/components/ListHeader.js
```diff
@@ -11,12 +9,10 @@ export const ListHeader = ({ queryVariables, data, updateListData }) => {
...
     data.dataProviders &&
-    _sortBy(_uniq(data.dataProviders.map((item) => item.dataProvider.name)), (item) =>
-      item.toUpperCase()
-    ).map((dataProvider, index) => ({
+    data.dataProviders.map((dataProvider, index) => ({
       id: index + 1,
-      value: dataProvider,
-      selected: dataProvider === queryVariables.dataProvider
+      value: dataProvider.name,
+      selected: dataProvider.name === queryVariables.dataProvider
```
SoleCincis, line 15: This line is to be sure that selected dataProvider on the app matches one of the listed dataProviders in the queryVariables ?

donni106: this line sets `selected` to `true` or `false` depending on the `dataProvider.name` of the current step in the loop and the to that point selected data provider in query variables, which is the selected one from the dropdown.

**case 1:**
```
// current object in the loop
{
  "name": "OSF"
}

// queryVariables to that point
{
  "limit": 15,
  "dataProvider": "OSF"
}

// result from the map
{
  id: 1,
  value: "OSF",
  selected: true
}
```

**case 2:**
```
// current object in the loop
{
  "name": "rbb|24"
}

// queryVariables to that point
{
  "limit": 15,
  "dataProvider": "OSF"
}

// result from the map
{
  id: 2,
  value: "rbb|14",
  selected: false
}
```

src/components/ListHeader.js
```diff
@@ -16,18 +16,35 @@ export const ListHeader = ({ data }) => {
...
+  const selectedDropdownData = dropdownData.find((entry) => entry.selected);
+
+  // https://medium.com/swlh/prevent-useeffects-callback-firing-during-initial-render-the-armchair-critic-f71bc0e03536
+  const initialRender = useRef(true);
+
+  // influence list data when changing selected dropdown value
+  // call update of the list with the selected data provider
+  useEffect(() => {
+    if (initialRender.current) {
+      initialRender.current = false;
```
SoleCincis, line 36: in this way we are preventing the default initialRender when useEffect is triggered

src/components/ListHeader.js
```diff
@@ -16,18 +16,35 @@ export const ListHeader = ({ data }) => {
...
+    } else {
+      // do not pass the value, if the id is 0, because we do not want to use "- Alle -" inside of
+      // updateListData
+      updateListData(!!selectedDropdownData.id && selectedDropdownData.value);
```
SoleCincis, line 40: how do we make sure that id > 0 ?

SoleCincis (approved):

src/components/ListHeader.js
donni106: `selectedDropdownData.id` is 0, 1, 2, 3, ...
checking the boolean interpretation of a integer results in the following:

```
!!0 => false
!!1 => true
!!2 => true
!!3 => true
```

so we can say `!!selectedDropdownData.id && selectedDropdownData.value` to ensure, that it is passing just values where id > 0

src/components/ListHeader.js
donni106: yes, what we got from the link a few lines above: https://medium.com/swlh/prevent-useeffects-callback-firing-during-initial-render-the-armchair-critic-f71bc0e03536

PR #94, opened 2020-09-24
Recognize orientation changes and provide new dimensions

donni106:
Used https://docs.expo.io/versions/latest/sdk/screen-orientation to implement a listener for orientation changes. Please see commits for further information.

Updated the `ImageCarousel` component to respond to orientation changes.

iPhone Landscape | iPad Landscape
------------ | -------------
Simulator Screen Shot - iPhone 8 Plus - 2020-09-24 at 15 14 24 | Simulator Screen Shot - iPad Pro (9 7-inch) - 2020-09-24 at 15 14 32

### DIFF
--- a/src/OrientationProvider.js
+++ b/src/OrientationProvider.js
@@ -9,7 +9,7 @@ const defaultDimensions = {
 };
 
 // default orientation should be portrait up, because we defined that in the expo configs. but we
-// want to be sure with checking the devive dimension.
+// want to be sure with checking the device dimension.
 const defaultOrientation =
   defaultDimensions.width < defaultDimensions.height ? 'portrait' : 'landscape';
 
--- a/src/OrientationProvider.js
+++ b/src/OrientationProvider.js
@@ -2,7 +2,7 @@ import PropTypes from 'prop-types';
 import React, { createContext, useEffect, useState } from 'react';
 import { Dimensions } from 'react-native';
 
-// TODO: import orientation package
+import * as ScreenOrientation from 'expo-screen-orientation';
 
 const defaultDimensions = {
   height: Dimensions.get('window').height,
@@ -17,10 +17,17 @@ export const OrientationProvider = ({ children }) => {
   const [dimensions, setDimensions] = useState(defaultDimensions);
 
   useEffect(() => {
-    // TODO: on orientation change set new dimensions
+    ScreenOrientation.addOrientationChangeListener(() => {
+      // https://docs.expo.io/versions/latest/sdk/screen-orientation/#screenorientationaddorientationchangelistenerlistener
+      setDimensions({
+        height: Dimensions.get('window').height,
+        width: Dimensions.get('window').width
+      });
+    });
 
     // returned function will be called on component unmount
-    // TODO: return () => unsubscribe();
+    // https://docs.expo.io/versions/latest/sdk/screen-orientation/#screenorientationremoveorientationchangelisteners
+    return () => ScreenOrientation.removeOrientationChangeListeners();
   }, []);
 
   return (
--- a/src/OrientationProvider.js
+++ b/src/OrientationProvider.js
@@ -1,0 +1,35 @@
+import PropTypes from 'prop-types';
+import React, { createContext, useEffect, useState } from 'react';
+import { Dimensions } from 'react-native';
+
+// TODO: import orientation package
+
+const defaultDimensions = {
+  height: Dimensions.get('window').height,
+  width: Dimensions.get('window').width
+};
+
+export const OrientationContext = createContext({
+  dimensions: defaultDimensions
+});
+
+export const OrientationProvider = ({ children }) => {
+  const [dimensions, setDimensions] = useState(defaultDimensions);
+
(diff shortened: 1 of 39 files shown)


marcometz (approved):
cool! Die Anpassungen, die dafür notwendig waren, sehen gar nicht so komplex aus :-)

src/OrientationProvider.js
```diff
@@ -1,28 +1,50 @@
...
+    ScreenOrientation.addOrientationChangeListener(({ orientationInfo }) => {
+      // we need to wait a short period of time before reading and setting the orientation and
+      // dimensions, because some devices may take longer then other to update the orientation.
+      setTimeout(() => {
+        // https://docs.expo.io/versions/latest/sdk/screen-orientation/#screenorientationaddorientationchangelistenerlistener
+        setOrientation(getOrientation(orientationInfo.orientation));
+        setDimensions({
+          height: Dimensions.get('window').height,
+          width: Dimensions.get('window').width
+        });
```
SoleCincis, line 46: we set the state for Orientation and Dimension, using addOrientationChangeListener method (from expo-screen-orientation) we set new Orientation using the previously declared getOrientation function then we set new Dimensions again using  Dimensions.get from RN .

src/OrientationProvider.js
```diff
@@ -31,7 +53,7 @@ export const OrientationProvider = ({ children }) => {
   }, []);
 
   return (
-    <OrientationContext.Provider value={{ dimensions }}>
+    <OrientationContext.Provider value={{ orientation, dimensions }}>
       {children}
     </OrientationContext.Provider>
```
SoleCincis, line 58: OrientationContext.Provider is passing its logic and orientation and dimensions props to all {children} component across the app , all children inside  <MainAppWithApolloProvider />

src/helpers/imageHelper.js
```diff
@@ -1,14 +1,13 @@
...
-export const imageHeight = (horizontal = false) => {
+export const imageHeight = (width) => {
   // image aspect ratio is 360x180, so for accurate ratio in our view we need to calculate
-  // a factor with our current device with for the image, to set a correct height
-  const factor = imageWidth(horizontal) / 360;
+  // a factor with our current device width for the image, to set a correct height
+  const factor = width / 360;
 
   return 180 * factor;
 };
```
SoleCincis, line 13: I can't formulate a sentence to describe what's happening here

src/components/ImagesCarousel.js
```diff
@@ -24,23 +24,30 @@ const TouchableImage = ({ navigation, item }) => {
...
+  const { orientation, dimensions } = useContext(OrientationContext);
 
   const renderItem = ({ item }) => {
     const { routeName, params } = item.picture;
 
+    // special image width for carousel, to be not full width on landscape
+    const width = orientation === 'landscape' ? dimensions.height : dimensions.width;
+    const imageStyle = {
+      height: imageHeight(width),
+      width
```
SoleCincis, line 48: imageHeight from imageHelper , what is it happening here ? from comment image will not be full width on landscape , how is this happening ?

src/OrientationProvider.js
```diff
@@ -1,28 +1,50 @@
...
 
   useEffect(() => {
-    ScreenOrientation.addOrientationChangeListener(() => {
-      // https://docs.expo.io/versions/latest/sdk/screen-orientation/#screenorientationaddorientationchangelistenerlistener
-      setDimensions({
-        height: Dimensions.get('window').height,
-        width: Dimensions.get('window').width
-      });
+    ScreenOrientation.addOrientationChangeListener(({ orientationInfo }) => {
+      // we need to wait a short period of time before reading and setting the orientation and
```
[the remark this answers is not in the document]
donni106: it's from the package and passed from the listener https://docs.expo.io/versions/latest/sdk/screen-orientation/#screenorientationaddorientationchangelistenerlistener
which is a `OrientationChangeEvent` https://docs.expo.io/versions/latest/sdk/screen-orientation/#screenorientationorientationchangeevent
which gives us to attriutes. one is `orientationInfo`

> orientationLock (OrientationLock): The current OrientationLock of the device.
> orientationInfo (ScreenOrientationInfo): The current ScreenOrientationInfo of the device.

src/helpers/imageHelper.js
donni106: a number is returned, which is calculated based on width.

if width is 360, than `factor` is 360 / 360 = 1. then the return is 180 * 1 = 180.
this means, if the width is 360, the height will be 180.
we calculate this to ensure that images always be in 2:1 ratio.

src/components/ImagesCarousel.js
donni106: imageStyle is set with the values passed.

height will be the height based on the width calculated in the helper, which i described in a comment before.
width will be the width depending on the orientation, which ensures, that the image has the width of the device in portrait and the height of the device in landscape. width in portrait equals height in landscape. so at eh end, images are always same sized on portrait and landscape, right?!

PR #97, opened 2020-09-25
Accessibility hints and labels recognized by VoiceOver (iOS) and TalkBack (Android)

SoleCincis:
- tested with voiceOver and compared to other App to test labels and hints level of understanding
- most of interactive elements of the app are accessible from VoiceOver now, using two different properties:
  - accessibilityLabel : gives a title to the element
  - accessibilityHint : gives a functionality to the element
- property accessibility it's not working like aspected for non interactive elements
  - see TODOS
- buttons are not accessible elements, intentionally , because every button title , it's always explanatory of its functionality and voiceOver can access it
- Images are not jet accessible, it would be good to implement a logic that has access to a short image description
- CardList Item not jet accessible would be good to implement a logic that can distinguish Orte from Touren
  - maybe for textList Item as well since there is no difference at the moment between News and Veranstaltungen items in index and preview 
- Dropdown and Carousel are not accessible elements jet

SVA-94

- VoiceOver (iOS) and TalkBack (Android) can read :
  - in TextLIstItem the specific title of that item + with a proper pause makes clear what kind of element , in this case a button
  - in CardListItem the specific category and the specific title of that item +  pause and what kind of element
  - in CategoryListItem the specific title of that item plus using 'count' logic , how many articles are available in this category + pause and what kind of element
  - in InfoCar and OperatingCompanyInfo every element , it says the title/name, the actual content, what kind of element and what kind of action "the element can do"
  - in Button the specific title + pause and what kind of element
  - in HomeScreen the specific heading + description + pause and what kind of element
  - in VersionNumber what kind of version have the App currently
  - in Link description + pause and what kind of element ( it seams that this component is not used )
  - Image, ServiceScreen and AboutScreen are not accessible I couldn't understand why

SVA-94

- semicolon to help voiceover and talkback readability
  - after 'Einträge' and 'App version'
- update accessible is the same as accessible={true}
- removed unnecessary brackets around contact."" in InfoCard and OperatingCompanyInfo
- added missing fax label in InfoCard
- accessibilityHint in HomeScreen , not necessary at the moment

SVA-94

### DIFF
--- a/src/components/CardListItem.js
+++ b/src/components/CardListItem.js
@@ -12,6 +12,7 @@ import { Touchable } from './Touchable';
 export const CardListItem = memo(({ navigation, horizontal, item, orientation, dimensions }) => {
   const { routeName, params, image, category, name } = item;
 
+  // TODO: accessibility logic needs to be implemented
   return (
     <Touchable
       onPress={() =>
--- a/src/components/CardListItem.js
+++ b/src/components/CardListItem.js
@@ -12,9 +12,10 @@ import { Touchable } from './Touchable';
 export const CardListItem = memo(({ navigation, horizontal, item, orientation, dimensions }) => {
   const { routeName, params, image, category, name } = item;
 
-  // TODO: accessibility logic needs to be implemented
+  // TODO: count articles logic could to be implemented
   return (
     <Touchable
+      accessibilityLabel={`(${category}) ${name} (Taste)`}
       onPress={() =>
         navigation.navigate({
           routeName,
--- a/src/components/CategoryListItem.js
+++ b/src/components/CategoryListItem.js
@@ -50,7 +50,7 @@ export class CategoryListItem extends React.PureComponent {
         }
         delayPressIn={0}
         Component={Touchable}
-        accessibilityLabel={`${title} (Anzahl verfügbarer Einträge ${count}) (Taste)`}
+        accessibilityLabel={`${title} (Anzahl verfügbarer Einträge: ${count}) (Taste)`}
       />
     );
   }
--- a/src/components/CategoryListItem.js
+++ b/src/components/CategoryListItem.js
@@ -32,8 +32,8 @@ export class CategoryListItem extends React.PureComponent {
           bottomDivider !== undefined
             ? bottomDivider
             : item.toursCount > 0
-            ? index < section.data.length - 1 // do not show a bottomDivider after last tours entry
-            : true
+              ? index < section.data.length - 1 // do not show a bottomDivider after last entry
+              : true
         }
         topDivider={topDivider !== undefined ? topDivider : false}
         containerStyle={{
--- a/src/components/CategoryListItem.js
+++ b/src/components/CategoryListItem.js
@@ -32,8 +32,8 @@ export class CategoryListItem extends React.PureComponent {
           bottomDivider !== undefined
             ? bottomDivider
             : item.toursCount > 0
-              ? index < section.data.length - 1 // do not show a bottomDivider after last tours entry
-              : true
+            ? index < section.data.length - 1 // do not show a bottomDivider after last tours entry
+            : true
         }
         topDivider={topDivider !== undefined ? topDivider : false}
         containerStyle={{
@@ -50,6 +50,7 @@ export class CategoryListItem extends React.PureComponent {
         }
         delayPressIn={0}
         Component={Touchable}
+        accessibilityLabel={`${title} (Anzahl verfügbarer Einträge ${count}) (Taste)`}
       />
     );
   }
--- a/src/components/Image.js
+++ b/src/components/Image.js
@@ -34,24 +34,18 @@ export const Image = ({ source, style, PlaceholderContent }) => {
   }, []);
 
   if (!globalSettings.showImageRights || !source.copyright) {
-    // TODO: how to make an image accessible? is the accessibilityLabel working? maybe put it at the surrounding View?
     return (
       <RNEImage
         source={uri ? (source.uri ? { uri } : uri) : null}
         style={style}
         PlaceholderContent={PlaceholderContent}
         placeholderStyle={{ backgroundColor: colors.transparent }}
-        // TODO: is accessible needed?
-        accessible
-        // TODO: needs correct label. we want to ue a captionText here, which needs to come as a prop from outside
-        //       only apply that caption text to accessibilityLabel if there is one
-        // the captionText can be available per source.captionText
-        accessibilityLabel={`${PlaceholderContent}`}
+        accessible={!!source.captionText}
+        accessibilityLabel={source.captionText}
       />
     );
   }
 
-  // TODO: how to make an image accessible? is the accessibilityLabel working? maybe put it at the surrounding View?
   return (
     <View>
       <RNEImage
@@ -59,12 +53,8 @@ export const Image = ({ source, style, PlaceholderContent }) => {
         style={style}
         PlaceholderContent={PlaceholderContent}
         placeholderStyle={{ backgroundColor: colors.transparent }}
-        // TODO: is accessible needed?
-        accessible
-        // TODO: needs correct label. we want to ue a captionText here, which needs to come as a prop from outside
-        //       only apply that caption text to accessibilityLabel if there is one
-        // the captionText can be available per source.captionText
-        accessibilityLabel={`${PlaceholderContent}`}
+        accessible={!!source.captionText}
+        accessibilityLabel={source.captionText}
       />
       <ImageRights imageRights={source.copyright} />
     </View>
--- a/src/components/Image.js
+++ b/src/components/Image.js
@@ -34,18 +34,24 @@ export const Image = ({ source, style, PlaceholderContent }) => {
   }, []);
 
   if (!globalSettings.showImageRights || !source.copyright) {
+    // TODO: how to make an image accessible? is the accessibilityLabel working? maybe put it at the surrounding View?
     return (
       <RNEImage
         source={uri ? (source.uri ? { uri } : uri) : null}
         style={style}
         PlaceholderContent={PlaceholderContent}
         placeholderStyle={{ backgroundColor: colors.transparent }}
+        // TODO: is accessible needed?
         accessible
+        // TODO: needs correct label. we want to ue a captionText here, which needs to come as a prop from outside
+        //       only apply that caption text to accessibilityLabel if there is one
+        // the captionText can be available per source.captionText
         accessibilityLabel={`${PlaceholderContent}`}
       />
     );
   }
-  // TODO Image is not accessible here, why ?
+
+  // TODO: how to make an image accessible? is the accessibilityLabel working? maybe put it at the surrounding View?
   return (
     <View>
       <RNEImage
@@ -53,7 +59,11 @@ export const Image = ({ source, style, PlaceholderContent }) => {
         style={style}
         PlaceholderContent={PlaceholderContent}
         placeholderStyle={{ backgroundColor: colors.transparent }}
+        // TODO: is accessible needed?
         accessible
+        // TODO: needs correct label. we want to ue a captionText here, which needs to come as a prop from outside
+        //       only apply that caption text to accessibilityLabel if there is one
+        // the captionText can be available per source.captionText
         accessibilityLabel={`${PlaceholderContent}`}
       />
       <ImageRights imageRights={source.copyright} />
--- a/src/components/Image.js
+++ b/src/components/Image.js
@@ -40,7 +40,7 @@ export const Image = ({ source, style, PlaceholderContent }) => {
         style={style}
         PlaceholderContent={PlaceholderContent}
         placeholderStyle={{ backgroundColor: colors.transparent }}
-        accessible={true}
+        accessible
         accessibilityLabel={`${PlaceholderContent}`}
       />
     );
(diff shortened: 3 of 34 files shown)


SoleCincis:
have to fix the issues

donni106:
you have create the PR to master and not to the base branch, where you started from. that was the reason for more commits than yours in this PR. i changed that on saturday. DeepCode's analysis has not started again after that, so their warnings are still for the code from the other commits I think. I haven't find a way to re-start DeepCode analysis. Maybe you find one. If not, i will see, when making the review, hopefully tomorrow.

src/components/CardListItem.js
```diff
@@ -14,6 +14,7 @@ export class CardListItem extends React.PureComponent {
     const { navigation, horizontal, item } = this.props;
     const { routeName, params, image, category, name } = item;
 
+    // TODO accessibility logic needs to be implemented
```
donni106, line 17: currently it is read strange with "in progress {category} {title}" without pause.
maybe we need a accessible "({category})  {title} (Taste)", right? to have nice pause while reading between category, title and taste

src/components/screens/InfoCard.js
```diff
@@ -31,7 +31,11 @@ const contactView = (contact) => (
     {!!contact.phone && (
       <InfoBox>
         <Icon xml={phoneIcon(colors.primary)} style={styles.margin} />
-        <TouchableOpacity onPress={() => openLink(`tel:${contact.phone}`)}>
+        <TouchableOpacity
+          accessibilityLabel="Telefonnummer"
```
donni106, line 35: better the phone number itself with (Taste)
and than a hint saying that the call will open. something like "Wechselt zur Telefon-App"

src/components/screens/InfoCard.js
```diff
@@ -40,7 +44,11 @@ const contactView = (contact) => (
     {!!contact.email && (
       <InfoBox>
         <Icon xml={mail(colors.primary)} style={styles.margin} />
-        <TouchableOpacity onPress={() => openLink(`mailto:${contact.email}`)}>
+        <TouchableOpacity
+          onPress={() => openLink(`mailto:${contact.email}`)}
+          accessibilityLabel="E-mail-Adresse"
+          accessibilityHint="Navigieren zur E-mail-App"
```
donni106, line 50: lets go with "Wechselt zu E-Mail-App" here, "navigieren" indicates moving inside of the app.

src/components/screens/InfoCard.js
```diff
@@ -40,7 +44,11 @@ const contactView = (contact) => (
     {!!contact.email && (
       <InfoBox>
         <Icon xml={mail(colors.primary)} style={styles.margin} />
-        <TouchableOpacity onPress={() => openLink(`mailto:${contact.email}`)}>
+        <TouchableOpacity
+          onPress={() => openLink(`mailto:${contact.email}`)}
+          accessibilityLabel="E-mail-Adresse"
```
donni106, line 49: should be the real address with (Taste) here,right?

src/components/screens/InfoCard.js
```diff
@@ -133,7 +145,11 @@ export const InfoCard = ({ addresses, category, contact, contacts, webUrls, open
         return (
           <InfoBox key={index}>
             <Icon xml={urlIcon(colors.primary)} style={styles.margin} />
-            <TouchableOpacity onPress={() => openLink(url, openWebScreen)}>
+            <TouchableOpacity
+              onPress={() => openLink(url, openWebScreen)}
+              accessibilityLabel="Webseite"
+              accessibilityHint="Navigieren zur Webseite"
```
donni106, line 151: "Öffnet Webseite in der aktuellen App"

src/screens/AboutScreen.js
```diff
@@ -62,15 +62,18 @@ export const AboutScreen = ({ navigation }) => {
...
 
+          // TODO accessibility not working on Title
           return (
             <ScrollView>
               {!!headlineAbout && (
-                <TitleContainer>
-                  <Title>{headlineAbout}</Title>
+                <TitleContainer accessible={true}>
+                  <Title accessibilityHint="Kopfzeile der Über die App Liste">
+                    {headlineAbout}
```
donni106, line 75: maybe for the title we add "headline (Überschrift) (Taste)"? with the addition of "Überschrift" for headlines?

src/components/CardListItem.js
```diff
@@ -14,7 +14,7 @@ export class CardListItem extends React.PureComponent {
     const { navigation, horizontal, item } = this.props;
     const { routeName, params, image, category, name } = item;
 
-    // TODO accessibility logic needs to be implemented
+    // TODO count articles logic could to be implemented
```
SoleCincis, line 17: this logic to count could also be applied in the TextList, so that the user can know the index of the element she/he clicked on

src/components/Image.js
```diff
@@ -39,17 +39,21 @@ export const Image = ({ source, style, PlaceholderContent }) => {
...
         PlaceholderContent={PlaceholderContent}
         placeholderStyle={{ backgroundColor: colors.transparent }}
+        accessible={true}
+        accessibilityLabel={`${PlaceholderContent}`}
       />
     );
   }
-
+  // TODO Image is not accessible here, why ?
   return (
```
SoleCincis, line 48: we need content for PlaceholderContent

src/screens/AboutScreen.js
SoleCincis: despite the accessibility method this element remains "invisible" to VoiceOver, this same logic worked in VersionNumber, even though VersionNumber does not have the onPress logic exactly like this title component

SoleCincis:
(!!!)

src/components/Image.js
donni106: What do you mean with PlaceholderContent?

SoleCincis:
**Card list preview**: when scrolling horizontally voice Over recognizes a page change and declares the change instead of reading the next card item, user must click a second time to hear the card.

src/components/Image.js
SoleCincis: PlaceholderContent I thought this is a sentence that describe a bit the image in case it's not visible for some reasons ... oder ?

src/components/CategoryListItem.js
```diff
@@ -50,6 +50,7 @@ export class CategoryListItem extends React.PureComponent {
         }
         delayPressIn={0}
         Component={Touchable}
+        accessibilityLabel={`${title} (Anzahl verfügbarer Einträge ${count}) (Taste)`}
```
donni106, line 53: nice additional info about the entries!
plese try to add a ":" to make voiceover and talback read the number better after the text?
`Anzahl verfügbarer Einträge: ${count}`

src/components/Image.js
donni106: i have to say, i do not know what `PlaceholderContent` is :)
but it seems to be not a correct object to pass in `accessibilityLabel`. this needs to be checked again!

src/components/VersionNumber.js
```diff
@@ -9,7 +9,12 @@ import { Wrapper } from './Wrapper.js';
 export const VersionNumber = () => {
   return (
     <Wrapper>
-      <RegularText small style={styles.version}>
+      <RegularText
+        small
+        style={styles.version}
+        accessible={true}
+        accessibilityLabel={`(App version:) ${appJson.expo.version} `}
```
donni106, line 16: i think it is enough to wrap the text in `(...)` or to add a `:` to have a short pause in reading.
there is also a whitespace at the end before closing `, please clean it up.

src/components/VersionNumber.js
```diff
@@ -9,7 +9,12 @@ import { Wrapper } from './Wrapper.js';
 export const VersionNumber = () => {
   return (
     <Wrapper>
-      <RegularText small style={styles.version}>
+      <RegularText
+        small
+        style={styles.version}
+        accessible={true}
```
donni106, line 15: **good to know**:
for passing a prop with the value `true`, it can be shortend with just `accessible`, without explicitly giving the value true.
we already do that in many places in the code.

`<RegularText accessible>` is the same as `<RegularText accessible={true}>`

src/components/screens/InfoCard.js
```diff
@@ -32,8 +32,7 @@ const contactView = (contact) => (
       <InfoBox>
         <Icon xml={phoneIcon(colors.primary)} style={styles.margin} />
         <TouchableOpacity
-          accessibilityLabel="Telefonnummer"
-          accessibilityHint="Navigieren zu anruf option"
+          accessibilityLabel={`(Telefonnummer) (${contact.phone}) (Taste) (Wechselt zur Telefon-App)`}
```
donni106, line 35: nice breaks. around `${contact.phone}` could be too much, as the text before and the text after is already in `(...)`?

src/components/screens/InfoCard.js
```diff
@@ -71,8 +69,17 @@ export const InfoCard = ({ addresses, category, contact, contacts, webUrls, open
...
-        <RegularText>{category.name}</RegularText>
+        <RNEIcon
+          accessible={true}
+          accessibilityLabel="Menü icon"
+          name="list"
+          type="material"
+          color={colors.primary}
+          iconStyle={styles.margin}
+        />
+        <RegularText accessible={true} accessibilityLabel={`(Kategorie) (${category.name}`}>
```
donni106, line 80: here you started a `(` but have not closed it `(${category.name}`.
as there is no text after it and the text before is already wrapped, the category name does not need to be wrapped in `(...)`

src/components/screens/OperatingCompanyInfo.js
```diff
@@ -104,22 +102,21 @@ export const OperatingCompanyInfo = ({ address, contact, name, webUrls, openWebS
...
-            <InfoBox accessible={true}>
+            <InfoBox>
               <RNEIcon
                 name="print"
                 type="material"
                 color={colors.primary}
                 iconStyle={{ marginRight: normalize(10) }}
               />
-              <RegularText primary accessibilityLabel="Fax-nummer">
+              <RegularText primary accessible={true} accessibilityLabel={`(Fax) (${contact.fax}`}>
```
donni106, line 119: nice that you have put `Fax` inside. otherwise it would be just a number without context

src/screens/AboutScreen.js
donni106: thanks. strange, yes. i have no idea, yet. let's have a check together next week.

src/screens/HomeScreen.js
```diff
@@ -165,7 +164,7 @@ export const HomeScreen = ({ navigation }) => {
                     }
                   })
                 }
-                accessibilityLabel="Kopfzeile der Vorschau der Touren- und Orteliste"
+                accessibilityLabel={`${headlinePointsOfInterestAndTours} (Überschrift) (Taste)`}
                 accessibilityHint="Navigiert zur Touren- und Orteliste"
```
donni106, line 168: forgot to remove the hint, or intended?

SoleCincis: forgot !

src/components/Image.js
SoleCincis: So my idea is to have an image + an image short description that is coming per server ( which our clients are providing ) with the aim of making picture accessible as well . What do you think ?

donni106: you are right, we do not have the information about the image. this needs to come aside with the image data.
let us check if we already get some image descriptions in the whole data, that we could use here.

src/screens/AboutScreen.js
SoleCincis: "The accessibility label on a component with accessible enabled is constructed by concatenating all children" 

from : https://sophieau.com/article/react-native-a11y/

donni106: works from a newer test today

src/components/CardListItem.js
donni106: good idea, we will put it on a roadmap for the future

src/components/Image.js
donni106: there is often a `captionText` for images, we can use

src/components/screens/NewsItem.js
```diff
@@ -208,13 +208,13 @@ export const NewsItem = ({ data, navigation }) => {
...
             <Touchable onPress={openWebScreen}>
-              <Title>{trimNewLines(title)}</Title>
+              <Title accessibilityLabel={`${title} (Überschrift)`}>{trimNewLines(title)}</Title>
             </Touchable>
           </TitleContainer>
         ) : (
           !!title && (
             <TitleContainer>
-              <Title>{trimNewLines(title)}</Title>
+              <Title accessibilityLabel={`${title} (Überschrift)`}>{trimNewLines(title)}</Title>
```
donni106, line 217: please add the `trimNewLines(title)` also in the `accessibilityLabel`, so we will not have problems with some new lines that can exist

src/components/screens/PointOfInterest.js
```diff
@@ -87,7 +87,7 @@ export const PointOfInterest = ({ data, navigation }) => {
         {/* {!!location && (
           <View>
             <TitleContainer>
-              <Title>{texts.pointOfInterest.location}</Title>
+              <Title accessibilityLabel={`${texts.pointOfInterest.location} (Überschrift)`}>{texts.pointOfInterest.location}</Title>
```
donni106, line 90: does the line length fit here, or better break it in three lines?

src/components/screens/TourCard.js
```diff
@@ -24,7 +24,7 @@ const addressOnPress = (address) => {
 export const TourCard = ({ addresses, lengthKm }) => (
   <View>
     <TitleContainer>
-      <Title>{texts.tour.tour}</Title>
+      <Title accessibilityLabel={`${texts.tour.tour} (Überschrift) `}>{texts.tour.tour}</Title>
```
donni106, line 27: small thing: a space too much after (Überschrift)

PR #98, opened 2020-09-28
Update icons in tab bar

SoleCincis:
- we are using tow different kind of Icons in our tabBar;
  - first two are 'svg' fils saved in the app ( home and service )
  - second two are 'Ionicons', which have a small white frame on the bottom on the icon
     - that's why they were having some margin that now I remove
- add some marginTop for all tabBar icons in order to have space between icons and top of tabBar
 - now icons in tabBar are aligned, and text looks centred
  - but still looks a bit off on iPad screen ( info on SVA-92 )
- PropTypes for name and focused were missing

 SVA-93

- on changes orientation, tab bar icons get some extra margin to the right
 - both (Icon and TabBarIcon) are wrapped in a view which as a landscapeStyle prop
   - and just on TabBarIcon component icon size increases a bit
    - here there is a bug , on landscape this two icons are cut on the right
- TabBarIcon component is not default function component anymore that's why now it's imported inside curly brackets

 SVA-93

- TODO is made so comment is gone
- focused prop in TabBarIcon it's a boolean data type not an object
- corrections on Icon are made based on Daniel's comment ( https://github.com/ikuseiGmbH/smart-village-app-app/pull/98/files/8700e6836f9a305fbfa210d62f3252d0d6019c5d#r497405526 )

 SVA-93

before : 
Simulator Screen Shot - iPhone SE (2nd generation) - 2020-09-23 at 14 05 43

now : 
IMG_0274

- Maybe we should use just one kind of icon in tabBar, to simplify stylization. 
here icons examples on debug, the see the difference  : 
svg: 
Screenshot 2020-09-28 at 17 10 34
ionicon: 
Screenshot 2020-09-28 at 17 10 20

SVA-93

### DIFF
--- a/src/components/Icon.js
+++ b/src/components/Icon.js
@@ -1,39 +1,55 @@
 import PropTypes from 'prop-types';
-import React, { useContext } from 'react';
-
-import { StyleSheet, View } from 'react-native';
+import React from 'react';
+import { View } from 'react-native';
 import { SvgXml } from 'react-native-svg';
+import { Ionicons } from '@expo/vector-icons';
 
-import { normalize } from '../config';
-import { OrientationContext } from '../OrientationProvider';
+import { colors, normalize } from '../config';
 
-export const Icon = ({ icon, width, height, style, landscapeStyle }) => {
-  const { orientation } = useContext(OrientationContext);
-  // const landscapeStyle = orientation === 'landscape' ? style : null;
+/**
+ * Smart icon component which can handle SVGs passed as `xml` prop or icon fonts, which are
+ * rendered with a given `name` prop.
+ */
+export const Icon = ({ xml, width, height, name, size, focused, style, iconStyle }) => {
+  const color = focused ? colors.accent : colors.primary;
 
   return (
-    <View style={[style, orientation === 'landscape' && landscapeStyle]}>
-      <SvgXml width={width} height={height} xml={icon} />
+    <View style={style}>
+      {xml && <SvgXml xml={xml} width={width} height={height} style={iconStyle} />}
+      {name && <Ionicons name={name} size={size} color={color} style={iconStyle} />}
     </View>
   );
 };
 
-const styles = StyleSheet.create({
-  marginIcon: {
-    marginRight: normalize(10)
+// thx to: https://stackoverflow.com/a/49682510/9956365
+const isRequired = (props, propName, componentName) => {
+  // ensure, that one of 'xml' or 'name' is given
+  if (!props.xml && !props.name) {
+    return new Error(`One of 'xml' or 'name' is required by '${componentName}' component.`);
   }
-});
+
+  // ensure, that only one of 'xml' or 'name' is passed at a time
+  if (props.xml && props.name) {
+    return new Error(`Only one of 'xml' or 'name' is required by '${componentName}' component.`);
+  }
+};
 
 Icon.propTypes = {
-  icon: PropTypes.string.isRequired,
+  xml: isRequired,
   width: PropTypes.number,
   height: PropTypes.number,
+  name: isRequired,
+  size: PropTypes.number,
+  focused: PropTypes.bool,
   style: PropTypes.object,
-  landscapeStyle: PropTypes.object
+  iconStyle: PropTypes.object
 };
 
 Icon.defaultProps = {
+  // width & height marks the size for svg
   width: normalize(24),
   height: normalize(24),
-  landscapeStyle: styles.marginIcon
+  // size is for font icon
+  size: normalize(26),
+  focused: false
 };
--- a/src/components/Icon.js
+++ b/src/components/Icon.js
@@ -7,12 +7,12 @@ import { SvgXml } from 'react-native-svg';
 import { normalize } from '../config';
 import { OrientationContext } from '../OrientationProvider';
 
-export const Icon = ({ icon, width, height, style }) => {
+export const Icon = ({ icon, width, height, style, landscapeStyle }) => {
   const { orientation } = useContext(OrientationContext);
-  const landscapeStyle = orientation === 'landscape' ? style : null;
+  // const landscapeStyle = orientation === 'landscape' ? style : null;
 
   return (
-    <View style={landscapeStyle}>
+    <View style={[style, orientation === 'landscape' && landscapeStyle]}>
       <SvgXml width={width} height={height} xml={icon} />
     </View>
   );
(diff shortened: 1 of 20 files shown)


donni106:
Isn't there no more space now between top of the tabbar and the Ionicon icons?

SoleCincis:
(!) strange behavior (!) 
sometimes app on landscape looks like this: 
Simulator Screen Shot - iPhone 11 Pro Max - 2020-09-28 at 19 58 37
 reasons still unknown.

donni106:
Can you please provide a clear overview of the Tabbar for portrait and the tabbar for landscape? How was it before and what is better now?

src/components/TabBarIcon.js
```diff
@@ -1,15 +1,21 @@
+import PropTypes from 'prop-types';
 import React from 'react';
 import { Ionicons } from '@expo/vector-icons';
 
 import { colors, normalize } from '../config';
 
+//TODO refactor TabBarIcon to be a function component that can use OrientationContext
```
donni106, line 7: what type of component is it now?

src/navigation/MainTabNavigator.js
```diff
@@ -132,7 +132,12 @@ const MainTabNavigator = createBottomTabNavigator(
     AboutStack
   },
   {
-    tabBarOptions: { activeTintColor: colors.accent, inactiveTintColor: colors.primary }
+    tabBarOptions: {
+      activeTintColor: colors.accent,
+      inactiveTintColor: colors.primary,
+      tabStyle: { marginTop: normalize(5) },
+      labelStyle: { margin: normalize(3) }
```
donni106, line 139: where do you have these attributes from? is there a link to a documentation?
what is the difference between `tabStyle` and `labelStyle`?

src/components/TabBarIcon.js
SoleCincis: it's a default functional component, but we need a "fancy" functional component to be able to use hooks
  which we will need to setup `useContext` for orientation stuff ( from your commits in https://github.com/ikuseiGmbH/smart-village-app-app/pull/94/commits )

donni106:
I needed to run CodeClimate manually https://codeclimate.com/github/ikuseiGmbH/smart-village-app-app/compare/hotfix/SVA-93-tabBar-aligne-icons-text

I don't know why. Maybe because it is a hotfix branch. Please create feature-Branches. You are working on features, not on bugs reported by any customer.

src/navigation/MainTabNavigator.js
SoleCincis: from official documentation:  https://reactnavigation.org/docs/3.x/bottom-tab-navigator/#tabbarlabel 
tabStyle = to style tabBar box 
labelStyle = to style just the text of tabBar 
in labelStyle there is a mistake , it should have a pudding not a margin (!)

src/components/TabBarIcon.js
donni106: but it is already a functional component (definition is `function` and not `class`). so maybe it needs to add orientation functionality, but no refactoring to functional component.

src/navigation/MainTabNavigator.js
SoleCincis, line 139: I should use OrientationContext here, when orientation === 'landscape' increase spacing between icon and label . 
Like in this example : 
IMG_0266

src/components/TabBarIcon.js
```diff
@@ -8,8 +9,12 @@ export default function TabBarIcon(props) {
     <Ionicons
       name={props.name}
       size={normalize(26)}
```
SoleCincis, line 12: Here I should use OrientationContext, when orientation=== 'landscape' then TabBarIcon should increase its size in order to looks like an svg icon which is slightly larger

SoleCincis:
this is how it looked before : IMG_161D26DB3CDB-1
this is how it looks now : IMG_0270

in my version unternehmen is cut because of I add some margin around the labels, which is a mistake , there should be some extra space for the label just on 'landscape' mode

SoleCincis:
Now on portrait is back to normal 
IMG_0272 
but on landscape tabBarIcon are still giving me some problems , the right side of last two icons is cut 
IMG_0276

src/components/TabBarIcon.js
donni106: the TODO comment can now be removed, as you already refactored

src/components/TabBarIcon.js
```diff
@@ -1,15 +1,35 @@
...
+};
+const styles = StyleSheet.create({
+  marginIcon: {
+    marginRight: normalize(10)
+  }
+});
+
+TabBarIcon.propTypes = {
+  name: PropTypes.string.isRequired,
+  focused: PropTypes.object,
```
donni106, line 29: there is a warning on the screen about the wrong PropType defined for `focused`. please correct that

src/screens/CompanyScreen.js
```diff
@@ -19,7 +19,7 @@ import {
 } from '../components';
 import { getQuery } from '../queries';
 import { graphqlFetchPolicy, refreshTimeFor } from '../helpers';
-import TabBarIcon from '../components/TabBarIcon';
+import { TabBarIcon } from '../components/TabBarIcon';
```
donni106, line 22: nice that you have think of updating every place of importing `TabBarIcon`

src/components/Icon.js
```diff
@@ -1,15 +1,28 @@
...
+
+import { StyleSheet, View } from 'react-native';
 import { SvgXml } from 'react-native-svg';
 
 import { normalize } from '../config';
+import { OrientationContext } from '../OrientationProvider';
+
+export const Icon = ({ icon, width, height, style }) => {
+  const { orientation } = useContext(OrientationContext);
+  const landscapeStyle = orientation === 'landscape' ? style : null;
```
donni106, line 12: using `style` for just for landscape changes the whole logic.
the style was there before in every case. now it is just applied to the container `View`, if in landscape mode.

this will lead to problems in components, where `style` is already used, for example in `Link`:

```
<Icon icon={link(colors.secondary)} style={styles.icon} />
```

that icon styles now only apply on landscape mode.

what you want is a new `landscapeStyle`, which is not related to the main style, which can be passed. it does not need to be a component prop i think, or do we need to pass landscape style from outside? maybe to have the ability to customize it. ok, so `landscapeStyle` can be added in the props:

```
({ icon, width, height, style, landscapeStyle })
```

in `View` than both styles need to be added with an array, while the landscape style only, if orientation is landscape:

```
<View style={[style, orientation === 'landscape' && landscapeStyle]}>
```

and as defaultProps you need to change `style`, which should have no default, to `landscapeStyle`.

have you understand? otherwise reach me on Slack.

donni106:
We have components `Icon` and `TabBarIcon`, which are nearly the same. `TabBarIcon` is also used in context outside of the tab bar. We should refactor that and integrate the `TabBarIcon` in `Icon`. Contact me, and we can do that in pair programming.

src/components/Icon.js
SoleCincis: all clear , I'm going to take care of it now .

donni106:
Updated in pair programming:

iOS | Android
----|----
Simulator Screen Shot - iPhone 8 Plus - 2020-09-30 at 16 00 05 | Screenshot_20200930-160417

donni106 (approved):

PR #105, opened 2020-10-02
feat(button): new style on landscape mode

SoleCincis:
- implemented OrientationContext , which knows when device is on portrait or landscape mode
- buttonStyle condition , when on landscape mode button is on 60% of its width and centred in the screen
  - if portrait we don't want to render new style for button so it must be an empty object

SVA-93

feat(button): corrections on style and condition for landscape mode

- buttonStyleLandscape is an appropriate name, more specific
  - it is also in alphabetical order
- refactored conditional logic , aim dry code

SVA-93

- now the condition for new style on landscape mode is placed in the button container
  - using alignSelf in order to see the container shrink to the size of its content
- thanks to this article: https://www.skptricks.com/2019/02/react-native-button-width-fit-to-content.html

SVA-93

### DIFF
--- a/src/components/Button.js
+++ b/src/components/Button.js
@@ -1,12 +1,14 @@
 import PropTypes from 'prop-types';
-import React from 'react';
+import React, { useContext } from 'react';
 import { StyleSheet } from 'react-native';
 import { Button as RNEButton } from 'react-native-elements';
 
 import { colors, normalize } from '../config';
 import { DiagonalGradient } from './DiagonalGradient';
+import { OrientationContext } from '../OrientationProvider';
 
 export const Button = ({ title, onPress }) => {
+  const { orientation } = useContext(OrientationContext);
   return (
     <RNEButton
       onPress={onPress}
@@ -14,6 +16,7 @@ export const Button = ({ title, onPress }) => {
       titleStyle={styles.titleStyle}
       containerStyle={styles.containerStyle}
       ViewComponent={DiagonalGradient}
+      buttonStyle={orientation === 'landscape' ? styles.buttonStyle : {}}
       useForeground
     />
   );
@@ -23,6 +26,10 @@ const styles = StyleSheet.create({
   titleStyle: {
     color: colors.lightestText,
     fontFamily: 'titillium-web-bold'
+  },
+  buttonStyle: {
+    width: '60%',
+    alignSelf: 'center'
   },
   containerStyle: {
     marginBottom: normalize(21)
--- a/src/components/Button.js
+++ b/src/components/Button.js
@@ -16,7 +16,7 @@ export const Button = ({ title, onPress }) => {
       titleStyle={styles.titleStyle}
       containerStyle={styles.containerStyle}
       ViewComponent={DiagonalGradient}
-      buttonStyle={orientation === 'landscape' ? styles.buttonStyle : {}}
+      buttonStyle={orientation === 'landscape' && styles.buttonStyleLandscape}
       useForeground
     />
   );
@@ -27,9 +27,9 @@ const styles = StyleSheet.create({
     color: colors.lightestText,
     fontFamily: 'titillium-web-bold'
   },
-  buttonStyle: {
-    width: '60%',
-    alignSelf: 'center'
+  buttonStyleLandscape: {
+    alignSelf: 'center',
+    width: '60%'
   },
   containerStyle: {
     marginBottom: normalize(21)
--- a/src/components/Button.js
+++ b/src/components/Button.js
@@ -9,14 +9,16 @@ import { OrientationContext } from '../OrientationProvider';
 
 export const Button = ({ title, onPress }) => {
   const { orientation } = useContext(OrientationContext);
+
   return (
     <RNEButton
       onPress={onPress}
       title={title}
-      titleStyle={styles.titleStyle}
-      containerStyle={
-        orientation === 'landscape' ? styles.containerStyleLandscape : styles.containerStyle
-      }
+      titleStyle={[styles.titleStyle, orientation === 'landscape' && styles.titleStyleLandscape]}
+      containerStyle={[
+        styles.containerStyle,
+        orientation === 'landscape' && styles.containerStyleLandscape
+      ]}
       ViewComponent={DiagonalGradient}
       useForeground
     />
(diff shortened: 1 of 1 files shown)


SoleCincis:
This is how the App looks like at the moment, button on 60% of its width: 
**iPhone SE** 
Iphone SE - 60%

 **iPhone 11 Pro Max** 
Simulator Screen Shot - iPhone 11 Pro Max - 60%

**iPad Pro (12.9-inch)**

src/components/Button.js
```diff
@@ -24,6 +27,10 @@ const styles = StyleSheet.create({
     color: colors.lightestText,
     fontFamily: 'titillium-web-bold'
   },
+  buttonStyle: {
+    width: '60%',
+    alignSelf: 'center'
```
donni106, line 32: we like attributes ordered alphabetically, right?

src/components/Button.js
```diff
@@ -24,6 +27,10 @@ const styles = StyleSheet.create({
     color: colors.lightestText,
     fontFamily: 'titillium-web-bold'
   },
+  buttonStyle: {
```
donni106, line 30: if it is just used for landscape, it should be part of the name here, to be clear

src/components/Button.js
```diff
@@ -1,19 +1,22 @@
...
 export const Button = ({ title, onPress }) => {
+  const { orientation } = useContext(OrientationContext);
   return (
     <RNEButton
       onPress={onPress}
       title={title}
       titleStyle={styles.titleStyle}
       containerStyle={styles.containerStyle}
       ViewComponent={DiagonalGradient}
+      buttonStyle={orientation === 'landscape' ? styles.buttonStyle : {}}
```
donni106, line 19: you could leave it undefined if you want, when the condition does not match. example: `orientation === 'landscape' && styles.buttonStyle`

SoleCincis:
This is a 50% variant : 
**iPhone SE** 
IMG_0280
**iPhone 11 Pro Max**
Simulator Screen Shot - iPhone 11 Pro Max - 2020-10-02 at 17 48 11

iPad Pro (12.9-inch) 
Simulator Screen Shot - iPad Pro (12 9-inch) (4th generation) - 2020-10-02 at 17 49 03

src/components/Button.js
SoleCincis: that's right .. !

src/components/Button.js
SoleCincis: right!

donni106:
> This is a 50% variant :
> **iPhone SE**
> IMG_0280
> **iPhone 11 Pro Max**
> Simulator Screen Shot - iPhone 11 Pro Max - 2020-10-02 at 17 48 11
> 
> iPad Pro (12.9-inch)
> Simulator Screen Shot - iPad Pro (12 9-inch) (4th generation) - 2020-10-02 at 17 49 03

Is there a way to let the button make it size depending on the text content? A percentage can be too big or too small for different button texts. Best would be, that the button width is the text width with a bit padding, no matter what its text is.
and for portrait its 100% of the screen width (as it is now by default).
Should we try to achieve that?

SoleCincis:
> > This is a 50% variant :
> > ...
> 
> Is there a way to let the button make it size depending on the text content? A percentage can be too big or too small for different button texts. Best would be, that the button width is the text width with a bit padding, no matter what its text is.
> and for portrait its 100% of the screen width (as it is now by default).
> Should we try to achieve that?

Yes, it sounds like a cool challenge !

SoleCincis:
current status for button: 
IMG_0301
IMG_0298
IMG_0299

SoleCincis:
I think this branch can be merged

src/components/Button.js
```diff
@@ -24,6 +28,11 @@ const styles = StyleSheet.create({
     color: colors.lightestText,
     fontFamily: 'titillium-web-bold'
   },
+  containerStyleLandscape: {
+    justifyContent: 'center',
+    alignItems: 'center',
+    marginBottom: normalize(21)
```
donni106, line 34: alphabetic order...
i will update it

src/components/Button.js
```diff
@@ -1,18 +1,22 @@
...
+import React, { useContext } from 'react';
 import { StyleSheet } from 'react-native';
 import { Button as RNEButton } from 'react-native-elements';
 
 import { colors, normalize } from '../config';
 import { DiagonalGradient } from './DiagonalGradient';
+import { OrientationContext } from '../OrientationProvider';
 
 export const Button = ({ title, onPress }) => {
+  const { orientation } = useContext(OrientationContext);
```
donni106, line 11: empty line before return, please
i will update it

src/components/Button.js
```diff
@@ -1,18 +1,22 @@
...
   return (
     <RNEButton
       onPress={onPress}
       title={title}
       titleStyle={styles.titleStyle}
-      containerStyle={styles.containerStyle}
+      containerStyle={
+        orientation === 'landscape' ? styles.containerStyleLandscape : styles.containerStyle
```
donni106, line 18: to avoid duplication in styles, where `marginBottom: normalize(21)` is defined twice, one can use an array here.
with the array one can set multiple style objects, as well as setting them conditionally.
we use that in other places already. this means we can say:
- `styles.container` should always apply
- `styles.containerLandscape` should only apply on landscape
this results in the following code:
```
containerStyle={[
    styles.containerStyle,
    orientation === 'landscape' && styles.containerStyleLandscape
  ]}
```

this way, just `styles.container` needs the bottom margin, because it is applied every time.
i will update it.

donni106:
> current status for button:
> ...
> IMG_0299

i am missing some horizontal padding. the text of buttons and the button ending are too close, aren't they?
this looks better:

Simulator Screen Shot - iPhone 11 - 2020-10-16 at 11 42 06

i will update it

src/components/Button.js
SoleCincis: thanks

src/components/Button.js
SoleCincis: thanks

src/components/Button.js
SoleCincis: love it!!

src/components/Button.js
```diff
@@ -9,14 +9,16 @@ import { OrientationContext } from '../OrientationProvider';
...
+
   return (
     <RNEButton
       onPress={onPress}
       title={title}
-      titleStyle={styles.titleStyle}
-      containerStyle={
-        orientation === 'landscape' ? styles.containerStyleLandscape : styles.containerStyle
-      }
+      titleStyle={[styles.titleStyle, orientation === 'landscape' && styles.titleStyleLandscape]}
```
donni106, line 17: here for title i used the same array stuff for applying styles

donni106 (approved):

src/components/Button.js
SoleCincis: thanks

PR #108, opened 2020-10-08
feat(cardListitem): new style condition for landscape mode

SoleCincis:
- passing orientation as prop
- created a condition for new style when orientation is triggered

SVA-93

- in Card List deleted a comment , there is not the place where to make changes
- WrapperWithOrientation is now implemented conditionally around cart item
- in addition imageWidth it should be on 50% when orientation is on landscape
  - attempt to implement OrientationContext based on ImagesCarousel
   - getting a red screen about an invalid hooks call

SVA-93

- refactored CardListItem into a functional component
- moved WrapperWithOrientation behind card component
  - that's where should stay in order to wrap the all card and made changes when orientation changes
- temporary deletion of horizontal logic for preview card list (which is visible on home screen only)
  -  it has been deleted due a error that comes up saying "undefined it's not an object"
- if landscape now we have a combo of:
  - dimensions.height which influences images width based on device window
  - and WrapperWithOrientation which centers the card within the screen ( adding padding left and right )

SVA-93

- refactored CardList in order to use hooks
- on CardListItem orientation and dimensions are required because the calculation of the width is depending their
  - in addition wrapper orientation is gone because passed to parent component CardList which takes care of orientation all at once and not every time for each cardList
- in SafeAreaView we have a style prop in order to use a different style when needed
 - this is applied on SafeAreaView in IndexScreen where we need to center the list
SVA-93

### DIFF
--- a/src/components/CardListItem.js
+++ b/src/components/CardListItem.js
@@ -1,5 +1,5 @@
 import PropTypes from 'prop-types';
-import React, { useContext } from 'react';
+import React from 'react';
 import { Card } from 'react-native-elements';
 import { Platform, StyleSheet, View } from 'react-native';
 
@@ -8,12 +8,9 @@ import { imageHeight, imageWidth } from '../helpers';
 import { Image } from './Image';
 import { RegularText, BoldText } from './Text';
 import { Touchable } from './Touchable';
-import { WrapperWithOrientation } from './Wrapper';
-import { OrientationContext } from '../OrientationProvider';
 
-export const CardListItem = ({ navigation, horizontal, item }) => {
+export const CardListItem = ({ navigation, horizontal, item, orientation, dimensions }) => {
   const { routeName, params, image, category, name } = item;
-  const { orientation } = useContext(OrientationContext);
 
   return (
     <Touchable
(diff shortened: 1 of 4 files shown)


src/components/CardListItem.js
```diff
@@ -53,9 +53,14 @@ export class CardListItem extends React.PureComponent {
...
 /* this works properly, we do not want that warning */
-const stylesWithProps = ({ horizontal }) => {
+const stylesWithProps = ({ horizontal, orientation }) => {
   // image width should be only 70% when rendering horizontal cards, otherwise substract paddings
-  const width = horizontal ? imageWidth() * 0.7 : imageWidth() - 2 * normalize(14);
+  // when orientation image width should be device width + double padding TODO: need a fix
+  const width = horizontal
+    ? imageWidth() * 0.7
+    : orientation
+    ? imageWidth() - 2 * normalize(30)
```
SoleCincis, line 62: it does not behave as expected, needs a fix!

SoleCincis:
Current status for card item : 
IMG_0304
IMG_0305 
Weird!

SoleCincis:
invalid hooks call in stylesWithProps: Simulator Screen Shot - iPhone 11 Pro Max - 2020-10-09 at 13 14 30

donni106 (approved):

SoleCincis (commented):
Thanks for your corrections Daniel !

PR #111, opened 2020-10-22
Wrapper on portrait mode for bigger screens

donni106:
feat: wrapper on portrait mode

    - implemented a logic for big screens on portrait mode
      - based on maximum dimensions.width of mobile on portrait mode, according to: http://www.javascriptkit.com/dhtmltutors/cssmediaqueries2.shtml
    - applied new dimension prop across the app
    - in card list item when portrait big screen width should behave like on landscape

    SVA-92

chore: add todos for wrapper on portrait mode

    - added todos for situations, that needs to be updated

    SVA-92

refactor(home carousel): rename Carousel to HomeCarousel

    - renamed carousel component used only on home screen to avoid
      naming conflicts with `Carousel` from react-native-snap-carousel

refactor(image sizes): adjust widths for different orientations

    - added detailed calculations for image/screen widths for several
      use cases (orientation/dimension)
    - chosen 414 for max full width, because this is the max screen size for
      default iOS and Android devices
      - https://screensiz.es
    - added `memo` for some components to avoid unnecessary re-renderings
      - https://dmitripavlutin.com/use-react-memo-wisely

    SVA-92

feat(orientation): adjust drawer for several cases

    - updated drawer component to have the header with close icon
      always visible
      - before it was scroll along with the entries
    - added calculations for the header of the drawer based on
      device platform and orientation

    SVA-92


src/components/CardListItem.js
```diff
@@ -62,10 +62,9 @@ const styles = StyleSheet.create({
 /* this works properly, we do not want that warning */
 const stylesWithProps = ({ horizontal, orientation, dimensions }) => {
   // image width should be only 70% when rendering horizontal cards, otherwise substract paddings
-
   const width = horizontal
     ? imageWidth() * 0.7
-    : orientation === 'landscape'
+    : orientation === 'landscape' || (orientation === 'portrait' && dimensions.width > 450)
```
donni106, line 67: we need that check on multiple places in the code and do not want to repeat that hardcoded 450.
i will think of a solution to have it somewhere placed generally.

donni106, line 67: we do not need to check for `orientation === 'portrait'`, because we do not care. if the device is greater than 450 in width, than we want to apply the wrapper.

src/components/CardList.js
```diff
@@ -34,30 +33,29 @@ export const CardList = ({ data, navigation, horizontal, refreshControl }) => {
...
+      )}
+      ListFooterComponent={
+        data.length > 10 &&
+        !listEndReached && (
+          <ActivityIndicator color={colors.accent} style={{ margin: normalize(14) }} />
+        )
+      }
+      onEndReached={() => setListEndReached(true)}
+      refreshControl={refreshControl}
+      style={stylesWithProps({ orientation, dimensions }).wrapper}
```
donni106, line 56: the wrapper component was making problems here, so I added 15% padding style here directly, like centering

SoleCincis (approved):

PR #112, opened 2020-10-23
feat(service boxes): add sizing and positioning based on orientation

donni106:
- removed `align-items: center` to always be full width inside the flex area
- added conditions for landscape mode and based on dimensions
  with re-rendering on orientation change

Bildschirmfoto 2020-10-23 um 16 57 33

### DIFF
--- a/src/components/ServiceBox.js
+++ b/src/components/ServiceBox.js
@@ -1,9 +1,25 @@
-import styled from 'styled-components/native';
+import styled, { css } from 'styled-components/native';
 
-import { normalize } from '../config';
+import { consts, normalize } from '../config';
 
 export const ServiceBox = styled.View`
-  align-items: center;
-  flex-basis: ${100 / 3}%;
+  flex-basis: ${100 / 3.3}%;
   margin: ${normalize(14)}px 0;
+
+  ${(props) =>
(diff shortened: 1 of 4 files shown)


SoleCincis (approved):

src/components/ServiceBox.js
```diff
@@ -1,9 +1,25 @@
...
+
+  ${(props) =>
+    props.dimensions.width > consts.DIMENSIONS.FULL_SCREEN_MAX_WIDTH &&
+    css`
+      /*
+        need to add slightly more than 5.3 like on landscape in order to re-render
+        on orientation change. with the same values in this condition and for landscape,
+        it seems that the styled component is memoized and a re-rendering does not take place.
+      */
+      flex-basis: ${100 / 5.301}%;
```
SoleCincis, line 17: actually I don't understand this numbers , what does it mean ${100 / 5.301}% ?

donni106: 100/3 means 33,3333% width for a service box
100/5 means 20% parts of full width

i added 0.3 in both cases in order to use a bit less space to have more spacing between service boxes.

the 5.301 is explained in the comment. 5.3 could not be used in both places, so one of them needed to be different.
so i have chosen a value, that is nearly the same but different

PR #113, opened 2020-10-23
Upgrade Expo SDK to 39

donni106:
chore: upgrade Expo SDK to 39

    - followed instructions here:
      https://dev.to/expo/expo-sdk-39-is-now-available-1lm8

    SVA-100

feat(keyboard in form): add keyboard avoiding view

    - added `KeyboardAvoidingView` in order to use the keyboard
      on iOS without any problems
      - without the view, the keyboard is overlaying elements in the bottom
        area of the screen
    - added `keyboardHidesTabBar` for Android, because a tab bar should
      not be visible above open keyboards
    - exported methods used in `CustomDrawerContentComponent` to use
      it as well in `FormScreen`

    SVA-100

fix(detail screen): add title in details for tour and point of interest

    - there is a title missing in details, because for that type
      of data it is `name` and not `title`
    - added manually a `title` for details used as backup data
      when offline for example


SoleCincis (approved):

src/screens/FormScreen.js
```diff
@@ -59,60 +68,69 @@ export const FormScreen = () => {
...
-              </View>
-            )}
-          </Mutation>
-        </WrapperWithOrientation>
-      </ScrollView>
+      <KeyboardAvoidingView
+        enabled={device.platform === 'ios'}
+        behavior={device.platform === 'ios' && 'padding'}
+        keyboardVerticalOffset={
+          device.platform === 'ios' && getHeaderHeight(orientation) + statusBarHeight(orientation)
```
SoleCincis, line 75: How and why getHeaderHeight and statusBarHeight can effect the Keyboard ?

donni106: good question. i figured out, that the `KeyboardAvoidingView` has this vertical offset. the [default value is 0](https://reactnative.dev/docs/keyboardavoidingview/#keyboardverticaloffset).
our header affects the available space for the keyboard. you can see a wrong offset with a white area like padding above the keyboard.
so we need the exact size of header and tell it the `KeyboardAvoidingView`. we get the size with the new helper methods.
for iOS we have the special case, that the header is different in portrait and landscape. on landscape, it is smaller. so we need this additional case to be handled, which happens here with all the informations of device and orientation.

SoleCincis: brilliant !

PR #117, opened 2020-10-30
feat(event filter): add category filter on index screen

donni106:
- updated `ListHeader` to be more generic
  - now different data can be processed, not only `dataProviders`
  - added logics based on query
- updated event record queries
- added `GET_EVENT_RECORDS_AND_CATEGORIES` to event record
  queries to request categories together with event records
  - query is used on initial query in `IndexScreen` for event records
- placed TODOs that are necessary after main server PR gets merged:
  https://github.com/ikuseiGmbH/smart-village-app-mainserver/pull/159
- removed get-method for list header component, as there is no condition,
  because the conditional stuff happens inside of the component

SVA-58

Simulator Screen Shot - iPhone 11 Pro Max - 2020-10-30 at 17 09 38

### DIFF
--- a/src/components/ListHeader.js
+++ b/src/components/ListHeader.js
@@ -2,26 +2,57 @@ import PropTypes from 'prop-types';
 import React, { useEffect, useRef, useState } from 'react';
 
 import { texts } from '../config';
+import { QUERY_TYPES } from '../queries';
 import { DropdownSelect } from './DropdownSelect';
 import { Wrapper } from './Wrapper';
 
-export const ListHeader = ({ queryVariables, data, updateListData }) => {
-  const dataProviders =
-    data &&
-    data.dataProviders &&
-    data.dataProviders.map((dataProvider, index) => ({
-      id: index + 1,
-      value: dataProvider.name,
-      selected: dataProvider.name === queryVariables.dataProvider
-    }));
+export const ListHeader = ({ query, queryVariables, data, updateListData }) => {
+  const dropdownLabel = {
+    [QUERY_TYPES.EVENT_RECORDS]: texts.categoryFilter.category,
+    [QUERY_TYPES.NEWS_ITEMS]: texts.categoryFilter.dataProvider
+  }[query];
+
+  const dropdownInitialData = (query) => {
+    switch (query) {
+    case QUERY_TYPES.EVENT_RECORDS:
+      return (
+        data &&
+          data.categories &&
+          data.categories
+            .filter((category) => !!category.eventRecordsCount) // TODO: pre-filter results on server?
+            .map((category, index) => ({
+              id: index + 1,
+              value: category.name,
+              selected: category.name === queryVariables.category
+            }))
+      );
+    case QUERY_TYPES.NEWS_ITEMS:
+      return (
+        data &&
+          data.dataProviders &&
+          data.dataProviders.map((dataProvider, index) => ({
+            id: index + 1,
+            value: dataProvider.name,
+            selected: dataProvider.name === queryVariables.dataProvider
+          }))
+      );
+    }
+  };
+
+  // check if there is something set in the certain `queryVariables`
+  // if not, - Alle - will be selected in the `dropdownData`
+  const selectedInitial = {
+    [QUERY_TYPES.EVENT_RECORDS]: !queryVariables.category,
+    [QUERY_TYPES.NEWS_ITEMS]: !queryVariables.dataProvider
+  }[query];
 
   const [dropdownData, setDropdownData] = useState([
     {
       id: 0,
       value: '- Alle -',
-      selected: !queryVariables.dataProvider
+      selected: selectedInitial
     },
-    ...dataProviders
+    ...(dropdownInitialData(query) || [])
   ]);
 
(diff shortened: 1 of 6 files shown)


src/screens/IndexScreen.js
```diff
@@ -209,27 +200,39 @@ export const IndexScreen = ({ navigation }) => {
   const fetchPolicy = graphqlFetchPolicy({ isConnected, isMainserverUp });
   const globalSettings = useContext(GlobalSettingsContext);
   const { filter = {} } = globalSettings;
-  const { news: showNewsFilter = false } = filter;
+  const { news: showNewsFilter = false, events: showEventsFilter = true } = filter;
```
SoleCincis, line 203: I don't get this code, what I get is  : 
filter is originally a empty object in globalSettings (line 202) 
filter can be news which is showNewsFilter = false or can be events  which is showEventsFilter = true .

src/queries/index.js
```diff
@@ -47,6 +53,7 @@ export const getQuery = (query, filterOptions = {}) => {
 export const getFetchMoreQuery = (query, filterOptions = {}) => {
   const FETCH_MORE_QUERIES = {
     [QUERY_TYPES.EVENT_RECORDS]: GET_EVENT_RECORDS,
+    // TODO: check if GET_NEWS_ITEMS would be enough here, as the query gets a dataProvider param or not, but the rest is the same (like for event records)
     [QUERY_TYPES.NEWS_ITEMS]: filterOptions.showNewsFilter
       ? GET_FILTERED_NEWS_ITEMS
       : GET_NEWS_ITEMS
```
SoleCincis, line 59: getFetchMoreQuery I thought is for "scroll up to refresh" .. but it seams is not for that. So why are you fetching more data ?

src/components/ListHeader.js
```diff
@@ -43,16 +74,13 @@ export const ListHeader = ({ queryVariables, data, updateListData }) => {
 
   return (
     <Wrapper>
-      <DropdownSelect
-        data={dropdownData}
-        setData={setDropdownData}
-        label={texts.categoryFilter.label}
-      />
+      <DropdownSelect data={dropdownData} setData={setDropdownData} label={dropdownLabel} />
```
SoleCincis, line 77: dropdownData is dropdownInitialData, isn't it ? 
Here I expected data={dropdownInitialData}

src/screens/IndexScreen.js
donni106: `globalSettings` comes from the server.
this will have a key `filter`. if not, `{}` is used as fallback.
in `filter` there should be `news` and `events`. if not, `false` is used as fallback for `news`, which will be renamed for further use to `showNewsFilter`. `true` is used as fallback for `events`, which will be renamed to `showEventsFilter`.

we use this javascript destructing of values in many places in our app. you need to get comfortable with it. maybe read about it in italian? https://developer.mozilla.org/it/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment

src/queries/index.js
donni106: "pull to refresh" is when you are on top of the list and pull the screen down. then the query for refresh is triggered.
"fetch more" is used when you scroll down through the list. always on reaching the end of the list more data will be fetched, i think 15 in our case. the number is set in the query variables.

src/components/ListHeader.js
donni106: `dropdownData` is the state variable we use for the select. `dropdownInitialData` is a method we use to calculate the dropdown data when the component gets rendered. maybe the naming can be optimized somehow. any suggestions?

src/screens/IndexScreen.js
```diff
@@ -265,8 +268,8 @@ export const IndexScreen = ({ navigation }) => {
 
         let ListHeaderComponent = null;
 
-        if (showNewsFilter) {
-          ListHeaderComponent = getListHeaderComponent(query, queryVariables, data, updateListData);
+        if (showFilter) {
+          ListHeaderComponent = <ListHeader {...{ query, queryVariables, data, updateListData }} />;
         }
```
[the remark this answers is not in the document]
donni106: it is defined on destruction. i explained it in the other comment https://github.com/ikuseiGmbH/smart-village-app-app/pull/117#discussion_r518766305

src/components/ListHeader.js
SoleCincis: the dropdown data when the component gets rendered should not go on our stat variable directly ?

donni106: it does: `dropdownInitialData(query)` is called in the `useState` for `dropdownData` to have it set directly.

donni106:
before merge the server needs to be checked to work correctly with the queries from the app.
there was a rollbar error message connected when testing locally: https://rollbar.com/SmartVillage/SmartVillage-Server/items/76

donni106:
> before merge the server needs to be checked to work correctly with the queries from the app.
> there was a rollbar error message connected when testing locally: https://rollbar.com/SmartVillage/SmartVillage-Server/items/76
> 
> also the dropdown filters needs to be checked when selecting "- Alle -" to perform an action.

✔️ done!

PR #124, opened 2020-11-24
initial implementation of push notifications

digorath:

The entry point of push notifications in the app is within `HomeScreen.js` in form of the `usePushNotifications` hook.
When the app is opened on a new device for the first time, an `Alert` should show, notifying the user that the app provides push notifications and gives them the option to opt in or opt out. On opt-in the system permission popup should show up.

Furthermore the app should notice system permission changes upon reentering. (not visible to user)

The Bad Belzig backend is set up for notifications and you can see if your device registers here: https://server.bad-belzig.smart-village.app/notification/devices

you can "toggle" the in app permissions by calling `setInAppPermission` with `true` or `false`.

This PR includes the introduction of TypeScript into the project.


src/pushNotifications/PermissionHandling.ts
```diff
@@ -0,0 +1,108 @@
...
+  inAppPermission && updatePushToken();
+
+  // this will only show the alert if inAppPermission is undefined (or null), but not if it is false
+  inAppPermission ?? showInitialPushAlert();
+};
+
+const registerForPushNotificationsAsync = async (): Promise<string | undefined> => {
+  const { data: token } = await Notifications.getExpoPushTokenAsync();
+
+  if (Platform.OS === 'android') {
```
donni106, line 44: we are using a `device` object from `configs` with `device.platform` instead of `Platform` directly, to avoid the additional `Platform` import in components.
as you already import `texts` from `../configs`, you can add `device` there and use it here.

src/pushNotifications/PermissionHandling.ts
```diff
@@ -0,0 +1,108 @@
...
+      lightColor: '#FF231F7C'
+    });
+  }
+
+  return token;
+};
+
+const handleSystemPermissions = async (): Promise<boolean> => {
+
+  if (!Constants.isDevice) return false;
```
donni106, line 58: should we add a comment here about the need of early return?

src/pushNotifications/TokenHandling.ts
```diff
@@ -0,0 +1,75 @@
...
+    })
+  };
+
+  if (accessToken) fetch(requestPath, fetchObj);
+};
+
+const addTokenToServer = async (token: string) => {
+  const accessToken = await SecureStore.getItemAsync('ACCESS_TOKEN');
+  const requestPath = secrets[namespace].serverUrl + secrets[namespace].rest.pushDevicesRegister;
+  const os = Platform.OS === 'ios' || Platform.OS === 'android' ? Platform.OS : 'undefined';
```
donni106, line 49: isn't a `Platform.OS` enough here? maybe `toString()` to have the correct type even if it is `undefined`?
`device.platform` should be used here, as pointed out [here](https://github.com/ikuseiGmbH/smart-village-app-app/pull/124/files#r529742692).

other possibility would be:
```
Platform.select({
  android: 'android',
  ios: 'ios',
  default: 'undefined'
})
```

src/queries/types.js
```diff
@@ -1,3 +1,5 @@
+import * as _ from 'lodash';
```
donni106, line 1: it is more lightweight to import specific methods from lodash instead of everything here.

```
import _camelCase from 'lodash/camelCase';
```

we are doing this already elsewhere in the codebase.
i can get you a link if you want to read more about that?!

src/queries/types.js
```diff
@@ -13,3 +15,17 @@ export const QUERY_TYPES = {
...
   TOURS: 'tours'
 };
+
+/**
+ * Parse a query type from json string if one matches.
+ * @argument {string} input
+ * @return {string | undefined}
+ */
+export const getQueryType = (input) => {
+  const camelCaseInput = _.camelCase(input);
```
donni106, line 25: `_camelCase` here if you change the import

src/queries/types.js
```diff
@@ -13,3 +15,17 @@ export const QUERY_TYPES = {
   TOUR: 'tour',
   TOURS: 'tours'
 };
+
+/**
+ * Parse a query type from json string if one matches.
+ * @argument {string} input
```
donni106, line 21: we use `@param` instead of its synonym `@argument`. we should keep it uniform.

src/screens/DetailScreen.js
```diff
@@ -159,7 +159,7 @@ DetailScreen.navigationOptions = ({ navigation, navigationOptions }) => {
     ),
     headerRight: (
       <WrapperRow>
-        {shareContent && (
+        {!!shareContent && (
```
donni106, line 162: this was necessary because `shareContent` should be a text, that we do not want to be rendered here, but pass to the `openShare` method. it `shareContent` could happen that React Native wants to render it here because it is a string, which would result in an error on iOS as it is not wrapped in a Text component.

would be nice to have that change pointed out in the commit message.

src/screens/HomeScreen.js
```diff
@@ -63,6 +64,27 @@ export const HomeScreen = ({ navigation }) => {
...
   const [refreshing, setRefreshing] = useState(false);
 
+  const interactionHandler = useCallback((response) => {
+    const data = response?.notification?.request?.content?.data;
+    const queryType = data?.query_type 
+      ? getQueryType(data.query_type) 
+      : undefined;
+
+    if (data?.id 
+      && queryType
```
donni106, line 74: this looks weirdly formatted, should be `if (data?.id && queryType)` in one line, right? is eslint not complaining?

src/pushNotifications/TokenHandling.ts
```diff
@@ -0,0 +1,75 @@
...
+  if (token) {
+    return SecureStore.setItemAsync(PushNotificationStorageKeys.PUSH_TOKEN, token);
+  } else {
+    return SecureStore.deleteItemAsync(PushNotificationStorageKeys.PUSH_TOKEN);
+  }
+};
+
+const getTokenFromStorage = async () => {
+  return SecureStore.getItemAsync(PushNotificationStorageKeys.PUSH_TOKEN);
+};
```
donni106, line 75: this could be one lined to

```
const getTokenFromStorage = async () => SecureStore.getItemAsync(PushNotificationStorageKeys.PUSH_TOKEN);
```

src/pushNotifications/TokenHandling.ts
```diff
@@ -0,0 +1,75 @@
...
+    })
+  };
+  if (accessToken) fetch(requestPath, fetchObj);
+};
+
+const storeTokenSecurely = async (token?: string) => {
+  if (token) {
+    return SecureStore.setItemAsync(PushNotificationStorageKeys.PUSH_TOKEN, token);
+  } else {
+    return SecureStore.deleteItemAsync(PushNotificationStorageKeys.PUSH_TOKEN);
```
donni106, line 69: a return value from `storeTokenSecurely` is not used anywhere. the `return`s can be remoevd here, right? and is an await missing, as the method is created with async?

src/pushNotifications/TokenHandling.ts
```diff
@@ -0,0 +1,75 @@
...
+const namespace = appJson.expo.slug as keyof typeof secrets;
+
+export enum PushNotificationStorageKeys {
+    PUSH_TOKEN = 'PUSH_TOKEN',
+    IN_APP_PERMISSION = 'IN_APP_PERMISSION',
+}
+
+export const handleIncomingToken = async (token?: string) => {
+  // eslint-disable-next-line no-console
+  console.log(token); // FIXME remove for production
```
donni106, line 17: i would suggest to remove the console log already before merge

src/pushNotifications/TokenHandling.ts
```diff
@@ -0,0 +1,75 @@
...
+
+const namespace = appJson.expo.slug as keyof typeof secrets;
+
+export enum PushNotificationStorageKeys {
+    PUSH_TOKEN = 'PUSH_TOKEN',
+    IN_APP_PERMISSION = 'IN_APP_PERMISSION',
+}
+
+export const handleIncomingToken = async (token?: string) => {
+  // eslint-disable-next-line no-console
```
donni106, line 16: it should not be disabled, as it is mentioned to be removed later anyways.
eslint complaining about it helps to not forget to remove it.

src/pushNotifications/TokenHandling.ts
```diff
@@ -0,0 +1,75 @@
...
+import { secrets } from '../config';
+import * as SecureStore from 'expo-secure-store';
+import * as appJson from '../../app.json';
+
+
+const namespace = appJson.expo.slug as keyof typeof secrets;
+
+export enum PushNotificationStorageKeys {
+    PUSH_TOKEN = 'PUSH_TOKEN',
+    IN_APP_PERMISSION = 'IN_APP_PERMISSION',
```
donni106, line 12: indentation should be always 2 spaces

src/pushNotifications/TokenHandling.ts
```diff
@@ -0,0 +1,75 @@
+import { Platform } from 'react-native';
+
+import { secrets } from '../config';
+import * as SecureStore from 'expo-secure-store';
+import * as appJson from '../../app.json';
+
+
```
donni106, line 7: please avoid unnecessary line breaks. eslint should automatically fix this.

src/pushNotifications/TokenHandling.ts
```diff
@@ -0,0 +1,75 @@
...
+const namespace = appJson.expo.slug as keyof typeof secrets;
```
donni106, line 8: you can import `namespace` directly from `configs`. or does that not work because of needing types?

src/pushNotifications/TokenHandling.ts
```diff
@@ -0,0 +1,75 @@
+import { Platform } from 'react-native';
+
+import { secrets } from '../config';
+import * as SecureStore from 'expo-secure-store';
+import * as appJson from '../../app.json';
```
donni106, line 5: please order imports in the following way:

```
import { Platform } from 'react-native';
import * as SecureStore from 'expo-secure-store';

import { secrets } from '../config';
import * as appJson from '../../app.json';
```

as we import from libraries first and local stuff second, divided by a line break, all over the codebase to maintain fast readability.

src/pushNotifications/PermissionHandling.ts
```diff
@@ -0,0 +1,108 @@
...
+    greetingBody,
+    [
+      {
+        text: decline,
+        onPress: () => setInAppPermission(false)
+      },
+      {
+        text: approve,
+        onPress: () => setInAppPermission(true),
+        style: 'cancel'
```
donni106, line 103: on iOS the cancel style is used for negative alert actions, so it should be used for `decline` here.
https://reactnative.dev/docs/alert

src/pushNotifications/PermissionHandling.ts
```diff
@@ -0,0 +1,108 @@
...
+    permissionMissingTitle,
+    permissionMissingBody,
+    undefined,
+    { cancelable: false }
+  );
+};
+
+const showInitialPushAlert = (): void => {
+  const { greetingBody, greetingTitle, approve, decline } = texts.pushNotifications;
+  Alert.alert(
```
donni106, line 92: we have the rule to always have a line break between `const`s and other logics. i think you too, but forgot it here?

src/pushNotifications/PermissionHandling.ts
```diff
@@ -0,0 +1,108 @@
...
+  }
+
+  return finalStatus === PermissionStatus.GRANTED;
+};
+
+export const updatePushToken = async () => {
+  await handleSystemPermissions()
+    .then((hasPermission) => {
+      if (hasPermission) return registerForPushNotificationsAsync();
+    })
```
donni106, line 75: when using `if` as one liner already, than you could consider one step further and one line everything with `&&`, right?

```
.then((hasPermission) => hasPermission && registerForPushNotificationsAsync())
```

src/pushNotifications/PermissionHandling.ts
```diff
@@ -0,0 +1,108 @@
...
+const handleSystemPermissions = async (): Promise<boolean> => {
+
+  if (!Constants.isDevice) return false;
+
+  const { status: existingStatus } = await Permissions.getAsync(Permissions.NOTIFICATIONS);
+
+  let finalStatus = existingStatus;
+  if (existingStatus === PermissionStatus.UNDETERMINED) {
+    const { status } = await Permissions.askAsync(Permissions.NOTIFICATIONS);
+    finalStatus = status;
```
donni106, line 65: maybe `askedStatus` makes it more clear, whats happening here?

const { status: askedStatus } = await Permissions.askAsync(Permissions.NOTIFICATIONS);
finalStatus = askedStatus;

src/pushNotifications/PermissionHandling.ts
```diff
@@ -0,0 +1,108 @@
...
+
+const registerForPushNotificationsAsync = async (): Promise<string | undefined> => {
+  const { data: token } = await Notifications.getExpoPushTokenAsync();
+
+  if (Platform.OS === 'android') {
+    Notifications.setNotificationChannelAsync('default', {
+      name: 'default',
+      importance: Notifications.AndroidImportance.MAX, // FIXME: check for reasonable value
+      vibrationPattern: [0, 250, 250, 250],
+      lightColor: '#FF231F7C'
```
donni106, line 49: usage of color needs to be checked for android.
anyways the color string should go in `colors`.

src/index.js
```diff
@@ -236,13 +237,16 @@ const MainAppWithApolloProvider = () => {
...
-
-    // set orientation to "default", to allow both portrait and landscape
-    ScreenOrientation.lockAsync(ScreenOrientation.OrientationLock.DEFAULT);
+    if(!loading) SplashScreen.hideAsync().then(
+      (result) => {
+        setIsSplashScreenVisible(!result);
+        // set orientation to "default", to allow both portrait and landscape
+        ScreenOrientation.lockAsync(ScreenOrientation.OrientationLock.DEFAULT);
+      }
+    );
```
donni106, line 246: is eslint not complaining here?
i think when using `if` with multiple lines and method chaining, the readability gets heavy. i would prefer:

```
if (!loading) {
  SplashScreen.hideAsync().then(
    (result) => {
      setIsSplashScreenVisible(!result);
      // set orientation to "default", to allow both portrait and landscape
      ScreenOrientation.lockAsync(ScreenOrientation.OrientationLock.DEFAULT);
    }
  );
}
```

or

```
!loading && SplashScreen.hideAsync().then(
  (result) => {
    setIsSplashScreenVisible(!result);
    // set orientation to "default", to allow both portrait and landscape
    ScreenOrientation.lockAsync(ScreenOrientation.OrientationLock.DEFAULT);
  }
);
```

what do you think?

donni106: why was this change made? would be nice to read about it in the corresponding commit message.

src/hooks/PushNotification.ts
```diff
@@ -0,0 +1,67 @@
...
+  }, []); // empty dependencies because it will only used once in the "mountEffect" below
+
+  useEffect(() => {
+    initializePushPermissions();
+
+    AppState.addEventListener('change', onGetActive);
+
+    // This listener is fired whenever a notification is received while the app is foregrounded
+    notificationListener.current
+      = notificationHandler && Notifications.addNotificationReceivedListener((notification) => {
```
donni106, line 39: haven't seen the `=` in the beginning of a line yet.
does eslint not complain about this?
i think this file needs to be re-formatted in a few places.

src/config/texts.js
```diff
@@ -41,6 +41,14 @@ export const texts = {
     operatingCompany: 'Anbieter',
     prices: 'Preise'
   },
+  pushNotifications: {
+    approve: 'Jetzt einschalten',
+    decline: 'Vielleicht später',
+    greetingBody: 'Damit dich wichtige Mitteilungen erreichen, würden wir dir gerne Benachrichtigungen schicken können. Du kannst diese Einstellung jederzeit in deinem persönlichen Bereich ändern.',
```
donni106, line 47: do you copied this from another app or defined it by yourself?
i always try to avoid "du" or "sie" with more generic sentences. maybe we need to consult @pwilimzig to find the right solution that works for all customers.

App.js
```diff
@@ -12,7 +12,7 @@ const AppWithFonts = () => {
   const { trackAppStart } = useMatomo();
 
   trackAppStart();
-
+
```
donni106, line 15: these spaces should be reverted

.eslintrc.yml
```diff
@@ -55,6 +58,8 @@ rules:
     - error
   react-native/no-color-literals:
     - error
+  '@typescript-eslint/no-unused-vars': 
+    - 2
```
donni106, line 62: why were these changes to `np-unused-vars` and `@typescript-eslint/no-unused-vars` added and why are they different?

package.json
```diff
@@ -4,7 +4,7 @@
     "android": "expo start --android",
     "eject": "expo eject",
     "ios": "expo start --ios",
-    "lint": "./node_modules/.bin/eslint __tests__ src",
+    "lint": "./node_modules/.bin/eslint __tests__ src --ext .js,.jsx,.ts,.tsx",
```
donni106, line 7: we do not have `jsx` and `tsx` files in the codebase. these two can be removed, right?

src/pushNotifications/PermissionHandling.ts
donni106: there are also two occurrences in `PermissionHandling.ts`

src/pushNotifications/TokenHandling.ts
digorath: Using `Platform.select` changes the type that can be inferred by ts from this
Screenshot 2020-11-25 at 09 09 59
to this
Screenshot 2020-11-25 at 09 10 29

This is not a big issue, due to the fact that we just need it to be a `string` to send it in the json though.

The reason this is mapped to just those three values is, because our the backend implementation handles these three cases, and not `"macos" | "windows" | "web"`.

I will change it to use `device.platform` instead of `Platform.OS`, but leave the structure as is.

src/pushNotifications/TokenHandling.ts
digorath: It could be one lined. do you prefer to have it in one long line instead?
I would prefer a format like this.

```ts
const getTokenFromStorage = async () =>
  SecureStore.getItemAsync(PushNotificationStorageKeys.PUSH_TOKEN);
```

src/queries/types.js
digorath: done in 5c4968a26c167de3f0343c155960812f53460942

src/queries/types.js
digorath: done in 5c4968a26c167de3f0343c155960812f53460942

src/pushNotifications/TokenHandling.ts
digorath: done in 5c4968a26c167de3f0343c155960812f53460942

eslint did not catch this

src/pushNotifications/TokenHandling.ts
digorath: done in 5c4968a26c167de3f0343c155960812f53460942

src/pushNotifications/PermissionHandling.ts
digorath: done in e2f9b849f1157cec42932f4fc81cf4ce13f20fc8

src/screens/HomeScreen.js
digorath: I think this is an artefact from having more conditions there. I changed it to be in one line in e2f9b849f1157cec42932f4fc81cf4ce13f20fc8

src/pushNotifications/TokenHandling.ts
digorath: removed in e2f9b849f1157cec42932f4fc81cf4ce13f20fc8

src/pushNotifications/TokenHandling.ts
digorath: ok. :)
it is removed now anyways

src/pushNotifications/TokenHandling.ts
digorath: didn't see eslint autofix did not catch this.
it is changed in e2f9b849f1157cec42932f4fc81cf4ce13f20fc8

src/pushNotifications/TokenHandling.ts
digorath: exactly.
if we import it from configs it is just a `string` value, which causes problems when accessing the secrets object with it as a key.

should we change the namespace file to be a ts file now, or keep it as is and do it sometime in the future?

src/pushNotifications/PermissionHandling.ts
digorath: done in e2f9b849f1157cec42932f4fc81cf4ce13f20fc8

src/pushNotifications/PermissionHandling.ts
digorath: done in e2f9b849f1157cec42932f4fc81cf4ce13f20fc8

src/index.js
digorath: changed to the second proposed alternative in e2f9b849f1157cec42932f4fc81cf4ce13f20fc8

I changed it because we unlocked the orientation by setting it to default while the splash screen was still visible.
When I tried it out and changed the orientation to landscape by holding my device horizontally, while the app was still in the splash screen, the orientation changed, while the splash screen kept its format, causing half of the screen to be visible behind the splash screen.

src/hooks/PushNotification.ts
digorath: no eslint does not complain here. same with any other operator. you can have

```ts
conditionA &&
conditionB
```
as well as 
```ts
conditionA
&& conditionB
```

I changed it to have the assignment operator in the end of the previous line in e2f9b849f1157cec42932f4fc81cf4ce13f20fc8

App.js
digorath: done in e2f9b849f1157cec42932f4fc81cf4ce13f20fc8

src/index.js
donni106: nice, thanks

.eslintrc.yml
digorath: the normal `no-unused-vars` rule must be disabled because it conflicts with the newly added one from typescript.

the typescript one is needed over the normal one because otherwise eslint will complain about unused variables in type definitions of functions

```ts
type NotificationHandler = (arg: Notifications.Notification) => void;
type ResponseHandler = (arg: Notifications.NotificationResponse) => void;
```

package.json
digorath: tbh I did not realise we have not `.jsx` files. 🤣 
I can remove that one, if you want.

We will have `.tsx` files in the project. Everytime we have a functional component it needs to be in a `.tsx` file and not a plain `.ts` file.

src/pushNotifications/PermissionHandling.ts
```diff
@@ -0,0 +1,112 @@
+import * as Permissions from 'expo-permissions';
+import * as Notifications from 'expo-notifications';
+import Constants from 'expo-constants';
+import { Alert } from 'react-native';
+import { addToStore, readFromStore } from '../helpers';
+import { PermissionStatus } from 'expo-permissions';
+import { device, texts } from '../config';
+import { handleIncomingToken, PushNotificationStorageKeys } from './TokenHandling';
```
donni106, line 8: here i found another case, where imports are mixed and not libraries first and local stuff second, divided by a line break.
what do you think about grouping all expo imports and pulling `from 'react-native` to the top?
in most cases we import stuff from prop types first, react second, react native third. do you think this is worth it?

```
import { Alert } from 'react-native';
import * as Permissions from 'expo-permissions';
import * as Notifications from 'expo-notifications';
import Constants from 'expo-constants';
import { PermissionStatus } from 'expo-permissions';

import { addToStore, readFromStore } from '../helpers';
import { device, texts } from '../config';
import { handleIncomingToken, PushNotificationStorageKeys } from './TokenHandling';
```

src/pushNotifications/TokenHandling.ts
donni106: yes, with line break, because of max line length of 100 characters.
what about the async await i questioned?

src/pushNotifications/TokenHandling.ts
donni106: sad. i will try things out regarding this.

src/pushNotifications/PermissionHandling.ts
digorath: updated it in 94bd1238af679bc33b72c53861317369ad81796d

we can do it if you prefer it that way. or are there any eslint rules/plugins that can sort everything neatly for us?

src/pushNotifications/TokenHandling.ts
donni106: all right. lets do it next time we need namespace in ts.

src/hooks/PushNotification.ts
donni106: thanks. i will have a look at eslint regarding this.

package.json
donni106: ok, then keep `.jsx` as well for now. i will read about it later.

src/pushNotifications/PermissionHandling.ts
donni106: my experience is, that it helps having same structures across different projects. you can scan files quickly this way.
i do not know about an eslint solution for this, maybe there is one nowadays? research needed.

src/pushNotifications/PermissionHandling.ts
digorath: changed in 1aa5ba1e0ecbb2bb407889e1de6a1ba1b72f696c

src/pushNotifications/TokenHandling.ts
```diff
@@ -0,0 +1,71 @@
...
+
+const namespace = appJson.expo.slug as keyof typeof secrets;
+
+export enum PushNotificationStorageKeys {
+  PUSH_TOKEN = 'PUSH_TOKEN',
+  IN_APP_PERMISSION = 'IN_APP_PERMISSION',
+}
+
+export const handleIncomingToken = async (token?: string) => {
```
donni106, line 13: can you please add a comment for describing the behavior here for our future selfs.

src/pushNotifications/PermissionHandling.ts
donni106: lets use `lightColor: colors.primary`

src/pushNotifications/PermissionHandling.ts
```diff
@@ -0,0 +1,112 @@
...
+  inAppPermission ?? showInitialPushAlert();
+};
+
+const registerForPushNotificationsAsync = async (): Promise<string | undefined> => {
+  const { data: token } = await Notifications.getExpoPushTokenAsync();
+
+  if (device.platform === 'android') {
+    Notifications.setNotificationChannelAsync('default', {
+      name: 'default',
+      importance: Notifications.AndroidImportance.MAX, // FIXME: check for reasonable value
```
donni106, line 48: lets use `DEFAULT` here, until we know, what this does?

```
export enum AndroidImportance {
  UNKNOWN,
  UNSPECIFIED,
  NONE,
  MIN,
  LOW,
  DEFAULT,
  HIGH,
  MAX,
}
```

https://docs.expo.io/versions/latest/sdk/notifications/#notificationchannel

src/config/texts.js
donni106: we will receive texts soon by him.

src/pushNotifications/PermissionHandling.ts
donni106: opened an issue for that: https://github.com/ikuseiGmbH/smart-village-app-app/issues/128

src/screens/DetailScreen.js
digorath: I will try to be more elaborate in future commits. :)

src/pushNotifications/TokenHandling.ts
digorath: removed return and async in cf1e55b43be95dc9d0326affe8b7ec9614a7b759

src/pushNotifications/PermissionHandling.ts
digorath: I will keep it as is for now

src/pushNotifications/PermissionHandling.ts
digorath: changed in 1673fd6321b2bcd779d45def158dbe766aecc6fc

also changed android importance to default

src/pushNotifications/PermissionHandling.ts
digorath: changed to default

src/pushNotifications/TokenHandling.ts
digorath: simplified and async removed in bafda5bb11fe80a9aaf472b5caec925eba81fa2b

src/config/texts.js
digorath: updated with the preliminary text we received from him

will resolve this for now.

src/pushNotifications/TokenHandling.ts
digorath: added comment in 079ae03f862b1547eecbaa6eb998264f235dac55

donni106 (approved):

donni106:
> ## DeepCode's analysis on [#079ae0](https://www.deepcode.ai/app/gh/ikuseiGmbH/smart-village-app-app/a0abd5beb28798bde1d0fab9282eb830b38ec60a/ikuseiGmbH/smart-village-app-app/079ae03f862b1547eecbaa6eb998264f235dac55/pr/_/%2F/code/?utm_source=gh_review&c=0&w=0&i=1&) found:
> * ℹ️ **1** minor issue. 👇
> 
> ## Top issues
> Description	Example fixes
> No catch method for promise. This may result in an unhandled promise rejection. Occurrences:
> * [TokenHandling.ts:17](https://github.com/ikuseiGmbH/smart-village-app-app/blob/079ae03f862b1547eecbaa6eb998264f235dac55/src/pushNotifications/TokenHandling.ts#L17)
> * [index.js:240](https://github.com/ikuseiGmbH/smart-village-app-app/blob/079ae03f862b1547eecbaa6eb998264f235dac55/src/index.js#L240)
> * [PermissionHandling.ts:23](https://github.com/ikuseiGmbH/smart-village-app-app/blob/079ae03f862b1547eecbaa6eb998264f235dac55/src/pushNotifications/PermissionHandling.ts#L23)
> 
> [🔧 Example fixes](https://www.deepcode.ai/app/gh/ikuseiGmbH/smart-village-app-app/a0abd5beb28798bde1d0fab9282eb830b38ec60a/ikuseiGmbH/smart-village-app-app/079ae03f862b1547eecbaa6eb998264f235dac55/pr/_/%2Fsrc%2FpushNotifications%2FTokenHandling.ts/javascript%2Fdc_interfile_project%2FPromiseNotCaughtGeneral/code/?utm_source=gh_review&)
> #### 👉 View analysis in [**DeepCode’s Dashboard**](https://www.deepcode.ai/app/gh/ikuseiGmbH/smart-village-app-app/a0abd5beb28798bde1d0fab9282eb830b38ec60a/ikuseiGmbH/smart-village-app-app/079ae03f862b1547eecbaa6eb998264f235dac55/pr/_/%2F/code/?utm_source=gh_review&c=0&w=0&i=1&) | [_Configure the bot_](https://www.deepcode.ai/app/gh/?ownerconfig=ikuseiGmbH)

we will not need to fix this. this shouldn't be a problem for now.

src/screens/HomeScreen.js
```diff
@@ -63,6 +64,25 @@ export const HomeScreen = ({ navigation }) => {
...
+    const queryType = data?.query_type 
+      ? getQueryType(data.query_type) 
+      : undefined;
+
+    if (data?.id && queryType) {
+      // navigate to the newsItem
+      navigation.navigate({
+        routeName: 'Detail',
+        params: {
+          query: QUERY_TYPES.NEWS_ITEM,
```
donni106, line 78: @digorath i found this! `query: QUERY_TYPES.NEWS_ITEM` is hardcoded but `queryType` should be used here, right?
`// navigate to the newsItem` was work in progress maybe and not finished? maybe a `// TODO: ...` would be helpful in this case.

furthermore the `params` is missing keys, which can be compared with the new methods in `queryDataParser.js`.
example for news items:

```
params: {
      title: newsItem.titleDetail,
      query: QUERY_TYPES.NEWS_ITEM,
      queryVariables: { id: `${newsItem.id}` },
      rootRouteName: 'NewsItems',
      shareContent: {
        message: shareMessage(newsItem, QUERY_TYPES.NEWS_ITEM)
      },
      details: newsItem
    },
```

the `rootRouteName` can be determined based on the `queryType`.

`shareContent` and `details` will be harder, as we do not have the data for a `newsItem` (or other detail object to open)

PR #125, opened 2020-11-24
combine settings and matomo

donni106:
This PR is a first version of combining Matomo analytics with the settings screen.
When toggling the switch for Matomo an Alert is asking the user and pointing out, that an app restart is necessary for changes to take effect.

For testing locally, the `__DEV__` const needs to be removed or inverted to `!__DEV__` in App.js.

activation alert | deactivation alert
-------------- | --------------
Simulator Screen Shot - iPhone 11 Pro Max - 2020-11-24 at 19 34 50 |  Simulator Screen Shot - iPhone 11 Pro Max - 2020-11-24 at 19 34 43

Also added tracking of navigation to the settings screen.

### DIFF
--- a/src/screens/SettingsScreen.js
+++ b/src/screens/SettingsScreen.js
@@ -98,7 +98,7 @@ export const SettingsScreen = () => {
                     [
                       {
                         text: 'Nein',
-                        onPress: () => revert(),
+                        onPress: revert,
                         style: 'cancel'
                       },
                       { text: 'Ja', onPress: createMatomoUserId }
@@ -114,7 +114,7 @@ export const SettingsScreen = () => {
                     [
                       {
                         text: 'Nein',
-                        onPress: () => revert(),
+                        onPress: revert,
                         style: 'cancel'
                       },
                       { text: 'Ja', onPress: removeMatomoUserId }
--- a/src/screens/SettingsScreen.js
+++ b/src/screens/SettingsScreen.js
@@ -2,7 +2,7 @@ import React, { useContext, useState } from 'react';
 import { RefreshControl, SectionList, StyleSheet, TouchableOpacity, View } from 'react-native';
 
 import { OrientationContext } from '../OrientationProvider';
-import { colors, device, normalize, texts } from '../config';
+import { colors, consts, device, normalize, texts } from '../config';
 import {
   Icon,
   RegularText,
@@ -18,6 +18,9 @@ import { arrowLeft } from '../icons';
 import { QUERY_TYPES } from '../queries';
 import { SettingsContext } from '../SettingsProvider';
 import { storageHelper } from '../helpers';
+import { useMatomoTrackScreenView } from '../hooks';
+
+const { MATOMO_TRACKING } = consts;
 
 const keyExtractor = (item, index) => `index${index}-id${item.id}`;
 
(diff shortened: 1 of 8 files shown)


src/screens/SettingsScreen.js
```diff
@@ -18,11 +25,111 @@ import {
...
+                      {
+                        text: 'Nein',
+                        onPress: () => revert(),
+                        style: 'cancel'
+                      },
+                      { text: 'Ja', onPress: createMatomoUserId }
+                    ],
+                    { cancelable: false }
+                  );
+                }, 300),
```
donni106, line 78: used timeout here to wait for the toggle animation in order to not have visual side effects

src/screens/SettingsScreen.js
```diff
@@ -18,11 +25,111 @@ import {
...
+              value: matomoValue,
+              onActivate: (revert) =>
+                setTimeout(() => {
+                  Alert.alert(
+                    texts.settingsTitles.analytics,
+                    texts.settingsContents.analytics.onActivate,
+                    [
+                      {
+                        text: 'Nein',
+                        onPress: () => revert(),
```
digorath, line 71: this should simply be
```ts
onPress: revert
```
I think

src/screens/SettingsScreen.js
```diff
@@ -18,11 +25,111 @@ import {
...
+          ]
+        },
+        {
+          data: [
+            {
+              title: texts.settingsTitles.analytics,
+              topDivider: true,
+              type: 'toggle',
+              value: matomoValue,
+              onActivate: (revert) =>
```
digorath, line 63: the way onActivate and inRevert are set up, they are not async.

we do however use `await` when we call them inside of `toggleSwitch` in `SettingsListItem.js`

digorath (changes requested):

src/screens/SettingsScreen.js
donni106: true

src/screens/SettingsScreen.js
donni106: you mean `onActivate` and `onDeactivate`? and your advice is to update them here?

digorath: opening an alert does not return a promise.

it is a synchronous function with void as return value.

however, inside of SettingsListItem we call it with `await` which does imply it is async

donni106: when i look at it now, they maybe do not need to be awaited there

donni106: removed `await` in https://github.com/ikuseiGmbH/smart-village-app-app/pull/125/commits/caa3d385adcf707c70b62b5b1e6d15099760a06a

we need to pay attention on merging, because it needs to be done in a later case in `ToggleListItem` as well, as the file got split.

src/screens/SettingsScreen.js
donni106: done in we need to pay attention on merging, because it needs to be done in a later case in `ToggleListItem` as well, as the file got split.

src/screens/SettingsScreen.js
digorath: I don't feel it is necessary. does it behave weirdly on android?

digorath (approved):

src/screens/SettingsScreen.js
donni106: i had the effect, that when toggling per switch click on iOS, it was not as smooth as expected sometimes, because the animation and alert are simultaneously.
on the other hand users wouldn't see the activity indicator. but maybe this is not necessary as the alert comes in.

donni106: removed in https://github.com/ikuseiGmbH/smart-village-app-app/pull/125/commits/662a758d48bc2bf7531d482b634f25ff62fbaeb2

digorath (approved):

Issue #120, opened 2020-11-11
Better known share icon on iOS


donni106:
The current share icon is an Android one. It would be better to have platform specific ones, like mentioned here https://github.com/FortAwesome/Font-Awesome/issues/14957 and discussed there https://github.com/FortAwesome/Font-Awesome/issues/10619

SoleCincis:
Can We close this issue ? :)

donni106:
sure, do it

PR #130, opened 2020-11-25
[WIP] ios share icon

SoleCincis:
- the new helix header colors the icon white
  - so now our 'color' variable in the Icon file has a third conditional color
- the share icon and the drawer icon do not get along probably because they are of 2 different types
  - after several attempts they still remain not perfectly online

Issue #120

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -243,13 +243,12 @@ const MainAppWithApolloProvider = () => {
   }, [initialGlobalSettings]);
 
   useEffect(() => {
-    !loading && SplashScreen.hideAsync().then(
-      (result) => {
-        setIsSplashScreenVisible(!result);
+    !loading &&
+      SplashScreen.hideAsync().then(() => {
+        setIsSplashScreenVisible(false);
         // set orientation to "default", to allow both portrait and landscape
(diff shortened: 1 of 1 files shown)


SoleCincis:
current status of detail screen : 
Screenshot 2020-11-25 at 20 39 14

donni106:
There are five commits that shouldn't be included. Your branch was not based on master I think, but on you previous work on lists.
Please update the branch to have the one necessary commit adding onto master @SoleCincis.

donni106:
> New share icon is not an svg, could this be a reason why the two icons are not inlined ?
> Screenshot 2020-11-25 at 21 02 57Screenshot 2020-11-25 at 21 03 07

I don't think that it is the difference between being svg and font icon. They are two different icons which need to be placed in a certain way to each other to be aligned. It is on you to make them fit.

donni106 (changes requested):
There are five commits that shouldn't be included. Your branch was not based on master I think, but on you previous work on lists.
Please update the branch to have the one necessary commit adding onto master @SoleCincis.

If your PR is WIP, as you put in to the title, it should be marked at draft or at least there need to be some TODOs, which points out, what is missing and will be done to finish. Now it is incomplete.

I will review the code, if everything is in place.

PR #131, opened 2020-11-26
Implement settings functionality for notifications

digorath:

The toggle should now be working properly aside from its wrong initial state.
The token should be added to and removed from the server by toggling.


donni106:
This is based on: https://github.com/ikuseiGmbH/smart-village-app-app/pull/124

src/pushNotifications/TokenHandling.ts
```diff
@@ -38,7 +48,13 @@ const removeTokenFromServer = async (token: string) => {
...
   };
 
-  if (accessToken) fetch(requestPath, fetchObj);
+  if (accessToken) 
+    // 204 means that it was a success on the server
+    // 404 means that the token was already not on the server and can be treated
+    //  as a success
+    return fetch(requestPath, fetchObj)
+      .then((response) => response.status === 204 || response.status === 404);
+  return false;
```
donni106, line 57: please use {} for multiline body in ifs.
the `return false` at the end should also be separated by a line break.
this way now, the readability suffers.

```
if (accessToken) {
  // 204 means that it was a success on the server
  // 404 means that the token was already not on the server and can be treated
  //  as a success
  return fetch(requestPath, fetchObj)
    .then((response) => response.status === 204 || response.status === 404);
}
  
return false;
```

donni106:
we know about @deepcode-ci-bot wish to have `catch` parts, but decided to not add them, because they are not necessary currently. we would need an global error handling to be able to do something with catched errors.

donni106 (approved):

PR #135, opened 2020-11-30
feat(icon): dried out the logic for different devices

SoleCincis:
- for better readability
- if using Platform we need to import an additional object from react native
  - which we don't need if we use just our device
  - this makes even more sense in Radiobutton where device is used multiple times in the file

(!) this change is not tested on android

### DIFF
--- a/src/components/Radiobutton.js
+++ b/src/components/Radiobutton.js
@@ -1,9 +1,9 @@
 import PropTypes from 'prop-types';
 import React from 'react';
-import { Platform, StyleSheet } from 'react-native';
+import { StyleSheet } from 'react-native';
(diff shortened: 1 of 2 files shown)


SoleCincis:
Codeclime issues are "Similar blocks of code" in Radiobutton, for some reasons I can't change it to wontfix

donni106 (approved):

donni106:
> Codeclime issues are "Similar blocks of code" in Radiobutton, for some reasons I can't change it to wontfix

seems that i have mis-changed github access settings when adding @digorath.
i updated it again, can you please check both in an upcoming PR, if you can edit the codeclimate issues.

donni106:
> (!) this change is not tested on android

thx for info. i have checked it now.

PR #137, opened 2020-12-02
Feature: open street maps for POIs

digorath:

Added a web view map for points of interest.

The map is shown on the details screen for a specific POI.

Furthermore for a selected category of POIs there now is the option to show it as a map instead. If selected you see the map with all the locations of the POIs. If you tap on one pin, the details are shown below.

In the process of adding it to the IndexScreen (category screen) the screen was refactored.

SVA-86

---

1. selected a category -> default view is list view
2. switched to map view (with no POI selected)
3. selected a POI
4. selected a POI from list view

|   	|   Android	| iOS  	|
|---	|---	|---	|
|  1 	|   Screenshot 2020-12-02 at 15 57 51	|   Screenshot 2020-12-02 at 16 01 08	|
|   2	|  Screenshot 2020-12-02 at 15 58 12 	|   Screenshot 2020-12-02 at 16 01 18	|
|   3	|   Screenshot 2020-12-02 at 15 58 23	|   Screenshot 2020-12-02 at 16 01 25	|
|   4	|   Screenshot 2020-12-02 at 15 58 47	|   Screenshot 2020-12-02 at 16 01 35	|


src/screens/HomeScreen.js
```diff
@@ -165,7 +165,6 @@ export const HomeScreen = ({ navigation }) => {
         }
       >
         <HomeCarousel navigation={navigation} />
-
```
donni106, line 168: was the line break removal intended?

src/components/map/LocationOverview.tsx
```diff
@@ -0,0 +1,88 @@
...
+            />}
+        </View>
+      </ScrollView>
+    </SafeAreaViewFlex>
+  );
+};
+
+const styles = StyleSheet.create({
+  map: {
+    height: normalize(200),
```
donni106, line 90: why have you chosen 200?
i would say we should make it 16:9 ratio depending on the device width, like images. this will work for the map at the bottom of a POI.

for the location overview the device orientation is not considered yet. therefore a 16:9 would be too big.
i think we should wrap it on landscape as well and then we can use 16:9 ratio also for the overview map.

src/components/screens/PointOfInterest.js
```diff
@@ -154,6 +148,20 @@ export const PointOfInterest = ({ data, navigation }) => {
...
 
+        {!hideMap && !!latitude && !!longitude && (
+          <View>
+            <TitleContainer>
+              <Title accessibilityLabel={`${texts.pointOfInterest.location} (Überschrift)`}>
+                {texts.pointOfInterest.location}
+              </Title>
+            </TitleContainer>
+            <WebViewMap locations={[{ position: { lat: latitude, lng: longitude } }]}
+              style={{ height: normalize(200) }}
```
donni106, line 160: a 16:9 ratio depending on the device would fit better for different device sizes. see my comment other comment on this at `LocationOverview.tsx`

src/components/map/LocationOverview.tsx
```diff
@@ -0,0 +1,88 @@
...
+    marginVertical: normalize(12)
```
donni106, line 91: i will remove the vertical margin when added the new map switch style, because this margin adds a weird looking white space between map and headline container

- [ ] remove marginVertical

src/components/map/LocationOverview.tsx
```diff
@@ -0,0 +1,88 @@
...
+    <SafeAreaViewFlex>
+      <ScrollView>
+        <WebViewMap
+          locations={mapToMapMarkers(overviewData)}
+          onMessageReceived={onMessageReceived}
+          style={styles.map}
+          zoom={10}
+        />
+        <View>
+          {detailsLoading ? <ActivityIndicator />
```
donni106: color is missing for the activity indicator

```
<ActivityIndicator color={colors.accent} />
```

donni106: best would be to have the indicator centered vertically on the white area underneath the map.
can be ignored for now, if it is too much hassle.

donni106 (changes requested):

src/components/MapSwitchHeader.tsx
```diff
@@ -0,0 +1,49 @@
...
+  setShowMap: (newValue: boolean) => void,
+  showMap: boolean
+}
+
+export const MapSwitchHeader = ({ setShowMap, showMap }: Props) => {
+  const onPressShowMap = useCallback(() => setShowMap(true), [setShowMap]);
+  const onPressShowList = useCallback(() => setShowMap(false), [setShowMap]);
+
+  return <View style={styles.container}>
+    <TouchableOpacity onPress={onPressShowList} style={styles.button}>
```
donni106: we have a `Touchable` component that should be used here, to use the correct components depending on the operating system. i will add that as well when adding the new component for styling.

src/components/MapSwitchHeader.tsx
```diff
@@ -0,0 +1,49 @@
...
+import { StyleSheet, TouchableOpacity, View } from 'react-native';
+import { colors, normalize } from '../config';
+import { BoldText } from './Text';
+
+type Props = {
+  setShowMap: (newValue: boolean) => void,
+  showMap: boolean
+}
+
+export const MapSwitchHeader = ({ setShowMap, showMap }: Props) => {
```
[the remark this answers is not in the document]
donni106: Android | iOS
--------|------
Screenshot_20201203-151703 | Simulator Screen Shot - iPhone 11 Pro Max - 2020-12-03 at 15 18 37

src/components/MapSwitchHeader.tsx
```diff
@@ -0,0 +1,49 @@
+import React, { useCallback } from 'react';
+import { StyleSheet, TouchableOpacity, View } from 'react-native';
+import { colors, normalize } from '../config';
```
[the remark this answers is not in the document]
donni106: done with updating the component

src/components/map/LocationOverview.tsx
donni106: updated the `ActivityIndicator` with removing the vertical margin.
centering vertically was not working with two tries.

src/screens/HomeScreen.js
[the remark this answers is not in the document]
digorath: no. I reverted it in 09bc9bb195bef32a5850d686ee55a4b1bfdaf3c9

src/components/screens/PointOfInterest.js
```diff
@@ -179,5 +187,6 @@ export const PointOfInterest = ({ data, navigation }) => {
 
 PointOfInterest.propTypes = {
   data: PropTypes.object.isRequired,
+  hideMap: PropTypes.bool,
```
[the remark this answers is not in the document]
digorath: done in 83a2366b2f98950f73af6a15ed7527a729fa4a93

src/components/map/WebViewMap.tsx
```diff
@@ -0,0 +1,57 @@
...
+            return {
+              ...{
+                icon: locationMarker(colors.primary),
+                iconAnchor: { x: 9, y: 20 }
+              },
+              ...location
+            };
+          })
+        ]}
+        mapCenterPosition={mapCenterPos}
```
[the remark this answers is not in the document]
digorath: I assume this is an artefact of my attempts to make recentering and moving of the map possible.
removed the unnecessary variable in a016f2b8efee7993f12be25d290d39e08488c39b

src/components/MapSwitchHeader.tsx
digorath: this introduced a typescript issue which I resolved with 8e4977015a7e291a01c41326667346eedfb06dbb

donni106: thanks. i tried but wasn't able to find a solution.
we need to check if your rewriting works everywhere. have you checked all screens, where `Touchable` is used?
is something like this still working, where a component is passed as prop: `Component={Touchable}`?

src/components/DropDownHeader.js
```diff
@@ -6,7 +6,7 @@ import { QUERY_TYPES } from '../queries';
 import { DropdownSelect } from './DropdownSelect';
 import { Wrapper } from './Wrapper';
 
-export const ListHeader = ({ query, queryVariables, data, updateListData }) => {
+export const DropDownHeader = ({ query, queryVariables, data, updateListData }) => {
```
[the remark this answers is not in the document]
digorath: done in 411c2a6805e160d8efdb4c06b865a1a0ffaac2ce

src/components/MapSwitchHeader.tsx
digorath: I double checked everything on iOS device and android emulator (my test device is still not available).

it is now a very simple functional component instead of a union type of two functional components that use a different set of props.

donni106: all right, should work, thanks

src/components/map/LocationOverview.tsx
digorath: done in c57107eb38b453d5499ca788e7a5cab02becb237
reworked wrapper as well in the process, so please check if you disagree with any of the changes

src/components/screens/PointOfInterest.js
digorath: see above (done in c57107eb38b453d5499ca788e7a5cab02becb237)

src/components/DropdownHeader.js
```diff
@@ -6,7 +6,7 @@ import { QUERY_TYPES } from '../queries';
 import { DropdownSelect } from './DropdownSelect';
 import { Wrapper } from './Wrapper';
 
-export const DropDownHeader = ({ query, queryVariables, data, updateListData }) => {
+export const DropdownHeader = ({ query, queryVariables, data, updateListData }) => {
```
donni106, line 9: did you use VSCode functionalities for renaming across files, or did you do it manually?

src/components/Wrapper.js
```diff
@@ -33,7 +34,8 @@ export const InfoBox = styled(WrapperRow)`
   margin-bottom: ${normalize(5)}px;
 `;
 
-export const WrapperWithOrientation = memo(({ orientation, dimensions, children }) => {
+export const WrapperWithOrientation = ({children}) => {
+  const {orientation, dimensions} = useContext(OrientationContext);
```
donni106, line 38: nice! very obvious to do it this way.

src/components/map/WebViewMap.tsx
```diff
@@ -0,0 +1,69 @@
...
+        mapCenterPosition={mapCenterPosition ?? locations?.[0]?.position}
+        zoom={zoom}
+      />
+    </View>
+  );
+};
+
+const styles = StyleSheet.create({
+  map: {
+    aspectRatio: 16 / 9,
```
donni106, line 66: :o nice! never used it. maybe that is the solution to go with also for ours images instead of calculating on our own?

src/components/screens/PointOfInterest.js
```diff
@@ -83,7 +88,7 @@ export const PointOfInterest = ({ data, navigation }) => {
     <View>
       {!!images && images.length > 1 && <ImagesCarousel data={images} />}
 
-      <WrapperWithOrientation orientation={orientation} dimensions={dimensions}>
+      <WrapperWithOrientation>
```
donni106, line 91: now the main components rely on the `OrientationContext` and the `WrapperWithOrientation` as well. can this lead to more re-renders than before, as both react to orientation changes?

src/components/Wrapper.js
donni106: but i see something not changed on landscape. when i log, there is no re-rendering when i turn the phone. this needs some investigation. it does not work the same way as in bad belzig app in testflight for example.
when i turn the phone inside an POI, the layout does not change.

src/components/map/WebViewMap.tsx
donni106: but this does not work for landscape. the map fills everything and the screen cannot be handled anymore.

Screenshot_20201208-195432

you can see in the testflight version for bad belzig for example, that images always stay the same size on landscape and portrait. the imagecarousel changes to a version previews to the left and right of the main image.

Screenshot_20201208-200328

the map should also not increase in size, or not that much. maybe this would work if the layout change on orientation change would apply in this branch. could be that the 100% width depend on the smaller content area and the map gets smaller. needs to be checked.

donni106 (changes requested):
pleae check the layout change on orientation change. in my local test this does not behave the way it should.

src/components/DropdownHeader.js
digorath: I used VSCode for the const, and renaming of the file.

But for git to register the file name change in capitalisation I removed and readded it manually.

src/components/screens/PointOfInterest.js
digorath: that should not be the case, as demonstrated here: https://snack.expo.io/@jk.tpwd/rerender-context

src/components/map/WebViewMap.tsx
digorath: the map is wrapped by a `WrapperWithOrientation` that has lesser width.

the issue lies with the orientation not being updated properly on android. (see above)

src/components/Wrapper.js
digorath: there is an underlying issue with the orientation context.

the orientation listener is either unregistered somewhere, or stops working entirely under android.
this happens in master as well.

donni106: i see the correct behaviour for android in master and for example live Bad Belzig App. we can talk about it in our call.

src/screens/IndexScreen.js
```diff
@@ -36,6 +37,7 @@ import {
   shareMessage,
   subtitle
 } from '../helpers';
+import { MapSwitchHeader } from '../components/MapSwitchHeader';
```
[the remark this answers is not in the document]
digorath: done in fbc0c825f370ca356cc88bf60acc0bd607869659

src/components/map/WebViewMap.tsx
```diff
@@ -0,0 +1,57 @@
...
+        ]}
+        mapMarkers={[
+          ...locations.map((location) => {
+            return {
+              ...{
+                icon: locationMarker(colors.primary),
+                iconAnchor: { x: 9, y: 20 }
+              },
+              ...location
+            };
```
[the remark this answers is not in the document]
digorath: changed it in fbc0c825f370ca356cc88bf60acc0bd607869659
the MapMarker type should always have an icon so no need for a fallback value

src/components/map/WebViewMap.tsx
```diff
@@ -0,0 +1,57 @@
+/* eslint-disable react-native/no-color-literals */
```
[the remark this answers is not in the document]
digorath: done in f635b582000d4f6535015877c27c22425713a21e

src/components/map/WebViewMap.tsx
```diff
@@ -0,0 +1,57 @@
...
+import React, { useCallback } from 'react';
+import { StyleProp, View, ViewStyle } from 'react-native';
+import { MapMarker, WebViewLeaflet, WebviewLeafletMessage } from 'react-native-webview-leaflet';
+import { colors } from '../../config';
```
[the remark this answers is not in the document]
digorath: done in d8b0e1412c67e75218f35242814cca6c5d4a7dfe

src/components/Wrapper.js
digorath: added a workaround for the size of the map in 6ad5fba4f34d194fe49bc6f978ee95c95cb9b373

digorath: as discussed there is a react-native 0.63 bug that causes the orientation update to get lost on android (refer [here](https://github.com/facebook/react-native/issues/29451))

that means that this is the best we can do with our limited time

src/components/map/WebViewMap.tsx
```diff
@@ -49,9 +49,16 @@ export const WebViewMap = ({
   );
 };
 
-const styles = StyleSheet.create({
-  map: {
-    aspectRatio: 16 / 9,
-    width: '100%'
+// this will only be set at the start of the app, so this will be the width of portrait mode
+// needs to be done due to react native 0.63 bug for android
```
donni106, line 54: marking with `TODO:` the next time helps to identify it across the app. even code climate points out comments with `TODO:`, `FIXME:`, ...
this way, it is harder to forget it somewhen.

src/components/map/WebViewMap.tsx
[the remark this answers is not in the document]
donni106: external/interal imports line break ;)
not necessary here now, as it will be fixed with the upcoming linting PR.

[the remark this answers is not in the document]
donni106: ah ok, is done in later commit.

src/components/screens/PointOfInterest.js
```diff
@@ -16,12 +16,14 @@ import { ImagesCarousel } from '../ImagesCarousel';
 import { OrientationContext } from '../../OrientationProvider';
 import { useMatomoTrackScreenView } from '../../hooks';
 import { matomoTrackingString } from '../../helpers';
+import { WebViewMap } from '../map';
```
donni106: Bildschirmfoto 2020-12-09 um 21 28 57

sometimes this can happen, where an explicit import is necessary. sorry if i missdirected something here.

done in https://github.com/ikuseiGmbH/smart-village-app-app/pull/137/commits/11f10e4f71a1949fa735808e88ca329e1f050f6f

donni106 (approved):

src/components/map/WebViewMap.tsx
donni106: talked with @digorath

PR #138, opened 2020-12-03
Unify ListComponent

digorath:

- refactor in preparation of the bookmark feature
  - extracted parsing logic from IndexScreen and HomeScreen
  - added ListComponent to handle selection by settings and query type
  - replaced getComponent structure in IndexScreen by ListComponent
  - replaced getListComponent structure in HomeScreen by ListComponent

at this point everything should still look and behave exactly the same 😄 

---

SVA-62

### DIFF
--- a/src/helpers/queryDataParser.js
+++ b/src/helpers/queryDataParser.js
@@ -126,7 +126,7 @@ const parseCategories = (data, skipLastDivider) => {
       title: category.name,
       query:
         category.pointsOfInterestCount > 0 ? QUERY_TYPES.POINTS_OF_INTEREST : QUERY_TYPES.TOURS,
-      queryVariables: { order: 'name_ASC', category: `${category.name}` },
+      queryVariables: { limit: 15, order: 'name_ASC', category: `${category.name}` },
       rootRouteName: category.pointsOfInterestCount > 0 ? 'PointsOfInterest' : 'Tours'
     },
     bottomDivider: skipLastDivider && index !== data.length - 1
--- a/src/helpers/queryDataParser.js
+++ b/src/helpers/queryDataParser.js
@@ -9,7 +9,7 @@ import { shareMessage } from './shareHelper';
 import { subtitle } from './textHelper';
 
 export const parseEventRecords = (data, skipLastDivider = false) => {
-  return data.map((eventRecord, index) => ({
+  return data?.map((eventRecord, index) => ({
     id: eventRecord.id,
     subtitle: subtitle(
       eventDate(eventRecord.listDate),
@@ -135,6 +135,7 @@ const parseCategories = (data, skipLastDivider = false) => {
 
 export const parseListItemsFromQuery = (query, data) => {
   if (!(data && data[query])) return;
+
   switch (query) {
   case QUERY_TYPES.EVENT_RECORDS:
     return parseEventRecords(data[query]);
--- a/src/helpers/queryDataParser.js
+++ b/src/helpers/queryDataParser.js
@@ -8,7 +8,7 @@ import { momentFormat } from './momentHelper';
 import { shareMessage } from './shareHelper';
 import { subtitle } from './textHelper';
 
-export const parseEventRecords = (data, skipLastDivider = false) => {
+export const parseEventRecords = (data, skipLastDivider) => {
   return data?.map((eventRecord, index) => ({
     id: eventRecord.id,
     subtitle: subtitle(
@@ -30,11 +30,11 @@ export const parseEventRecords = (data, skipLastDivider = false) => {
       },
       details: eventRecord
     },
-    bottomDivider: skipLastDivider ? index !== data.length - 1 : undefined
+    bottomDivider: skipLastDivider && index !== data.length - 1
   }));
 };
 
-export const parseNewsItems = (data, skipLastDivider = false) => {
+export const parseNewsItems = (data, skipLastDivider) => {
   return data?.map((newsItem, index) => ({
     id: newsItem.id,
     subtitle: subtitle(momentFormat(newsItem.publishedAt), newsItem.dataProvider?.name),
@@ -59,11 +59,11 @@ export const parseNewsItems = (data, skipLastDivider = false) => {
       },
       details: newsItem
     },
-    bottomDivider: skipLastDivider ? index !== data.length - 1 : undefined
+    bottomDivider: skipLastDivider && index !== data.length - 1
   }));
 };
 
-export const parsePointOfInterest = (data, skipLastDivider = false) => {
+export const parsePointOfInterest = (data, skipLastDivider) => {
   return data?.map((pointOfInterest, index) => ({
     id: pointOfInterest.id,
     title: pointOfInterest.name,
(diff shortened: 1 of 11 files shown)


src/helpers/queryDataParser.js
```diff
@@ -0,0 +1,150 @@
...
+
+import { QUERY_TYPES } from '../queries';
+import { eventDate } from './dateTimeHelper';
+import { mainImageOfMediaContents } from './imageHelper';
+import { momentFormat } from './momentHelper';
+import { shareMessage } from './shareHelper';
+import { subtitle } from './textHelper';
+
+export const parseEventRecords = (data, skipLastDivider = false) => {
+  return data.map((eventRecord, index) => ({
```
donni106, line 12: missing `?` for `data` like in the other methodes. i will add it, as i am in the file right now.

- [ ] data?

src/helpers/queryDataParser.js
```diff
@@ -0,0 +1,151 @@
...
+      title: 'Veranstaltung',
+      query: QUERY_TYPES.EVENT_RECORD,
+      queryVariables: { id: `${eventRecord.id}` },
+      rootRouteName: 'EventRecords',
+      shareContent: {
+        message: shareMessage(eventRecord, QUERY_TYPES.EVENT_RECORD)
+      },
+      details: eventRecord
+    },
+    bottomDivider: skipLastDivider ? index !== data.length - 1 : undefined
```
donni106, line 33: we can use following for `bottomDivider` in all parse methods to implicitly return `undefined` which simplifies in my eyes:

```
bottomDivider: skipLastDivider && index !== data.length - 1
```

what do you think?

src/helpers/queryDataParser.js
```diff
@@ -0,0 +1,151 @@
...
+import { momentFormat } from './momentHelper';
+import { shareMessage } from './shareHelper';
+import { subtitle } from './textHelper';
+
+export const parseEventRecords = (data, skipLastDivider = false) => {
+  return data?.map((eventRecord, index) => ({
+    id: eventRecord.id,
+    subtitle: subtitle(
+      eventDate(eventRecord.listDate),
+      eventRecord.addresses?.[0]?.addition || eventRecord.addresses?.[0]?.city
```
donni106, line 16: it wasn't easy to catch the change to optional chaining here from the diffs. refactoring this is missing in the commit message, which should have been pointed out there to avoid problems.

src/screens/HomeScreen.js
```diff
@@ -399,47 +295,18 @@ export const HomeScreen = ({ navigation }) => {
...
 
                 const eventRecordsListType = listTypesSettings[QUERY_TYPES.EVENT_RECORDS];
-                const ListComponent = getListComponent(eventRecordsListType);
 
                 return (
                   <View>
                     <ListComponent
                       navigation={navigation}
                       data={eventRecords}
-                      leftImage={eventRecordsListType === LIST_TYPES.IMAGE_TEXT_LIST}
```
donni106, line 442: nice that this `leftImage` is not necessary anymore

donni106 (commented):
i appreciate the refactoring! thanks, looks good. i added linting and a small correction.

if `CardList` i have a warning popping up:

Simulator Screen Shot - iPhone 11 Pro Max - 2020-12-03 at 17 24 18

> [Unhandled promise rejection: Invariant Violation: Expecting a parsed GraphQL document. Perhaps you need to wrap the query string in a "gql" tag? http://docs.apollostack.com/apollo-client/core.html#gql]

i didn't get the reason yet.
can you please take an additional look.

src/helpers/queryDataParser.js
digorath: yes it is simply making sure that we do not do `undefined[0]` by optional chaining.

src/screens/HomeScreen.js
digorath: I think it was unnecessary before, due to the way the `ImageTextList` was either selected In the case were the condition is `true` and then ignores the prop altogether.

src/helpers/queryDataParser.js
digorath: I changed it in 1ba7bec2cc34621ee768b59c4fb706fa19475e62

But on a side note: why do we use bottom dividers on the list entries and not top dividers?

By using top dividers we could enable them on each item without the need to check for indices if I see it correctly.

that would also fix the inconsistency with tours and POIs.

digorath:
> i appreciate the refactoring! thanks, looks good. i added linting and a small correction.
> 
> if `CardList` i have a warning popping up:
> 
> Simulator Screen Shot - iPhone 11 Pro Max - 2020-12-03 at 17 24 18
> 
> > [Unhandled promise rejection: Invariant Violation: Expecting a parsed GraphQL document. Perhaps you need to wrap the query string in a "gql" tag? http://docs.apollostack.com/apollo-client/core.html#gql]
> 
> i didn't get the reason yet.
> can you please take an additional look.

I can't seem to reliably reproduce it. can you provide any specific steps to reproduce it? "if `CardList`" was a bit short ;)

If you cannot give a reliable method of reproducing this warning, I will take an in depth look on my own.

digorath:
> > i appreciate the refactoring! thanks, looks good. i added linting and a small correction.
> > if `CardList` i have a warning popping up:
> > Simulator Screen Shot - iPhone 11 Pro Max - 2020-12-03 at 17 24 18
> > > [Unhandled promise rejection: Invariant Violation: Expecting a parsed GraphQL document. Perhaps you need to wrap the query string in a "gql" tag? http://docs.apollostack.com/apollo-client/core.html#gql]
> > 
> > 
> > i didn't get the reason yet.
> > can you please take an additional look.
> 
> I can't seem to reliably reproduce it. can you provide any specific steps to reproduce it? "if `CardList`" was a bit short ;)
> 
> If you cannot give a reliable method of reproducing this warning, I will take an in depth look on my own.

I just checked it again, and found a way to reproduce it, but it is not specific to this branch.

For me one warning like this pops up only on specific categories, so I assume there is something wrong with the data in those categories.

One category that causes this warning is "Angelgewässer", while "Aussichtstürme" seems to be unaffected.

digorath:
> > > i appreciate the refactoring! thanks, looks good. i added linting and a small correction.
> > > if `CardList` i have a warning popping up:
> > > Simulator Screen Shot - iPhone 11 Pro Max - 2020-12-03 at 17 24 18
> > > > [Unhandled promise rejection: Invariant Violation: Expecting a parsed GraphQL document. Perhaps you need to wrap the query string in a "gql" tag? http://docs.apollostack.com/apollo-client/core.html#gql]
> > > 
> > > 
> > > i didn't get the reason yet.
> > > can you please take an additional look.
> > 
> > 
> > I can't seem to reliably reproduce it. can you provide any specific steps to reproduce it? "if `CardList`" was a bit short ;)
> > If you cannot give a reliable method of reproducing this warning, I will take an in depth look on my own.
> 
> I just checked it again, and found a way to reproduce it, but it is not specific to this branch.
> 
> For me one warning like this pops up only on specific categories, so I assume there is something wrong with the data in those categories.
> 
> One category that causes this warning is "Angelgewässer", while "Aussichtstürme" seems to be unaffected.

the issue lies [here](https://github.com/ikuseiGmbH/smart-village-app-app/blob/3c997c41c11311cad9e90180eb6aa4c2100f2191/src/queries/index.js#L55) and is not introduced by this PR

src/helpers/queryDataParser.js
donni106: if the first item has a top divider, it looks weird combined with the <TitleContainer><Title>... so we would need to have to condition other way around.

donni106:
I will have a deeper look for fixing it and plan to merge it for a 1.5.0 release in january.

PR #141, opened 2020-12-08
fix(startup): add checks for empty objects

donni106:
- setup of global settings was buggy when there is nothing stored on the device yet
- added lodashs `_isEmpty` because check for empty objects were not implemented correctly


digorath (approved):
Looks good to me

PR #142, opened 2020-12-08
Addition of a  useRefreshTime hook

### LINKED ISSUE #121: Refactor logics for refresh time intervals to own hook useRefreshTime(...)
We have refresh time logics in several screens, which are handled by an `useEffect` each in the same way with different params. There could be a new hook handling this generally.

digorath:

As described in #121 there was a lot of code duplication for the refreshTime.
It is now centralised in a new custom hook.

There should be no visual changes.

The default value for the second argument was chosen because there was only one place that differed from it.

---

Fixes #121

### DIFF
--- a/src/hooks/RefreshTime.ts
+++ b/src/hooks/RefreshTime.ts
@@ -1,0 +1,26 @@
+import { useCallback, useEffect, useState } from 'react';
+
+import { consts } from '../config';
+import { refreshTimeFor } from '../helpers';
+
+export const useRefreshTime = (
+  refreshTimeKey: string,
+  refreshInterval = consts.REFRESH_INTERVALS.STATIC_CONTENT
+) => {
+  const [refreshTime, setRefreshTime] = useState<number>();
+
+  const getRefreshTime = useCallback(async () => {
+    const time = await refreshTimeFor(
+      refreshTimeKey,
+      refreshInterval
+    );
+
+    setRefreshTime(time);
+  }, [refreshTimeKey, setRefreshTime]);
+
+  useEffect(() => {
+    getRefreshTime();
+  }, [getRefreshTime]);
+
+  return refreshTime;
+};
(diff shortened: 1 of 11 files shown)


src/hooks/RefreshTime.ts
```diff
@@ -0,0 +1,26 @@
+import { useCallback, useEffect, useState } from 'react';
+
+import { consts } from '../config';
+import { refreshTimeFor } from '../helpers';
+
+export const useRefreshTime = (
+  refreshTimeKey: string,
+  refreshInterval = consts.REFRESH_INTERVALS.STATIC_CONTENT
```
donni106, line 8: maybe we should export `const ONCE_A_DAY = 'ONCE_A_DAY';` in `consts` to use it here, as `STATIC_CONTENT` is logically used for data of type static contents. the value is the same, but the meaning is a different.

donni106:
I plan to merge this for 1.4.2 release next week.

src/hooks/RefreshTime.ts
donni106: done in https://github.com/ikuseiGmbH/smart-village-app-app/pull/142/commits/b525e8503899998113202a45cec465cccab2b50a

donni106 (approved):

SoleCincis:
Why useRefreshTime? 
We don't want to copy and paste the same logic within all the components. So what we do instead, is to extract the logic. And move it to its own function. So any component that may need the logic, can import it in, and use it.

SoleCincis:
So refreshTime and RefreshControl both refetch data , just first one is doing it with an interval of once per day and other one is doing it on user action ( pull screen down ) ? 
If yes could we use this new custom hook somehow in RefreshControl as well ?

SoleCincis (commented):
doubts have been clarified.

PR #146, opened 2020-12-14
bookmarks

digorath:

Implemeted the new bookmarks feature.

The following components were added:
* New BookmarkHeader for DetailScreen
* New Screens:
  * BookmarkScreen
  * BookmarkCategoryScreen

The BookmarkScreen can be reached through the new headerLeft on the home screen.
The navigation to the settings screen was moved to the headerRight of the bookmark screen.

The BookmarkCategoryScreen behaves similar to the Index Screen but provides no filter options.
Filtering for a provider that one has no bookmarks of caused the whole list including the filter header to disappear.

---

SVA-62

Blocked by #138


donni106:
@digorath please rebase this branch onto the latest master to get all updates of version 1.4.2 and resolve conflicts

donni106:
if i bookmark an entry, i can see it in my list. then i go into the entry from the list and remove it from the bookmarks. then i navigate back to the list and the entry is still present there. i assume that users would be confused, as they just removed it the moment before.

donni106:
bookmarks are not working offline. without network connection the app crashed for me.

src/components/index.js
```diff
@@ -1,6 +1,7 @@
 export * from './map';
 export * from './screens';
 
+export * from './bookmarks';
```
donni106, line 4: here i think was my thought to structure the folders first separated by a line break from the files, that are directly in /components. so the bookmarks line should go up, right?

src/config/consts.js
```diff
@@ -22,19 +22,21 @@ export const consts = {
...
   MATOMO_TRACKING: {
     SCREEN_VIEW: {
-      HOME: 'Home',
-      NEWS_ITEMS: 'News',
-      EVENT_RECORDS: 'Events',
-      POINTS_OF_INTEREST: 'Points of interest',
-      TOURS: 'Tours',
-      POINTS_OF_INTEREST_AND_TOURS: 'Points of interest and tours',
-      SERVICE: 'Service',
+      BOOKMARK_CATEGORY: 'Bookmark Category',
```
donni106, line 25: im englischen `category` klein

src/helpers/index.js
```diff
@@ -1,3 +1,4 @@
+export * from './bookmarkHelpers';
```
donni106, line 1: filename singular please to align with the others

src/navigation/AppStackNavigator.js
```diff
@@ -11,6 +12,7 @@ import {
 import { defaultStackNavigatorConfig } from './defaultStackNavigatorConfig';
 
 import { texts } from '../config';
+import { BookmarkCategoryScreen } from '../screens/BookmarkCategoryScreen';
```
donni106, line 15: from /..screens?

donni106:
so the data marked is sorted like in the "normal" screens, right? with the published date for news for example.
it is not the order the user marked something.
it was not written to do it in another way, but there can be different opinions on sorting in a bookmark list. maybe this needs to be a filter somewhen in the future.

src/helpers/bookmarkHelpers.ts
```diff
@@ -0,0 +1,84 @@
...
+  );
+
+  return !!bookmarks?.[key]?.find((item) => item === id);
+};
+
+export const toggleBookmark = async (itemType: string, id: string, category?: number) => {
+  const bookmarks: BookmarkList | undefined = await readFromStore(storeKey);
+  const listQueryType = getListQueryType(itemType);
+
+  // FIXME: add error handling
```
donni106, line 51: `FIXME:` 👍  
(well, i would define it as TODO, as it is not existing yet compared to a FIXME for something not working correctly)

donni106 (changes requested):
thanks. looks quite good. i encountered some problems in using, that needs to be tackled.

digorath:
> bookmarks are not working offline. without network connection the app crashed for me.
> 
> Screenshot_20201216-123811

this should be fixed with 8d8944110b55f037579c3bdcea418e4eb29961ec

src/components/index.js
digorath: you are right. changed it in 8a84e9e25ce56257a4c54ca67a967bcc32fcc310

src/helpers/index.js
digorath: changed it in 8a84e9e25ce56257a4c54ca67a967bcc32fcc310

src/config/consts.js
digorath: changed it in 8a84e9e25ce56257a4c54ca67a967bcc32fcc310

src/navigation/AppStackNavigator.js
digorath: changed it in 8a84e9e25ce56257a4c54ca67a967bcc32fcc310

digorath:
> so the data marked is sorted like in the "normal" screens, right? with the published date for news for example.
> it is not the order the user marked something.
> it was not written to do it in another way, but there can be different opinions on sorting in a bookmark list. maybe this needs to be a filter somewhen in the future.

for now I changed it in 8d8944110b55f037579c3bdcea418e4eb29961ec to show the most recent 3 bookmarked items in the overview screen.
the category/type screen will show them sorted with the default sorting order for now.

src/config/texts.js
```diff
@@ -9,6 +17,9 @@ export const texts = {
     pointsOfInterest: 'Orte',
     tours: 'Touren'
   },
+  errors: {
+    noData: 'Bitte überprüfen sie ihre Internetverbindung.'
```
donni106, line 21: `sie ihre` -> `Sie Ihre`

src/components/bookmarks/BookmarkSection.tsx
```diff
@@ -0,0 +1,132 @@
...
+
+  if (loading) {
+    return (
+      <LoadingContainer>
+        <ActivityIndicator color={colors.accent} />
+      </LoadingContainer>
+    );
+  }
+
+  if (!data) return null;
```
donni106, line 110: I think i reach this null, when I bookmark things while i am offline. I got a white screen.

src/screens/BookmarkCategoryScreen.js
```diff
@@ -0,0 +1,107 @@
...
+      <Wrapper>
+        <RegularText>{texts.bookmarks.noBookmarksinCategory}</RegularText>
+      </Wrapper>
+    );
+  }
+
+  if (!data) {
+    return (
+      <Wrapper>
+        <RegularText>{texts.errors.noData}</RegularText>
```
donni106, line 70: I haven't reached this while using offline.

donni106:
> if i bookmark an entry, i can see it in my list. then i go into the entry from the list and remove it from the bookmarks. then i navigate back to the list and the entry is still present there. i assume that users would be confused, as they just removed it the moment before.

this does work as expected now.
done in https://github.com/ikuseiGmbH/smart-village-app-app/pull/146/commits/a20d08bb2a2127b124bcdd0d983f4731243bebbd @digorath?

src/helpers/bookmarkHelper.ts
```diff
@@ -0,0 +1,67 @@
...
+// than the other item types in every other component
+// this way the hierarchy in the storage remains flat and homogeneous
+export const getKeyFromTypeAndCategory = (itemType: string, category?: number) =>
+  category ? `${itemType}-${category}` : itemType;
+
+export const toggleBookmark = async (itemType: string, id: string, category?: number) => {
+  const bookmarks: BookmarkList | undefined = await readFromStore(BOOKMARKS_STORE_KEY);
+  const listQueryType = getListQueryType(itemType);
+
+  // FIXME: add error handling
```
donni106, line 35: what error handling is being thought of here?

donni106 (commented):
It seems to be necessary to discuss behavior for offline usage.

digorath:
> > if i bookmark an entry, i can see it in my list. then i go into the entry from the list and remove it from the bookmarks. then i navigate back to the list and the entry is still present there. i assume that users would be confused, as they just removed it the moment before.
> 
> this does work as expected now.
> done in [a20d08b](https://github.com/ikuseiGmbH/smart-village-app-app/commit/a20d08bb2a2127b124bcdd0d983f4731243bebbd) @digorath?

yes

src/helpers/bookmarkHelper.ts
digorath: once we have some sort of error reporting, we can report this case. until then we can't do much about it.

src/components/bookmarks/BookmarkSection.tsx
digorath: true. this is actually an annoying one to fix as an error text would potentially be shown multiple times if added here.
we could pass in an onNoData callback to the sections from the overview screen, and have it be handled there. what do you think?

src/config/texts.js
digorath: done in fc785872ca155350a23f1886ecfd91d39ae4d089

src/components/bookmarks/BookmarkSection.tsx
donni106: let us call about it with more context

donni106: this place will be used for the no data callback.

donni106 (approved):

PR #150, opened 2020-12-18
fix(push notification): navigation on interaction

digorath:

previously we always used `QUERY_TYPES.NEWS_ITEM` for the navigation triggered by a push notification.

This can cause errors if push notifications for other item types are sent out.

### DIFF
--- a/src/components/screens/EventRecord.js
+++ b/src/components/screens/EventRecord.js
@@ -53,12 +53,13 @@ export const EventRecord = ({ data, navigation }) => {
   } = data;
   const link = webUrls && webUrls.length && webUrls[0].url;
   const rootRouteName = navigation.getParam('rootRouteName', '');
+  const headerTitle = navigation.getParam('title', '');
   // action to open source urls
   const openWebScreen = (webUrl) =>
     navigation.navigate({
       routeName: 'Web',
       params: {
-        title: 'Veranstaltung',
+        title: headerTitle,
         webUrl: !!webUrl && typeof webUrl === 'string' ? webUrl : link,
         rootRouteName
       }
--- a/src/components/screens/NewsItem.js
+++ b/src/components/screens/NewsItem.js
@@ -51,12 +51,13 @@ export const NewsItem = ({ data, navigation }) => {
   // the title of a news item is either a given main title or the title from the first content block
   const title = mainTitle || (!!contentBlocks && !!contentBlocks.length && contentBlocks[0].title);
   const rootRouteName = navigation.getParam('rootRouteName', '');
+  const headerTitle = navigation.getParam('title', '');
   // action to open source urls
   const openWebScreen = (webUrl) =>
     navigation.navigate({
       routeName: 'Web',
       params: {
-        title: 'Nachricht',
+        title: headerTitle,
         webUrl: !!webUrl && typeof webUrl === 'string' ? webUrl : link,
         rootRouteName
       }
--- a/src/components/screens/Tour.js
+++ b/src/components/screens/Tour.js
@@ -37,12 +37,13 @@ export const Tour = ({ data, navigation }) => {
     webUrls
   } = data;
   const rootRouteName = navigation.getParam('rootRouteName', '');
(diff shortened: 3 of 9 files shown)


donni106:
We are missing some navigation params to make the detail screen behave as navigating from list entries:

from push:
```js
navigation.navigate({
  routeName: 'Detail',
  params: {
    query: queryType,
    queryVariables: { id: data.id }
  }
});
```

from lists (example for a event record):
```js
navigation.navigate({
  title: 'Veranstaltung',
  query: QUERY_TYPES.EVENT_RECORD,
  queryVariables: { id: `${eventRecord.id}` },
  rootRouteName: 'EventRecords',
  shareContent: {
    message: shareMessage(eventRecord, QUERY_TYPES.EVENT_RECORD)
  },
  details: eventRecord
});
```

`title` is necessary to display a header title on the detail screen, which is the german expression for the content type. for news items this is complicated, as the title is based on the news item category. otherwise it is a string based on the `queryType`.
`rootRouteName` maybe is not needed anymore. this need to be checked in terms of react navigation update. otherwise it is a string based on the `queryType`.
`shareContent` is needed to have the share functionality in the detail screen. the `shareMessage` params need to contain the item itself, which is not available in push case, where there is only the id.
we can ignore `details` in push case, as we do not need to that offline backup there.

the share stuff is the biggest problem. do we need to pass more data with the push notification or refactor the entire share functionality? otherwise we could fetch all necessary data with the id inside the `interactionHandler`, so everything needed is present for navigation?

digorath:
> We are missing some navigation params to make the detail screen behave as navigating from list entries:
> 
> from push:
> 
> ```js
> navigation.navigate({
>   routeName: 'Detail',
>   params: {
>     query: queryType,
>     queryVariables: { id: data.id }
>   }
> });
> ```
> 
> from lists (example for a event record):
> 
> ```js
> navigation.navigate({
>   title: 'Veranstaltung',
>   query: QUERY_TYPES.EVENT_RECORD,
>   queryVariables: { id: `${eventRecord.id}` },
>   rootRouteName: 'EventRecords',
>   shareContent: {
>     message: shareMessage(eventRecord, QUERY_TYPES.EVENT_RECORD)
>   },
>   details: eventRecord
> });
> ```
> 
> `title` is necessary to display a header title on the detail screen, which is the german expression for the content type. for news items this is complicated, as the title is based on the news item category. otherwise it is a string based on the `queryType`.
> `rootRouteName` maybe is not needed anymore. this need to be checked in terms of react navigation update. otherwise it is a string based on the `queryType`.
> `shareContent` is needed to have the share functionality in the detail screen. the `shareMessage` params need to contain the item itself, which is not available in push case, where there is only the id.
> we can ignore `details` in push case, as we do not need to that offline backup there.
> 
> the share stuff is the biggest problem. do we need to pass more data with the push notification or refactor the entire share functionality? otherwise we could fetch all necessary data with the id inside the `interactionHandler`, so everything needed is present for navigation?

I would recommend not putting it into the interaction handler as this would cause the navigation to receive a noticable delay, and we would need to include offline handling in there as well.

with react-navigation 5 we can dynamically set the header info after navigating, which we should do at that point anyways, as we would reduce data duplication in the navigation params as well as remove the bad practice of passing in entire data structures into the params instead of ids.

donni106:
we decided to have no sharing on detail screens from push notifications, as we cannot handle that properly with the current code structure limited by the older react navigation version.
this can be tackled after react navigation update, which will give us more opportunities.

for `title` and `rootRouteName` we could add methods to return values based on the `queryType`.
for news items, the `title` is based on a category, which we do not have at that point, so we can only use `Nachricht` in that case. but even better than nothing, or what do you say?

donni106:
with the update https://github.com/ikuseiGmbH/smart-village-app-app/pull/150/commits/b59a91e7f3a89ba279e61d2dc87f3eaee731874f now there is always a `title` for a detail screen and in a drawer a corresponding root route would be marked as bold as well.

donni106 (approved):

PR #153, opened 2020-12-22
fix(location overview): fetch policy

digorath:

- added fetchPolicy and skip parameter to queries

### DIFF
--- a/src/components/map/LocationOverview.tsx
+++ b/src/components/map/LocationOverview.tsx
@@ -1,11 +1,13 @@
-import React, { useCallback, useState } from 'react';
+import React, { useCallback, useContext, useState } from 'react';
 import { useQuery } from 'react-apollo';
 import { ActivityIndicator, View } from 'react-native';
 import { WebviewLeafletMessage } from 'react-native-webview-leaflet';
 import { NavigationScreenProps, ScrollView } from 'react-navigation';
 
 import { colors } from '../../config';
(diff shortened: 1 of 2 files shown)


donni106:
Why were these new params added? What is happening now, which was missing before?

src/components/map/LocationOverview.tsx
```diff
@@ -36,14 +38,21 @@ const mapToMapMarkers = (data: any) => {
 };
 
 export const LocationOverview = ({ navigation, category }: Props) => {
+  const { isConnected, isMainserverUp } = useContext(NetworkContext);
   const [selectedPointOfInterest, setSelectedPointOfInterest] = useState<string>();
 
+  const fetchPolicy = graphqlFetchPolicy({ isConnected, isMainserverUp, refreshTime: undefined});
```
donni106, line 44: this could be `const fetchPolicy = graphqlFetchPolicy({ isConnected, isMainserverUp })` like in other places already. The `refreshTime: undefined` is not necessary, if there is no refresh time value to pass.

digorath: it is necessary in this case because typescript does not know that this is an optional key of the argument.

digorath:
> Why were these new params added? What is happening now, which was missing before?

the fetchpolicy was added so that we use caching here, and do not fire a query while being offline either.

PR #154, opened 2021-01-05
Epic/widgets

digorath:

This PR will accumulate all the changes for the widgets.
By itself it only provides the framework

---

SVA-80, SVA-82, SVA-83


src/components/widgets/Widgets.tsx
```diff
@@ -0,0 +1,38 @@
...
+
+  const filteredWidgets = getExistingWidgets(widgets);
+
+  if (!filteredWidgets?.length) return null;
+
+  const widgetComponents = filteredWidgets.map((Component, index) => {
+    return <Component key={index} navigation={navigation} />;
+  });
+
+  return <View style={styles.container}>{widgetComponents}</View>;
```
donni106, line 30: we have several `Wrapper` styled components for container `View`s, which also makes sense here.
there is no existing wrapper for `row` and `space-between`, only a `WrapperRow`. i will create a `WrapperRowSpaceBetween` and use it here. so we do not need the `styles` and `StyleSheet` and `View` in this file.

donni106: done in https://github.com/ikuseiGmbH/smart-village-app-app/pull/154/commits/26661f8e177cf56258f9073ed6b8a217e7e19319

donni106: even better is the ability to add a prop for the justify to not need to create multiple components for that.
like for the text component where several props can individualize the text component.
so we will have `<WrapperRow center>` and `<WrapperRow space-between>` and so on.

donni106: done in https://github.com/ikuseiGmbH/smart-village-app-app/pull/154/commits/7076d4c2a7eaf37592fe9ef50033565dff4c864b

digorath: :+1:

donni106 (approved):

PR #156, opened 2021-01-05
feature: event widget

digorath:

Added an event widget to the home screen
- it shows the number of events that are active today.
- interaction will lead to the index screen for the events, filtered for today.
- changed condition to in index screen to still render header and list if data length is 0
  - otherwise the filter logic needs to be entirely reworked for the event widget

---

SVA-82

### DIFF
--- a/src/components/widgets/EventWidget.tsx
+++ b/src/components/widgets/EventWidget.tsx
@@ -1,0 +1,78 @@
+import moment from 'moment';
+import React, { useCallback, useContext } from 'react';
+import { useQuery } from 'react-apollo';
+import { StyleSheet, View } from 'react-native';
+import { NavigationScreenProp } from 'react-navigation';
+
+import { colors, consts, normalize, texts } from '../../config';
+import { graphqlFetchPolicy } from '../../helpers';
+import { useRefreshTime } from '../../hooks';
+import { calendar } from '../../icons';
+import { NetworkContext } from '../../NetworkProvider';
+import { getQuery, QUERY_TYPES } from '../../queries';
+import { Icon } from '../Icon';
+import { BoldText, RegularText } from '../Text';
+import { Touchable } from '../Touchable';
+import { Wrapper } from '../Wrapper';
+
+type Props = {
+  navigation: NavigationScreenProp<never>;
+};
+
+export const EventWidget = ({ navigation }: Props) => {
+  const refreshTime = useRefreshTime('event-widget', consts.REFRESH_INTERVALS.ONCE_A_DAY);
+  const { isConnected, isMainserverUp } = useContext(NetworkContext);
+
+  const fetchPolicy = graphqlFetchPolicy({ isConnected, isMainserverUp, refreshTime });
+
+  const currentDate = moment().format('YYYY-MM-DD');
+  const queryVariables = {
+    dateRange: [currentDate, currentDate]
+  };
+
+  const { data, loading } = useQuery(getQuery(QUERY_TYPES.EVENT_RECORDS), {
+    fetchPolicy,
+    variables: queryVariables
+  });
+
+  const onPress = useCallback(() => {
+    navigation.navigate('Index', {
+      title: 'Veranstaltungen',
+      query: QUERY_TYPES.EVENT_RECORDS,
+      queryVariables,
+      rootRouteName: 'EventRecords'
+    });
+  }, [navigation]);
+
+  const eventCount = data?.eventRecords?.length;
+
+  return (
+    <Touchable onPress={onPress}>
+      <Wrapper>
+        <View style={styles.container}>
+          <Icon style={styles.icon} xml={calendar(colors.primary)} />
+          <BoldText style={styles.count}>{loading ? ' ' : eventCount}</BoldText>
+        </View>
+        <RegularText style={styles.text}>{texts.homeTitles.events}</RegularText>
+      </Wrapper>
+    </Touchable>
+  );
+};
+
+const styles = StyleSheet.create({
+  container: {
+    flexDirection: 'row',
+    justifyContent: 'center'
+  },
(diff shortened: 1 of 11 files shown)


src/components/widgets/EventWidget.tsx
```diff
@@ -0,0 +1,78 @@
...
+  }, [navigation]);
+
+  const eventCount = data?.eventRecords?.length;
+
+  return (
+    <Touchable onPress={onPress}>
+      <Wrapper>
+        <View style={styles.container}>
+          <Icon style={styles.icon} xml={calendar(colors.primary)} />
+          <BoldText style={styles.count}>{loading ? ' ' : eventCount}</BoldText>
```
donni106, line 54: you can add the prop `primary` for setting the color instead of in stylesheet

src/components/widgets/EventWidget.tsx
```diff
@@ -0,0 +1,78 @@
...
+        </View>
+        <RegularText style={styles.text}>{texts.homeTitles.events}</RegularText>
```
donni106, line 56: you can add the prop primary for setting the color instead of in stylesheet

src/components/widgets/EventWidget.tsx
```diff
@@ -0,0 +1,78 @@
...
+};
+
+const styles = StyleSheet.create({
+  container: {
+    flexDirection: 'row',
+    justifyContent: 'center'
+  },
+  count: {
+    color: colors.primary,
+    fontSize: normalize(20),
```
donni106, line 69: in the screenshot from the customer, the texts are smaller than default texts/texts in lists and bold, here they are bigger and normal weight.

src/components/widgets/EventWidget.tsx
```diff
@@ -0,0 +1,78 @@
...
+    paddingTop: normalize(4)
```
donni106, line 70: is 4 choosen to vertically center the number next to the icon?

src/components/widgets/EventWidget.tsx
```diff
@@ -0,0 +1,78 @@
...
+      rootRouteName: 'EventRecords'
+    });
+  }, [navigation]);
+
+  const eventCount = data?.eventRecords?.length;
+
+  return (
+    <Touchable onPress={onPress}>
+      <Wrapper>
+        <View style={styles.container}>
```
donni106, line 52: instead of this view with stylesheet you can use `<WrapperRow center>` after rebase on `epic/widgets` branch

src/screens/IndexScreen.js
```diff
@@ -168,30 +168,26 @@ export const IndexScreen = ({ navigation }) => {
...
-                    onRefresh={() => refresh(refetch)}
-                    colors={[colors.accent]}
-                    tintColor={colors.accent}
-                  />
-                }
-              />
+              <View>
+                {showFilter ? (
+                  <DropdownHeader {...{ query, queryVariables, data, updateListData }} />
+                ) : null}
```
donni106, line 175: to have it a bit shorter and avoid the `: null`, this should work: `{showFilter && <DropdownHeader {...{ query, queryVariables, data, updateListData }} />}`

donni106: having the `DropdownHeader` not as list header but as component before the list results in different behavior. with that the dropdown is sticky and only the list scrolls. when scrolling the list at the very end, the last item is not visible, because the list has not the full screen space.
this should be fixed.

i think of reverting the change and render the dropdown instead of `null` if there are no list items.

donni106 (changes requested):
please see the comments on code lines for refactorings and fixing the dropdown if there are no list items.

src/screens/IndexScreen.js
donni106: another bad side effect is, that the pull to refresh is under the dropdown and the list does not span the whole screen.
if there is just one event, this results in strange scrolling in that small area:

Simulator Screen Shot - iPhone 11 Pro Max - 2021-01-08 at 12 16 50

src/components/widgets/EventWidget.tsx
digorath: done in c9ab08812de7540f17593e724e8cbe47fe520152

src/components/widgets/EventWidget.tsx
digorath: done in c9ab08812de7540f17593e724e8cbe47fe520152

src/screens/IndexScreen.js
digorath: 7ac953e72ba1ba13b6f9bb652023493dd73e8a48 and 6ab0a0ddcce8c24cabc377ef520c67fcf10dbcba should fix the issue.

It is not great that cardlist and the other lists behave inconsistently in terms of styling. we should reconsider that. in the future

src/components/widgets/EventWidget.tsx
digorath: done in c9ab08812de7540f17593e724e8cbe47fe520152

src/components/widgets/EventWidget.tsx
digorath: changed it to small and bold in 556fcc822429334750529d374ae079338316286e

src/components/widgets/EventWidget.tsx
digorath: yes. unfortunately the normal methods of alignment do not work with the way the text is set up, once we require a bigger size.

changed it in a28b18eee39539aa6c7fdf6e07938cccad007cba to work with props of BoldText instead

src/screens/IndexScreen.js
donni106: all right, thanks. talk about it tomorrow.

donni106 (approved):

PR #157, opened 2021-01-05
Feature: construction sites widget

digorath:

Added a widget for construction sites
- it shows the number of construction sites
- interaction will lead to a list of construction sites
- detail page of a construction site can show
  - image
  - title + category
  - location
  - directional information
  - cause/reason of the construction site
  - further information
  - information about restrictions that are in place
  - map with location

---

SVA-83


digorath:
do we need to improve the refreshing behaviour?

src/screens/ConstructionSiteOverviewScreen.js
```diff
@@ -0,0 +1,83 @@
...
+          ? formattedStartDate
+          : `${formattedStartDate} - ${formattedEndDate}`;
+
+      return (
+        <ListItem
+          onPress={onPress}
+          rightIcon={<Icon xml={arrowRight(colors.darkText)} />}
+          subtitle={<RegularText>{formattedDates}</RegularText>}
+          title={<BoldText>{extendedTitle}</BoldText>}
+          topDivider
```
donni106, line 41: last element is missing a bottom divider, like there is in other list screens

src/navigation/AppStackNavigator.js
```diff
@@ -21,6 +23,18 @@ const AppStackNavigator = (headerRight = true) =>
           title: props.navigation.getParam('title', texts.screenTitles.home)
         })
       },
+      ConstructionSiteDetail: {
+        screen: ConstructionSiteDetailScreen,
+        navigationOptions: () => ({
+          title: texts.screenTitles.constructionSite
```
donni106, line 29: detail title should be singular, like in other detail screens

src/screens/ConstructionSiteOverviewScreen.js
```diff
@@ -0,0 +1,83 @@
...
+      const formattedEndDate = endDate ? momentFormat(endDate) : undefined;
+      const formattedDates =
+        !formattedEndDate || formattedEndDate === formattedStartDate
+          ? formattedStartDate
+          : `${formattedStartDate} - ${formattedEndDate}`;
+
+      return (
+        <ListItem
+          onPress={onPress}
+          rightIcon={<Icon xml={arrowRight(colors.darkText)} />}
```
donni106, line 38: arrow icon should be primary color, like in all other lists

src/screens/ConstructionSiteOverviewScreen.js
```diff
@@ -0,0 +1,83 @@
...
+          subtitle={<RegularText>{formattedDates}</RegularText>}
+          title={<BoldText>{extendedTitle}</BoldText>}
```
donni106, line 40: to match existing list items, the first line of the list item should be a small normal text with the date and the second line the bold title.
currently it is switched.

|construction|event|
|-------------|-----|
|Bildschirmfoto 2021-01-08 um 12 42 33|Bildschirmfoto 2021-01-08 um 12 42 24|

src/components/widgets/ConstructionSiteWidget.tsx
```diff
@@ -0,0 +1,75 @@
...
+export const ConstructionSiteWidget = ({ navigation }: Props) => {
+  const { constructionSites, setConstructionSites } = useContext(ConstructionSiteContext);
+  const { isConnected, isMainserverUp } = useContext(NetworkContext);
+
+  const fetchPolicy = graphqlFetchPolicy({ isConnected, isMainserverUp, refreshTime: undefined });
+
+  const { data } = useQuery(getQuery(QUERY_TYPES.PUBLIC_JSON_FILE), {
+    variables: { name: 'constructionSites' },
+    fetchPolicy,
+    skip: !isConnected || !isMainserverUp
```
donni106, line 30: why do you added `skip` here? if the fetch policy is `cache-only` based on `isConnected` and `isMainserverUp` the query should work as expected, or am i wrong?
do we need to add that `skip` in every other query now?

src/ConstructionSiteProvider.tsx
```diff
@@ -0,0 +1,17 @@
...
+  constructionSites: ConstructionSite[];
+  setConstructionSites: React.Dispatch<React.SetStateAction<ConstructionSite[]>>;
+}>({ constructionSites: [], setConstructionSites: () => null });
+
+export const ConstructionSiteProvider = ({ children }: { children: React.ReactNode }) => {
+  const [constructionSites, setConstructionSites] = useState<ConstructionSite[]>([]);
+  return (
+    <ConstructionSiteContext.Provider value={{ constructionSites, setConstructionSites }}>
+      {children}
+    </ConstructionSiteContext.Provider>
```
donni106, line 15: this provider is added in order to async query the data and rerender the widget, am i right?

src/components/widgets/ConstructionSiteWidget.tsx
```diff
@@ -0,0 +1,75 @@
...
+        <View style={styles.container}>
+          <Icon style={styles.icon} xml={constructionSite(colors.primary)} />
+          <BoldText style={styles.count}>{constructionSites.length}</BoldText>
+        </View>
+        <RegularText style={styles.text}>Baustellen</RegularText>
+      </Wrapper>
+    </Touchable>
+  );
+};
+
```
donni106, line 58: for this file the same five points apply like commented in the [event widget PR](https://github.com/ikuseiGmbH/smart-village-app-app/pull/156/files#diff-80a14a3a3ec573140c4b751e0e10b89043deb97bca78ef4da931f08f17030e54)

src/config/consts.js
```diff
@@ -22,6 +22,8 @@ export const consts = {
 
   MATOMO_TRACKING: {
     SCREEN_VIEW: {
+      CONSTRUCTION_SITE_DETAIL: 'Construction site detail',
+      CONSTRUCTION_SITE_OVERVIEW: 'Construction site overview',
```
donni106, line 26: i think we should avoid "overview" and "detail", as we do not have that for other pages. maybe we can use some plural like "Constructions"? maybe we need some namespace like "Widget / ..." as well. was this string already discussed somehow? if not, we should add it to the meeting agenda. 

please add it to the matomo doc in confluence at the end.

src/helpers/jsonValidation.ts
```diff
@@ -0,0 +1,63 @@
+import _isArray from 'lodash/isArray';
+import _isObjectLike from 'lodash/isObjectLike';
+import { ConstructionSite } from '../types';
```
donni106, line 3: should be auto linted with a line break with our new eslint/prettier config, right?

src/helpers/jsonValidation.ts
```diff
@@ -0,0 +1,63 @@
...
+    isGeoLocationOrUndefined(location)
+  );
+};
+
+export const filterForValidConstructionSites = (json: unknown): ConstructionSite[] => {
+  if (!_isArray(json)) return [];
+  return json.filter((item) => {
+    if (!item) return false;
+    return checkConstructionSiteTypes(item);
+  });
```
donni106, line 62: there are several places in this file with close `return`s to each other, which is heavy to read. let us plan a call with @SoleCincis in order to find a way to improve the readability in this file?!

src/screens/ConstructionSiteDetailScreen.js
```diff
@@ -0,0 +1,162 @@
...
+  TitleContainer,
+  WebViewMap,
+  Wrapper,
+  WrapperWithOrientation
+} from '../components';
+import { arrowLeft, locationIconAnchor, location as locationIcon } from '../icons';
+import { useMatomoTrackScreenView } from '../hooks';
+import { ConstructionSiteContext } from '../ConstructionSiteProvider';
+import { momentFormat } from '../helpers';
+import moment from 'moment';
```
donni106, line 22: moment should be auto linted at the top block with our new eslint/prettier setup, right?

src/screens/ConstructionSiteDetailScreen.js
```diff
@@ -0,0 +1,162 @@
...
+      : `${formattedStartDate} ${startTime}`;
+  }
+
+  return formattedEndDate ? `${formattedStartDate} - ${formattedEndDate}` : formattedStartDate;
+};
+
+export const ConstructionSiteDetailScreen = ({ navigation }) => {
+  const { constructionSites } = useContext(ConstructionSiteContext);
+
+  const index = navigation.getParam('index');
```
donni106, line 45: this ensures, that the screen does not break if there is a navigation without index param? can this happen somehow? how can i test it?

src/screens/ConstructionSiteDetailScreen.js
```diff
@@ -0,0 +1,162 @@
...
+              <View style={styles.verticalPadding}>
+                <BoldText>Weitere Informationen: </BoldText>
+                <RegularText>{description}</RegularText>
+              </View>
+            )}
+            {!!restrictions?.length && (
+              <View style={styles.verticalPadding}>
+                <BoldText>Aktuelle Einschränkungen: </BoldText>
+                {restrictions.map((restriction) => (
+                  <RegularText key={'restriction-' + index}>- {restriction}</RegularText>
```
donni106, line 109: use template-string here like in other places `restriction-${index}`? is eslint complaining here?

src/screens/ConstructionSiteDetailScreen.js
```diff
@@ -0,0 +1,162 @@
...
+        </TouchableOpacity>
+      </View>
+    )
+  };
+};
+
+const styles = StyleSheet.create({
+  informationContainer: {
+    flexDirection: 'row',
+    flexWrap: 'wrap'
```
donni106, line 150: we could think about creating a styled-components wrapper for this container.
we already have a `WrapperWrap` with the following styles, which maybe works?

```
flex-direction: row;
flex-wrap: wrap;
justify-content: space-between;
width: 100%;
```

otherwise we can adjust/add things.

donni106 (changes requested):
please see the comments on code lines for refactorings and fixes.

src/navigation/AppStackNavigator.js
digorath: done in cd9815e1e0e6c8923c384d1382e3843d21f2bad3

src/screens/ConstructionSiteOverviewScreen.js
digorath: done in ea6d205ceae89ec2fc4ba6934cf94eb28fd31757

src/screens/ConstructionSiteOverviewScreen.js
digorath: done in f77e0d4bd76de4c86fe68e8d94782584b22209f8

src/components/widgets/ConstructionSiteWidget.tsx
digorath: you are right. the skip is unnecessary here, and would actually cause the cached version in the case of the user having no connection to not be used either.

i removed the skip in d4af0ca08d76b2fca85d3b9d39a6f148eeabf4c9

src/screens/ConstructionSiteOverviewScreen.js
digorath: fixed in 64f19e3bd5412b6013b5b80c9fc238d2e7d1253e

src/ConstructionSiteProvider.tsx
digorath: this provider is being added to not pass through the entire data objects as navigation params but instead to be able to work with identifiers/indices if needed.

src/components/widgets/ConstructionSiteWidget.tsx
digorath: i added a todo.

i will fix it as soon as the event widget is merged to the epic and i updated this branch with it

src/config/consts.js
digorath: we can discuss it in the meeting, as it was not discussed at any other place yet.

I will add it to the confluence site at the end.

src/helpers/jsonValidation.ts
digorath: the eslint import plugin is not merged yet due to issues with code climate.

i fixed it in 026308b2533032adc0542d2e374c7c51e81a27b4

src/screens/ConstructionSiteDetailScreen.js
digorath: a) i do not follow the first question. can you please rephrase?
b) eslint does not complain, should it?

src/screens/ConstructionSiteDetailScreen.js
digorath: i personally find it awkward having the information be pushed to the right edge of the screen/view due to it being space between.

besides that it should work.

i could add a 'spaceBetween' prop to WrapperWrap and use that prop everywhere but here.

What are your thoughts?

src/screens/ConstructionSiteDetailScreen.js
digorath: you can add a dummy element somewhere that lets you navigate to the screen with wrong props.

currently it should not be able to go wrong, because we have only one place where we navigate to the screen.

It could happen due to more navigations being added later on, or in the process of refactoring.

src/screens/ConstructionSiteDetailScreen.js
digorath: same as the other one. fixed it in 026308b2533032adc0542d2e374c7c51e81a27b4

src/screens/ConstructionSiteDetailScreen.js
donni106: do we want that for other routes as well?

src/screens/ConstructionSiteDetailScreen.js
digorath: changed it in 576046e1ec323823d7efca37960b378005f213d7

donni106: i thought i know messages from eslint about that. but maybe from another project.
so we are cool with having both options over the code base here, right?

src/screens/ConstructionSiteDetailScreen.js
donni106: yes sure, should not use `space-between` here, if this breaks the layout. that was not my intention.

src/helpers/jsonValidation.ts
donni106: sure, but we want proper grouped imports, even if CodeClimate has not answered with a solution yet

src/screens/ConstructionSiteOverviewScreen.js
donni106: new screenshot would be nice to not have to run it again on my side

donni106 (commented):
thanks! we will finish this PR after talking about some details tomorrow and on wednesday

src/screens/ConstructionSiteDetailScreen.js
digorath: :+1:

src/helpers/jsonValidation.ts
digorath: meeting is scheduled 💪

src/config/consts.js
donni106: Bildschirmfoto 2021-01-12 um 12 53 43

src/screens/ConstructionSiteDetailScreen.js
donni106: keep it with the `informationContainer` as is.
we do not want to update every occurrence of `WrapperWrap` now with figuring out a best way with props.
and `WrapperRow` is missing the `wrap` part, so there is no wrapper component existing, that can be used here, right?!

digorath: after i had now changed it i saw this comment, so you can decide whether or not to keep the last commit :-P

src/config/consts.js
digorath: :+1:

src/helpers/jsonValidation.ts
digorath: we can refactor it at a later date after the meeting. i suggest merging it as is for now

src/components/Wrapper.js
```diff
@@ -38,8 +44,13 @@ export const WrapperRow = styled.View`
...
 export const WrapperWrap = styled(WrapperRow)`
   flex-wrap: wrap;
-  justify-content: space-between;
   width: 100%;
+
+  ${(props) =>
+    props.spaceBetween &&
+    css`
+      justify-content: space-between;
+    `};
```
donni106, line 53: as `WrapperRow` already has a `spaceBetween` prop and `WrapperWrap` is based on it, the prop does not need to be declared here again.

src/screens/ConstructionSiteDetailScreen.js
donni106: https://github.com/ikuseiGmbH/smart-village-app-app/pull/157/commits/505a20527fa5051b0ce55ed3db9b7259464e5654

ok. than we keep your work for sure. these four places are fine i would say.

src/helpers/jsonValidation.ts
donni106: ok, i linked this discussion in my calendar event for tomorrow.

src/components/Wrapper.js
digorath: done in e7c8e34880bbaf13da77dedcb934806cd58899c5

donni106 (approved):

PR #160, opened 2021-01-13
fix(parsing public json content): catch errors on JSON.parse

donni106:
- added try/catch for every `JSON.parse` of random public json content from main server, which can have format issues, to ensure that the app does not crash
- used optional chaining to reduce possibilities of types for the returned value
  - if one of `data` or `data.publicJsonFile` or `data.publicJsonFile.content` is undefined, `JSON.parse` errors with a SyntaxError as well and we land in the catch -> main variable stays `[]`

SVA-133


digorath (approved):
code diff looks clean

PR #162, opened 2021-01-14
finish weather widget and weather screens

donni106:
- refactorings
- style adjustments
- fix for offline (polling)

please see the commit messages for detail

SVA-80

---

screens for new landscape mode after orientation change:

Simulator Screen Shot - iPhone 11 Pro Max - 2021-01-14 at 19 52 44-down

### DIFF
--- a/src/screens/WeatherScreen.js
+++ b/src/screens/WeatherScreen.js
@@ -1,24 +1,29 @@
 import React, { useContext } from 'react';
 import { useQuery } from 'react-apollo';
-import { FlatList, StyleSheet, TouchableOpacity, View } from 'react-native';
+import { ActivityIndicator, FlatList, StyleSheet, TouchableOpacity, View } from 'react-native';
 import { ScrollView } from 'react-native';
 
 import {
   DailyWeather,
   HourlyWeather,
   Icon,
+  LoadingContainer,
   SafeAreaViewFlex,
   Title,
   TitleContainer,
-  WeatherAlert
+  TitleShadow,
+  WeatherAlert,
+  WrapperWithOrientation
 } from '../components';
-import { colors, consts, normalize } from '../config';
+import { colors, consts, device, normalize, texts } from '../config';
 import { graphqlFetchPolicy } from '../helpers';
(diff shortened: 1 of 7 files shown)


src/screens/WeatherScreen.js
```diff
@@ -61,25 +75,30 @@ export const WeatherScreen = () => {
...
-            <TitleContainer>
-              <Title>Aktuelles Wetter</Title>
-            </TitleContainer>
+            <WrapperWithOrientation>
+              <TitleContainer accessibilityLabel={`${texts.weather.currentHeadline} (Überschrift)`}>
+                <Title>{texts.weather.currentHeadline}</Title>
+              </TitleContainer>
+              {device.platform === 'ios' && <TitleShadow />}
+            </WrapperWithOrientation>
             <FlatList
```
digorath, line 102: I think it is awkward to have the hourly weather be wider than the rest.

I personally would just wrap the whole screen in the wrapper with orientation. What do you think?

digorath (changes requested):
just a small thing. rest looks good

src/screens/WeatherScreen.js
donni106: it is weird as well to have something scrolling horizontally kind of endless, but cut at some point on the screen without visual ending/border. we would have white space left and right and in the middle there are elements horizontally scrolling but "magically" wrapped. one would swipe the elements into and from the nowhere. that is why we keep the device frame as visual borders.

for that reason, we also have a special case for the carousel, where the prev and next images will be visible already with some shading, to be able to cover the whole width, as the element is "endless"

was this understandable enough?
can you think of

digorath: i understand your point.

we can leave it as is then.

I want to point out though, that on your screenshot the cutoff still happens, just closer to the edge of the screen. I assume it has to do with some safe area inset.

digorath (approved):

src/screens/WeatherScreen.js
donni106: on devices with notch, yes. we cannot change that. but for devices without notch, it is the device edge.

PR #163, opened 2021-01-15
fix(map): hide POI map in offline

digorath:

- linting changes in LocationOverview.tsx
- hide the map for pois in offline mode, to prevent rendering a white rectangle

### DIFF
--- a/src/components/screens/PointOfInterest.js
+++ b/src/components/screens/PointOfInterest.js
@@ -18,12 +18,14 @@ import { useMatomoTrackScreenView } from '../../hooks';
 import { matomoTrackingString } from '../../helpers';
 import { WebViewMap } from '../map/WebViewMap';
 import { location, locationIconAnchor } from '../../icons';
+import { NetworkContext } from '../../NetworkProvider';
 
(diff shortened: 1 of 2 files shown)


src/components/screens/PointOfInterest.js
```diff
@@ -148,7 +150,7 @@ export const PointOfInterest = ({ data, hideMap, navigation }) => {
           </View>
         )}
 
-        {!hideMap && !!latitude && !!longitude && (
+        {!hideMap && !!latitude && !!longitude && isConnected && isMainserverUp && (
```
donni106, line 153: i think `isConnected` is enough?
the map does not rely on our mainserver, or am i wrong?
if there are `latitude` and `longitude`, we can get a map if online, right?

digorath: added a comment to this in the code, that explains the value of including the check for `isMainserverUp`

donni106: we keep both. but both checks are not enough 100%, because the openstreetmap server can be down as well...

donni106:
✅

PR #164, opened 2021-01-19
customizable image component format

donni106:
refactor(image aspect ratio): move default aspect ratio to consts

- changed the ratio to 2:1, because 360:180 is the same as 2:1
- having the image aspect ratio in a central place, we can edit the value on one place, instead of searching for the method

feat(image aspect ratio): set IMAGE_ASPECT_RATIO with value from global settings

- added setter for `IMAGE_ASPECT_RATIO` in src/index.js if `imageAspectRatio` is present in global settings
  - added method for parsing and checking the value with returning the defaults, if something is not expected
- added comment in `consts`, hinting that the value can differ
- moved reading the value from `consts` into `imageHeight` to get the correct updated value when calling the method

feat(image aspect ratio): add alternative for home carousel

- added param for `imageHeight` method to be able to pass an alternative aspect ratio
  - fallback is the default aspect ratio if none is given
- added prop `aspectRatio` for `ImagesCarousel` to pass an alternative aspect ratio
  - used `aspectRatio` to calculate the height per `imageHeight` method
- used the new prop in `HomeCarousel` to pass an alternative aspect ratio, that can be set per main server global settings
- refactored return in `Image` to reduce duplication with moving the condition for image rights in the render part
- refactored default image styles in `Image` to ensure the correct aspect ratio
  - we need to call the default styles in a method to ensure correct defaults for image aspect ratio, which could be overwritten bei server global settings. otherwise (as default prop) the style would be set before the overwriting occurred.

refactor(map aspect ratio): match given aspect ratio for images

- adjust aspect ratio for map, because it should have the same aspect ratio as images

SVA-135

---

On the main server there can be the following addition in the global settings:

```
"imageAspectRatio": "2:1",
"homeCarousel": {
  "imageAspectRatio": "1:1"
}
```

which will result in:

Simulator Screen Shot - iPhone 11 Pro Max - 2021-01-19 at 15 19 39

furthermore, here are examples for different aspect ratios for home and rest:

Simulator Screen Shot - iPhone 11 Pro Max - 2021-01-19 at 19 06 04-side
Simulator Screen Shot - iPhone 11 Pro Max - 2021-01-19 at 19 06 26-side

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -15,7 +15,7 @@ import _isEmpty from 'lodash/isEmpty';
 
 import { auth } from './auth';
 import { colors, consts, device, namespace, secrets, texts } from './config';
-import { graphqlFetchPolicy, storageHelper } from './helpers';
+import { graphqlFetchPolicy, parsedImageAspectRatio, storageHelper } from './helpers';
 import { getQuery, QUERY_TYPES } from './queries';
 import { NetworkProvider } from './NetworkProvider';
 import NetInfo from './NetInfo';
@@ -297,6 +297,10 @@ const MainAppWithApolloProvider = () => {
 
   if (initialGlobalSettings.navigation === consts.TABS) {
     AppContainer = createAppContainer(MainTabNavigator);
+  }
+
+  if (initialGlobalSettings.imageAspectRatio) {
+    consts.IMAGE_ASPECT_RATIO = parsedImageAspectRatio(initialGlobalSettings.imageAspectRatio);
   }
 
   return (
--- a/src/components/Image.js
+++ b/src/components/Image.js
@@ -1,6 +1,6 @@
 import PropTypes from 'prop-types';
 import React, { useContext, useEffect, useState } from 'react';
-import { ActivityIndicator, View } from 'react-native';
+import { ActivityIndicator, StyleSheet } from 'react-native';
 import { CacheManager } from 'react-native-expo-image-cache';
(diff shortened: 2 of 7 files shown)


src/index.js
```diff
@@ -299,6 +299,10 @@ const MainAppWithApolloProvider = () => {
     AppContainer = createAppContainer(MainTabNavigator);
   }
 
+  if (initialGlobalSettings.imageAspectRatio) {
+    consts.IMAGE_ASPECT_RATIO = parsedImageAspectRatio(initialGlobalSettings.imageAspectRatio);
+  }
```
digorath, line 304: I would have expected this to be somewhere near line 192, where we set the settings depending on the content of the json file. what do you think?

digorath (changes requested):

src/index.js
donni106: it is also working there. but i put it down here, because we already do something conditionally based on keys from global settings here (preparing the right navigation). so i decided to set the image aspect ratio from the settings key here as well.
around 192 there are setups for the app in general. in that case processing the server data for global settings and saving it to the local storage. i found it too detailed and out of context, to have a setter for a detailed functionality there.
we should wait for setting up the settings entirely (line 255) before working with the values. thats why i landed in 302.

digorath: ok :+1:

digorath (approved):

PR #165, opened 2021-01-20
fix(image loading spinner android): update react-native-elements from 1.1.0 to 1.2.7

donni106:
- problem: https://github.com/react-native-elements/react-native-elements/issues/2000
- latest v1 is 1.2.7: https://github.com/react-native-elements/react-native-elements/blob/next/CHANGELOG.md#v127
- when updating like this, we needed to make changes where using `Image`
  - not all styles will be applied anymore, you need to change styles to props
  - we had the need of moving `resizeMode` and `borderRadius` from inside of style prop
    to be props itself
    - default for `resizeMode` was `cover` before in the code of react-native-elements,
      as it is the default for React Native `Image` component https://reactnative.dev/docs/image#resizemode
  - refactored all occurrences

SVA-141

### DIFF
--- a/src/screens/ServiceScreen.js
+++ b/src/screens/ServiceScreen.js
@@ -135,13 +135,13 @@ export const ServiceScreen = ({ navigation }) => {
                                   source={{ uri: item.icon }}
                                   style={styles.serviceImage}
                                   PlaceholderContent={null}
+                                  resizeMode="contain"
                                 />
                               )}
                               <BoldText
                                 small
                                 primary
                                 center
-                                accessible
                                 accessibilityLabel={`${item.title} (Taste)`}
                               >
                                 {item.title}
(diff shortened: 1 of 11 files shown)


src/screens/ServiceScreen.js
```diff
@@ -170,8 +170,7 @@ const styles = StyleSheet.create({
   serviceImage: {
     alignSelf: 'center',
     height: normalize(40),
-    marginBottom: normalize(7),
-    resizeMode: 'contain',
+    marginBottom: normalize(17),
```
digorath, line 173: Was this increase of the margin intended?

digorath (commented):
just one thing to check.
everything else looks fine.

src/screens/ServiceScreen.js
donni106: no! i have tested around. very good catch!

digorath (approved):
looks good now

PR #166, opened 2021-01-21
fix(safe area insets): upgrade safe area view

digorath:
forcefully upgraded react-native-safe-area-view to support iPhone 12

Before

Screenshot 2021-01-21 at 16 24 26

After

Screenshot 2021-01-21 at 16 23 19

SVA-145

### DIFF
--- a/package.json
+++ b/package.json
@@ -89,5 +89,8 @@
     "react-test-renderer": "16.11.0",
     "set-value": "^2.0.1",
     "typescript": "~4.0.0"
+  },
+  "resolutions": {
+    "react-native-safe-area-view": "2.0.0"
   }
 }
--- a/src/index.js
+++ b/src/index.js
@@ -13,6 +13,7 @@ import { InMemoryCache } from 'apollo-cache-inmemory';
 import { persistCache } from 'apollo-cache-persist';
 import _reduce from 'lodash/reduce';
 import _isEmpty from 'lodash/isEmpty';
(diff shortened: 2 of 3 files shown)


donni106:
i have tested simulators

|iPhone 11 Pro Max - 13.0|iPhone 5s - 11.0.1|iPad Pro (12.9-inch) (3rd generation)|
|---|---|---|
|:white_check_mark:|:white_check_mark:|:white_check_mark:|

and real devices

|iPhone 6 - 12.4.9|Android 7 Phone|Android 9 Tablet|
|---|---|---|
|:white_check_mark:|:white_check_mark:|:white_check_mark:|

i could not find any problems with related to safe area.

donni106 (approved):

digorath:
Tested on simulators/emulators:

| iPhone 12 Pro | iPhone 12 Pro Max | Android 11 Pixel 4 |
|---|---|---|
| ✅ | ✅  | ✅  |

Tested on physical devices:
| iPhone  | Android 10 |
|---|---|
| ✅ | ✅  |

on the android 11 pixel 4 emulator the top navbar was too wide
Screenshot 2021-01-22 at 11 13 37

That might be emulator related, or might be an real issue. it is the same in master already and is not affected by the update.

donni106:
Android 11 Pixel 4 emulator - before vs. after:

Screenshot_1611315007-side

is think we can tolerate the difference for now.
the iPhone 12 problem was more significant.

PR #168, opened 2021-01-21
fix(images): correct rendering of images

donni106:
- image sizes depend on orientation of the device,
  therefore we already had a method to set correct widths
  - this logics are now placed in `imageWidth` to be used elsewhere
    if necessary
- with updated react-native-elements image component the `alignSelf` was
  missing in some cases
  - this needs to be a `containerStyle` with the newer version
- adjusted every usage of `Image`
- the map also needed adjustments concerning height & width

---

i hope i reached every code.
i think it behaves again like expected and in versions before 1.5.1.

Simulator Screen Shot - iPhone 11 Pro Max - 2021-01-21 at 18 45 43-side

Simulator Screen Shot - iPhone 11 Pro Max - 2021-01-21 at 18 45 48-down

---

refactor(image width): use min of dimensions height and width

- we could simplify the logic for correct image width with using the min value of dimensions height and width
  - this made it possible to get rid of `orientation` and `dimensions` props in several components without changing render results
  - this also made it possible to get rid of `OrientationContext` in some components

### DIFF
--- a/src/components/map/WebViewMap.tsx
+++ b/src/components/map/WebViewMap.tsx
@@ -1,10 +1,9 @@
-import React, { useCallback, useContext } from 'react';
+import React, { useCallback } from 'react';
 import { StyleProp, StyleSheet, View, ViewStyle } from 'react-native';
 import { MapMarker, WebViewLeaflet, WebviewLeafletMessage } from 'react-native-webview-leaflet';
 
 import { colors } from '../../config';
 import { imageHeight, imageWidth } from '../../helpers';
-import { OrientationContext } from '../../OrientationProvider';
 
 type Props = {
   locations?: MapMarker[];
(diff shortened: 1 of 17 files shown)


src/components/map/WebViewMap.tsx
```diff
@@ -57,12 +59,15 @@ export const WebViewMap = ({
 // we need to call the default styles in a method to ensure correct defaults for image aspect ratio,
 // which could be overwritten bei server global settings. otherwise (as default prop) the style
 // would be set before the overwriting occurred.
-const stylesForMap = () =>
-  StyleSheet.create({
+const stylesForMap = (orientation: any, dimensions: any) => {
```
donni106, line 62: what can we type instead of `any` here?

digorath (changes requested):
You can also get rid of the dimensions/orientation prop in CardListItem entirely.

The rest looks fine.

digorath (approved):
👍

PR #170, opened 2021-01-26
Render missing web urls

donni106:
feat(InfoCard): merge all available web urls to render

- added attributes to queries to request all necessary data
- created a method to merge all available web urls to render in `InfoCard` for events, points of interest and tours

feat(web urls): filter descriptions starting with "url"

- added a check for descriptions starting with "url" because we do not want to present web urls behind an "url" or "urlInformation" or "urlVideo" or "urlVideoPreview"
  - in that case, we render the just the `url` instead of the `description`

SVA-146 (_please see comment in the ticket for screenshots_)

### DIFF
--- a/src/components/screens/InfoCard.js
+++ b/src/components/screens/InfoCard.js
@@ -175,8 +175,11 @@ export const InfoCard = ({ addresses, category, contact, contacts, webUrls, open
             } (Taste) (Öffnet Webseite in der aktuellen App)`}
             onPress={() => openLink(url, openWebScreen)}
           >
-            {!description && <RegularText primary>{url}</RegularText>}
-            {!!description && <RegularText primary>{description}</RegularText>}
+            {!description || !!description?.startsWith('url') ? (
+              <RegularText primary>{url}</RegularText>
+            ) : (
+              <RegularText primary>{description}</RegularText>
+            )}
           </TouchableOpacity>
         </InfoBox>
       );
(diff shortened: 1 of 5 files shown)


src/components/screens/InfoCard.js
```diff
@@ -67,6 +67,24 @@ const contactView = (contact) => (
   </View>
 );
 
+const mergeWebUrls = ({ webUrls, contact, contacts }) => {
+  const mergedWebUrls = [...webUrls];
```
digorath, line 71: is `webUrls` ensured to not be `undefined` here?
trying to spread `undefined` causes an error.

you can avoid it with
```
const mergedWebUrls = [...(webUrls ?? [])];
```
or with
```
const mergedWebUrls = webUrls ? [...webUrls] : [];
```
of which i prefer the second version for readability reasons.

digorath (changes requested):
One small thing to fix.

src/components/screens/InfoCard.js
donni106: done in https://github.com/ikuseiGmbH/smart-village-app-app/pull/170/commits/d919904fe6442110d647dfd00b57a9d280e445f2

digorath (approved):

SoleCincis (approved):

src/queries/eventRecords.js
```diff
@@ -102,6 +102,22 @@ export const GET_EVENT_RECORDS_AND_CATEGORIES = gql`
...
+        fax
+        webUrls {
+          url
+          description
+        }
+      }
+      webUrls: urls {
+        url
+        description
+      }
```
donni106, line 120: we missed someting:
https://github.com/ikuseiGmbH/smart-village-app-app/pull/195/files

PR #171, opened 2021-01-27
Add form as route for drawer

donni106:
- added Form to AppStackNavigator to be able to navigate to the form screen with drawer mode apps

### DIFF
--- a/src/navigation/AppStackNavigator.js
+++ b/src/navigation/AppStackNavigator.js
@@ -6,6 +6,7 @@ import {
   ConstructionSiteDetailScreen,
   ConstructionSiteOverviewScreen,
   DetailScreen,
+  FormScreen,
   HomeScreen,
   HtmlScreen,
   IndexScreen,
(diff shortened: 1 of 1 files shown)


digorath (commented):
How will the behaviour be for older app versions if we make use of that? Whether or not we use this is handled server sided, right?

donni106:
> How will the behaviour be for older app versions if we make use of that? Whether or not we use this is handled server sided, right?

good question! older app versions will see the navigation entry, because the navigation json is unversioned. when clicking an entry with unknown route screen, the navigation closes and nothing more happens.

digorath (approved):

donni106:
> > How will the behaviour be for older app versions if we make use of that? Whether or not we use this is handled server sided, right?
> 
> good question! older app versions will see the navigation entry, because the navigation json is unversioned. when clicking an entry with unknown route screen, the navigation closes and nothing more happens.

@digorath and i had a good call about versioning of static contents from server and i opened a new ticket SVA-150.

SoleCincis (approved):

PR #172, opened 2021-01-27
Switch rendering of tours and points of interest

donni106:
- switched the sections for the list on the category list screen, to have points of interest first and tours last
- renamed every "Touren und Orte" to "Orte und Touren" because we show them in that order on the category screen now


digorath (approved):

digorath:
the `noBookmarksYet` text could be aligned with it as well.

SoleCincis (approved):

PR #174, opened 2021-01-27
Optimize images carousel

donni106:
chore(home carousel): add pull to refresh

- add missing `refreshing` prop from home screen to force refetch of data

refactor(images carousel): optimize code for carousel items

- created `ImagesCarouselItem` from `TouchableImage`, which returns a simple image or a touchable image that can navigate to a given `routeName` with `params`
- reduced the conditions for return of `renderItem` in `ImagesCarousel` with the result, that there is always a return of `ImagesCarouselItem` whether with extra query or not
  - added optional chaining for the remained condition to shorten the checks
- reduced the conditions in `renderItem` for `itemWidth` with using the  new `imageWidth()` helper method, that is taking care of the correct  size
- added `useCallback` for `renderItem` in `ImagesCarousel` to avoid unintended re-renders
- added early return of `null` if `item` is `undefined` to avoid errors in further context
- avoid rendering a carousel if there is just one element to show
  - if there is one entry in the data, we do not want to render a whole carousel, we than just need the one item to render

SVA-81

### DIFF
--- a/src/components/ImagesCarousel.js
+++ b/src/components/ImagesCarousel.js
@@ -1,61 +1,30 @@
 import PropTypes from 'prop-types';
 import Carousel from 'react-native-snap-carousel';
-import React, { useContext } from 'react';
-import { ActivityIndicator, StyleSheet, TouchableOpacity } from 'react-native';
+import React, { useCallback, useContext } from 'react';
+import { ActivityIndicator, StyleSheet } from 'react-native';
 import { Query } from 'react-apollo';
 
 import { colors } from '../config';
-import { shareMessage } from '../helpers';
+import { imageWidth, shareMessage } from '../helpers';
 import { getQuery } from '../queries';
-import { Image } from './Image';
+import { ImagesCarouselItem } from './ImagesCarouselItem';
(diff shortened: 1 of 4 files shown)


digorath (approved):
Looks good to me.

I was thinking if there is an easy way to get rid of the duplication of the return statement in ImagesCarousel.js (inside the Query tag and right below it) using `useQuery` but i dont see an obvious way to do that without restructuring the whole component.

If you can think of an elegant way, it could be worthwhile to just do that as well in this PR, otherwise we can just merge this as an improvement already.

SoleCincis (approved):

donni106:
> Looks good to me.
> 
> I was thinking if there is an easy way to get rid of the duplication of the return statement in ImagesCarousel.js (inside the Query tag and right below it) using `useQuery` but i dont see an obvious way to do that without restructuring the whole component.
> 
> If you can think of an elegant way, it could be worthwhile to just do that as well in this PR, otherwise we can just merge this as an improvement already.

we cannot make it now to restructure the component, so we go with this current improvments. i add a link to your proposal in SVA-109 @digorath.

PR #175, opened 2021-01-28
Show message on slider item

donni106:
feat(images carousel): render an items message onto an image

- created `ImageMessage` for rendering a `message` in a `HtmlView` absolutely positioned on the bottom left with some spacing to the container edge
- added `message` prop to `ImagesCarouselItem` to pass through to `Image`
- wrapped `ImagesCarouselItem` with `memo` to reduce re-renderings
  - tested it with console logs, that were decreased
- added `message` prop to `Image` and render `ImageMessage` if `message` is present

SVA-81

---

1. `"message": "<h1 style=\"color: #ffffff; margin-bottom: 0\">nur Januar</h1><h2 style=\"color: #ffffff\">ganztägig</h2>"`
2. `"message": "<div style=\"background-color: rgba(20, 20, 20, 0.4); padding: 4px\"><p style=\"background-color: transparent; margin-bottom: 0px; color: #ffffff\">immer, 10:00 - 12:00 und 14:00 - 16:00</p></div>"`

Simulator Screen Shot - iPhone 11 Pro Max - 2021-01-28 at 13 55 53-side

### DIFF
--- a/src/components/ImageMessage.js
+++ b/src/components/ImageMessage.js
@@ -1,0 +1,24 @@
+import PropTypes from 'prop-types';
+import React from 'react';
+import { StyleSheet, View } from 'react-native';
+
+import { normalize } from '../config';
+import { HtmlView } from './HtmlView';
+
+export const ImageMessage = ({ message }) => (
+  <View style={styles.containerStyle}>
+    <HtmlView html={message} />
+  </View>
+);
+
+const styles = StyleSheet.create({
+  containerStyle: {
+    bottom: normalize(24),
+    position: 'absolute',
+    left: normalize(14)
+  }
+});
+
+ImageMessage.propTypes = {
+  message: PropTypes.string.isRequired
+};
--- a/src/components/Image.js
+++ b/src/components/Image.js
@@ -7,10 +7,12 @@ import { Image as RNEImage } from 'react-native-elements';
 import { colors } from '../config';
 import { imageHeight, imageWidth } from '../helpers';
 import { SettingsContext } from '../SettingsProvider';
+import { ImageMessage } from './ImageMessage';
 import { ImageRights } from './ImageRights';
 
 export const Image = ({
   source,
+  message,
   style,
   containerStyle,
   PlaceholderContent,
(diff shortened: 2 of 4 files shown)


src/components/ImageMessage.js
```diff
@@ -0,0 +1,25 @@
...
+export const ImageMessage = ({ message }) => (
+  <View style={styles.containerStyle}>
+    <HtmlView html={message} />
+  </View>
+);
+
+const styles = StyleSheet.create({
+  containerStyle: {
+    bottom: normalize(24),
+    paddingHorizontal: normalize(4),
```
digorath, line 17: what does the paddingHorizontal do, that you cannot already do with increasing `bottom` to `normalize(28)` in this case?

digorath (commented):
Code looks fine. It works fine. 👍 

Just one small question that might entail a minor change.

SoleCincis (approved):

src/components/ImageMessage.js
donni106: `horizontal` is "left-right", so a spacing from the container edges left and right. as you can see in the second example, this can be useful when having background color. sure, one can add that as well with html and nested divs, but we can provide is here already to not have to experiment too much in the html.

digorath: uhm.. ye. left right xD
question still stands with `left: normalize(...)`

but this means that we have a little bit earlier line break for long texts right?

donni106: if we increase the left, this does not affect the inner content spacing. so the background color and the text would start on the same pixel.

digorath: but we have no background color, am i wrong?

donni106: we can have like in example 2. and i think it will be a common use case, as just text on images are hard to read.
but yes, we should outsource it to the one who is adding the content.
as i  can see, i also added padding in the background div container haha `<div style=\"background-color: rgba(20, 20, 20, 0.4); padding: 4px\">` so the `paddingHorizontal` here could be removed, ok.

donni106: done with https://github.com/ikuseiGmbH/smart-village-app-app/commit/e3b37da961f680deae9a4f7661148e6741f389f5

digorath (approved):

PR #176, opened 2021-01-28
Show just in time slider items

donni106:
- created date and time helper methods to identify items to show based on their `dates` object with or without given date and/or time periods
  - used `moment` for easier working with date and time values
- added a filter for `carouselData` in `ImagesCarousel` for present items and items with active date/time periods
  - this made the check for a present `item` in `renderItem` superfluous
- renamed `carouselImages` to `homeCarouselData` in `HomeCarousel`, because it is more on point as the data is not only about images anymore

SVA-81

---

Any ideas on optimizing logics or readability of the date and time conditions?

### DIFF
--- a/babel.config.js
+++ b/babel.config.js
@@ -1,6 +1,7 @@
-module.exports = function(api) {
+module.exports = function (api) {
   api.cache(true);
   return {
     presets: ['babel-preset-expo'],
+    plugins: ['@babel/plugin-proposal-logical-assignment-operators']
   };
 };
--- a/package.json
+++ b/package.json
@@ -67,6 +67,7 @@
     "styled-components": "~4.3.2"
   },
   "devDependencies": {
+    "@babel/plugin-proposal-logical-assignment-operators": "^7.12.1",
     "@types/lodash": "^4.14.165",
     "@types/react": "~16.9.35",
     "@types/react-dom": "~16.9.8",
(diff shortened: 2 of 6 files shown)


digorath (commented):
what do you think of

```
export const isActive2 = (item) => {
  const { dates } = item;

  let active = true;

  if (dates?.length) {
    const now = moment();

    dates.forEach((date) => {
      const { dateStart, dateEnd, timeStart, timeEnd } = date;

      active = !dateStart || moment(dateStart).startOf('day').isBefore(now);

      active &&= !dateEnd || moment(dateEnd).endOf('day').isAfter(now);

      active &&= !timeStart || moment(timeStart, 'HH:mm').isBefore(now, 'HH:mm');

      active &&= !timeEnd || moment(timeEnd, 'HH:mm').isAfter(now, 'HH:mm');
    });

    return active;
  }
  return active;
};
```

for that to work we would need to enable logical assignements (`&&=`) or replace them by `= active && (...)`

I personally find it a bit harder to follow with start and end times being used together in the expressions instead of checking for starting conditions and ending conditions seperately.

i feel like the conditions are less clear to read that way at least.

donni106:
> what do you think of
> ...

very good, thanks! this should cover all the cases with less code in a more readable way. i will integrate it, test it and update the PR.

SoleCincis (approved):

digorath (approved):

Issue #167, opened 2021-01-21
some icons on landscape iPhones are cut on some devices


donni106:
![Simulator Screen Shot - iPhone 11 Pro Max - 2021-01-21 at 16 45 41-down](https://user-images.githubusercontent.com/1942953/105375366-f1bb2680-5c08-11eb-9753-50497f375dba.png)

SoleCincis:
### After some test on available devices I can conclude that this bug is present on landscape screens with a height between  67.3 mm - 77.8 mm

iPhone SE ( 4-inch, horizontal height 58.6 mm, width 123.8 mm - smallest screen that I can test on )  
(!) **No issues** 
![IMG_20E83162DFB6-1](https://user-images.githubusercontent.com/43641321/106285222-80045d80-6244-11eb-9c3c-b2ab915ce116.jpeg)
iPhone SE 2nd generation  (  4.7-inch, horizontal height 67.3 mm, width 138.4 mm ) 
**slightly cropped icon**
![Simulator Screen Shot - iPhone SE (2nd generation) - 2021-01-29 at 15 44 47](https://user-images.githubusercontent.com/43641321/106291105-81855400-624b-11eb-8158-701facfc4edc.png)
iPhone 8 ( 4.7-inch, horizontal height 67.3 mm, width 138.4 mm ) 
**slightly cropped icon**
![iPhone 8 - ](https://user-images.githubusercontent.com/43641321/106285996-729ba300-6245-11eb-8732-a421b52733cd.png)
iPhone 8Plus ( 5.5-inch, horizontal height 78.1 mm, width 158.4 mm ) 
**partially cropped icon**
![Simulator Screen Shot - iPhone 8 Plus - 2021-01-29 at 15 21 03](https://user-images.githubusercontent.com/43641321/106286306-c9a17800-6245-11eb-9cec-dc245485c9a4.png)
 iPhone 11 Pro Max ( 6.5-inch, horizontal height 77.8 mm,  width 158.0 mm ) 
**partially cropped icon**
![Simulator Screen Shot - iPhone 11 Pro Max - 2021-01-29 at 15 23 45](https://user-images.githubusercontent.com/43641321/106291760-50595380-624c-11eb-9ad2-0542078a46ac.png)
iPad Pro (9.7-inch, horizontal height 169.5 mm, width 240 mm - smallest iPad screen that I can test on)
(!) **No issues** 
[...]

PR #178, opened 2021-01-29
fix(share icon and settings icon) removed paddingVertical

SoleCincis:
On IOS devices : 
- paddingVertical was apparently unnecessary and created a bug
- without it the bug does not occur and there is no break in all other devices
- added a new prop hitSlop. As the removed paddings hitSlop defines how far a touch event can start away from the view
  - resulting in a "smooth touch" experience
 
#167

Has to be tested on Android devices

### DIFF
--- a/src/components/Icon.js
+++ b/src/components/Icon.js
@@ -16,7 +16,7 @@ export const Icon = ({ xml, width, height, name, size, focused, iconColor, style
   const color = iconColor || (focused ? colors.accent : colors.primary);
 
   return (
-    <View style={style}>
+    <View style={style} hitSlop={{ top: 12, bottom: 12 }}>
       {xml && <SvgXml xml={xml} width={width} height={height} style={iconStyle} />}
       {name && <Ionicons name={name} size={size} color={color} style={iconStyle} />}
     </View>
--- a/src/screens/BB-BUS/DetailScreen.js
+++ b/src/screens/BB-BUS/DetailScreen.js
@@ -385,13 +385,11 @@ const styles = StyleSheet.create({
   },
   iconLeft: {
     paddingLeft: normalize(14),
-    paddingRight: normalize(7),
-    paddingVertical: normalize(4)
+    paddingRight: normalize(7)
   },
   iconRight: {
     paddingLeft: normalize(7),
-    paddingRight: normalize(14),
-    paddingVertical: normalize(4)
+    paddingRight: normalize(14)
   },
   sectionTitle: {
     backgroundColor: colors.transparent,
--- a/src/screens/BookmarkScreen.js
+++ b/src/screens/BookmarkScreen.js
@@ -121,13 +121,11 @@ const styles = StyleSheet.create({
   },
   iconLeft: {
     paddingLeft: normalize(14),
-    paddingRight: normalize(7),
-    paddingVertical: normalize(4)
+    paddingRight: normalize(7)
   },
   iconRight: {
     paddingLeft: normalize(7),
-    paddingRight: normalize(14),
-    paddingVertical: normalize(4)
+    paddingRight: normalize(14)
   }
 });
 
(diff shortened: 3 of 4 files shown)


SoleCincis:
**Bug fix example on iPhone 11 Pro Max** 
Simulator Screen Shot - iPhone 11 Pro Max - 2021-01-29 at 19 12 10 Simulator Screen Shot - iPhone 11 Pro Max - 2021-01-29 at 19 12 31

SoleCincis:
iPad Pro (12.9-inch) (4th generation): example of the non-breaking on device that did not have the bug
Simulator Screen Shot - iPad Pro (12 9-inch) (4th generation) - 2021-01-29 at 19 10 46
Simulator Screen Shot - iPad Pro (12 9-inch) (4th generation) - 2021-01-29 at 19 10 46

digorath (commented):
First:
* It looks fine under android as well.
* Code changes look fine

Second:
I added the paddings as I felt that the icons where hard to click, especially under android. Since the padding causes problems it has to go, but maybe we should add some top and bottom hitslop to the icons instead?

SoleCincis:
> First:
> 
> * It looks fine under android as well.
> * Code changes look fine
> 
> Second:
> I added the paddings as I felt that the icons where hard to click, especially under android. Since the padding causes problems it has to go, but maybe we should add some top and bottom hitslop to the icons instead?

I'm playing around with something like this,  Is this what you had in mind ? 

Screenshot 2021-02-01 at 14 23 06

src/components/Icon.js
```diff
@@ -16,7 +16,7 @@ export const Icon = ({ xml, width, height, name, size, focused, iconColor, style
   const color = iconColor || (focused ? colors.accent : colors.primary);
 
   return (
-    <View style={style}>
+    <View style={style} hitSlop={{ top: 12, bottom: 12 }}>
```
donni106, line 19: does this work for everywhere we use `Icon`? or are there maybe overlapping click areas somewhere?

SoleCincis: the result of my tests is positive, meaning on the smallest screen (at my disposal) iPhone SE, there are no overlaps. However, I have problems in making tests using the secrets of other cities, I cannot see some sections, such as the service area on the home screen, which is essential for tests this new feature . In more simple words if use Bad Belzig secrets I don't see the same public app , I can't see Bad Belzig data but for example service area is not visible. 
What am I doing wrong ?

donni106: don't know from that description. changing secrets should work live and direct and the app refreshes a few times.
if not, stop yarn, change secrets, start yarn.
pull to refresh can also be a fix, when it seems to show old data.

SoleCincis: sorry it is a really bad description .

digorath (approved):

src/components/Icon.js
donni106: your or mine? :D can i do or tell something more?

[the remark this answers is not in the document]
SoleCincis: it is apparently a prop of view : https://reactnative.dev/docs/view#hitslop

SoleCincis: > the result of my tests is positive, meaning on the smallest screen (at my disposal) iPhone SE, there are no overlaps. However, I have problems using the secrets of different cities, I cannot see some sections, such as the service area on the home screen, which is essential for tests this new feature . In more simple words if use Bad Belzig secrets I don't see the same public app , I can see Bad Belzig data but for example service area is not visible.
> What am I doing wrong ?

mine was horrible... ! 

I think it time to merge this branch , what do you think ? :D

donni106: ok

donni106 (approved):

src/components/Icon.js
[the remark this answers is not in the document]
digorath: yes it is, the question was if that additional hitslop for the internal element has any interaction with the parent element.
i.e. does the parent element register for touch events of the children like that, or does it only register touch events on its own occupied area (increased its own hitslop).
and the answer is: even if the hitslop of the child element exceeds the bounds of the parent element, the parent element registers touch events on it.

PR #182, opened 2021-02-03
Added logo to POIs

digorath:

The logo of the data provider is now visible on POIs in the same way as it is for tours.

Screenshot 2021-02-03 at 12 20 05

---

SVA-159

### DIFF
--- a/src/components/screens/PointOfInterest.js
+++ b/src/components/screens/PointOfInterest.js
@@ -1,24 +1,25 @@
+import _filter from 'lodash/filter';
 import PropTypes from 'prop-types';
 import React, { useContext } from 'react';
 import { StyleSheet, View } from 'react-native';
-import _filter from 'lodash/filter';
 
 import { colors, consts, device, texts } from '../../config';
-import { HtmlView } from '../HtmlView';
-import { Image } from '../Image';
-import { Title, TitleContainer, TitleShadow } from '../Title';
-import { Wrapper, WrapperWithOrientation } from '../Wrapper';
-import { PriceCard } from './PriceCard';
-import { InfoCard } from './InfoCard';
-import { OperatingCompanyInfo } from './OperatingCompanyInfo';
-import { OpeningTimesCard } from './OpeningTimesCard';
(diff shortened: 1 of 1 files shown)


src/components/screens/PointOfInterest.js
```diff
@@ -55,6 +56,8 @@ export const PointOfInterest = ({ data, hideMap, navigation }) => {
         rootRouteName
       }
     });
+
+  const logo = dataProvider && dataProvider.logo && dataProvider.logo.url;
```
donni106, line 60: this was intended copy & paste from tours instead of refactoring to optional chaining at the same time?

src/components/screens/PointOfInterest.js
```diff
@@ -1,24 +1,25 @@
...
 import { OpeningTimesCard } from './OpeningTimesCard';
-import { ImagesCarousel } from '../ImagesCarousel';
-import { useMatomoTrackScreenView } from '../../hooks';
-import { matomoTrackingString } from '../../helpers';
-import { WebViewMap } from '../map/WebViewMap';
-import { location, locationIconAnchor } from '../../icons';
-import { NetworkContext } from '../../NetworkProvider';
-import { TMBNotice } from '../TMB/Notice';
+import { OperatingCompanyInfo } from './OperatingCompanyInfo';
+import { PriceCard } from './PriceCard';
```
donni106, line 22: imports restructured because of eslint import plugin?
do we want that in this commit or maybe do not stash this part and fix all imports over the code base at all with one branch?

digorath: imports restructured because i noticed they are messy. not a plugin, just vscode ;)

i did it in this file as i touched the file anyways.

src/components/screens/PointOfInterest.js
digorath: i wanted to touch as little as possible, as both components are really similar and should be refactored to reduce code duplication.

this way it remains easier to see the similarities between the files when it comes to that.

i did not want to touch both files.

donni106 (approved):

PR #183, opened 2021-02-05
Fix/sva 172 homescreen refreshing

digorath:

- Refactored the homescreen to reduce code redundancy.
- Added useHomeRefresh hook to add to components that have internal functions that need to be called when the pull-to-refresh is triggered.
  - in our case those are all refetches of apollo queries for now
- added useHomeRefresh hook to widgets and new HomeSection component that was the result of the refactoring.

## Known "issue"

There are still remaining components that still use the "old" refreshing method with the "refreshing" prop.
These can be refactored at a later point.

SVA-172

### DIFF
--- a/src/components/HomeSection.tsx
+++ b/src/components/HomeSection.tsx
@@ -16,9 +16,9 @@ import { Touchable } from './Touchable';
 import { Wrapper } from './Wrapper';
 
 type Props = {
-  categoryTitle: string;
-  categoryTitleDetail?: string;
-  categoryButton: string;
+  title: string;
+  titleDetail?: string;
+  buttonTitle: string;
   fetchPolicy:
     | 'cache-first'
     | 'network-only'
@@ -26,7 +26,7 @@ type Props = {
     | 'no-cache'
     | 'standby'
     | 'cache-and-network';
-  navigateToCategory: () => void;
+  navigate: () => void;
   navigation: NavigationScreenProp<never>;
   query: string;
   queryParser?: (data: unknown) => unknown[];
@@ -34,11 +34,11 @@ type Props = {
 };
 
 export const HomeSection = ({
-  categoryButton,
-  categoryTitle,
-  categoryTitleDetail,
+  buttonTitle,
+  title,
+  titleDetail,
   fetchPolicy,
-  navigateToCategory,
+  navigate,
   navigation,
   query,
   queryParser,
@@ -63,17 +63,15 @@ export const HomeSection = ({
   }
 
   const items =
-    queryParser?.(data) ?? parseListItemsFromQuery(query, data, true, categoryTitleDetail ?? '');
+    queryParser?.(data) ?? parseListItemsFromQuery(query, data, true, titleDetail ?? '');
 
   if (!items || !items.length) return null;
 
   return (
     <>
       <TitleContainer>
-        <Touchable onPress={navigateToCategory}>
-          <Title accessibilityLabel={`${categoryTitle} (Überschrift) (Taste)`}>
-            {categoryTitle}
-          </Title>
+        <Touchable onPress={navigate}>
+          <Title accessibilityLabel={`${title} (Überschrift) (Taste)`}>{title}</Title>
         </Touchable>
       </TitleContainer>
       {device.platform === 'ios' && <TitleShadow />}
@@ -87,7 +85,7 @@ export const HomeSection = ({
         />
 
         <Wrapper>
-          <Button title={categoryButton} onPress={navigateToCategory} />
+          <Button title={buttonTitle} onPress={navigate} />
         </Wrapper>
       </View>
     </>
--- a/src/components/HomeSection.tsx
+++ b/src/components/HomeSection.tsx
@@ -5,6 +5,7 @@ import { NavigationScreenProp } from 'react-navigation';
 
 import { colors, consts, device } from '../config';
 import { parseListItemsFromQuery } from '../helpers';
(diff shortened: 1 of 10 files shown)


src/components/HomeSection.tsx
```diff
@@ -0,0 +1,95 @@
...
+type Props = {
+  categoryTitle: string;
+  categoryTitleDetail?: string;
+  categoryButton: string;
+  fetchPolicy:
+    | 'cache-first'
+    | 'network-only'
+    | 'cache-only'
+    | 'no-cache'
+    | 'standby'
```
donni106, line 27: where do you have these strings from? `standby` is new for me. i do not think that our apollo version does already know that

src/components/HomeSection.tsx
```diff
@@ -0,0 +1,95 @@
...
+    queryParser?.(data) ?? parseListItemsFromQuery(query, data, true, categoryTitleDetail ?? '');
+
+  if (!items || !items.length) return null;
+
+  return (
+    <>
+      <TitleContainer>
+        <Touchable onPress={navigateToCategory}>
+          <Title accessibilityLabel={`${categoryTitle} (Überschrift) (Taste)`}>
+            {categoryTitle}
```
donni106, line 75: the wording `category` here and for button is not very accurate. The `category` exists for news items but not for events, point of interests and tours.
i think wer should abstract more here:

```
categoryButton,
categoryTitle,
categoryTitleDetail,
fetchPolicy,
navigateToCategory
```

to 

```
buttonTitle,
title,
titleDetail,
fetchPolicy,
navigate
```

src/components/screens/HomeCarousel.js
```diff
@@ -12,71 +12,58 @@ import { LoadingContainer } from '../LoadingContainer';
...
-          return (
-            <LoadingContainer>
-              <ActivityIndicator color={colors.accent} />
-            </LoadingContainer>
-          );
-        }
+  const { data, loading, refetch } = useQuery(getQuery(QUERY_TYPES.PUBLIC_JSON_FILE), {
+    variables: { name: 'homeCarousel' },
+    fetchPolicy,
+    skip: !refreshTime
```
donni106, line 32: this `skip` is new here but i cannot find out why?!
"skip if there is no refresh time"? what case did you think of here?

src/components/HomeSection.tsx
digorath: done in fdbc6c498dd9b3fcbce163062c82fbf2f16e180d

src/components/HomeSection.tsx
digorath: Screenshot 2021-02-08 at 14 00 30

I took them from the prop types that the useQuery hook expects, so that hook should support it.

src/components/screens/HomeCarousel.js
digorath: the case i thought of was "refresh time is undefined", which will happen on every initial render ;)

due to the fact how the refresh time is loaded from async storage, it is always undefined at first.

previously we had an early return of a loadingspinner, if the refreshTime was undefined, before the query was executed.

because we now use the hook (and the amount and order of hooks needs to remain consistent between renders) we need to execute the hook with the skip property instead, if the refreshTime is undefined.

donni106: clear description i couldn't think of/find by myself with looking on the diff.
thanks. agree.

if we wouldn't skip, the query would request fresh data before knowing, that there is a cached version from within the refresh time, right? (if there is)

digorath: exactly. the reasoning for the skip is exactly the same as we had before with the early return

donni106 (approved):

src/screens/HomeScreen.js
```diff
@@ -159,193 +142,54 @@ export const HomeScreen = ({ navigation }) => {
...
-                          />
-                        </Wrapper>
-                      </View>
-                    );
-                  }}
-                </Query>
-              </Fragment>
+            ({ categoryButton, categoryId, categoryTitle, categoryTitleDetail }, index) => (
+              <HomeSection
+                key={index}
```
donni106, line 152: was war der grund für die `key` änderung? hab mal gelesen, dass nur ein index nicht optimal ist und glaube im gesamten bundle nicht eindeutig genug ist. auf die map hier bezogen is der zwar eindeutig, aber versuche extra immer noch was klar global eindeutigeres hinzuzufügen.

donni106: https://de.reactjs.org/docs/lists-and-keys.html#keys

> Wir empfehlen dir nicht die Indizes für die Keys zu verwenden, da sich die Reihenfolge der Listeneinträge verändern kann. Dies kann sich negativ auf die Performance auswirken und zu Problemen mit dem Komponenten-State führen. Im Artikel von Robin Pokorny kannst du eine ausführlichere Erklärung über die negativen Auswirkungen beim Verwenden des Index als Key finden. Wenn du Listenelemente keinen expliziten Key zuweist, verwendet React standardmäßig die Indizes als Key.
>
> Wenn du dich dafür interessiert mehr zu lernen, ist hier eine ausführliche Erklärung, warum Keys nötig sind.

digorath: I did not see the benefit of adding the category title to the key.
do you think there will be fluctuation in categories that might cause issues down the line?
should we check other occurences of the index being the key as well?

the keys only need to be unique between siblings, so i do not think there is any value in making them globally unique.

PR #185, opened 2021-02-09
Refactor NewsItem and extract ImageSection/MediaSection from items

digorath:

- extracted image parsing from mediacontents and the display of images from news items, POIs, tours and event records to ImageSection
- extracted story section from NewsItem and used new MediaSection as well as the overall extracted ImageSection


src/components/ImageSection.js
```diff
@@ -0,0 +1,54 @@
...
+export const ImageSection = ({ mediaContents }) => {
+  let images = [];
+  !!mediaContents?.length &&
+    _filter(
+      mediaContents,
+      (mediaContent) =>
+        mediaContent.contentType === 'image' || mediaContent.contentType === 'thumbnail'
+    ).forEach((item) => {
+      !!item.sourceUrl &&
+        !!item.sourceUrl.url &&
```
donni106, line 19: we could optional chain here as well

src/components/ImageSection.js
```diff
@@ -0,0 +1,54 @@
...
+import PropTypes from 'prop-types';
+import React from 'react';
+import { StyleSheet } from 'react-native';
+
+import { Image } from './Image';
+import { ImagesCarousel } from './ImagesCarousel';
+import { WrapperWithOrientation } from './Wrapper';
+
+export const ImageSection = ({ mediaContents }) => {
+  let images = [];
```
donni106, line 11: i think this can be changed to `const` right, as we `push` and not reassign `images`?

src/components/screens/NewsItem.js
```diff
@@ -248,23 +86,22 @@ export const NewsItem = ({ data, navigation }) => {
           {!!logo && <Logo source={{ uri: logo }} />}
         </Wrapper>
 
-        {!!story && story}
+        {!!contentBlocks?.length &&
+          contentBlocks.map((contentBlock, index) => (
+            <StorySection
+              key={`${index}`}
```
donni106, line 92: if its just the `index` it does not need to be stringified. we could dicuss again the topic of just having an index or more like it was before with `${index}-${contentBlock.id}`. we started to mention this in a different PR. maybe we need to do a strict research and vote for the way we want to go with from now?

src/components/StorySection.js
```diff
@@ -0,0 +1,62 @@
...
+import { ImageSection } from './ImageSection';
+import { MediaSection } from './MediaSection';
+import { BoldText } from './Text';
+import { Wrapper, WrapperHorizontal } from './Wrapper';
+
+// eslint-disable-next-line complexity
+export const StorySection = ({ contentBlock, index, openWebScreen, settings }) => {
+  return (
+    <>
+      {index > 0 && !!contentBlock.title && (
```
donni106, line 16: please take the comments with the refactoring.

>// skip the title for the first content block because it is used as the main title

otherwise it will be harder to remember, why we skip the first index

src/components/StorySection.js
```diff
@@ -0,0 +1,62 @@
...
+                  ? contentBlock.intro
+                  : `<p>${contentBlock.intro}</p>`
+              }</div>`
+            )}
+            tagsStyles={{ div: { fontFamily: 'titillium-web-bold' } }}
+            openWebScreen={openWebScreen}
+          />
+        </WrapperHorizontal>
+      )}
+      {index > 0 && !!contentBlock.mediaContents && (
```
donni106, line 36: this had a comment before as well

> // skip images for the first content block because they are rendered as main images

do you think it is unnecessary?

donni106 (commented):
nice sectioning

src/components/ImageSection.js
digorath: done in 940791c14a151ff5ce3b8ce57dbf9832b86a57ac

src/components/ImageSection.js
digorath: yes you are right. done in 940791c14a151ff5ce3b8ce57dbf9832b86a57ac

src/components/screens/NewsItem.js
digorath: i reintroduced `contentBlock.id` as it is not harmful and in unexpected edge cases beneficial

src/components/StorySection.js
digorath: added it again

src/components/StorySection.js
digorath: hmm. depends on how you reach the file and look through it. and since it depends on that, i reintroduced the comments ;)

donni106 (approved):

PR #186, opened 2021-02-09
fix(image copyrights): crashing on empty string

digorath:
- the empty string is evaluated to false
  - this caused it to be tried to to be rendered (outside of text component)


donni106 (approved):
👍

PR #188, opened 2021-02-09
refactor(home screen): useHomeRefresh

digorath:

- added useHomeRefresh to remaining components

### DIFF
--- a/src/components/screens/Service.js
+++ b/src/components/screens/Service.js
@@ -1,7 +1,7 @@
 import PropTypes from 'prop-types';
 import React, { useContext } from 'react';
 import { StyleSheet, TouchableOpacity, View } from 'react-native';
-import { Query } from 'react-apollo';
+import { useQuery } from 'react-apollo';
 
 import { NetworkContext } from '../../NetworkProvider';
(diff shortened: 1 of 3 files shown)


src/components/screens/Service.js
```diff
@@ -17,108 +17,96 @@ import { getQuery, QUERY_TYPES } from '../../queries';
...
+                      />
+                    ) : (
+                      <Image
+                        source={{ uri: item.icon }}
+                        style={styles.serviceImage}
+                        PlaceholderContent={null}
+                        resizeMode="contain"
+                      />
+                    )}
+                    <BoldText small lightest center accessibilityLabel={`${item.title} (Taste)`}>
```
donni106, line 99: these props needs caution when merging in different releases!

donni106 (approved):
the description of the PR or the commit could be more explicit. one does not remember for sure everything about the implemented `useHomeRefresh` . i needed to search a bit in my memories.

src/components/screens/Service.js
digorath: ok. this cause conflcts though, whenever a release has used other props there, so it should be easy to catch.

PR #192, opened 2021-02-11
Fix/ no link for incomplete address

digorath:

Addresses are no longer clickable if not at least either the city, the street or the zip of it are strings of positive length.

The refactoring got rid of a lot of duplicated code.
For information on refactoring see commit message.

SVA-179


src/components/infoCard/AddressSection.tsx
```diff
@@ -0,0 +1,73 @@
...
+
+  // we do not check for address.kind here as we passed in a specific address, that we want to show
+  if (address) filteredAddresses.unshift(address);
+
+  return (
+    <>
+      {filteredAddresses.map((item, index) => {
+        const address = formatAddress(item);
+
+        const isPressable = item.city?.length || item.street?.length || item.zip?.length;
```
donni106, line 41: is it enough to have one of those? or do we want the address click just in case of a complete address?

src/components/infoCard/InfoCard.tsx
```diff
@@ -0,0 +1,106 @@
...
+  address,
+  addresses,
+  category,
+  contact,
+  contacts,
+  name,
+  webUrls,
+  openWebScreen
+}: Props) => (
+  <View>
```
donni106, line 65: this `View` was a `Wrapper` before or am i wrong? does this change rendering visually?

src/components/infoCard/InfoCard.tsx
```diff
@@ -0,0 +1,106 @@
...
+  category?: { name?: string };
+  name?: string;
+  openWebScreen: (link: string) => void;
+};
+
+const mergeWebUrls = ({ webUrls, contact, contacts }: WebUrlProps) => {
+  const mergedWebUrls = webUrls ? [...webUrls] : [];
+
+  if (contact?.www) {
+    mergedWebUrls.unshift({ url: contact.www });
```
donni106, line 32: i cannot see where this `contact.www` comes from and why it is `unshift`ed?

src/components/BB-BUS/Authority.js
```diff
@@ -40,71 +43,7 @@ export const Authority = ({ data, accordion, onPress, bottomDivider, openWebScre
...
-              return {
-                phone: !!phone && phone.value,
-                fax: !!fax && fax.value,
-                email: !!email && email.value,
-                www: !!www && www.value
-              };
-            })()}
-            openWebScreen={openWebScreen}
-          />
+          <InfoCard address={address} contact={contact} openWebScreen={openWebScreen} />
```
donni106, line 46: wow, this was exactly the same except of the need for `getAddress` and `getContact`?

donni106 (commented):
nice thanks. hard to review this all new files (a lot of scrolling and jumping around to compare). makes the code base smarter.
i hope merging to different releases will be smooth. could become a part you will take care of.

src/components/infoCard/AddressSection.tsx
digorath: i think it is reasonable to open the map then, because the map app will then search for a location based on the given information.
do you think we should be more strict here?

src/components/BB-BUS/Authority.js
digorath: i am not exactly sure what you mean, but there was a lot of duplicated code between `Authority.js` and `Persons.js`

src/components/infoCard/AddressSection.tsx
donni106: don't care. lets see, what map apps will be able to do and if these cases will occur.

src/components/BB-BUS/Authority.js
donni106: for bus bb these two methods are necessary to provide proper objects because of the weird nesting of directus graphql responses, right? without them it wouldn't work. so i was happy, that everything else works the same.

src/components/infoCard/InfoCard.tsx
donni106: but this `{!!name && ( ...` stuff was in operating company before inside of a `<Wrapper>`

Bildschirmfoto 2021-02-19 um 11 21 43

digorath: and it still is in `OperatingCompany` as the new `InfoCard` is wrapped inside a `Wrapper` there.

donni106 (approved):

PR #194, opened 2021-02-12
feat(weather): pull to refresh

digorath:

- added pull to refresh for weather screen
- added proper component if there is no data

### DIFF
--- a/src/screens/WeatherScreen.js
+++ b/src/screens/WeatherScreen.js
@@ -1,17 +1,19 @@
 import React, { useContext } from 'react';
 import { useQuery } from 'react-apollo';
-import { ActivityIndicator, FlatList, ScrollView, View } from 'react-native';
+import { ActivityIndicator, FlatList, RefreshControl, ScrollView, View } from 'react-native';
 
 import {
   DailyWeather,
   HeaderLeft,
   HourlyWeather,
   LoadingContainer,
+  RegularText,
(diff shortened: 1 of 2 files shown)


donni106:
a screenshot for the case of no data would be helpful for quicker review

src/screens/WeatherScreen.js
```diff
@@ -52,26 +54,49 @@ export const WeatherScreen = () => {
     fetchPolicy === 'network-only'
       ? { fetchPolicy, pollInterval: POLL_INTERVALS.WEATHER }
       : { fetchPolicy };
-  const { data, loading } = useQuery(getQuery(QUERY_TYPES.WEATHER_MAP), queryVariables);
+  const { data, loading, refetch } = useQuery(getQuery(QUERY_TYPES.WEATHER_MAP), queryVariables);
 
   useMatomoTrackScreenView(MATOMO_TRACKING.SCREEN_VIEW.WEATHER);
 
-  if (loading) {
+  if (loading && !data?.weatherMap) {
```
donni106, line 61: is `loading` not enough. what about if loading and there is data? we would see the previous data until the data changes, right?

digorath: your seceond/third question is kind of the answer to the first one.
`loading` would be enough for it being stable, but if we have previous data, we would like to show it together with an activity indicator on top.

PR #197, opened 2021-02-17
fix(dropdown): position in landscape mode

digorath:

- added orientation handling for dropdown

SVA-180

----

Labeled as draft as I am unsure, if this is what we actually want/need here.

### DIFF
--- a/src/components/DropdownSelect.js
+++ b/src/components/DropdownSelect.js
@@ -1,5 +1,5 @@
 import PropTypes from 'prop-types';
-import React, { memo, useEffect, useState } from 'react';
+import React, { useCallback, useContext, useEffect, useState } from 'react';
 import { StyleSheet, View } from 'react-native';
 import Dropdown from 'react-native-modal-dropdown';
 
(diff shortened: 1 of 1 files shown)


donni106:
can you add a screenshot please, here or in the ticket, to check it visually @digorath

donni106:
is it possible to put just the dropdowns modal in an orientation sensitive wrapper?

digorath:
> is it possible to put just the dropdowns modal in an orientation sensitive wrapper?

yes, i can remove the wrapper with orientation from the remaining screen elements.

digorath:
Screenshot 2021-02-17 at 18 04 25

src/components/DropdownSelect.js
```diff
@@ -9,69 +9,92 @@ import { Wrapper, WrapperRow, WrapperHorizontal } from './Wrapper';
...
-    const selectedData = data.find((entry) => entry.selected);
-    const selectedIndex = data.findIndex((entry) => entry.selected);
-    const preselect = (index) => this[`dropdown${label}`].select(index);
+  const marginHorizontal =
+    orientation === 'portrait' ? normalize(14) : device.height * 0.15 + normalize(14);
 
-    useEffect(() => {
-      preselect(selectedIndex);
-    }, [selectedData]);
+  const adjustFrame = useCallback(
```
donni106, line 28: this adjusts the dropdown modal on orientation change now?

digorath: the marginHorizontal inside (which is set 2 lines above) is the deciding factor, yes

src/components/DropdownSelect.js
```diff
@@ -9,69 +9,92 @@ import { Wrapper, WrapperRow, WrapperHorizontal } from './Wrapper';
...
+  searchPlaceholder
+}) => {
+  const { orientation } = useContext(OrientationContext);
 
-    const [arrow, setArrow] = useState('down');
-    const selectedData = data.find((entry) => entry.selected);
-    const selectedIndex = data.findIndex((entry) => entry.selected);
-    const preselect = (index) => this[`dropdown${label}`].select(index);
+  const marginHorizontal =
+    orientation === 'portrait' ? normalize(14) : device.height * 0.15 + normalize(14);
```
donni106, line 26: what is the calculation of `device.height * 0.15` based on here?

donni106: ok i see, `0.15` is the 15% landscape wrapper padding

donni106: i was wondering, why the dropdown modal wasn't full width on landscape.

changing to

```
// 0.15 = 15% landscape wrapper padding
const marginHorizontal = orientation === 'portrait' ? normalize(14) : device.width * 0.15;
```

results in

Simulator Screen Shot - iPhone 11 Pro Max - 2021-03-02 at 11 52 24

digorath: and... you prefer that?
i am not quite sure what to make of this comment? xD

donni106: hmmm...but i think that list has no landscape wrapper. we need to reduce the full width by safe area size. otherwise devices without safe areas have no full width here...which could be fine as well.

donni106: with `const { left: safeAreaLeft, right: safeAreaRight } = useSafeAreaInsets();` (from `react-native-safe-area-context`) we can get the safe are insets:

```
"insets": Object {
  "bottom": 21,
  "left": 44,
  "right": 44,
  "top": 0,
}
```

donni106: so we need to end up with

```
const marginHorizontal = orientation === 'portrait' ? normalize(14) : safeAreaLeft + normalize(14);
```

donni106: Simulator Screen Shot - iPhone 11 Pro Max - 2021-03-02 at 12 23 54

IMG_2519B2B94384-1

donni106 (approved):
@digorath now you have a final look on the changes please

src/components/DropdownSelect.js
donni106: see `adjustFrame` here https://github.com/siemiatj/react-native-modal-dropdown#props

src/components/DropdownSelect.js
digorath: looks good to me on both devices as it is now

digorath:
@donni106 looks good to me now

PR #198, opened 2021-02-19
Feature: Cross link business account data

donni106:
on point of interest detail screens, if a data provider is a business account, we want to cross link data of other types.
this PR introduces an event records section. it is workin progress, there are TODOs and a few steps to go. but please check the code already to be able to discuss about the structure that is setup, so we can continue with the best possible solution.

SVA-78

Simulator Screen Shot - iPhone 11 Pro Max - 2021-02-19 at 19 49 04


src/components/HomeSection.tsx
```diff
@@ -69,13 +68,7 @@ export const HomeSection = ({
...
   return (
     <>
-      <TitleContainer>
-        <Touchable onPress={navigate}>
-          <Title accessibilityLabel={`${title} (Überschrift) (Taste)`}>{title}</Title>
-        </Touchable>
-      </TitleContainer>
-      {device.platform === 'ios' && <TitleShadow />}
-
+      <SectionHeader title={title} onPress={navigate} />
```
donni106, line 71: this can be done on much more places across the code base

src/components/screens/PointOfInterest.js
```diff
@@ -138,6 +160,17 @@ export const PointOfInterest = ({ data, hideMap, navigation }) => {
           </Wrapper>
         )}
 
+        {!!businessAccount &&
+          !!crossDataTypes?.length &&
```
digorath, line 164: do you have more complex plans in mind for the crossDataTypes? as of now we set it as a constant local variable that always has to have length `3`.

src/queries/pointsOfInterest.js
```diff
@@ -147,3 +149,79 @@ export const GET_POINT_OF_INTEREST = gql`
     }
   }
 `;
+
+// TODO: add newsItems and tours
+// TODO: which data is really necessary in each?
+export const GET_POINT_OF_INTEREST_CROSS_DATA = gql`
```
digorath, line 155: i assumed we would have `newsItems`, `pointsOfInterest`, `tours`, `genericItems` as fields/properties of the POI query. am i misunderstanding the concept here, or why do we have a separate query for this?

src/screens/LunchScreen.js
```diff
@@ -124,11 +122,11 @@ export const LunchScreen = ({ navigation }) => {
...
         refreshControl={
           <RefreshControl
-            colors={[colors.accent]}
-            onRefresh={() => refetch?.()}
             // using this refreshing prop causes the loading spinner to animate in from top and
             // push the list down, whenever we change the date
             refreshing={false}
+            onRefresh={() => refetch?.()}
+            colors={[colors.accent]}
             tintColor={colors.accent}
```
digorath, line 130: what is the reason for this resorting of props?

src/components/screens/PointOfInterest.js
```diff
@@ -19,12 +20,14 @@ import { TMBNotice } from '../TMB/Notice';
...
 import { PriceCard } from './PriceCard';
+import { getQuery, QUERY_TYPES } from '../../queries';
+import { CrossDataSection } from '../CrossDataSection';
 
 const { MATOMO_TRACKING } = consts;
 
 /* eslint-disable complexity */
 /* NOTE: we need to check a lot for presence, so this is that complex */
-export const PointOfInterest = ({ data, hideMap, navigation }) => {
+export const PointOfInterest = ({ data, hideMap, navigation, fetchPolicy }) => {
```
digorath, line 30: passing in an external fetchpolicy, which is dependent on the refresh time for another query, does not seem like the right way to go here. maybe we should instead get a new refreshtime internally depending on the id and crossDataType?

src/screens/LunchScreen.js
donni106: see commit message:

> ordered props for `RefreshControl` like everywhere else for `RefreshControl`

src/queries/pointsOfInterest.js
donni106: no, the server provides the corresponding ids and the app needs to query them. this does not work simultaneously. i will send you a link to the slack conversation about it.

src/components/screens/PointOfInterest.js
donni106: its data for the point of interest, so i thought it can be a proper way to use the fetch policy for the corresponding point of interest. so so the cross data refresh with the point of interest. but i can see also cases, where the point of interest should not update while the events, news, ... needs a refresh. 

will see if there is a time to add an own refresh time and fetch policy here. if you would find the time, you could add that here as well.

src/components/screens/PointOfInterest.js
donni106: we could map directly, right. this is a very secure check for the `crossDataTypes` to not break the app, if somebody removes the variable without recognizing by others.

src/components/screens/PointOfInterest.js
digorath: added fetchpolicy for the crossdata itself, but it is located somewhere else now anyways.

src/components/HomeSection.tsx
digorath: yes. i have had it on my list for march for quite some time now 💪

digorath:
The requirements of this feature have shifted to the following:
* there should be a way to navigate to an additional screen where the data of the data provider can be reached from

What has changed:
* I unified the different occurences of sections of list data as far as seemed reasonable
* A new screen was added for the data provider
* queries
  * inconsistencies in regards to requesting or not requesting the id of data caused caching issues.
  * added filter options for a number of queries
* added an additional filter option for the location overview

Screenshot 2021-02-26 at 14 08 44
Screenshot 2021-02-26 at 14 08 37

digorath:
@donni106 I seem to be unable to request a review from you as this PR was initially opened by you.

So I will do it manually now: 'Please review!'

src/components/bookmarks/BookmarkSection.tsx
```diff
@@ -84,27 +64,17 @@ export const BookmarkSection = ({
...
+    <DataListSection
+      loading={loading}
+      navigation={navigation}
+      query={query}
+      buttonTitle={texts.bookmarks.showAll}
+      navigate={onPressShowMore}
+      sectionData={data}
+      sectionTitleDetail={categoryTitleDetail}
+      sectionTitle={sectionTitle}
+      showButton={ids.length > 3}
```
donni106, line 77: props here could be ordered the same way as in `HomeSection`

```
<DataListSection
  loading={loading}
  navigation={navigation}
  query={query}
  buttonTitle={buttonTitle}
  sectionData={data}
  navigate={navigate}
  sectionTitle={title}
  sectionTitleDetail={titleDetail}
  showButton
/>
```

or other way around

src/helpers/queryDataParser.js
```diff
@@ -139,12 +139,25 @@ const parseCategories = (data, skipLastDivider) => {
...
+  const pointsOfInterest = parsePointOfInterest(data?.[QUERY_TYPES.POINTS_OF_INTEREST]);
+  const tours = parseTours(data?.[QUERY_TYPES.TOURS]);
+
+  return {
+    [QUERY_TYPES.EVENT_RECORDS]: eventRecords,
+    [QUERY_TYPES.NEWS_ITEMS]: newsItems,
+    [QUERY_TYPES.POINTS_OF_INTEREST]: pointsOfInterest,
+    [QUERY_TYPES.TOURS]: tours
+  };
+};
```
donni106, line 159: why have you decided to create local consts?
could return directly and fit in 100 line length:

```
const parseCrossData = (data) => ({
  [QUERY_TYPES.EVENT_RECORDS]: parseNewsItems(data?.[QUERY_TYPES.EVENT_RECORDS]),
  [QUERY_TYPES.NEWS_ITEMS]: parseNewsItems(data?.[QUERY_TYPES.NEWS_ITEMS]),
  [QUERY_TYPES.POINTS_OF_INTEREST]: parsePointOfInterest(data?.[QUERY_TYPES.POINTS_OF_INTEREST]),
  [QUERY_TYPES.TOURS]: parseTours(data?.[QUERY_TYPES.TOURS])
});
```

src/helpers/queryHelper.ts
```diff
@@ -0,0 +1,30 @@
...
+    [QUERY_TYPES.EVENT_RECORD]: ROOT_ROUTE_NAMES.EVENT_RECORDS,
+    [QUERY_TYPES.NEWS_ITEM]: ROOT_ROUTE_NAMES.NEWS_ITEMS,
+    [QUERY_TYPES.POINT_OF_INTEREST]: ROOT_ROUTE_NAMES.POINTS_OF_INTEREST_AND_TOURS,
+    [QUERY_TYPES.TOUR]: ROOT_ROUTE_NAMES.POINTS_OF_INTEREST_AND_TOURS
+  }[query]);
+
+export const getTitleForQuery = (query: string) => {
+  switch (query) {
+    case QUERY_TYPES.NEWS_ITEMS:
+      return texts.homeCategoriesNews.categoryTitle;
```
donni106, line 20: the text for news items is a fallback if there is none from global settings, right?
so it can differ here if there are different texts existing in global settings, right?

src/components/CrossDataSection.tsx
```diff
@@ -0,0 +1,77 @@
...
+      <LoadingContainer>
+        <ActivityIndicator color={colors.accent} />
+      </LoadingContainer>
+    );
+  }
+
+  return crossDataTypes.map((crossDataType, index) => (
+    <DataListSection
+      key={`${index}-${crossDataType}`}
+      horizontal={false}
```
donni106, line 68: shouldn't points of interest and tours be displayed as horizontal slider? or depend on the list settings? all types can be displayed as chosen in list type settings here, or not?

src/components/CrossDataSection.tsx
```diff
@@ -0,0 +1,77 @@
...
+      query={crossDataType}
+      sectionData={data}
+      navigation={navigation}
+      buttonTitle={texts.dataProvider.showAll}
+      navigate={getNavigationFunction(navigation, dataProviderName, crossDataType)}
+      showButton
+    />
```
donni106, line 75: i am missing a divider between "Orte" entries

Simulator Screen Shot - iPhone 11 Pro Max - 2021-03-03 at 19 16 20

donni106, line 75: there shouldn't be a button if the entries of a section are less then the limit, because the index view will not have additional entries.

src/components/CrossDataSection.tsx
```diff
@@ -0,0 +1,77 @@
...
+  navigation: NavigationScreenProp<never>,
+  dataProviderName: string,
+  query: string
+) => {
+  return () =>
+    navigation.push('Index', {
+      queryVariables: { dataProvider: dataProviderName },
+      query,
+      title: getTitleForQuery(query)
+    });
```
donni106, line 37: the filter does not correspond to the data visible on the index screen. for filtered cross data, the dropdown header seems to need adjustments.

Simulator Screen Shot - iPhone 11 Pro Max - 2021-03-03 at 19 37 11-side

there is already a condition to not show a filter at all if there is just one entry inside.

```
// do not show a filter, if there is just one entry (Alle + Item = 2)
if (dropdownData.length <= 2) return null;
```

as cross data are always of the same data provider, the filter on data provider makes no sense anyways.

donni106, line 37: the category filter does not correspond to the data visible on the index screen. for filtered cross data, the dropdown header seems to need adjustments.

Simulator Screen Shot - iPhone 11 Pro Max - 2021-03-03 at 19 45 14-side

there are more categories to filter than categories from entries available on the screen.

src/components/CrossDataSection.tsx
```diff
@@ -45,7 +46,7 @@ export const CrossDataSection = ({ dataProviderId, dataProviderName, navigation
   const variables = {
     dataProviderId,
     orderEventRecords: 'listDate_ASC',
-    limit: 3
+    limit: 4
```
donni106, line 49: why 4 now?

src/screens/IndexScreen.js
```diff
@@ -107,10 +113,10 @@ export const IndexScreen = ({ navigation }) => {
...
   if (!query) return null;
 
-  const showFilter = {
-    [QUERY_TYPES.EVENT_RECORDS]: showEventsFilter,
-    [QUERY_TYPES.NEWS_ITEMS]: showNewsFilter
-  }[query];
+  // const showFilter = {
+  //   [QUERY_TYPES.EVENT_RECORDS]: showEventsFilter,
+  //   [QUERY_TYPES.NEWS_ITEMS]: showNewsFilter
+  // }[query];
```
donni106, line 119: comment can be removed?

src/helpers/queryDataParser.js
```diff
@@ -145,7 +145,7 @@ const parsePointsOfInterestAndTours = (data) => {
 };
 
 const parseCrossData = (data) => {
-  const eventRecords = parseNewsItems(data?.[QUERY_TYPES.EVENT_RECORDS]);
+  const eventRecords = parseEventRecords(data?.[QUERY_TYPES.EVENT_RECORDS]);
```
donni106, line 148: nice catch!

src/helpers/queryHelper.ts
digorath: that is correct. however this was necessary due to us ignoring news categories there.
i added categories for news now.

src/screens/IndexScreen.js
digorath: done

src/components/CrossDataSection.tsx
digorath: because i will not know if there are more than 3 and need to show the button, if i only fetch at most 3 here.

src/helpers/queryDataParser.js
digorath: it is removed altogether now

digorath (commented):
Please check again @donni106

src/components/CrossDataSection.tsx
donni106: ok, sure.
could be written somewhere.

src/screens/BookmarkScreen.js
```diff
@@ -15,8 +15,8 @@ import {
 import { colors, consts, device, normalize, texts } from '../config';
 import { getKeyFromTypeAndCategory } from '../helpers';
 import { useBookmarks, useMatomoTrackScreenView } from '../hooks';
+import { useNewsCategories } from '../hooks/NewsCategories';
```
donni106, line 18: could be an import from /hooks directly, right?

src/components/CrossDataSection.tsx
donni106: aaah and because of the missing images i thought of text list items. all right, i see.

src/components/CrossDataSection.tsx
donni106: ok

donni106: ok, so we will wait until somebody asks for it

src/components/CrossDataSection.tsx
donni106: > i removed the forced vertical display here

where? haven't found that. wanted to see, if it is also done for tours.

src/components/CrossDataSection.tsx
donni106: list type settings will be considered, right? my tests on that for points of interest and events worked out.

digorath: yes they are

src/components/CrossDataSection.tsx
digorath: as discussed this was done by `horizontal={false}` which is now removed

src/screens/BookmarkScreen.js
digorath: done. (as in the other file in which it is used.)

digorath (approved):

PR #200, opened 2021-02-22
chore(push notifications): enable disabling through server settings

digorath:

Some communities do not immediately want to have push notifications enabled. so i added the option to disable it through server settings.
this way the permission popup and the settings should not show, when the server sided settings do explicitly set push notifications to a disabled state.

SVA-185

### DIFF
--- a/src/hooks/PushNotification.ts
+++ b/src/hooks/PushNotification.ts
@@ -16,8 +16,15 @@ type ResponseHandler = (arg: Notifications.NotificationResponse) => void;
 export const usePushNotifications = (
   notificationHandler?: NotificationHandler,
   interactionHandler?: ResponseHandler,
-  behavior?: Notifications.NotificationBehavior
+  behavior?: Notifications.NotificationBehavior,
+  active?: boolean
 ): void => {
+  // this causes the active state to never change between rerenders.
+  // like this enabling or disabling the pushNotifications requires an app restart.
+  const [isActive] = useState(active);
+
+  if (isActive === false) return;
+
   const notificationListener = useRef<Subscription>();
   const responseListener = useRef<Subscription>();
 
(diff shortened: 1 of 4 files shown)


src/screens/SettingsScreen.js
```diff
@@ -80,12 +80,15 @@ export const SettingsScreen = () => {
       });
 
     const updateSectionedData = async () => {
-      const { settings = { matomo: false } } = globalSettings;
-      const pushPermission = await readFromStore(PushNotificationStorageKeys.IN_APP_PERMISSION);
+      const { settings = { matomo: false }, pushNotifications } = globalSettings;
```
donni106, line 83: do we want to have `matomo` nested in `settings` but `pushNotifications` not? why should there be a difference?

digorath: updated it to be in line with other "on/off" features

src/hooks/PushNotification.ts
```diff
@@ -16,8 +16,15 @@ type ResponseHandler = (arg: Notifications.NotificationResponse) => void;
 export const usePushNotifications = (
   notificationHandler?: NotificationHandler,
   interactionHandler?: ResponseHandler,
-  behavior?: Notifications.NotificationBehavior
+  behavior?: Notifications.NotificationBehavior,
+  active?: boolean
 ): void => {
+  // this causes the active state to never change between rerenders.
+  // like this enabling or disabling the pushNotifications requires an app restart.
```
donni106, line 23: we should add the information on activation/deactivation like for matomo: `...wird [...] mit dem nächsten Neustart der App wirksam.`

digorath: this is happening when the SERVER settings get changed.
do you think there is a reasonable way to notify a user about that?

donni106: aha, haven't read it in a way, that the server is meant here.

donni106 (approved):

PR #201, opened 2021-02-23
feat(image): add refresh interval

digorath:

- added option to set a refresh interval for the image component
- added parsing of refresh interval for ImagesCarousel

## How to test

The first image on the internal test server has a refresh interval set. So let the app run against that server and try it out :)

SVA-182

### DIFF
--- a/src/components/Image.js
+++ b/src/components/Image.js
@@ -9,6 +9,17 @@ import { imageHeight, imageWidth } from '../helpers';
 import { SettingsContext } from '../SettingsProvider';
 import { ImageMessage } from './ImageMessage';
 import { ImageRights } from './ImageRights';
+import { useInterval } from '../hooks';
+
+const addQueryParam = (url, param) => {
+  if (!url?.length) return;
+
+  if (url.endsWith('/')) {
+    url = url.slice(0, url.length - 1);
+  }
+
+  return url.includes('?') ? `${url}&${param}` : `${url}?${param}`;
+};
 
 export const Image = ({
   source,
@@ -18,11 +29,14 @@ export const Image = ({
   PlaceholderContent,
   aspectRatio,
   resizeMode,
-  borderRadius
+  borderRadius,
+  refreshInterval
 }) => {
   const [uri, setUri] = useState(null);
   const { globalSettings } = useContext(SettingsContext);
+  const refreshCount = useInterval(refreshInterval);
 
+  // only use cache when refreshInterval is undefined
   // if there is a source.uri to fetch, do it with the CacheManager and set the local path to show.
   // if there is no uri, the source itself should be already a local path, so set it immediately.
   useEffect(() => {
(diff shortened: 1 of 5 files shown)


src/components/Image.js
```diff
@@ -33,15 +43,22 @@ export const Image = ({
...
+            )
+        : mounted && setUri(source);
+    } else {
+      // add an artificial query param to the end of the url to trigger a rerender and refetch
+      mounted && setUri(addQueryParam(source.uri ?? source, `svaRefreshCount=${count}`));
+    }
 
     return () => (mounted = false);
-  }, [source, setUri]);
+  }, [count, refreshInterval, source, setUri]);
```
donni106, line 61: i find the variable name `count` too general. one do not know a quick read, what is counted. what about `svaRefreshCount`? your already made your thoughts about a naming :)

src/components/Image.js
```diff
@@ -18,11 +25,14 @@ export const Image = ({
...
   resizeMode,
-  borderRadius
+  borderRadius,
+  refreshInterval
 }) => {
   const [uri, setUri] = useState(null);
   const { globalSettings } = useContext(SettingsContext);
+  const count = useInterval(refreshInterval);
 
+  // only use cache when refreshInterval is undefined
```
donni106, line 35: 👍

src/components/Image.js
```diff
@@ -33,15 +43,22 @@ export const Image = ({
...
+      source.uri
+        ? CacheManager.get(source.uri)
+            .getPath()
+            .then((path) => mounted && setUri(path))
+            .catch((err) =>
+              console.warn('An error occurred with cache management for an image', err)
+            )
+        : mounted && setUri(source);
+    } else {
+      // add an artificial query param to the end of the url to trigger a rerender and refetch
```
donni106, line 56: 👍

src/components/Image.js
```diff
@@ -9,6 +9,13 @@ import { imageHeight, imageWidth } from '../helpers';
 import { SettingsContext } from '../SettingsProvider';
 import { ImageMessage } from './ImageMessage';
 import { ImageRights } from './ImageRights';
+import { useInterval } from '../hooks';
+
+const addQueryParam = (url, param) => {
+  if (!url?.length) return;
+
+  return url.includes('?') ? `${url}&${param}` : `${url}?${param}`;
```
donni106, line 17: there could be the case, where the `url` ends with `/`, right? maybe we need to catch that, because `.../?...` can behave differently to `...?...`. i do not know how to test right now. let's go with this solution until we face a problem.

src/components/Image.js
digorath: i set it to `refreshCount`. adding the `sva` prefix feels unnecessary for the internal variable. i only added it to the query param name to make it considerably less likely that that param actually has any influence

donni106 (approved):

src/components/Image.js
digorath: if the url ends on a slash, that slash is removed now, as it was not really difficult to do so.

src/components/Image.js
donni106: good choice

PR #203, opened 2021-02-25
Feature/sva 77 jobs and ads

digorath:

* added bookmark logic for generic items
* added queryparsing for generic items
* added generic item query
* added jobs and ads in form of generic items
  * added new common component for jobs and ads


src/components/Offer.js
```diff
@@ -0,0 +1,154 @@
...
+    publicationDate,
+    sourceUrl,
+    title
+  } = data;
+
+  useMatomoTrackScreenView(
+    matomoTrackingString([
+      genericType === GenericType.Job
+        ? MATOMO_TRACKING.SCREEN_VIEW.JOB_OFFER
+        : MATOMO_TRACKING.SCREEN_VIEW.COMMERCIAL_OFFER,
```
donni106, line 41: - [ ] confluence matomo tracking

src/helpers/dateTimeHelper.ts
```diff
@@ -0,0 +1,28 @@
...
+ *
+ * @return the formated event date string in format 'DD.MM.YYYY'
+ */
+export const eventDate = (dateStart?: string, dateEnd?: string) => {
+  if (!dateStart && !dateEnd) {
+    return '';
+  } else {
+    // by the previous check one of the two needs to be defined for moment format
+    // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+    return momentFormat((dateStart || dateEnd)!);
```
donni106, line 18: what does the `!` at the end? 
how does the comment fit when writing `by the previous check one of the two needs to be defined`? the previous check is the if statement? moment is called for both values existing, right? i do not get it completely here!?

why have you added an explicit `else` here?

src/helpers/genericTypeHelper.ts
```diff
@@ -0,0 +1,26 @@
+import { consts, texts } from '../config';
+import { GenericType } from '../types';
+
+export const getGenericItemDetailTitle = (genericType: GenericType): string => {
+  switch (genericType) {
+    case GenericType.Commercial:
+      return texts.commercial.commercial;
+    case GenericType.Job:
+      return texts.job.job;
```
donni106, line 9: this will grow with upcoming generic types, right?

src/helpers/genericTypeHelper.ts
```diff
@@ -0,0 +1,26 @@
...
+};
+
+const { ROOT_ROUTE_NAMES } = consts;
+
+export const getGenericItemRootRouteName = (genericType: GenericType): string => {
+  switch (genericType) {
+    case GenericType.Commercial:
+      return ROOT_ROUTE_NAMES.COMMERCIALS;
+    case GenericType.Job:
+      return ROOT_ROUTE_NAMES.JOBS;
```
donni106, line 22: this will grow with upcoming generic types, right?

src/helpers/queryDataParser.js
```diff
@@ -3,14 +3,29 @@ import _shuffle from 'lodash/shuffle';
...
+    const dateEnd = item?.dates?.[0]?.dateEnd;
+    const hasNotEnded = dateEnd ? isTodayOrLater(dateEnd) : true;
+
+    const publicationDate = item?.publicationDate;
+    const isPublished = publicationDate ? isBeforeEndOfToday(publicationDate) : true;
+
+    return hasNotEnded && isPublished;
+  }
+  return true;
+};
```
donni106, line 27: nice, we can use this, until the server is able to return the already filtered data

src/queries/index.js
```diff
@@ -31,6 +32,8 @@ export const getQuery = (query, filterOptions = {}) => {
     [QUERY_TYPES.EVENT_RECORDS]: filterOptions.showEventsFilter
       ? GET_EVENT_RECORDS_AND_CATEGORIES
       : GET_EVENT_RECORDS,
+    [QUERY_TYPES.GENERIC_ITEM]: GET_GENERIC_ITEMS,
```
donni106, line 35: plural `GET_GENERIC_ITEMS` is correct here?

src/screens/IndexScreen.js
```diff
@@ -149,7 +151,7 @@ export const IndexScreen = ({ navigation }) => {
                 query: getFetchMoreQuery(query),
                 variables: {
                   ...queryVariables,
-                  offset: listItems.length
+                  offset: data?.[query]?.length
```
donni106, line 154: why was this necessary and `listItems` changed with `data?.[query]?`? can the length of the one differ from the other?

src/helpers/dateTimeHelper.ts
digorath: the `!` tells typescript that that value is not undefined (it is usually better to avoid that, because it introduces possible loss of typeSafety.

the `momentFormat` is called if EITHER of them is undefined, which causes the disjunction to be always result in a defined value. however typescripts inference does not catch that at least one of them has to be defined if the code is written this way.

i changed it now, to make it clearer.

src/helpers/genericTypeHelper.ts
digorath: yes, this will grow with each new generic type

src/helpers/genericTypeHelper.ts
digorath: as does this

src/queries/index.js
digorath: added the singular query now

src/screens/IndexScreen.js
digorath: this length can now differ in the case of generic items, where we now filter the listitems for the dates.

this should have done `data.query.length` from the start, because that is what actually defines the offset.

src/helpers/queryDataParser.js
digorath: ✅

src/screens/IndexScreen.js
donni106: the list here shows `listItems` and fetches more data on scroll. so the offset needs to be the number of list items rendered. `listItems` is the number we see on the screen, right? and `data[query]` has a number returned by the server, which gets filtered before rendering, or have i misunderstood? so `data[query]` can have more entries than `listItems`, is this correct? and when refetching, we need to apply the number of already fetched data, because when applying the filtered (smaller) number, we could run into repeating items?

donni106: if the server gives already filtered data for generic items as well, `data[query].length` would equal `listItems.lengths` again in all cases, right?

digorath: yes.

the listItems length was previously always the same, but does not necessarily need to correspond to the amount of fetched data items. (in theory we could render multiple items per fetched dataEntry as well)
so we should always used the amount we fetched/received previously as the offset for the new fetch.

It could not only cause duplicate items, but also cause infinite refetching and appending as well, for example:
we fetch 10 items per fetchMore and only the first ever item gets not filtered.
that way we would always fetch with offset 1 without ever reaching the end of the available data.
and if item 12 was be valid again it would also never be shown either.

__tests__/dummyData/GenericItemDummyData.js
```diff
@@ -0,0 +1,81 @@
...
+  ),
+  dates: {
+    id: 'datesId',
+    dateEnd: '2021-03-24 15:54:31 +0100'
+  },
+  publicationDate: '2021-02-24 15:54:31 +0100',
+  payload: {
+    employmentType: 'Vollzeit, Remote'
+  }
+};
```
donni106, line 81: `Unable to resolve "../../__tests__/dummyData/GenericItemDummyData" from "src/components/Offer.js"`

when using it in `Offer` instead of `data`

src/components/Offer.js
```diff
@@ -0,0 +1,147 @@
...
+          )
+        )}
+        {device.platform === 'ios' && <TitleShadow />}
+        {!!dataProvider?.name && (
+          <Wrapper>
+            <RegularText small>{dataProvider.name}</RegularText>
+          </Wrapper>
+        )}
+
+        {typeof payload?.employmentType === 'string' && (
```
donni106, line 98: missing check for presence, which results in rendering the containers for `employmentType = ''`

donni106: done in ad999fb84c028adedd0bb0927d2a9938629ef713

src/components/Offer.js
```diff
@@ -0,0 +1,147 @@
...
+          <Wrapper>
+            <WrapperWrap>
+              <BoldText>{texts.job.employmentType}</BoldText>
```
donni106, line 101: `texts.job.employmentType` is missing. what should the value be? if there is none, remove the line please

src/components/Offer.js
```diff
@@ -0,0 +1,147 @@
...
+      title
+    ])
+  );
+
+  const link = sourceUrl?.url;
+  const rootRouteName = navigation.getParam('rootRouteName', '');
+  const headerTitle = navigation.getParam('title', '');
+
+  // action to open source urls
+  const openWebScreen = (webUrl) =>
```
donni106, line 52: this unsexy method seems not to be accurate or not used everywhere. when clicking "der alfigste fred..." links, the browser opens and not our internal web screen.

src/components/screens/OperatingCompany.tsx
```diff
@@ -29,11 +31,11 @@ export const OperatingCompany = ({ title, operatingCompany, openWebScreen }: Pro
         <Title accessibilityLabel={`${title} (Überschrift)`}>{title}</Title>
       </TitleContainer>
       {device.platform === 'ios' && <TitleShadow />}
+      {!!logo && <Logo source={{ uri: logo }} />}
```
donni106, line 34: logo rendering should be the same as in existing components, inside the wrapper, `Event` for example:

```
<Wrapper>
  {!!logo && <Logo source={{ uri: logo }} />}

  <InfoCard
    category={category}
    addresses={addresses}
    contacts={contacts}
    webUrls={webUrls}
    openWebScreen={openWebScreen}
  />
</Wrapper>
```

donni106: done in e7cc835bf306052d8dfa8e6ec5856e9a44507369

__tests__/dummyData/GenericItemDummyData.js
digorath: as placing it there is a manual process for testing anyway, i think it is actually beneficial to not have it bundled in the app.

src/components/Offer.js
digorath: it was missing for the first info card.

src/components/Offer.js
digorath: i added the string now. it somehow got lost along the way oO

src/components/Offer.js
digorath: i made the check for string length more explicit now, since the first type check already checks for presence (aka type being neither undefined nor null).

__tests__/dummyData/GenericItemDummyData.js
donni106: ok

src/components/Offer.js
donni106: a quick screenshot here would help me not to need to run everything to see that small change

donni106 (approved):

PR #207, opened 2021-03-02
Location Overview: better zoom and placeholder while no POI is selected

digorath:

- added placeholder text while no POI is selected to make functionality apparent
- added a default zoom level

SVA-188

### DIFF
--- a/src/components/map/WebViewMap.tsx
+++ b/src/components/map/WebViewMap.tsx
@@ -21,7 +21,7 @@ export const WebViewMap = ({
   mapCenterPosition,
   onMessageReceived,
   style,
-  zoom
+  zoom = 10
 }: Props) => {
   const messageHandler = useCallback(
     (message) => {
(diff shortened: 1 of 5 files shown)


src/components/map/WebViewMap.tsx
```diff
@@ -21,7 +21,7 @@ export const WebViewMap = ({
   mapCenterPosition,
   onMessageReceived,
   style,
-  zoom
+  zoom = 10
```
donni106, line 24: how come 10?

example with that value:

Simulator Screen Shot - iPhone 11 Pro Max - 2021-03-03 at 18 39 50

when i am using the bad belzig app, it hould be clear, that the location is somewhere there. so a more close view could be useful, maybe:

Simulator Screen Shot - iPhone 11 Pro Max - 2021-03-03 at 18 40 20

or something in between?

Simulator Screen Shot - iPhone 11 Pro Max - 2021-03-03 at 18 44 25

digorath: increased it to 14 for POIs and construction sites

donni106 (approved):

PR #208, opened 2021-03-05
feature(waste reminder): added reminder settings

digorath:

- restructured reminder settings components
- added WasteReminder files to push notification folder
- added parsing for reminder settings
- added rest api calls for reminder settings
  - new keys for secrets are required
- added error handling for reminder settings
- added permission alert for reminder settings

SVA-79

## Important info

added new const to secrets file:

```ts
export const staticRestSuffix = {
  wasteReminderRegister: '/notification/wastes.json',
  wasteReminderDelete: '/notification/wastes/'
};
```


donni106:
> ## DeepCode's analysis on [#693e55](https://www.deepcode.ai/app/gh/ikuseiGmbH/smart-village-app-app/df338110c114be7e10b8c1ed5a118063db518578/ikuseiGmbH/smart-village-app-app/693e555ba701a69f7ce1037a1f401d2cf5ef8c08/pr/_/%2F/code/?utm_source=gh_review&c=0&w=0&i=2&) found:
> ℹ️ **2** minor issues. 👇
> 
> * unexpected label on statement Occurrences
> * Identifier 'errorOccured' is never reassigned; use 'const' instead of 'let'. Occurrences


what about these two deepcode advices?

src/config/texts.js
```diff
@@ -183,11 +186,15 @@ export const texts = {
...
     configureReminder: 'Erinnerungen einstellen',
+    errorOnUpdateBody: 'Beim aktualisieren Ihrer Einstellungen ist ein Fehler aufgetreten.',
+    errorOnUpdateTitle: 'Fehler',
     hint: 'Bitte geben Sie Ihre Straße an.',
     onDayBeforeCollection: 'Am Vortag',
     onDayOfCollection: 'Am Tag der Abholung',
     reminder: 'Erinnerungen',
     reminderTime: 'Zu welcher Uhrzeit möchten Sie benachrichtigt werden?',
+    unableToLoad:
+      'Beim laden Ihrer Einstellungen ist ein Fehler aufgetreten. Bitte überprüfen Sie Ihre Verbindung und versuchen Sie es erneut.',
```
donni106, line 197: - [ ] laden -> Laden

donni106:
Screenshots are missing to see directly whats going on and whats the differences between iOS and android.

donni106:
> ```ts
> wasteReminderRegistration: '/notification/wastes.json',
> wasteReminderDeletion: '/notification/wastes/'
> ```

for push it was:

```ts
pushDevicesRegister: '/notification/devices.json',
pushDevicesDelete: '/notification/devices/remove.json'
```

there are differences between `registration` and `register` as well as `deletion` and `delete`. could be aligned.
and the `wasteReminderDeletion` endpoint is missing something a `remove.json`?

src/components/wasteCalendar/WasteReminderSettings.tsx
```diff
@@ -0,0 +1,319 @@
...
+      setError(false);
+      const storedSettings = await getReminderSettings();
+
+      if (!areValidReminderSettings(storedSettings)) {
+        setError(true);
+        setLoading(false);
+      } else {
+        const newSettings: ReminderSettings = parseReminderSettings(storedSettings, street);
+
+        // TODO: fix label for stored settings of inactive types ans show them as well
```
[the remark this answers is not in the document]
digorath: this todo is still there, and could be done at some point. this is only affecting the case when a reminder setting is set for a waste type, and then the waste types for that street change afterwards.

example.
user sets up reminder for waste type "bio" and street "A" . -> all collection dates of type "bio" get removed from  street "A" in the database -> the option for "bio" is no longer visible in the settings section for street "A", and will remain configured with the previous configuration.

donni106:
big screenshot because of ability to compare with our existing settings

Simulator Screen Shot - iPhone 11 Pro Max - 2021-03-10 at 13 55 25-tile

there are inconsistencies currently:

in the existing settings every "row" is pressable. in waste reminder settings just the toggles are pressable. a user should be able to press always the title as well as the toggles. visually the row with toggles should be also full with and with bold text like in the existing settings, right?
the bold text for the user to say what to do can be a regular text advice, like the error message.

the radiobuttons on the waste settings are visible directly instead of the selected value. this can be ok, because there are jsut two of them. but now we have radios on the left vs. radios on the right. further more the rows have more height, which can be the reason that they are visible directly. here, the whole row is pressable, which is good.

in existing settings the whole row is also clickable for list settings for example, with the edit indicator in the top right. i would suggest the same for the time which would result in the same section. bold headline and edit icon indicator fully pressable to open up the timepicker.
in the time picker itself i was wondering about the "Abbrechen" and "Ok" pressables, which we do not have right now somewhere else, or am i missing something? they are very squeezed between the two lines. this way we have a lot of different heights with objects between lines in this screen. do we need the "Ok"? couldn't it be saved with "Einstellungen aktualisieren"? so we would just need the cancel button, which we could place somewhere else then? i think two buttons would be too much, so the decision for some less attractive was good. maybe we could chat about such concepts and solutions before developing?

donni106:
the toggle on android behaves differently to the one in the existing settings. the toggle is animating on android for push setting but not here. could this be adjusted?

digorath:
> > ## DeepCode's analysis on [#693e55](https://www.deepcode.ai/app/gh/ikuseiGmbH/smart-village-app-app/df338110c114be7e10b8c1ed5a118063db518578/ikuseiGmbH/smart-village-app-app/693e555ba701a69f7ce1037a1f401d2cf5ef8c08/pr/_/%2F/code/?utm_source=gh_review&c=0&w=0&i=2&) found:
> > ℹ️ **2** minor issues. 👇
> > 
> > * unexpected label on statement Occurrences
> > * Identifier 'errorOccured' is never reassigned; use 'const' instead of 'let'. Occurrences
> 
> what about these two deepcode advices?

Those advices are inaccurate, because Deepcode is confused by the optional assignement operator.

src/components/wasteCalendar/WasteReminderSettings.tsx
```diff
@@ -0,0 +1,319 @@
...
+        />
+      }
+    >
+      <WrapperWithOrientation>
+        <Wrapper>
+          {error ? (
+            <RegularText>{texts.wasteCalendar.unableToLoad}</RegularText>
+          ) : (
+            <>
+              <Title>{texts.wasteCalendar.reminder}</Title>
```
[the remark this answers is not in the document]
digorath: added one (used new sectionheader component, which was recently introduced) and pulled it out of the wrapper, to make it in line with other uses of the sectionheader

src/config/texts.js
```diff
@@ -183,11 +186,15 @@ export const texts = {
   },
   wasteCalendar: {
     configureReminder: 'Erinnerungen einstellen',
+    errorOnUpdateBody: 'Beim aktualisieren Ihrer Einstellungen ist ein Fehler aufgetreten.',
```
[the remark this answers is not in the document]
digorath: done

src/config/texts.js
digorath: done

digorath:
> > ```ts
> > wasteReminderRegistration: '/notification/wastes.json',
> > wasteReminderDeletion: '/notification/wastes/'
> > ```
> 
> for push it was:
> 
> ```ts
> pushDevicesRegister: '/notification/devices.json',
> pushDevicesDelete: '/notification/devices/remove.json'
> ```
> 
> there are differences between `registration` and `register` as well as `deletion` and `delete`. could be aligned.
> and the `wasteReminderDeletion` endpoint is missing something a `remove.json`?

changed it to be in line with keys for general push notifications

donni106:
Simulator Screen Shot - iPhone 11 Pro Max - 2021-03-10 at 18 19 10

opened the simulator in this state and pulled to refresh, i saw the settings screen for a second. so something gets active quite short and deactivated again when checked, that the connection went wrong?

src/components/wasteCalendar/WasteReminderSettings.tsx
```diff
@@ -0,0 +1,361 @@
...
+
+  useEffect(() => {
+    setSwitchValue(active);
+  }, [active, setSwitchValue]);
+
+  return (
+    <ListItem
+      title={<RegularText>{categoryName}</RegularText>}
+      bottomDivider
+      rightIcon={<Switch switchValue={switchValue ?? false} toggleSwitch={toggleSwitch} />}
```
donni106, line 99: on android the switch is not animating here. didn't we fix that in our call?

src/components/wasteCalendar/ReminderSettingsReducer.ts
```diff
@@ -43,7 +43,7 @@ export const reminderSettingsReducer: React.Reducer<ReminderSettings, ReminderSe
     case ReminderSettingsActionType.UPDATE_ACTIVE_TYPE: {
       const { key, value } = action.payload;
       const newState = { ...state };
-      newState.activeTypes[key] = value;
+      newState.activeTypes[key] = { ...newState.activeTypes[key], active: value };
```
donni106, line 46: can `newState.activeTypes` be null or undefined, so that `[key]` would break?

src/components/wasteCalendar/WasteReminderSettings.tsx
```diff
@@ -0,0 +1,361 @@
...
+
+        setLoading(false);
+      }
+    },
+    [dispatch, setError, setLoading, types]
+  );
+
+  const onAbortIOS = useCallback(() => {
+    setShowDatePicker(false);
+    setLocalSelectedTime(state.reminderTime);
```
donni106, line 156: this resets the selected time, right? because with `onDatePickerChange` the `newTime` is already set per `setLocalSelectedTime` and when cancelling the picker/modal, the previous value needs to be stored?

src/components/wasteCalendar/WasteReminderSettings.tsx
```diff
@@ -0,0 +1,361 @@
...
+          // update setting
+          const id = await updateReminderSettings({
+            ...locationData,
+            onDayBefore: state.onDayBefore,
+            reminderTime: `${state.reminderTime.getHours()}:${state.reminderTime.getMinutes()}`,
+            wasteType: typeKey
+          });
+
+          // save new id
+          newState.activeTypes[typeKey].storeId = id;
```
donni106, line 215: can `newState.activeTypes` be null or undefined, so that `[typeKey]` would break?

src/components/wasteCalendar/WasteReminderSettings.tsx
```diff
@@ -0,0 +1,361 @@
...
+
+  const updateSettings = useCallback(async () => {
+    const newState = { ...state };
+    let errorOccured = false;
+
+    await Promise.all(
+      Object.keys(state.activeTypes).map(async (typeKey) => {
+        const entry = state.activeTypes[typeKey];
+        // delete inactive entries
+        if (!entry.active && entry.storeId) {
```
donni106, line 199: better optional chain `.active` and `.storeId` here, because entry could be nullish?

src/components/wasteCalendar/WasteReminderSettings.tsx
```diff
@@ -0,0 +1,361 @@
...
+          const success = await deleteReminderSetting(entry.storeId);
+
+          errorOccured ||= !success;
+        }
+
+        if (entry.active) {
```
donni106, line 205: better optional chain `.active` here, because entry could be nullish?

src/components/wasteCalendar/WasteReminderSettings.tsx
```diff
@@ -0,0 +1,361 @@
...
+    () => dispatch({ type: ReminderSettingsActionType.UPDATE_ON_DAY_BEFORE, payload: false }),
+    [dispatch]
+  );
+
+  const updateSettings = useCallback(async () => {
+    const newState = { ...state };
+    let errorOccured = false;
+
+    await Promise.all(
+      Object.keys(state.activeTypes).map(async (typeKey) => {
```
donni106, line 196: `forEach` instead of `map` or is the resulting array of the map used somehow?

src/components/wasteCalendar/WasteReminderSettings.tsx
```diff
@@ -0,0 +1,361 @@
...
+import { areValidReminderSettings, parseReminderSettings } from '../../jsonValidation';
+import { SectionHeader } from '../SectionHeader';
+
+const showErrorAlert = () => {
+  Alert.alert(texts.wasteCalendar.errorOnUpdateTitle, texts.wasteCalendar.errorOnUpdateBody);
+};
+
+const initialReminderTime = new Date();
+initialReminderTime.setHours(9);
+initialReminderTime.setMinutes(0);
```
donni106, line 43: did we know at what times the waste gets collected? can 9:00 be too late on the same day if somebody does not change the initial value so we should suggest an earlier time?

src/jsonValidation/reminderSettingsValidation.ts
```diff
@@ -0,0 +1,68 @@
...
+  };
+
+  if (filteredData.length) {
+    result.onDayBefore = filteredData[0].notify_days_before > 0;
+    result.reminderTime = new Date(filteredData[0].notify_at);
+  }
+
+  filteredData.forEach((item) => {
+    result.activeTypes[item.notify_for_waste_type] = { active: true, storeId: item.id };
+  });
```
donni106, line 65: this could go in the `if (filteredData.length) {` as there is nothing to forEach if length is 0. it does not break, but if there is the check already, it could be together.

src/pushNotifications/PermissionHandling.ts
```diff
@@ -107,3 +111,26 @@ const showInitialPushAlert = (): void => {
...
+      text: approve,
+      onPress: async () => {
+        await setInAppPermission(true);
+        approveCallback();
+      }
+    },
+    {
+      text: abort,
+      style: 'cancel'
+    }
```
donni106, line 134: this is the only alert without `{ cancelable: false });` now. intended?
and the order of "approve" and "cancel" is also different to the other alerts. there are cancel first and approve second.

src/pushNotifications/WasteReminder.ts
```diff
@@ -0,0 +1,126 @@
...
+  onDayBefore,
+  street,
+  wasteType,
+  zip
+}: SettingInfo) => {
+  const accessToken = await SecureStore.getItemAsync('ACCESS_TOKEN');
+  const pushToken = await getPushTokenFromStorage();
+  const requestPath = secrets[namespace].serverUrl + staticRestSuffix.wasteReminderRegister;
+  const os =
+    device.platform === 'ios' || device.platform === 'android' ? device.platform : 'undefined';
```
donni106, line 30: shouldn't `device.platform` be enough, as it returns the right one or something nullish otherwise?

src/pushNotifications/WasteReminder.ts
```diff
@@ -0,0 +1,126 @@
...
+  if (accessToken && pushToken) {
+    return fetch(requestPath, fetchObj)
+      .then((response) => {
+        if (response.ok) {
+          return response.json();
+        }
+      })
+      .catch(() => {
+        return undefined;
+      });
```
donni106, line 91: why `.then` and `.catch` here and in the method before the `await`-way? is there a specific reason for handling it differently?

src/components/wasteCalendar/WasteReminderSettings.tsx
digorath: we did....
and we did not...
we set the component to be the touchable, which caused it to be animated.
it seems that changing the component fixes it with fast refresh. so i have no idea how to go about it at the moment.

src/components/wasteCalendar/ReminderSettingsReducer.ts
digorath: it can not unless we do not initialize it when we call `useReducer`. and that will cause a type error, if we try to pass in a faulty initial value

src/components/wasteCalendar/WasteReminderSettings.tsx
digorath: they are entries from the state which are typed, and only ever interacted with in a typed way.
so either there is an entry for a key, which is of type 
```ts
{
  active: boolean;
  storeId?: number | undefined;
}
```
or a key is missing altogether

src/components/wasteCalendar/WasteReminderSettings.tsx
digorath: we need a map here, to return an array of promises which is the argument for `Promise.all`

src/components/wasteCalendar/WasteReminderSettings.tsx
digorath: same as above

src/components/wasteCalendar/WasteReminderSettings.tsx
digorath: we do not know the time, and i just defaulted to some value.
the user setting up the reminder will know when they need the reminder. i do not see the benefit of adding any specific time there

src/pushNotifications/WasteReminder.ts
digorath: i initially had no `await` syntax at all here, and changed it to that for the update function specifically, because there was a need of also awaiting and the `json()`promise and do an additional step with the result.

at that point it was more readable to switch to the `await` syntax.

src/pushNotifications/WasteReminder.ts
digorath: this way we pass `'undefined'` as a value to the backend in case something unhandled platform is used.
Screenshot 2021-03-11 at 10 50 08
Screenshot 2021-03-11 at 10 50 20

src/pushNotifications/PermissionHandling.ts
digorath: I removed the `{ cancelable: false }` from the `PermissionMissingAlert` as well now, as it is not required in either of these two, and moved the cancel button to be the first in the array.

src/jsonValidation/reminderSettingsValidation.ts
digorath: changed it

src/components/wasteCalendar/WasteReminderSettings.tsx
donni106: ok, strange but fine

donni106:
> i couldn't manage it to open the waste settings to set a reminder. running locally with `serverUrl: 'https://server.int-development.smart-village.app'`

i had the old secrets. with the updated one it worked.

donni106 (approved):
we need some feedback after on press to save the settings. i think `navigate.goBack()` would be a good one.

digorath:
> we need some feedback after on press to save the settings. i think `navigate.goBack()` or an `Alert` would be a good one.

added an alert on successful synchronisation of settings. this way we now always get an alert when we press the button. either the success one or the error one.

donni106 (approved):

PR #214, opened 2021-03-15
Snapshot tests with Jest 

SoleCincis:
**chore: snapshot tests fists attempts** 

- Test.test is the only test passing for far ,
- installed countless packages to solve endless failing test
- in package.json ignoring all dependencies that were causing a test fail

**chore: working snapshot test for simple components** 

- button, Icon and Link are successful tests producing snapshots
- CrossData is my first complex component
  - mocking hooks learn in process
- in OpeningTimesCard missing "/native" was throwing an error
  - that's why "jest-styled-components" isn't necessary anymore to solve this issue
- Wapper.test is there to show what we can  without jest-styled-components

**chore: test improvement** 

- correction on button test description invert is a styled prop not for landscape
- WrapperRow differentiation in the description of the different props and addition of the missing tests
- Icons addition of the missing tests , the tests pass but each has a warning that I have not been able to get rid of
   - skip svg icon test because my attempts to mock svg failed, so test in not passing
- moduleNameMapper in package.json should mock svg : https://github.com/kristerkari/react-native-svg-transformer#usage-with-jest

**feat: calling toJSON** 

-calling toJSON just to follow "explicit is better than implicit"

**chore(jest package.json): jest package uninstalled** 

- https://reactnativetesting.io/unit/setup.html


donni106:
> that's why "jest-styled-components" isn't necessary anymore to solve this issue

wasn't the package used for a different approach? please tell me what the package does and how it can be useful or not in our case.

__tests__/components/Link.test.js
```diff
@@ -0,0 +1,11 @@
+import React from 'react';
+import renderer from 'react-test-renderer';
+
+import { Link } from '../../src/components';
+
+describe('Link', () => {
+  it('renders a default Link', () => {
+    const tree = renderer.create(<Link description={'description'} url={'url'} />).toJSON();
```
donni106, line 8: here ist a `.toJSON()` and in other tests not. what is the difference in usage here and there?

__tests__/components/Icon.test.js
```diff
@@ -0,0 +1,11 @@
+import React from 'react';
+import renderer from 'react-test-renderer';
+
+import { Icon } from '../../src/components';
+
+describe('Icon', () => {
+  it('renders a default Icon', () => {
+    const tree = renderer.create(<Icon name={'name'} />);
```
donni106, line 8: `name` is not an icon key, right? does jest create different snapshots for different icons? we could than create tests for our main icons we use across the app.

src/components/screens/OpeningTimesCard.js
```diff
@@ -1,7 +1,7 @@
 import PropTypes from 'prop-types';
 import React from 'react';
 import { View, StyleSheet } from 'react-native';
-import styled from 'styled-components';
+import styled from 'styled-components/native';
```
donni106, line 4: what was the error here without `/native`? do you know why this fixes it? could this be necessary in other components as well?

package.json
```diff
@@ -71,6 +73,7 @@
   },
   "devDependencies": {
     "@babel/plugin-proposal-logical-assignment-operators": "^7.12.1",
+    "@babel/preset-typescript": "^7.13.0",
```
donni106, line 76: > installed countless packages to solve endless failing test

this is what i found in a commit message, but i see just one package that was installed here?

package.json
```diff
@@ -6,14 +6,16 @@
...
-    "test": "jest",
+    "test": "node_modules/.bin/jest",
+    "test:debug": "node --inspect-brk node_modules/jest/bin/jest.js --runInBand",
+    "test:coverage": "node_modules/.bin/jest --coverage",
     "build:android": "expo build:android",
     "build:ios": "expo build:ios"
   },
   "jest": {
-    "preset": "react-native",
+    "preset": "jest-expo",
```
donni106, line 16: was this change found somewhere in the expo docs?

package.json
```diff
@@ -6,14 +6,16 @@
     "ios": "expo start --ios",
     "lint": "./node_modules/.bin/eslint __tests__ src --ext .js,.jsx,.ts,.tsx",
     "start": "expo start",
-    "test": "jest",
+    "test": "node_modules/.bin/jest",
```
donni106, line 9: how does this differentiate from `"test": "jest"`?

__tests__/components/Button.test.js
```diff
@@ -0,0 +1,13 @@
+import React from 'react';
+import renderer from 'react-test-renderer';
+
+import { Button } from '../../src/components';
+
+describe('Button', () => {
+  const onPress = { navigate: jest.fn() };
+
+  it('renders a default button', () => {
+    const tree = renderer.create(<Button title={'title'} onPress={onPress} />);
```
donni106, line 10: the Button component can have the optional prop `invert`. we should have a second test with that case as well.

donni106, line 10: for string props, you do not need javascript braces, so best practice would be `title="title"` here

__tests__/components/CrossData.test.js
```diff
@@ -0,0 +1,17 @@
+import React, { useContext } from 'react';
+import renderer from 'react-test-renderer';
+
+import { CrossData } from '../../src/components';
+
+describe('CrossData', () => {
```
donni106, line 6: there is this test but no `snap` file with a snapshot. if this test should not be considered or not run, it could be marked with `skip` https://codewithhugo.com/run-skip-single-jest-test/

__tests__/components/Link.test.js
```diff
@@ -0,0 +1,11 @@
...
+    expect(tree).toMatchSnapshot();
+  });
```
donni106, line 10: does the Link component have other cases, that could be added to test?

donni106:
Can you add a screenshot please with the result of running jest @SoleCincis

SoleCincis:
> > that's why "jest-styled-components" isn't necessary anymore to solve this issue
> 
> wasn't the package used for a different approach? please tell us what the package does and how it can be useful or not in our case @SoleCincis

"jest-styled-components package improves the snapshot testing experience and provides a brand new matcher to make expectations on the style rules.!"  , it sounds good to me and it could be good to have it , if the aim of our testing it's about testing the style.
Initially I install it just to solve this error: 
Screenshot 2021-03-18 at 12 30 34

src/components/screens/OpeningTimesCard.js
SoleCincis: > what was the error here without `/native`? do you know why this fixes it? could this be necessary in other components as well?

The screenshot above shows the error, which tells that styled-components alone it's undefined , styled-components/native complete path is required, in order to use StyleSheet.create. It seem an anomaly present only in this file.

digorath: did you select the wrong screenshot?
this error is not the one that I initially found and fixed.

__tests__/components/CrossData.test.js
SoleCincis: I'm not running all the tests together, just the individual test files - but thanks for reminding me of .skip will help me soon!

SoleCincis:
Positive test results ( with function prop warning ) 
**BUTTON**
Screenshot 2021-03-19 at 16 51 16
**LINK**
Screenshot 2021-03-19 at 16 59 33

package.json
SoleCincis: https://jestjs.io/docs/25.x/tutorial-react-native no docu has "react-native" . 
This change comes from montas package.json

__tests__/components/Button.test.js
```diff
@@ -0,0 +1,13 @@
+import React from 'react';
+import renderer from 'react-test-renderer';
+
+import { Button } from '../../src/components';
+
+describe('Button', () => {
+  const onPress = { navigate: jest.fn() };
```
donni106, line 7: you should declare a function instead of an object here for `onPress` to avoid the prop warnings. if we would pass an object inside of the app, the buttons would not work on press.
Button from react native requires a function https://reactnative.dev/docs/button#requiredonpress

package.json
SoleCincis: I'm sorry this message is not accurate. I have done a bunch of changes and lost track of what was deleted and what stayed . I would be happy to squash the first commit message with the second

__tests__/components/Link.test.js
SoleCincis: thanks, was a mistake , corrected in 9d3b4be5b92af2f9b3b9669be38ae70ad29230e1

__tests__/components/Button.test.js
SoleCincis: done in 9d3b4be5b92af2f9b3b9669be38ae70ad29230e1

SoleCincis: done in 9d3b4be

__tests__/components/Link.test.js
SoleCincis: yes, thanks . done in 9d3b4be

__tests__/components/Button.test.js
SoleCincis: thanks! done in 9d3b4be

SoleCincis:
All green now! 
Screenshot 2021-03-19 at 18 02 20
Screenshot 2021-03-19 at 18 01 55
Screenshot 2021-03-19 at 17 41 29

src/components/screens/OpeningTimesCard.js
SoleCincis: Yes you are right ! this is the error you have seen: 
Screenshot 2021-03-19 at 18 18 08
but the screenshot above shows an error that I got as well from this issue

__tests__/components/Icon.test.js
SoleCincis: correct brackets in 9d3b4be5b92af2f9b3b9669be38ae70ad29230e1. Jest does not show different snapshots for different icons. What is your idea; to write a test for each icon that is constant in the app? such as left arrow, burger menu, settings .. ?

__tests__/components/Link.test.js
donni106: what was the mistake? do you know the difference in using `toJSON()` or not here?

__tests__/components/Icon.test.js
donni106: brackets are still `<Icon name={'name'} />` with that commit?

i thought of having different tests for different Icons with different `name` props. but when you say that having different names result in the same snapshot, than this is maybe useless.

on the other hand our icon component can be used in a lot more cases regarding the props. to i could think of a bunch more tests here to care different cases. can you have a look and think of other icon tests to add?

__tests__/components/Button.test.js
```diff
@@ -0,0 +1,19 @@
...
+import renderer from 'react-test-renderer';
+
+import { Button } from '../../src/components';
+
+describe('Button', () => {
+  const onPress = () => {
+    return;
+  };
+
+  it('renders a landscape button', () => {
```
donni106, line 11: why do you think that invert has something to do with landscape? its wrong. please check what invert does. i think the snapshots could tell. if not, try out two buttons (one invert one not) in the app.

__tests__/components/Wrapper.test.js
```diff
@@ -0,0 +1,66 @@
...
+    const tree = renderer.create(<WrapperLandscape noFlex />);
+    expect(tree).toMatchSnapshot();
+  });
+
+  it('renders WrapperRow', async () => {
+    const tree = renderer.create(<WrapperRow center />);
+    expect(tree).toMatchSnapshot();
+  });
+
+  it('renders WrapperRow', async () => {
```
donni106, line 43: when you read the output of jest there are two test cases with the same name here. it... should not be identical. it should always be a string that describes the test and the two wrapper row tests are different.

__tests__/components/Button.test.js
SoleCincis: yes right, Outline is one of the three default styles of RNEButton. Correction made in 61c542c168e0cf5d5715f440ab76cfabfb8ccf7e

__tests__/components/Wrapper.test.js
SoleCincis: Correction made in 61c542c168e0cf5d5715f440ab76cfabfb8ccf7e

__tests__/components/Icon.test.js
SoleCincis: "i thought of having different tests for different Icons with different name props. but when you say that having different names result in the same snapshot, than this is maybe useless" 
Here a visual confirmation: 
Screenshot 2021-03-22 at 15 49 32
Screenshot 2021-03-22 at 15 49 20

A few more icon cases here 61c542c168e0cf5d5715f440ab76cfabfb8ccf7e 
among with the following warning : 
Screenshot 2021-03-22 at 18 25 10

__tests__/components/Link.test.js
SoleCincis: https://jestjs.io/docs/25.x/cli#--json 
Prints the test results in JSON and this is not what we need ?

package.json
donni106: do you know the difference regarding the two paths?

package.json
donni106: > no docu has "react-native"

how do you mean that? the one you have linked has the "react-native" preset.

__tests__/components/Icon.test.js
donni106: isn't it strange, that all icons result in the same snapshot content? with different props this should variate somehow or what is the expectation?

__tests__/components/Link.test.js
donni106: `toJSON()` in code and `--json` for cli has nothing to do with each other. have you read about jest and `toJSON()`?

SoleCincis: In [jest docs](https://jestjs.io/docs/snapshot-testing#snapshot-testing-with-jest) it's not really explained. 
And  [stackoverflow](https://stackoverflow.com/questions/61284054/using-tojson-or-not-with-snapshot-testing-with-jest) has this explanation; **"I would suggest calling toJSON just to follow "explicit is better than implicit" principle and the fact that the inner behavior is subject to change."** 
So I actually think it's a better practice and more specific to use toJSON() in snapshot testing although I'm not 100% sure I understand what does it do.

package.json
SoleCincis: "jest" is installed globally on my machine ? and "node_modules/.bin/jest" is locally just in my project node modules?

__tests__/components/Icon.test.js
digorath: for now we will keep it as is.
these tests will likely not catch many changes, and we should look into why the snapshots are not having more information in the future.

 for now i prioritize moving along with the testing in general before getting hung up on this

package.json
digorath: the expo documentation has the `jest-expo` preset: https://docs.expo.io/guides/testing-with-jest/

the documentation also says to install `jest-expo`, which we have, but we also have just `jest`.

please see if the `jest` package is actually needed, and remove it, if it is unnecessary

digorath (changes requested):
Please see to the discussed points. :)

__tests__/components/Link.test.js
SoleCincis: ef1041b4785f639c1b81b3ffb5cd5d8338149e2e

package.json
SoleCincis: 215113225b7d1604db7f8eeff6b7cf951ee91319

package.json
donni106: should we define `./node_modules..` like in the `"lint"` command, to ensure that both work the same regarding their path of execution?

donni106:
this is the result when i run `yarn test`:o on my machine

Bildschirmfoto 2021-06-22 um 18 33 25

can we address the 7 failed tests in this PR?

donni106 (commented):
this is the result when i run `yarn test` on my machine :o

Bildschirmfoto 2021-06-22 um 18 33 25

can we address the 7 failed tests in this PR?

package.json
SoleCincis: so adding the dot but also this part `__tests__ src --ext .js,.jsx,.ts,.tsx"` ?

donni106: no these options are part of eslint cli and not jest. you can read about them here https://eslint.org/docs/user-guide/command-line-interface

SoleCincis:
This is the latest output , skipping all the tests that throw an error: 
Screenshot 2021-06-23 at 14 02 43Screenshot 2021-06-23 at 14 02 50

donni106 (approved):
rebased and updated everything possible

Bildschirmfoto 2021-06-23 um 15 37 27

__tests__/__snapshots__/App.test.js.snap
```diff
@@ -0,0 +1,3 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`App renders correctly 1`] = `null`;
```
donni106, line 3: don't know how much sense these null snapshots make, but they pass :D

src/SettingsProvider.js
```diff
@@ -1,7 +1,13 @@
...
 import React, { createContext, useState } from 'react';
 
-export const SettingsContext = createContext();
+export const SettingsContext = createContext({
+  globalSettings: {
+    filter: {},
+    sections: {},
+    widgets: []
+  }
+});
```
donni106, line 10: good catch from tests, that these were undefined initially...i think this cannot break any logics for the app. please have a short thought about it @digorath

digorath: widgets doesn't need to be defined, the component handles the undefined case itself.

as far as i could see the sections always have `{}` as its default value, wherever they are used. same with filter.

But why is `navigation` set as an empty object in the new versions? it should be either `consts.DRAWER` or `consts.TABS`

digorath:
just one minor thing, and then this can be merged.
all unskipped tests pass for me as well

src/SettingsProvider.js
donni106: what does the test run say when removing widgets here?
sections was necessary, otherwise it failed.
good point to set navigation to drawer. but an empty initial value would result in that as well through code, right?

```
if (_isEmpty(globalSettings)) {
  globalSettings = {
    navigation: consts.DRAWER
  };
}
```

donni106:
```
globalSettings: {
  filter: {},
  navigation: consts.DRAWER,
  sections: {}
}
```

makes the tests also pass. take this?

digorath: 👍

digorath (approved):
approved after the discussed settings context changes

src/SettingsProvider.js
donni106: 122ee8e

PR #218, opened 2021-04-26
chore(offer): replace provider name with logo

digorath:
- added dataprovider logo to generic item query
- updated offer to use SectionHeader component
- removed unnecessary arg from onPress function
  - the function was only ever used as onPress, which does not get a string as an argument

SVA-202

before:

Screenshot 2021-04-26 at 14 38 25

after:

Screenshot 2021-04-26 at 14 37 13

### DIFF
--- a/src/components/Offer.js
+++ b/src/components/Offer.js
@@ -52,12 +52,12 @@ export const Offer = ({ data, navigation }) => {
 
   // action to open source urls
   const openWebScreen = useCallback(
-    () =>
+    (webUrl) =>
       navigation.navigate({
         routeName: 'Web',
         params: {
           title: headerTitle,
-          webUrl: link,
+          webUrl: !!webUrl && typeof webUrl === 'string' ? webUrl : link,
           rootRouteName
         }
       }),
--- a/src/components/Offer.js
+++ b/src/components/Offer.js
@@ -2,18 +2,18 @@ import PropTypes from 'prop-types';
 import React, { useCallback } from 'react';
 import { View } from 'react-native';
 
-import { consts, device, texts } from '../config';
-import { matomoTrackingString, momentFormat, trimNewLines } from '../helpers';
+import { consts, texts } from '../config';
+import { matomoTrackingString, momentFormat } from '../helpers';
 import { useMatomoTrackScreenView } from '../hooks';
 import { GenericType } from '../types';
 import { DataProviderButton } from './DataProviderButton';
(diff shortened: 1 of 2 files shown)


src/components/Offer.js
```diff
@@ -48,15 +48,16 @@ export const Offer = ({ data, navigation }) => {
   const link = sourceUrl?.url;
   const rootRouteName = navigation.getParam('rootRouteName', '');
   const headerTitle = navigation.getParam('title', '');
+  const dataProviderLogo = dataProvider?.logo?.url;
 
   // action to open source urls
   const openWebScreen = useCallback(
-    (webUrl) =>
+    () =>
```
donni106, line 55: > * removed unnecessary arg from onPress function
>   * the function was only ever used as onPress, which does not get a string as an argument

does this also apply when passing down to other components? maybe it will be called with a param somewhere there?

```
{contentBlocks?.map((contentBlock, index) => {
  return (
    <StorySection
      contentBlock={contentBlock}
      index={index}
      key={`${contentBlock.id}-${index}`}
      openWebScreen={openWebScreen}
    />
  );
})}
{!!contact && (
  <Wrapper>
    <InfoCard contact={contact} openWebScreen={openWebScreen} />
  </Wrapper>
)}
<OperatingCompany
  title={texts.pointOfInterest.operatingCompany}
  logo={logo}
  operatingCompany={operatingCompany}
  openWebScreen={openWebScreen}
/>
```

src/queries/genericItem.js
```diff
@@ -77,6 +77,10 @@ export const GET_GENERIC_ITEMS = gql`
       }
       dataProvider {
         id
+        logo {
+          id
```
donni106, line 81: do we need the logo id in the queries that were updated?

digorath: we previously had an issue with the apollo cache, where the cache would misbehave if one query included the id, while another did not.

to resolve that we decided to include all ids, so that we can not run into this issue again if we add another query that includes the id at a later point.

src/components/Offer.js
digorath: you are right, it is for example used in `InfoCard`.

I missed that it was provided as a prop to multiple children.

I reverted the change.

Good catch!

src/queries/genericItem.js
donni106: forgot about that. thanks for pointing out.

donni106 (approved):

PR #226, opened 2021-05-12
Chore/sva 214 further changes after feedback

digorath:

* filter out organizations without current members from the person screen dropdown
* show current faction of a person in its preview
* show current faction of a person on its detail page

### Screenshots
Screenshot 2021-05-12 at 14 32 03
Screenshot 2021-05-12 at 15 23 06
Screenshot 2021-05-12 at 15 22 19

---

SVA-214

### DIFF
--- a/src/components/TextListItem.tsx
+++ b/src/components/TextListItem.tsx
@@ -42,10 +42,12 @@ export const TextListItem: NamedExoticComponent<Props> & {
 }>(({ navigation, item, noSubtitle, leftImage }) => {
   const { routeName, params, subtitle, title, bottomDivider, topDivider, picture } = item;
 
+  const titleText = <BoldText>{trimNewLines(title)}</BoldText>;
+
   return (
     <ListItem
-      title={noSubtitle || !subtitle ? undefined : <RegularText small>{subtitle}</RegularText>}
-      subtitle={<BoldText>{trimNewLines(title)}</BoldText>}
+      title={noSubtitle || !subtitle ? titleText : <RegularText small>{subtitle}</RegularText>}
+      subtitle={noSubtitle || !subtitle ? undefined : titleText}
       bottomDivider={bottomDivider !== undefined ? bottomDivider : true}
       topDivider={topDivider !== undefined ? topDivider : false}
       containerStyle={styles.container}
--- a/src/components/oParl/Person.tsx
+++ b/src/components/oParl/Person.tsx
@@ -4,6 +4,7 @@ import { NavigationScreenProp } from 'react-navigation';
 import { texts } from '../../config';
 import { PersonData } from '../../types';
 import { WrapperHorizontal } from '../Wrapper';
+import { getOrganizationNameString } from './oParlHelpers';
 import { Row, SimpleRow } from './Row';
 import {
   KeywordSection,
@@ -46,6 +47,10 @@ export const Person = ({ data, navigation }: Props) => {
     web
   } = data;
 
+  const faction = membership?.find(
+    (membership) => membership.organization?.classification === 'Fraktion' && !membership.endDate
+  );
+
   return (
     <>
       <Row left={personTexts.name} right={name} leftWidth={leftWidth} fullText />
@@ -54,6 +59,18 @@ export const Person = ({ data, navigation }: Props) => {
       <Row left={personTexts.formOfAddress} right={formOfAddress} leftWidth={leftWidth} />
       <Row left={personTexts.givenName} right={givenName} leftWidth={leftWidth} />
       <Row left={personTexts.familyName} right={familyName} leftWidth={leftWidth} />
+      <Row
+        left={personTexts.faction}
+        right={faction?.organization && getOrganizationNameString(faction.organization)}
+        leftWidth={leftWidth}
+        onPress={() =>
+          navigation.push('OParlDetail', {
+            id: faction?.id,
+            type: faction?.type,
+            title: texts.oparl.membership.membership
+          })
+        }
+      />
       <Row left={personTexts.gender} right={gender} leftWidth={leftWidth} />
       <Row
         left={personTexts.email}
--- a/src/components/oParl/previews/PersonPreview.tsx
+++ b/src/components/oParl/previews/PersonPreview.tsx
@@ -17,15 +17,15 @@ export const PersonPreview = ({ data, navigation }: Props) => {
 
   const title = getFullName(texts.oparl.person.person, data);
   const faction = membership?.find(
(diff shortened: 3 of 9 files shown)


src/components/TextListItem.tsx
```diff
@@ -42,10 +42,12 @@ export const TextListItem: NamedExoticComponent<Props> & {
...
   const { routeName, params, subtitle, title, bottomDivider, topDivider, picture } = item;
 
+  const titleText = <BoldText>{trimNewLines(title)}</BoldText>;
+
   return (
     <ListItem
-      title={noSubtitle || !subtitle ? undefined : <RegularText small>{subtitle}</RegularText>}
-      subtitle={<BoldText>{trimNewLines(title)}</BoldText>}
+      title={noSubtitle || !subtitle ? titleText : <RegularText small>{subtitle}</RegularText>}
+      subtitle={noSubtitle || !subtitle ? undefined : titleText}
```
donni106, line 50: i do not get this fully right now. can you make a real sentence to help me when to show a `title` and when a `subtitle`?

digorath: Previously we had always shown the `title`-prop as the subtitle, and the `subtitle`-prop as the title, if it was present and if the `noSubtitle`-flag was not set.

I now changed the logic to show the `title`-prop as the title instead of as the subtitle, if there is no `subtitle`-prop that we want to show (i.e. it is `undefined` or the `noSubtitle`-flag is set).

the condition for the ternary remains the same, and I added the same ternary condition to the `subtitle` of `ListItem` as well.

This way the `titleText` is always shown, and only as a subtitle if we want to show a subtitle as well.

src/screens/OParl/OParlPersonsScreen.tsx
```diff
@@ -69,6 +72,7 @@ const useListData = (
     if (organization) {
       setListData(
         organization.membership
+          ?.filter((mem) => !mem.endDate)
```
donni106, line 75: `mem` could be `membership` as well here like in the next line, because it is the same, right?

src/screens/OParl/OParlOverviewScreen.tsx
```diff
@@ -39,11 +39,11 @@ export const OParlOverviewScreen = ({ navigation }: Props) => {
     <SafeAreaViewFlex>
       <ScrollView>
         <WrapperWithOrientation>
-          <Tile
+          {/* <Tile
             icon={<Icon xml={oParlSearch(colors.primary)} />}
             title={overviewTexts.search}
             onPress={() => navigation.navigate('OParlSearch', { title: overviewTexts.search })}
-          />
+          /> */}
```
donni106, line 46: delete this instead commenting out?

src/components/oParl/previews/PersonPreview.tsx
```diff
@@ -4,26 +4,32 @@ import { texts } from '../../../config';
...
-    <OParlPreviewEntry
-      id={id}
-      type={type}
-      title={title}
-      topDivider={true}
-      navigation={navigation}
-      screenTitle={texts.oparl.person.person}
-    />
-  );
+  const params = { id, type, title: texts.oparl.person.person };
```
donni106, line 23: `person.person`? could there be a different subkey for better readability? i have not checked the texts file to see myself, sorry for that.

src/components/oParl/Person.tsx
```diff
@@ -46,6 +47,10 @@ export const Person = ({ data, navigation }: Props) => {
     web
   } = data;
 
+  const faction = membership?.find(
+    (membership) => membership.organization?.classification === 'Fraktion' && !membership.endDate
+  );
```
donni106, line 52: a faction here is this compared to a faction in /PersonPreview.tsx 

```
const faction = membership?.find(
    (membership) => membership.organization?.classification === 'Fraktion' && !membership.endDate
  )?.organization;
```

which confuses me. it could be the organization here as well, right? as that is what is used in `right` prop of the `Row`

src/components/oParl/Person.tsx
```diff
@@ -54,6 +59,18 @@ export const Person = ({ data, navigation }: Props) => {
...
       <Row left={personTexts.familyName} right={familyName} leftWidth={leftWidth} />
+      <Row
+        left={personTexts.faction}
+        right={faction?.organization && getOrganizationNameString(faction.organization)}
+        leftWidth={leftWidth}
+        onPress={() =>
+          navigation.push('OParlDetail', {
+            id: faction?.id,
+            type: faction?.type,
+            title: texts.oparl.membership.membership
```
donni106, line 70: `membership.membership`? could there be a different subkey for better readability? i have not checked the texts file to see myself, sorry for that.

src/components/oParl/Person.tsx
donni106: hm ok i see in `onPress` even more is used from a faction, which is different to the other file.
than maybe do the explicit `.organization` as well in the other file to have no differences in nearly the same?

src/components/oParl/previews/PersonPreview.tsx
donni106: i had a look.

```
person: {
      affix: 'Namenszusatz:',
      body: 'Körperschaft',
      email: 'Email:',
      faction: 'Fraktion:',
      familyName: 'Familienname:',
      formOfAddress: 'Anrede:',
      gender: 'Geschlecht:',
      givenName: 'Vorname:',
      license: 'Lizenz: ',
      life: 'Leben: ',
      lifeSource: 'Quelle zu Lebensangaben: ',
      location: 'Ort',
      membership: 'Mitgliedschaften',
      name: 'Name:',
      person: 'Person',
      persons: 'Personen',
      phone: 'Telefon:',
      status: 'Status:',
      title: 'Titel:'
    },
```

should be fine. could think of `singular` and `plural` or something, but no need to change that, i would say.

donni106 (approved):

src/screens/OParl/OParlPersonsScreen.tsx
digorath: renamed to mem

src/screens/OParl/OParlOverviewScreen.tsx
digorath: deleted

src/components/oParl/previews/PersonPreview.tsx
digorath: :ok:

src/components/oParl/Person.tsx
digorath: i changed it to be aligned in bit files.

PR #229, opened 2021-05-26
fix(oparl android): crash on overview screen

digorath:
- with the removal of the diagonal gradient the touchable had more than one child
  - the `TouchableNativeFeedback` requires exactly one child
- reintroduced a height for the wrapping views because the screen seemed very empty with only 3 entries being cramped at the top

## How to test:
* [x] Android portrait mode
* [x] iOS portrait mode
* [x] Android landscape mode
* [x] iOS landscape mode


donni106 (approved):
good catches.
have not checked it visually in a phone or simulator.

PR #236, opened 2021-06-17
Survey: Detail (first layout) and Overview

digorath:

- added survey overview and detail screens
  - survey detail screen will be expanded on in a future pull request
  - just displaying the data is the extent of the current ticket
- added survey queries
  - [x] hook up the code, once api is actually available

---

SVA2-9, SVA2-19, SVA2-14

---

## How to test:
* [ ] Android portrait mode
* [ ] iOS portrait mode
* [ ] Android landscape mode
* [ ] iOS landscape mode

## Screenshots:
Screenshot 2021-06-18 at 11 25 05
Screenshot 2021-06-18 at 11 25 15
Screenshot 2021-06-18 at 11 25 09


src/helpers/index.js
```diff
@@ -18,4 +18,5 @@ export * from './refreshIntervalHelper';
 export * from './searchHelper';
 export * from './shareHelper';
 export * from './storageHelper';
+export * from './surveyHelpers';
```
donni106, line 21: plural while others are singular?

src/helpers/surveyHelpers.tsx
```diff
@@ -0,0 +1,13 @@
...
+  localeObj?: Record<string, string | undefined>
+) => {
+  if (!localeObj) {
+    return;
+  }
+
+  return languages
+    .map((language) => localeObj?.[language])
+    .filter((langTitle) => langTitle?.length)
+    .join(' / ');
```
donni106, line 12: instead of joining with / the result should be a new line if there is polish text and italic style for polish.
i do not know if this is handled in a follow up already or if it is even possible in an easy way with our existing structures

src/hooks/surveyHooks.ts
```diff
@@ -0,0 +1,2 @@
+// TODO: implement properly
+export const useSurveyLanguages = () => ['de', 'pl'];
```
donni106, line 2: this todo is handled already in a follow up PR or still todo?

digorath: this is a todo that is still a todo because there is no endpoint yet as far as i know

src/navigation/AppStackNavigator.js
```diff
@@ -58,6 +60,7 @@ export const AppStackNavigator = (headerRight = true) => {
         options={screenOptionsWithSettings(headerRight)}
       />
       <Stack.Screen name="BookmarkCategory" component={BookmarkCategoryScreen} />
+      <Stack.Screen name="Category" component={IndexScreen} />
```
donni106, line 63: shouldn't be included here?

src/helpers/surveyHelpers.tsx
digorath: it is done partially.

for the `SectionHeader`s it is still with the ` / ` to avoid two `SectionHeader`s following each other.
for the overview it is still with a ` /  ` as well.

I was allowed to do layout changes in the call about the wireframe, in case some things do not fit properly.

That is also the reason for the dates being in two lines, instead of left and right (future PR)

src/navigation/AppStackNavigator.js
digorath: yes it should. i moved it up to sort by `name`

src/helpers/surveyHelpers.tsx
digorath: (also SVA2-19 does not speak of any wireframe yet, as this ticket was pre wireframe)

src/screens/SurveyOverviewScreen.tsx
```diff
@@ -0,0 +1,83 @@
...
+import { combineLanguages } from '../helpers';
+import { usePullToRefetch, useSurveyLanguages } from '../hooks';
+import { SURVEYS } from '../queries/survey';
+import { Survey } from '../types';
+
+const useSurveySections = () => {
+  const { data: surveys, loading, refetch } = useQuery<{
+    ongoing: Survey[];
+    archived: Survey[];
+  }>(SURVEYS, { fetchPolicy: 'cache-and-network' });
```
donni106, line 19: is the fetch policy safe? we cannot test changing a survey title for example, but i am not sure, if this would be recognized.
for "default" graphql queries (`graphqlFetchPolicy` in `graphqlHelper`) we use `network-only` if no cache refresh logic kicks in (`cache-first`)

src/screens/SurveyOverviewScreen.tsx
```diff
@@ -0,0 +1,83 @@
...
+
+  const sectionData = [];
```
donni106, line 21: rename to `surveySections` to be able to return `{ loading, refetch, surveySections };`? or is `sectionData` somehow better for readability?

src/screens/SurveyOverviewScreen.tsx
digorath: we will see, once we have count manipulation

src/types/Survey.ts
```diff
@@ -0,0 +1,15 @@
+export type Survey = {
+  id: string;
+  title?: Record<string, string | undefined>;
+  questionTitle: Record<string, string | undefined>;
+  description?: Record<string, string | undefined>;
+  dates: {
+    dateStart: string;
+    dateEnd: string;
+  };
```
donni106, line 9: is string for dates correct? as it is not used here elsewhere i assume this could be changed later (and is probably already, right?)

src/screens/SurveyOverviewScreen.tsx
digorath: hmm. can't think of a good reason to not rename it, so i will do it

src/types/Survey.ts
donni106: on the other hand, what the api returns is string, jo

Bildschirmfoto 2021-06-18 um 16 37 47

src/screens/SurveyOverviewScreen.tsx
```diff
@@ -0,0 +1,83 @@
...
+  section: { title, data }
+}: {
+  section: { title?: string; data: Survey[] };
+}) => {
+  if (!data.length || !title) return null;
+
+  return <SectionHeader title={title} />;
+};
+
+export const SurveyOverviewScreen = () => {
```
donni106, line 56: i think here is some landscape stuff missing:

Bildschirmfoto 2021-06-18 um 16 39 31

src/types/Survey.ts
digorath: string is correct as this is the type we get from the query, e.g. `2020-12-20`

there is a change from `dates` -> `date` in a future PR because the field name was actually incorrect

src/helpers/surveyHelpers.tsx
donni106: > That is also the reason for the dates being in two lines, instead of left and right (future PR)

i assumed already that this will be necessary when seen the wireframe first time

src/hooks/surveyHooks.ts
donni106: all right

src/screens/SurveyOverviewScreen.tsx
digorath: yes and no.
the safe area view is missing, but it not having a cutoff on both sides is in line with the tours and locations view

Screenshot 2021-06-18 at 16 42 08

src/navigation/AppStackNavigator.js
donni106: ah i thought it is added here out of context. my fault.

src/types/Survey.ts
donni106: ah true, i see now as well

src/helpers/index.js
digorath: changed to singular

src/screens/SurveyOverviewScreen.tsx
digorath: added safe area view now

src/screens/SurveyOverviewScreen.tsx
digorath: done

src/screens/SurveyOverviewScreen.tsx
donni106: nice

Bildschirmfoto 2021-06-18 um 17 00 49

donni106 (approved):

PR #237, opened 2021-06-18
Survey: response options

digorath:

- Overhauled the detail screen
  - added unique selection logic to response options 

- [x] retarget to epic once #236 is merged

---

SVA2-11

Screenshot 2021-06-18 at 12 17 21

---

## How to test:
* [ ] Android portrait mode
* [ ] iOS portrait mode
* [ ] Android landscape mode
* [ ] iOS landscape mode

## Screenshots:

Screenshot 2021-06-18 at 11 56 10
Screenshot 2021-06-18 at 11 58 39
Screenshot 2021-06-18 at 11 56 24
Screenshot 2021-06-18 at 11 56 19

### DIFF
--- a/src/components/SurveyAnswer.tsx
+++ b/src/components/SurveyAnswer.tsx
@@ -1,0 +1,59 @@
+import React, { useCallback } from 'react';
+import { StyleSheet, View } from 'react-native';
+import { colors } from '../config';
+import { useSurveyLanguages } from '../hooks';
+import { ResponseOption } from '../types';
+import { Radiobutton } from './Radiobutton';
+import { BoldText, RegularText } from './Text';
+import { Touchable } from './Touchable';
+import { Wrapper, WrapperRow } from './Wrapper';
+
+type Props = {
+  faded: boolean;
+  index: number;
+  responseOption: ResponseOption;
+  selected: boolean;
+  setSelection: (id: string) => void;
+};
+
+export const SurveyAnswer = ({ faded, index, responseOption, selected, setSelection }: Props) => {
+  const languages = useSurveyLanguages();
+
+  const { id } = responseOption;
+  const onPress = useCallback(() => setSelection(id), [id, setSelection]);
+
+  const fadeStyle = { opacity: faded ? 0.5 : 1 };
+
+  return (
+    <Touchable onPress={onPress}>
+      <Wrapper style={[styles.noPaddingBottom, fadeStyle]}>
+        <View style={styles.border}>
+          <WrapperRow>
+            <Wrapper style={styles.radioButtonContainer}>
+              <Radiobutton selected={selected} />
+            </Wrapper>
+            <Wrapper>
+              <BoldText>Antwort {String.fromCharCode(65 + index)}:</BoldText>
+              <RegularText>{responseOption.title[languages[0]]}</RegularText>
+              <BoldText italic>Odpowiedź {String.fromCharCode(65 + index)}:</BoldText>
+              <RegularText italic>{responseOption.title[languages[0]]}</RegularText>
+            </Wrapper>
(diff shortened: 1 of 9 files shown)


src/config/texts.js
```diff
@@ -404,7 +404,29 @@ export const texts = {
...
+    answerLabelPrefix: {
+      de: 'Antwort',
+      pl: 'Odpowiedź'
+    },
+    dateEnd: {
+      de: 'Abschluss der Umfrage:',
+      pl: 'Wypełnienie ankiety:'
+    },
+    dateStart: {
+      de: 'Start der Umfrage:',
```
donni106, line 417: `Startdatum der Umfrage` in ticket

src/components/Text.js
```diff
@@ -95,6 +107,11 @@ export const RegularText = styled(Text)`
 
 export const BoldText = styled(RegularText)`
   font-family: titillium-web-bold;
+  ${(props) =>
+    props.italic &&
+    css`
+      font-family: titillium-web-bold-italic;
+    `};
```
donni106, line 114: as `BoldText` is based on `RegularText` `italic` is already working, since it's part of regular.
can be removed here.

digorath: it is still needed because the italic prop of regular text uses the regular font weight and not the bold font weight

donni106: have you tried out? you can do `<BoldText italic ...` because of the dependence

src/components/SurveyAnswer.tsx
```diff
@@ -0,0 +1,66 @@
...
+
+const getAnswerLabel = (lang: 'de' | 'pl', index: number) => {
+  return `${texts.survey.answerLabelPrefix[lang]} ${String.fromCharCode(65 + index)}:`;
+};
+
+export const SurveyAnswer = ({ faded, index, responseOption, selected, setSelection }: Props) => {
+  const languages = useSurveyLanguages();
+
+  const { id } = responseOption;
+  const onPress = useCallback(() => setSelection(id), [id, setSelection]);
```
donni106, line 27: should it be a part already to not make selections in archived surveys?

digorath: i dont see that being part of any ticket

src/components/Text.js
digorath: ok. i tried it out now and i actually overlooked a thing.

yes the `italic` prop gets passed down.

no it does not do anything.

the font-family is already overwritten by the font-family set `BoldText` which the `italic` prop for `BoldText` then overwrites again.

so removing this results in the text becoming normal bold text which is no longer italic, even if we set the `italic` prop.

donni106: thanks for detailed explanation.
bold and italic are both set per font-family which i overlooked!
so `titillium-web-bold-italic` is introduced. i have not read before, that it is "bold-italic". i thought of a combination of the bold style and an italic style.

thanks. no everything makes sense!

donni106 (approved):
✔️ if every discussion is resolved

src/components/SurveyAnswer.tsx
donni106: cool, will you address that in the daily?

src/config/texts.js
digorath: as discussed in daily: `Start der Umfrage:` is what we want here

src/components/Text.js
digorath: :+1:

src/components/SurveyAnswer.tsx
digorath: addressed.

not part of this ticket, but should be done.

will do another PR that adds this logic in the end

PR #263, opened 2021-07-07
fix(drawer content): spilling over to the bottom

digorath:

- the drawer content was overflowing the screen to the bottom
  - adding a flex to the safe area view improved the behaviour

## How to test:
* [x] Android portrait mode
* [x] iOS portrait mode
* [x] Android landscape mode
* [x] iOS landscape mode

### DIFF
--- a/src/navigation/CustomDrawerContentComponent.js
+++ b/src/navigation/CustomDrawerContentComponent.js
@@ -3,10 +3,9 @@ import React, { useContext } from 'react';
 import { Platform, StyleSheet, TouchableOpacity, View } from 'react-native';
 import { Icon } from 'react-native-elements';
 import { getStatusBarHeight } from 'react-native-status-bar-height';
-import { SafeAreaView } from 'react-native-safe-area-context';
 
 import { colors, device } from '../config';
-import { DiagonalGradient } from '../components';
+import { DiagonalGradient, SafeAreaViewFlex } from '../components';
 import { OrientationContext } from '../OrientationProvider';
 
 import DrawerNavigatorItems from './DrawerNavigatorItems';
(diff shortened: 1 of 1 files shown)


src/navigation/CustomDrawerContentComponent.js
```diff
@@ -20,8 +20,8 @@ export const CustomDrawerContentComponent = ({ navigation, drawerRoutes, state }
 
   return (
     <DiagonalGradient>
-      <SafeAreaView>
-        <View style={stylesWithProps({ orientation }).header}>
+      <SafeAreaView style={styles.safeAreaView}>
```
donni106, line 23: could be our `SafeAreaViewFlex`

src/navigation/CustomDrawerContentComponent.js
```diff
@@ -20,8 +20,8 @@ export const CustomDrawerContentComponent = ({ navigation, drawerRoutes, state }
...
+        <View style={styles.header}>
```
donni106, line 24: right that the style is not depending on orientation. but that change could be reverted to have less code changes?
on the other hand it makes sense to separate the styles depending on orientation and not depending. hm

donni106 (approved):
checked in multiple devices sharing android screenshots with @digorath per slack.

src/navigation/CustomDrawerContentComponent.js
digorath: true. changed

src/navigation/CustomDrawerContentComponent.js
digorath: i reverted the style changes for now.

PR #270, opened 2021-07-09
Refactor/sva 247 centralize icons

digorath:

- Added new Icon components
- replaced the old Icon and TabBarIcon components
- Icons are now defined in a central place within the config
  - this allows for more easy maintainance of multiple branches with differing icons/iconsets as well as replacing an icon that is used in multiple places

---

SVA-247

---

## How to test:
* [ ] Android portrait mode
* [ ] iOS portrait mode
* [ ] Android landscape mode
* [ ] iOS landscape mode

## Screenshots:

- No screenshots given as everything should still look the same


digorath:
- [ ] check if some props can be pulled into the icon config

donni106 (approved):
> No screenshots given as everything should still look the same

agree

PR #273, opened 2021-07-12
feature(generic items): construction sites

digorath:

- use construction sites from generic items
  - added a (partial) type for generic items
  - added parsing from generic item to construction site type
  - changed components to no longer use construction sites from static contents
- added more specific query for construction sites
- removed ConstructionSiteProvider
- renamed unused function in momentHelper and using it now

---

SVA-190

---

## TODO:

- [x] add missing location description, once it is provided
- [x] update description to use HTML component, if required


src/hooks/constructionSites.ts
```diff
@@ -0,0 +1,91 @@
...
+): {
+  constructionSites: ConstructionSite[];
+  loading: boolean;
+  refresh: () => void;
+  refreshing: boolean;
+} => {
+  const { isConnected, isMainserverUp } = useContext(NetworkContext);
+  const [refreshing, setRefreshing] = useState(false);
+
+  const fetchPolicy = graphqlFetchPolicy({ isConnected, isMainserverUp });
```
donni106, line 61: do we explicitly do not want to have a `refreshTime` for constructions?

src/hooks/constructionSites.ts
```diff
@@ -0,0 +1,91 @@
...
+    cause: constructionSite.payload.cause,
+    description: constructionSite.contentBlocks?.[0]?.body,
+    direction: constructionSite.payload.direction,
+    endDate: constructionSite.dates?.[0]?.dateEnd,
+    imageUri: constructionSite.mediaContents?.[0]?.sourceUrl.url,
+    location,
+    locationDescription: constructionSite.locations?.[0]?.name,
+    restrictions: constructionSite.payload.restrictions?.map((value) => value.description)
+  };
+};
```
donni106, line 48: could this better be part of `queryDataParser` helper to reduce the code to the hook?

src/queries/constructionSite.js
```diff
@@ -0,0 +1,50 @@
...
+          url
+        }
+      }
+      dates {
+        id
+        dateEnd
+        dateStart
+        timeDescription
+        timeEnd
+        timeStart
```
donni106, line 45: do we need that time values?
are there other values, which we do not use on the construction screen?

src/screens/ConstructionSiteDetailScreen.js
```diff
@@ -58,16 +60,28 @@ export const ConstructionSiteDetailScreen = ({ route }) => {
...
-
-  useMatomoTrackScreenView(`${MATOMO_TRACKING.SCREEN_VIEW.CONSTRUCTION_SITE_DETAIL} / ${title}`);
+  } = constructionSites[0];
 
   const extendedTitle = (category ? `${category} - ` : '') + title;
   const formattedDates = formatDates(startDate, endDate);
 
+  if (loading && !refreshing) {
+    return <LoadingSpinner loading />;
+  }
```
donni106, line 70: i am not seeing the spinner when visiting detail screens. can you see it?

src/screens/ConstructionSiteDetailScreen.js
```diff
@@ -58,16 +60,28 @@ export const ConstructionSiteDetailScreen = ({ route }) => {
...
+
   return (
     <SafeAreaViewFlex>
-      <ScrollView keyboardShouldPersistTaps="handled">
+      <ScrollView
+        keyboardShouldPersistTaps="handled"
```
donni106, line 75: we do not have the need for some keyboard stuff on this screen, i guess.

src/types/GenericItem.ts
```diff
@@ -0,0 +1,31 @@
+import { ContentBlock } from './ContentBlock';
+import { DataProvider } from './DataProvider';
+import { SVA_Date } from './Date';
+import { MediaContent } from './MediaContent';
+import { WebUrl } from './WebUrl';
+
+// This type is not complete. Only the currently used fields have been added.
```
donni106, line 7: could this comment better be flagged with `TODO:`, `HINT:`, `NOTE:` or something to be more catchy for our future selfs?

src/screens/ConstructionSiteOverviewScreen.js
```diff
@@ -36,9 +35,9 @@ export const ConstructionSiteOverviewScreen = ({ navigation }) => {
           : `${formattedStartDate} - ${formattedEndDate}`;
 
       const propItem = {
-        title,
+        title: title,
```
donni106, line 38: just `title` would be straight forward, right?

src/screens/ConstructionSiteDetailScreen.js
```diff
@@ -39,12 +39,14 @@ const formatDates = (startDate, endDate) => {
...
 export const ConstructionSiteDetailScreen = ({ route }) => {
-  const { constructionSites } = useContext(ConstructionSiteContext);
+  const id = route.params?.id;
+  const { constructionSites, loading, refresh, refreshing } = useConstructionSites(id);
 
-  const index = route.params?.index;
+  useMatomoTrackScreenView(`${MATOMO_TRACKING.SCREEN_VIEW.CONSTRUCTION_SITE_DETAIL} / ${id}`);
 
-  if (typeof index !== 'number') return null;
+  if (!id || !constructionSites.length) return null;
```
donni106, line 49: we should not track a screen with an `id` that could be undefined

donni106 (changes requested):
looks good n general but please see the details in the open conversations

src/screens/ConstructionSiteOverviewScreen.js
digorath: ✔️

src/screens/ConstructionSiteDetailScreen.js
digorath: i added a fallback here, which should generally never be used. we only navigate here from the overview screen, and by selecting a specific construction site, which has to have an id.

src/screens/ConstructionSiteDetailScreen.js
digorath: for me the loading is generally already finished, by the time it has properly transitioned. if we set `true` as the condition here, it properly shows the loading spinner.

did you encounter an empty screen, instead of a loadingspinner at some point, or did you also have the data present immediately?

src/screens/ConstructionSiteDetailScreen.js
digorath: removed it 🪓

src/queries/constructionSite.js
digorath: i don't think we make use of them at the moment, so i removed them for now.

the copyright and captiontext of the mediacontent were also not being used, but instead of removing them i added them to the image component.

src/hooks/constructionSites.ts
digorath: i did not like putting them into the queryDataParser file as well, because that is already rather big.

instead i moved the queryDataParser file to a new subdirectory `parser` and renamed it to `listItemParser` and added a new file for the construction site parsing.

src/hooks/constructionSites.ts
digorath: i don't think we do. so i added it

src/types/GenericItem.ts
digorath: added `HINT`

src/screens/ConstructionSiteDetailScreen.js
donni106: i did some checks and also had the data on screen directly. the state of visible loading spinner did never occur, which is fine i would say.

why is there `!refreshing` included in the condition? we have just `loading` in other places. isn't that enough here as well?

donni106 (approved):
✔️ but please see two open conversations

src/screens/ConstructionSiteDetailScreen.js
digorath: you are right, we do not need the `!refreshing` condition here. i will remove it

src/screens/ConstructionSiteDetailScreen.js
digorath: in theory we could, but if we navigate to this screen by accident through other means (static contents or w/e) we might run into problems, because we can not guarantee the id being present.

so in my previous comment i should have phrased it as "we intend to only ever navigate there from the overview screen"

so in our expected cases the `!id` condition will always be false and not influence the outcome of the check. it is still reasonable to keep it there to avoid possible crashes, in the case of the app not being configured properly

PR #277, opened 2021-07-20
Feature: survey widget

digorath:

- added new survey widget
  - the widget shows the current number of ongoing surveys and navigates to the survey overview if clicked
---

SVA-256


donni106 (approved):
Screenshot 2021-07-20 at 12 22 12

PR #283, opened 2021-07-23
chore(html table): exchange dependency

digorath:

- the react-native-render-html-table-bridge is outdated and deprecated
  - instead added a version of @native-html/table-plugin that is compatible with our version of react-native-render-html
- adjusted some stylings to make it close to how it was before

## How to test:
* [x] Android portrait mode
* [x] iOS portrait mode
* [x] Android landscape mode
* [x] iOS landscape mode

### DIFF
--- a/src/components/HtmlView.js
+++ b/src/components/HtmlView.js
@@ -24,14 +24,30 @@ const cssRules =
     trOddBackground: colors.lightestText,
     trOddColor: colors.darkText,
     trEvenBackground: colors.lightestText,
-    trEvenColor: colors.darkText,
-    outerBorderWidthPx: 1,
-    outerBorderColor: '#b5b5b5',
-    columnsBorderWidthPx: 1
+    trEvenColor: colors.darkText
   }) +
   `
 :root {
   font-size: ${normalize(13)}px;
+}
+th {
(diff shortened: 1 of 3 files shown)


src/components/HtmlView.js
```diff
@@ -2,23 +2,23 @@ import PropTypes from 'prop-types';
...
 
-const tableCssRules =
+const cssRules =
   cssRulesFromSpecs({
     ...defaultTableStylesSpecs,
     linkColor: colors.primary,
-    // TODO: font family was not working at the moment with following implementation
-    //       https://github.com/jsamr/react-native-render-html-table-bridge/blob/master/readme.md#how-to-load-custom-fonts
+    // TODO: font family documentation
+    //      https://github.com/native-html/plugins/tree/rnrh/4.x#how-to-load-custom-fonts
```
donni106, line 21: i was not able to make it work and asked for more information here https://github.com/native-html/plugins/issues/13#issuecomment-885781518

donni106 (approved):
we could create an issue for the fonts and see what the time brings? the rest works fine as before.

src/components/HtmlView.js
digorath: created #293 for it

PR #292, opened 2021-07-30
ts config: apply expo setup

digorath:
- removed "EXPO_NO_TYPESCRIPT_SETUP" from yarn start
  - applied expo setup for typescript
  - auto imports work fine again with current vscode

Please double check with your vscode @donni106


donni106 (approved):
> Please double check with your vscode @donni106

i cannot remember exactly the problem that occurred, but "everything" works fine for me with this branch.

i have added a eslint disable comment that was pointed out when running `yarn lint`.

PR #295, opened 2021-07-30
chore(shareContent): improve type of openShare

### LINKED ISSUE #294: openShare type handling
Currently the type of `shareContent` in `shareHelper.js -> openShare` is handled cleanly.

This causes there to be a wrong type in shareHeader (and as soon as it is merged in the new version of `useOpenWebScreen`)

Please improve the behaviour.

digorath:
- added the correct type to openShare and fixed the types for ShareHeader and useOpenWebScreen

closes #294

### DIFF
--- a/src/helpers/shareHelper.js
+++ b/src/helpers/shareHelper.js
@@ -6,6 +6,9 @@ import { QUERY_TYPES } from '../queries';
 import { momentFormat } from './momentHelper';
 
 // https://facebook.github.io/react-native/docs/share
+/**
+ * @param {{ title?: string; message: string; } | { title?: string; url: string; }} shareContent
+ */
 export const openShare = async ({ message, title, url }) => {
   try {
     const result = await Share.share(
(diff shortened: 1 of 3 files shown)


src/helpers/shareHelper.js
```diff
@@ -6,6 +6,9 @@ import { QUERY_TYPES } from '../queries';
 import { momentFormat } from './momentHelper';
 
 // https://facebook.github.io/react-native/docs/share
+/**
+ * @param {{ title?: string; message: string; } | { title?: string; url: string; }} shareContent
+ */
```
SoleCincis, line 11: `|` This operator performs a bitwise OR operation, so what is the differences with` ||`

why is it conditional now:  title + message or title + url ?
and why types are commented out , is this logic still work in progress ?

digorath: https://jsdoc.app/about-getting-started.html

it is neither a bitwise or, not an or. it is a union type.

this is not work in progress either, and it is right that it is in the form of this comment (that is how jsdoc works)

digorath: we also use it here, for example: https://github.com/ikuseiGmbH/smart-village-app-app/blob/master/src/helpers/momentHelper.js#L4

donni106 (approved):

PR #311, opened 2021-08-27
fix: add order to event widget query

donni106:
the order was missing in the query to be the same as in the events index screen

### DIFF
--- a/src/components/widgets/EventWidget.tsx
+++ b/src/components/widgets/EventWidget.tsx
@@ -22,7 +22,8 @@ export const EventWidget = ({ text }: WidgetProps) => {
 
   const currentDate = moment().format('YYYY-MM-DD');
   const queryVariables = {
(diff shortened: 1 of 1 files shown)


digorath (approved):
i thought this was the default value. 🤔 

is there some kind of documentation anywhere (except for the server code itself)?

should we add these strings (`listDate_ASC`) to some kind of enum in the future?

donni106:
> is there some kind of documentation anywhere (except for the server code itself)?

no. if we would take a look at the graphql schema, this would also be unclear. maybe there is a way to document the default in the graphql schema, or to make clear, that this option is not the default.

> should we add these strings (`listDate_ASC`) to some kind of enum in the future?

good idea. there are a fix number of order variations for each model, which we could perfectly keep in sync across the main-server and mobile-app code that way.

PR #321, opened 2021-09-10
feat(encounter): scanner screen

digorath:

- replaced expo barcode scanner with expo camera
  - barcode scanner was not be able to properly resized for android (fixed aspect ratio)
- added new screen for scanning encouter qr codes
- added validation for encounter user type
- added dummy hooks for creating an encounter

---

SVA-287

### DIFF
--- a/src/screens/EncounterScannerScreen.tsx
+++ b/src/screens/EncounterScannerScreen.tsx
@@ -1,7 +1,7 @@
 import { BarCodeScanningResult, Camera } from 'expo-camera';
 import * as Linking from 'expo-linking';
-import React, { useEffect, useState } from 'react';
-import { useCallback } from 'react';
+import * as ScreenOrientation from 'expo-screen-orientation';
+import React, { useCallback, useEffect, useRef, useState } from 'react';
 import { Alert, ScrollView, StyleSheet } from 'react-native';
 
 import appJson from '../../app.json';
@@ -16,8 +16,17 @@ import {
 } from '../components';
 import { LoadingSpinner } from '../components/LoadingSpinner';
 import { device, texts } from '../config';
+import { sleep } from '../helpers';
 import { useCreateEncounter, useEncounterUser } from '../hooks';
 import { User } from '../types';
+
+const useOrientationLock = () =>
+  useEffect(() => {
+    ScreenOrientation.lockAsync(ScreenOrientation.OrientationLock.PORTRAIT_UP);
+    return () => {
+      (() => ScreenOrientation.lockAsync(ScreenOrientation.OrientationLock.DEFAULT))();
+    };
+  }, []);
 
 const showErrorAlert = () =>
   Alert.alert(texts.encounter.errorScanTitle, texts.encounter.errorScanBody);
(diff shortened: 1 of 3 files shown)


src/hooks/encounter/encounter.ts
```diff
@@ -0,0 +1,58 @@
...
+  loading: boolean;
+} => {
+  const [loading, setLoading] = useState(false);
+
+  const createEncounter = useCallback(async (qrId: string, userId: string) => {
+    setLoading(true);
+    try {
+      const res = await createEncounterAsync(qrId, userId);
+
+      if (!res.data || res.error) {
```
donni106, line 45: i think it will be necessary to check for the status code to decide wether there was success or an error

digorath: i will revisit all error handling upon implementing the api calls.

i don't know whether i will include the error logic here, or already inside of the `createEncounterAsync` function.

digorath: changed in #339 
`createEncounterAsync` will now return a parsed and validated result or nothing. so we only need to check for presence of `res` here. (the used data is no longer nested as `data` but is directly returned)

src/screens/EncounterScannerScreen.tsx
```diff
@@ -0,0 +1,137 @@
...
+import { useCreateEncounter, useEncounterUser } from '../hooks';
+import { User } from '../types';
+
+const showErrorAlert = () =>
+  Alert.alert(texts.encounter.errorScanTitle, texts.encounter.errorScanBody);
+
+const parseQrCode = (data: string): string | undefined => {
+  const result = Linking.parse(data);
+
+  if (result.scheme === (__DEV__ ? 'exp' : appJson.expo.scheme) && result.path === 'encounter') {
```
donni106, line 27: i miss an explanation for `exp`. why this in dev mode? i have a clue but cannot be sure.

src/screens/EncounterScannerScreen.tsx
```diff
@@ -0,0 +1,137 @@
...
+          <BoldText>{texts.encounter.scannerSubTitle}</BoldText>
+        </Wrapper>
+      </WrapperWithOrientation>
+      {isScanning ? (
+        <Wrapper style={styles.scannerContainer}>
+          <Camera
+            ratio="1:1"
+            onBarCodeScanned={handleBarCodeScanned}
+            style={styles.scanner}
+            // barCodeTypes={['org.iso.QRCode']}
```
donni106, line 114: comment needed for later?

digorath: nope, that is a comment that is a remainder of a previous version.
this prop is not supported here.

i removed it now

src/screens/EncounterScannerScreen.tsx
```diff
@@ -0,0 +1,137 @@
...
+    [navigation]
+  );
+
+  const { loading: loadingCreateEncounter, createEncounter } = useCreateEncounter(
+    onCreationSuccess,
+    showErrorAlert
+  );
+
+  const handleBarCodeScanned = ({ data, type }: BarCodeScanningResult) => {
+    if (type !== (device.platform === 'android' ? 256 : 'org.iso.QRCode')) {
```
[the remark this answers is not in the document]
digorath: i got this from scanning the qr code, that we ourselves created.

this way we do not react to codes of a different type, if one shows up in the camera view

src/screens/EncounterScannerScreen.tsx
```diff
@@ -0,0 +1,137 @@
...
+export const EncounterScannerScreen = ({ navigation }: { navigation: any }) => {
+  const { loading: loadingUser, userId } = useEncounterUser();
+  const [isScanning, setIsScanning] = useState(true);
+  const [hasPermission, setHasPermission] = useState<boolean>();
+
+  const onCreationSuccess = useCallback(
+    (data: User) => {
+      // TODO: add navigation here instead of console.log
+      // navigation.navigate(ScreenName.EncounterUserDetails, { data });
+      console.log(data);
```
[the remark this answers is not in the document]
digorath: ✔️

src/screens/EncounterScannerScreen.tsx
digorath: i added a comment for explanation

PR #324, opened 2021-09-14
Feature/sva 287 encounter history and user data

digorath:

Please explain your changes in addition to your commit messages.

---

**Add issue number, if available:**

Fixes #

---

## How to test:
* [ ] Android portrait mode
* [ ] iOS portrait mode
* [ ] Android landscape mode
* [ ] iOS landscape mode

## Screenshots:

**Add screenshots, if they are helpful**


src/screens/EncounterDataScreen.tsx
```diff
@@ -0,0 +1,214 @@
...
+  const selectImage = useCallback(async () => {
+    const { status } = await requestMediaLibraryPermissionsAsync();
+
+    if (status !== PermissionStatus.GRANTED) {
+      Alert.alert(texts.errors.image.title, texts.errors.image.body);
+      return;
+    }
+
+    const result = await launchImageLibraryAsync({
+      mediaTypes: MediaTypeOptions.Images,
```
donni106, line 60: an info somewhere or link to https://docs.expo.dev/versions/latest/sdk/imagepicker/#imagepickermediatypeoptions would have been nice.
so we do not want videos selectable but just images, what this option realizes.

src/screens/EncounterDataScreen.tsx
```diff
@@ -0,0 +1,214 @@
...
+
+  useEffect(() => {
+    if (userData.loading) {
+      return;
+    }
+
+    try {
+      setBirthDate(new Date(userData.birthDate));
+    } catch (e) {
+      console.warn('error when parsing the birthdate of the encounter user');
```
donni106, line 79: [small spelling stuff] birthdate -> birth date or birthday

src/components/Image.js
```diff
@@ -48,41 +48,60 @@ export const Image = ({
     // -> https://juliangaramendy.dev/use-promise-subscription/
     let mounted = true;
 
-    if (refreshInterval === undefined) {
-      source.uri
-        ? CacheManager.get(source.uri)
+    effect: {
+      // we do not want the refreshInterval or the caching to affect required static images or local images
+      if (!sourceProp.uri || sourceProp.uri.startsWith('file:///')) {
```
donni106, line 53: is `'file:///'` multi platform and applies to both iOS and Android?

src/components/Image.js
```diff
@@ -23,7 +23,7 @@ const addQueryParam = (url, param) => {
 };
 
 export const Image = ({
-  source,
+  source: sourceProp,
```
donni106, line 26: i have difficulties with the naming of `sourceProp`, but i understand the reason.
maybe we could rename the state variable `source` to something else to be able to keep `source` here maybe `imageSource`? or we could use `...props` here like in `ImageWithBadge` and access `props.source` everywhere where `sourceProp` is used now?

src/components/Image.js
```diff
@@ -33,7 +33,7 @@ export const Image = ({
   borderRadius,
   refreshInterval
 }) => {
-  const [uri, setUri] = useState(null);
+  const [source, setSource] = useState(null);
```
donni106, line 36: i have problems to think over this completely, but would setting `sourceProp` as initial value here make sense? but i guess there are cases, where this does not apply because of the later condition in the effect, right?

donni106 (changes requested):

src/screens/EncounterDataScreen.tsx
```diff
@@ -0,0 +1,214 @@
...
+              style={styles.inputField}
+              value={familyName}
+            />
+          </Wrapper>
+          <Wrapper style={styles.noPaddingTop}>
+            <Label>{texts.encounter.birthDate}</Label>
+            <Pressable
+              onPress={() => {
+                setIsDatePickerVisible(false);
+                setIsDatePickerVisible(true);
```
donni106, line 135: necessary way to make it work 100%, with setting false/true here?

src/screens/EncounterDataScreen.tsx
```diff
@@ -0,0 +1,214 @@
...
+    setPhone(userData.phone);
+  }, [userData]);
+
+  if (userData.loading) {
+    return <LoadingSpinner loading />;
+  }
+
+  return (
+    <SafeAreaViewFlex>
+      <ScrollView>
```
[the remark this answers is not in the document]
donni106: yes

PR #334, opened 2021-09-16
refactor(circular view): new component

digorath:

- extracted shared logic for styling circular views and reused the new component

### DIFF
--- a/src/components/CircularView.tsx
+++ b/src/components/CircularView.tsx
@@ -1,0 +1,24 @@
+import React, { useMemo } from 'react';
+import { StyleSheet, View } from 'react-native';
+
+type Props = View['props'] & { size: number };
+
+export const CircularView = ({ size, style, ...otherProps }: Props) => {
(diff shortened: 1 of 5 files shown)


src/components/CircularView.tsx
```diff
@@ -0,0 +1,24 @@
...
+};
+
+const getStyle = (size: number) =>
+  StyleSheet.create({
+    // eslint does not detect the function being called at the top
+    // eslint-disable-next-line react-native/no-unused-styles
+    circle: {
+      alignItems: 'center',
+      aspectRatio: 1,
+      borderRadius: size / 2,
```
donni106, line 19: could there be sizes, where `/ 2` will not result in circles here?
i think this should not depend on the size but always be something like 50%? is that possible, to always ensure a circle?

donni106 (changes requested):

src/components/CircularView.tsx
digorath: why do you think this should not result in a circle for some `size` values?

PR #341, opened 2021-09-28
fix(commercial): filter by date

digorath:

- added an array of types to filter instead of checking each type individually
- added `GenericType.Commercial` to the array

---

SVA-269

### DIFF
--- a/src/helpers/parser/listItemParser.js
+++ b/src/helpers/parser/listItemParser.js
@@ -13,10 +13,10 @@ import { subtitle } from '../textHelper';
 
 const { ROOT_ROUTE_NAMES } = consts;
 
-const dataFilteredGenericTypes = [GenericType.Job, GenericType.Commercial];
+const GENERIC_TYPES_WITH_DATES = [GenericType.Job, GenericType.Commercial];
 
 const filterGenericItems = (item) => {
-  if (dataFilteredGenericTypes.includes(item?.genericType)) {
+  if (GENERIC_TYPES_WITH_DATES.includes(item?.genericType)) {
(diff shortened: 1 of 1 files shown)


src/helpers/parser/listItemParser.js
```diff
@@ -13,8 +13,10 @@ import { subtitle } from '../textHelper';
 
 const { ROOT_ROUTE_NAMES } = consts;
 
+const dataFilteredGenericTypes = [GenericType.Job, GenericType.Commercial];
```
donni106, line 16: isn't this a constant and could be `DATA_FILTERED_GENERIC_TYPES`?
this would make `if (dataFilteredGenericTypes.includes(item?.genericType)) {` more readable in my opinion: `if (DATA_FILTERED_GENERIC_TYPES.includes(item?.genericType)) {`

donni106: in addition, `GENERIC_TYPES` would be enough, right? to make the question "if generic types includes the items generic type"

digorath: 1. yes it could be a constant
2. it was supposed to be `date....` not `data....`, and with that in mind i don't like reducing the name to `GENERIC_TYPES`.

I will fix it once we agree on a naming :)

donni106: the array contains two generic type strings. where do you see a date reference?

donni106: because of the later context, right?
so what about `GENERIC_TYPES_WITH_DATES`?

digorath: ✅

donni106 (approved):

PR #342, opened 2021-09-28
fix(lunch): wrong default for poiId

digorath:

During the react navigation upgrade a wrong default was added to the `poiId` of the lunches, resulting in wrong filtering behaviour.

---

SVA-266

### DIFF
--- a/src/screens/LunchScreen.js
+++ b/src/screens/LunchScreen.js
@@ -30,7 +30,7 @@ import { getQuery, QUERY_TYPES } from '../queries';
 const { MATOMO_TRACKING } = consts;
 
 export const LunchScreen = ({ navigation, route }) => {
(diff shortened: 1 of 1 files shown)


src/screens/LunchScreen.js
```diff
@@ -116,6 +116,7 @@ export const LunchScreen = ({ navigation, route }) => {
 
   return (
     <SafeAreaViewFlex>
+      <RegularText>{JSON.stringify(poiId)}</RegularText>
```
donni106, line 119: Is this logging and should be removed?

donni106 (changes requested):

src/screens/LunchScreen.js
digorath: Yes.

donni106 (approved):

PR #343, opened 2021-09-29
Feature/sva 287 encounter a11y

digorath:

Added accessibility labels to various places throughout the new screens

---

SVA-297


src/screens/EncounterDataScreen.tsx
```diff
@@ -98,14 +99,19 @@ export const EncounterDataScreen = ({ navigation }: StackScreenProps<any>) => {
...
+                style={styles.editIconContainer}
+              >
                 <Icon.EditSetting color={colors.placeholder} />
               </TouchableOpacity>
             </WrapperRow>
           </Wrapper>
           <Wrapper style={styles.noPaddingTop}>
             <Label>{texts.encounter.firstName}</Label>
             <TextInput
+              accessibilityLabel={`${a11yLabels.firstName} ${a11yLabels.textInput}: ${firstName}`}
```
donni106, line 114: will this be read out every time the input value changes?

digorath: good question. if it does it is the same way for the `FormScreen`. I will check

digorath: it does not read on every change.
it did however read "undefined" because it was not initialised with an empty string. so i fixed it ;)

donni106 (approved):

PR #348, opened 2021-09-30
feat(encounter): added welcome screen

digorath:

- updated usePullToRefetch
  - it no longer relies on the loading state of the query
- updated encounter user type
  - there is no longer a UserWithId type
  - instead the hook now returns the userId as an extra value

---

SVA-287

## Screenshots:

Screenshot 2021-09-30 at 14 12 55
Screenshot 2021-09-30 at 14 18 03


digorath:
@donni106 i added a `useStaticContent` hook locally in the `welcome` component. this could be improved, extracted and reused in many places. what are your thoughts?

src/components/encounter/EncounterWelcome.tsx
```diff
@@ -0,0 +1,142 @@
...
+  });
+
+  const { data, error: queryError, loading, refetch } = useQuery(
+    getQuery(QUERY_TYPES.PUBLIC_JSON_FILE),
+    {
+      variables: { name: publicJsonFile },
+      fetchPolicy,
+      skip: !refreshTime
+    }
+  );
```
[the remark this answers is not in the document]
donni106: maybe the whole `refreshTime`, `fetchPolicy` and `useQuery` part can be outsourced from components?

src/components/encounter/EncounterWelcome.tsx
```diff
@@ -0,0 +1,142 @@
...
+    } catch (error) {
+      setError(true);
+      console.warn(error, data);
+    }
+  }, [data, parseJson]);
+
+  return {
+    data: publicJsonFileData,
+    error: !!(queryError || error),
+    loading: loading || (!publicJsonFileData && !error),
```
donni106, line 82: reading the second line i like the way that `loading: loading ...` marks the "extended" loading condition. the first line is the other way around for no functional reason. let's do `error: !!(error || queryError)` as well to point out, that we pass through the value with additional condition?

src/hooks/encounter/user.ts
```diff
@@ -3,41 +3,44 @@ import { useCallback, useContext, useEffect, useState } from 'react';
...
+  const [userId, setUserId] = useState<string>();
   const [refreshing, setRefreshing] = useState(false);
 
   const loadUserCallback = useCallback(async () => {
     setRefreshing(true);
     setError(false);
 
     try {
-      setUser(await loadUser());
+      const storedId = await getEncounterUserId();
```
donni106, line 38: `storedId` regarded out of context very generic. as we handle different ids (user, encounters, ...) this can lead to misunderstanding if not traversing the code, maybe... `const storedEncounterUserId = await getEncounterUserId();` or `const storedUserId = await getEncounterUserId();` would be more precise, right?

src/screens/EncounterHomeScreen.tsx
```diff
@@ -62,6 +64,10 @@ export const EncounterHomeScreen = ({ navigation }: any) => {
     return <LoadingSpinner loading />;
   }
 
+  if (!userId) {
+    return <EncounterWelcome navigation={navigation} />;
```
donni106, line 68: so we have the welcome screen just if there is no user already present, right?

src/components/encounter/EncounterWelcome.tsx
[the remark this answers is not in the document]
donni106: created #360 for somewhen

PR #352, opened 2021-10-01
Feature/sva 298 sort pois by distance

digorath:

- the app will now react to a new server provided setting for location services
  - the server setting has to be located besides the settings for matomo etc. in global settings -> settings
  - the key for the new setting is globalSettings -> settings -> locationService
  - if enabled the app will sort the POIs of a selected category by proximity to the users current location
  - locationService can be disabled in app by a new setting on the settings screen, if it is enabled by the server sided config
 
---

SVA-298

### DIFF
--- a/src/config/texts.js
+++ b/src/config/texts.js
@@ -385,6 +385,10 @@ export const texts = {
       onDeactivate:
         'Soll Matomo Analytics deaktiviert werden? Die Deaktivierung von Matomo Analytics wird mit dem nächsten Neustart der App wirksam.',
       yes: 'Ja'
+    },
+    locationService: {
+      onSystemPermissionMissing:
+        'Um diese Einstellung zu aktivieren, muss zunächst die Berechtigung die Ortungsdienste ihres Gerätes zu verwenden in den Systemeinstellungen erteilt werden.'
     }
   },
   settingsScreen: {
@@ -401,6 +405,7 @@ export const texts = {
       sectionTitle: 'Listen-Layouts',
       textList: 'Textliste'
     },
+    locationService: 'Ortungsdienste',
     pushNotifications: 'Push-Benachrichtigungen'
   },
   survey: {
--- a/src/helpers/positionHelper.ts
+++ b/src/helpers/positionHelper.ts
@@ -1,0 +1,55 @@
+type LatLon = {
+  latitude: number;
+  longitude: number;
+};
+
+const calculateDistance = (a: LatLon, b: LatLon) =>
+  Math.sqrt(Math.pow(a.latitude - b.latitude, 2) + Math.pow(a.longitude - b.longitude, 2));
+
+const sortByDistancesFromPoint = <T>(
+  inputArray: T[],
+  getLatLon: (value: T) => LatLon | undefined,
+  position: LatLon
+): T[] => {
+  const sortingArray = inputArray.map((value, index) => {
+    const latLon = getLatLon(value);
+
+    return {
+      distance: latLon && calculateDistance(latLon, position),
+      index
+    };
+  });
+
+  sortingArray.sort((a, b) => {
+    if (a.distance !== undefined && b.distance !== undefined) {
+      // if both are defined, compare the distances
+      return a.distance - b.distance;
+    } else if (a.distance !== undefined) {
+      // if only the first one is defined treat it as being closer
+      return -1;
+    } else if (b.distance !== undefined) {
+      // if only the second one is defined treat it as being closer
+      return 1;
+    } else {
+      // if both are undefined, treat them as equal
+      return 0;
+    }
+  });
+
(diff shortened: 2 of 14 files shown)


src/config/texts.js
```diff
@@ -385,6 +385,10 @@ export const texts = {
       onDeactivate:
         'Soll Matomo Analytics deaktiviert werden? Die Deaktivierung von Matomo Analytics wird mit dem nächsten Neustart der App wirksam.',
       yes: 'Ja'
+    },
+    locationService: {
+      onSystemPermissionMissing:
+        'Um diese Einstellung zu aktivieren, muss zunächst die Berechtigung die Ortungsdienste ihres Gerätes zu verwenden in den Systemeinstellungen erteilt werden.'
```
donni106, line 391: could be simpler and without du/sie...

> Um diese Einstellung zu aktivieren muss zunächst die Berechtigung für Ortungsdienste in den Systemeinstellungen erteilt werden.

like that?

donni106:
got multiple errors/warnings on my android 8.1:

1. [Unhandled promise rejection: Error: Location provider is unavailable. Make sure that location services are enabled.]
2. [Unhandled promise rejection: Error: Encountered an exception while calling native method: Exception occurred while executing exported method setNotificationChannelAsync on module ExpoNotificationChannelManager: Unknown color]

src/hooks/locationHooks.ts
```diff
@@ -0,0 +1,61 @@
...
+  const { status } = await Location.requestForegroundPermissionsAsync();
+
+  await setAndSyncLocationSettings({ sortPOIs: status === Location.PermissionStatus.GRANTED });
+
+  if (status === Location.PermissionStatus.GRANTED) {
+    return await Location.getCurrentPositionAsync({});
+  }
+};
+
+export const useLocationSettings = () => {
```
donni106, line 20: got multiple errors/warnings on my android 8.1:

1. [Unhandled promise rejection: Error: Location provider is unavailable. Make sure that location services are enabled.]
2. [Unhandled promise rejection: Error: Encountered an exception while calling native method: Exception occurred while executing exported method setNotificationChannelAsync on module ExpoNotificationChannelManager: Unknown color]

(i do not know if this applies to the code marked, but i wanted to have conversations possible after this post)

src/helpers/positionHelper.ts
```diff
@@ -0,0 +1,55 @@
+type LatLon = {
+  latitude: number;
+  longitude: number;
+};
+
+const calculateDistance = (a: LatLon, b: LatLon) =>
+  Math.sqrt(Math.pow(a.latitude - b.latitude, 2) + Math.pow(a.longitude - b.longitude, 2));
```
donni106, line 7: do you have a source for this calculation?

digorath: no, this is just rectangular approximation.
but thanks for pointing it out.

due to us being relatively far north, I added a factor of ~0.63 (north of germany would be ~0.58, south of germany 0.68) to the longitude values (see https://www.movable-type.co.uk/scripts/latlong.html -> equirectangular approximation)
if we ever want to use this feature globally (not restricted to germany) then we need to revisit this.
this is a decent enough approximation for our use case though.

I added a reference and explanation of the calculation to the code

src/config/texts.js
digorath: changed

src/hooks/locationHooks.ts
digorath: 2. should probably not have anything to do with this. and is more likely to be an issue with push notifications.

1. what is your current version of expo go on your testing device? there seems to have been an issue with google changing stuff, but for me it works fine. new versions of expo go should fix the issue. if that does not fix it, then we need to investigate further.

digorath: related links:
https://github.com/expo/expo/issues/9377
https://github.com/expo/expo/issues/14248

digorath: there might be further issues with android: currently the `getCurrentPositionAsync' call does not resolve, on my testing device if i turn off wifi. i am doing further research

digorath:
@donni106 about the unkown color for the notifications:
 * i could reproduce the issue.
 * it seems to be related to the fact that the color needs to be of the format `#rrggbb` while we were providing `rgb(rrr,ggg,bbb)`
 
 we should add a "rgbToHex" method in there

src/hooks/locationHooks.ts
digorath: it might be due to my device/android not properly supporting offline location, as the google maps app will also no longer update the location when wifi is turned off.
as of now, i am unsure if that is depending on the device, or if that is a "feature" of android in general.
do you have any idea?

PS: about the notification thing please look [here](https://github.com/ikuseiGmbH/smart-village-app-app/pull/352#issuecomment-934331027)

PR #359, opened 2021-10-06
Fix/android notification channel color

digorath:

- added color parsing for the android notification channel color
  - the color needs to be provided in a 6 digit hex format, e.g. `#1230f0`
  - we now have 3 supported color formats: `rgb(r,g,b)`, `#rgb` and `#rrggbb`

---

More info in this PR: https://github.com/ikuseiGmbH/smart-village-app-app/pull/352#issuecomment-934331027


digorath:
I updated snapshots of other test in the process

PR #381, opened 2021-10-28
Feature: allow for server provided text in the tmb notice

digorath:
previously the text was hard coded into the app. that text is still used as a fallback, but can be overwritten by a text provided through the mainserver.

SVA-171


donni106 (approved):
i made an update to use a new field `notice` without any hardcoded TMB stuff. text is/has to be from data provider now and can be adjusted per main server at any time:

IMG_504F2A096027-1

digorath (commented):
The extra commit looks good to me. 🚀

donni106:
all servers are prepared. this can be included in 2.3.0 without any further todos.

PR #386, opened 2021-11-02
Upgrade to expo sdk version 43

digorath:
- upgraded corresponding dependencies
- updated node version specified in .nvmrc
- removed react-native-safe-area-view resolution
  - this package is no longer used by us
- updated react-native-webview-leaflet fork to be compatible with new changes

SVA-351

- [x] check tests


.nvmrc
```diff
@@ -1 +1 @@
-14.3.0
+v14.18.1
```
donni106, line 1: > updated node version specified in .nvmrc

can i ask why? this is not clear for me on first sight.

src/hooks/HomeRefresh.ts
```diff
@@ -1,5 +1,5 @@
+import { DeviceEventEmitter } from 'expo-modules-core';
```
donni106, line 1: why is it necessary to get that module from expo now instead of react-native like before? who or what was changing for this?

donni106 (commented):
thanks. seems manageable. a link directly here in the commit to https://blog.expo.dev/expo-sdk-43-aa9b3c7d5541 would have been straightforward

package.json
```diff
@@ -43,29 +43,29 @@
...
     "graphql-tag": "^2.12.5",
     "lodash": "~4.17.21",
     "matomo-tracker-react-native": "^0.3.0",
     "moment": "~2.29.1",
     "qs": "^6.10.1",
-    "react": "16.13.1",
+    "react": "17.0.1",
     "react-apollo": "~3.1.5",
-    "react-native": "https://github.com/expo/react-native/archive/sdk-42.0.0.tar.gz",
+    "react-native": "0.64.2",
```
donni106, line 68: > No more react-native fork in Expo managed apps. In SDK 42 and lower, your package.json includes something like: “react-native”: “https://github.com/expo/react-native/archive/sdk-42.0.0.tar.gz". In SDK 43 and higher, it’s just this: “react-native”: “0.64.2”.

i like that very much!

.nvmrc
digorath: `error jest-regex-util@27.0.6: The engine "node" is incompatible with this module. Expected version "^10.13.0 || ^12.13.0 || ^14.15.0 || >=15.0.0". Got "14.3.0"`

`v14.18.1` is the latest `v14.x.x` lts version. there is no `v15.x.x` lts version and expo has not yet been testet with `v16.x.x` versions

src/hooks/HomeRefresh.ts
digorath: https://blog.expo.dev/whats-new-in-expo-modules-infrastructure-7a7cdda81ebc

`react-native-unimodules` is deprecated (and removed) now, and the previous import took it from there.

donni106: got it

donni106 (approved):

package.json
```diff
@@ -65,7 +65,7 @@
     "qs": "^6.10.1",
     "react": "17.0.1",
     "react-apollo": "~3.1.5",
-    "react-native": "0.64.2",
+    "react-native": "^0.64.3",
```
donni106, line 68: > Note: this post was updated on November 9th, 2021 to indicate that the React Native version has been bumped from 0.64.2 to 0.64.3.

https://blog.expo.dev/expo-sdk-43-aa9b3c7d5541

package.json
```diff
@@ -18,7 +18,7 @@
       "\\.svg": "<rootDir>/__mocks__/svgMock.js"
     },
     "transformIgnorePatterns": [
-      "node_modules/(?!(jest-)?react-native|react-clone-referenced-element|@react-native-community|expo(nent)?|@expo(nent)?/.*|react-navigation|@react-navigation/.*|@unimodules/.*|unimodules|sentry-expo|native-base|@sentry/.*|matomo-tracker-react-native|ramda)"
+      "node_modules/(?!(jest-)?@?react-native|react-clone-referenced-element|@react-native-community|expo(nent)?|@expo(nent)?/.*|react-navigation|@react-navigation/.*|@unimodules/.*|unimodules|sentry-expo|native-base|@sentry/.*|matomo-tracker-react-native|ramda)"
```
donni106, line 21: did you read this somewhere? how come that this is now needed?

jest-setup.js
```diff
@@ -1,5 +1,8 @@
 import 'whatwg-fetch';
 import 'react-native-gesture-handler/jestSetup';
+import mockAsyncStorage from '@react-native-async-storage/async-storage/jest/async-storage-mock';
+
+jest.mock('@react-native-async-storage/async-storage', () => mockAsyncStorage);
```
donni106, line 5: what if not doing this?

.nvmrc
donni106: why with `v` now? and not just `14.18.1` as before? in .yvmrc we also add version number without `v`

package.json
digorath: react native types where changed around and not having this leads to errors because of type declarations being present in places where there previously where none.

jest-setup.js
digorath: the tests fail because the async storage is not properly defined for jest, if we do not mock it.

.nvmrc
digorath: no particular reason. it came from me copying the name of the lts version, which includes the `v`. nvm acts the same way, regardless of the `v` being there or not

donni106:
> @donni106 can you confirm that tests are running and not failing with the new setup?

yes

Bildschirmfoto 2021-11-15 um 17 26 01

PR #387, opened 2021-11-03
feat(image): set original source on cache fail

### LINKED ISSUE #361: Fall back to original source of an image on cache fail
i did notice, however, that we could add a `setSource` call in the catch block to set the `source`-state to the actual prop, in case the cache fails.
this way we would then get the warning that something went wrong, but instead of never setting the image uri, we could fall back to not using the cache.

_Originally posted by @digorath in https://github.com/ikuseiGmbH/smart-village-app-app/pull/324#discussion_r723393049_

digorath:
- previously the image would never load, when the caching failed
  - now it tries to fetch the remote directly

---

closes #361

### DIFF
--- a/src/components/Image.js
+++ b/src/components/Image.js
@@ -72,13 +72,14 @@ export const Image = ({
             .then((path) => {
               mounted && setSource({ uri: path });
             })
-            .catch((err) =>
+            .catch((err) => {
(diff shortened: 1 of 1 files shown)


donni106 (approved):
🦖

PR #388, opened 2021-11-03
BB Bus: always use cache first

digorath:
- added new const for refresh intervals: NEVER
- if this is selected as the refresh interval the request will always be 'cache-first'

SVA-194

### DIFF
--- a/src/helpers/refreshIntervalHelper.js
+++ b/src/helpers/refreshIntervalHelper.js
@@ -52,6 +52,10 @@ export const refreshTimeFor = async (refreshTimeKey, refreshInterval) => {
 
       break;
     }
+    case consts.REFRESH_INTERVALS.NEVER: {
+      // always return tomorrow, so that we never refresh
+      return moment().add(1, 'days').unix();
+    }
     default:
       break;
   }
--- a/src/helpers/refreshIntervalHelper.js
+++ b/src/helpers/refreshIntervalHelper.js
@@ -18,6 +18,11 @@ import { addToStore, readFromStore } from './storageHelper';
  * @return {Promise<number>} refresh time in seconds from 01.01.1970 00:00:00 UTC
  */
 export const refreshTimeFor = async (refreshTimeKey, refreshInterval) => {
+  if (refreshInterval === consts.REFRESH_INTERVALS.NEVER) {
+    // always return tomorrow, so that we never refresh
+    return moment().add(1, 'days').unix();
+  }
+
   const refreshIntervals = await readFromStore('refresh-intervals');
   const now = moment().unix(); // now in seconds from 01.01.1970 00:00:00 UTC
   const refreshTime =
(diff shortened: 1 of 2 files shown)


src/helpers/refreshIntervalHelper.js
```diff
@@ -52,6 +52,10 @@ export const refreshTimeFor = async (refreshTimeKey, refreshInterval) => {
 
       break;
     }
+    case consts.REFRESH_INTERVALS.NEVER: {
+      // always return tomorrow, so that we never refresh
+      return moment().add(1, 'days').unix();
+    }
```
donni106, line 58: do we need to delete possibly existing refresh time from AsyncStorage in this case? it could be, that there is something saved from the past, right, through the `ONCE_A_DAY` case here that was used before?

src/helpers/refreshIntervalHelper.js
```diff
@@ -52,6 +52,10 @@ export const refreshTimeFor = async (refreshTimeKey, refreshInterval) => {
 
       break;
     }
+    case consts.REFRESH_INTERVALS.NEVER: {
+      // always return tomorrow, so that we never refresh
+      return moment().add(1, 'days').unix();
```
donni106, line 57: i wonder if we should refactor the method to "always return in the same way". now we have that breaking cases and the return at the end of the method. in addition here is a return happening right in place of the case.

adjusting it could mean making `refreshTime` a `let` instead of `const` and to set it here with break additionally. but i do not think that this is more readable. at the end it does the same.

another thought is to return `now` here for `NEVER`. could this work.
when checking for `fetchPolicy` we do:

```
if (now < refreshTime) {
  // use cached data from phone depending on update interval
  return 'cache-first';
  }
```

`now` there is created after the `now` here and therefore returning `now` for `NEVER` would result in `now > refreshTime`, right? so it would be `network-only`. with your solution it works as expected. i cannot find a smarter solution.

src/helpers/refreshIntervalHelper.js
digorath: i don't think we need to delete those refresh times. we do not clean up previously stored refresh times for any other "switching" of refresh times either. in the case of `NEVER`, we simply ignore the previously stored refresh time.

src/helpers/refreshIntervalHelper.js
digorath: 1. i think changing it to return `now` seems more prone to error in future refactorings, as the order of creating the `now` variables matter. if we would for example refactor it somewhen to first create the now to check against, and then get the refresh time from the helper, we would suddenly have `NEVER` turn into an `ALWAYS`.
2. `NEVER` will probably remain the only special case. and the way it behaves, we could simply put it directly at the start of the function, causing it to never even load data from the async storage, but instead return immediately.
the rest of the cases would then be handled in the way they did previously. it would move the "special" handling to the top level, instead of being "hidden" in the switch statement.

src/helpers/refreshIntervalHelper.js
donni106: right, we have no benefit of including cleanup logics

src/helpers/refreshIntervalHelper.js
donni106: i ❤️  the 2. approach

digorath: ✅

donni106 (approved):

PR #393, opened 2021-11-11
Encounter: added more info and poi category

digorath:
- extracted `useStaticContent` hook and made it usable for html and json contents
- added a button to the home screen that leads to a category of pois
- added possibility to have additional info html content on the registered home screen

SVA-352

- [x] Fix query for category once it is available

## Screenshots:
Simulator Screen Shot - iPhone 12 Pro - 2021-11-11 at 18 27 47

### DIFF
--- a/src/hooks/staticContent.ts
+++ b/src/hooks/staticContent.ts
@@ -1,0 +1,88 @@
+import { useCallback, useContext, useMemo, useState } from 'react';
+import { useQuery } from 'react-apollo';
+
+import { graphqlFetchPolicy } from '../helpers';
+import { NetworkContext } from '../NetworkProvider';
+import { getQuery, QUERY_TYPES } from '../queries';
+
+import { useRefreshTime } from './TimeHooks';
+
+type StaticContentArgs<T = unknown> =
+  | {
+      name: string;
+      type: 'json';
+      refreshTimeKey?: string;
+      refreshInterval?: string;
+      parseFromJson?: (json: unknown) => T;
+    }
+  | {
+      name: string;
+      type: 'html';
+      refreshTimeKey?: string;
+      refreshInterval?: string;
+      parseFromJson?: undefined;
+    };
+
+export const useStaticContent = <T>({
+  name,
+  type,
+  parseFromJson,
+  refreshInterval,
+  refreshTimeKey
+}: StaticContentArgs<T>): {
+  data: T;
+  error: boolean;
+  loading: boolean;
+  refetch: () => Promise<unknown>;
+} => {
+  const { isConnected, isMainserverUp } = useContext(NetworkContext);
+  const [error, setError] = useState(false);
+
+  const refreshTime = useRefreshTime(refreshTimeKey ?? name, refreshInterval);
+
+  const fetchPolicy = graphqlFetchPolicy({
+    isConnected,
+    isMainserverUp,
+    refreshTime
+  });
+
+  const { data, error: queryError, loading, refetch } = useQuery(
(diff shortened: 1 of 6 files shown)


src/screens/EncounterHomeScreen.tsx
```diff
@@ -20,16 +21,27 @@ import {
...
 const INFO_ICON_SIZE = normalize(16);
 
 const a11yLabels = consts.a11yLabel;
 
 // eslint-disable-next-line @typescript-eslint/no-explicit-any
+// eslint-disable-next-line complexity
 export const EncounterHomeScreen = ({ navigation, route }: any) => {
+  // @ts-expect-error settings are not properly typed
+  const encounterCategoryId = useContext(SettingsContext).globalSettings.settings?.encounter
+    ?.categoryId;
```
donni106, line 44: so we need to add a new settings config on the main server with the value of the encounter category, right?

src/screens/EncounterHomeScreen.tsx
```diff
@@ -62,6 +80,14 @@ export const EncounterHomeScreen = ({ navigation, route }: any) => {
     });
   }, [navigation]);
 
+  const onPressToCategory = useCallback(() => {
+    navigation.navigate(ScreenName.Index, {
+      query: QUERY_TYPES.POINTS_OF_INTEREST,
+      queryVariables: { categoryId: encounterCategoryId },
```
donni106, line 86: do we miss `limit: 15, order: 'name_ASC'` here, like in `parseCategories` in `listItemParser.js`?

src/screens/EncounterHomeScreen.tsx
```diff
@@ -146,6 +177,13 @@ export const EncounterHomeScreen = ({ navigation, route }: any) => {
               }}
               title={texts.encounter.myData}
             />
+            {encounterCategoryId && (
+              <Button invert title={texts.encounter.toCategory} onPress={onPressToCategory} />
+            )}
+            {additionalInfo?.length && (
```
donni106, line 183: do we need `!!` cast to booleans  here in some cases for the conditions to not make react native think to render something? is this still a thing or maybe unnecessary since some newer version?

src/hooks/staticContent.ts
```diff
@@ -0,0 +1,88 @@
...
+  const fetchPolicy = graphqlFetchPolicy({
+    isConnected,
+    isMainserverUp,
+    refreshTime
+  });
+
+  const { data, error: queryError, loading, refetch } = useQuery(
+    getQuery(type === 'json' ? QUERY_TYPES.PUBLIC_JSON_FILE : QUERY_TYPES.PUBLIC_HTML_FILE),
+    {
+      variables: { name: name },
```
donni106, line 52: just `name`, as we love it short?

src/screens/EncounterHomeScreen.tsx
digorath: correct. i added it to the documentation in confluence

src/screens/EncounterHomeScreen.tsx
digorath: it should not be needed unless someone specifically adds an empty string into the setting.

i added it to both line 181 and 183, just to avoid having that possibility.

src/screens/EncounterHomeScreen.tsx
digorath: it is probably better to add it here as well. so i've added it ;)

src/hooks/staticContent.ts
digorath: done

donni106 (approved):

PR #398, opened 2021-11-18
Feature: app intro

digorath:
- added new app intro feature
- this app intro will be displayed at app start until completed once
- it can handle the required permission alerts in a more structured way

- the splashscreen and orientation lock handling was moved to the new `OnboardingManager` component
- the handling of system and matomo permissions are now being handled by the new `OnboardingManager` (and if an app intro is configured through that as well)

---

SVA-305

---

## How to test:

- for testing multiple scenarios it is best to not save that the intro was completed at first by commenting out line 58 of the `OnboardingManager`
- by manipulating the conditions in lines 66 and 80 different situations can be simulated

### DIFF
--- a/src/OnboardingManager.tsx
+++ b/src/OnboardingManager.tsx
@@ -36,6 +36,7 @@ const useInitializeAfterOnboarding = (onboardingComplete: boolean) => {
         Initializers[Initializer.PushNotifications]();
       }
 
+      // set orientation to "default", to allow both portrait and landscape
       ScreenOrientation.lockAsync(ScreenOrientation.OrientationLock.DEFAULT);
     }
   }, [onboardingComplete]);
--- a/src/OnboardingManager.tsx
+++ b/src/OnboardingManager.tsx
@@ -2,7 +2,7 @@ import * as SplashScreen from 'expo-splash-screen';
 import React, { useContext, useEffect, useState } from 'react';
 
 import { addToStore, readFromStore } from './helpers';
-import { Initializers } from './helpers/initializationHelper';
+import { Initializer, Initializers } from './helpers/initializationHelper';
 import { AppIntroScreen } from './screens';
 import { SettingsContext } from './SettingsProvider';
 
@@ -15,7 +15,11 @@ export const OnboardingManager = ({ children }: { children: React.ReactNode }) =
   const {
     globalSettings: {
       // @ts-expect-error settings context is not properly typed
-      settings: { onboarding: onboardingActive }
+      settings: {
+        locationService: locationServiceActive,
+        onboarding: onboardingActive,
+        pushNotifications: pushNotificationsActive
+      }
     }
   } = useContext(SettingsContext);
 
@@ -55,8 +59,13 @@ export const OnboardingManager = ({ children }: { children: React.ReactNode }) =
   // this effect ensures that all settings will be properly initialized, even when onboarding was completed before the settings where available, or an error occured
   useEffect(() => {
     if (onboardingStatus === 'complete') {
-      // TODO: filter out settings disabled by globalsettings
-      Initializers.forEach((init) => init());
+      // TODO: add matomo
+      if (locationServiceActive) {
+        Initializers[Initializer.LocationService]();
+      }
+      if (pushNotificationsActive) {
+        Initializers[Initializer.PushNotifications]();
+      }
     }
   }, [onboardingStatus]);
 
--- a/src/OnboardingManager.tsx
+++ b/src/OnboardingManager.tsx
@@ -1,12 +1,42 @@
 import * as SplashScreen from 'expo-splash-screen';
 import React, { useContext, useEffect, useState } from 'react';
 
+import { CustomMatomoProvider } from './CustomMatomoProvider';
 import { addToStore, readFromStore } from './helpers';
 import { Initializer, Initializers } from './helpers/initializationHelper';
 import { AppIntroScreen } from './screens';
 import { SettingsContext } from './SettingsProvider';
 
 const ONBOARDING_STORE_KEY = 'ONBOARDING_STORE_KEY';
+
+// this hook ensures that all settings will be properly initialized, even when onboarding was completed before the settings where available, or an error occured
+const useInitializeAfterOnboarding = (onboardingComplete: boolean) => {
+  const {
+    globalSettings: {
+      // @ts-expect-error settings context is not properly typed
+      settings: {
+        locationService: locationServiceActive,
+        pushNotifications: pushNotificationsActive,
+        matomo: matomoActive
+      }
+    }
+  } = useContext(SettingsContext);
+
+  // this effect ensures that all settings will be properly initialized, even when onboarding was completed before the settings where available, or an error occured
+  useEffect(() => {
+    if (onboardingComplete) {
+      if (locationServiceActive) {
+        Initializers[Initializer.LocationService]();
+      }
+      if (matomoActive) {
+        Initializers[Initializer.MatomoTracking]();
+      }
+      if (pushNotificationsActive) {
+        Initializers[Initializer.PushNotifications]();
+      }
+    }
+  }, [onboardingComplete]);
+};
 
 export const OnboardingManager = ({ children }: { children: React.ReactNode }) => {
   const [onboardingStatus, setOnboardingStatus] = useState<'loading' | 'complete' | 'incomplete'>(
@@ -15,11 +45,7 @@ export const OnboardingManager = ({ children }: { children: React.ReactNode }) =
   const {
     globalSettings: {
       // @ts-expect-error settings context is not properly typed
-      settings: {
-        locationService: locationServiceActive,
-        onboarding: onboardingActive,
-        pushNotifications: pushNotificationsActive
-      }
+      settings: { onboarding: onboardingActive }
     }
   } = useContext(SettingsContext);
 
@@ -56,18 +82,7 @@ export const OnboardingManager = ({ children }: { children: React.ReactNode }) =
     }
   }, []);
 
-  // this effect ensures that all settings will be properly initialized, even when onboarding was completed before the settings where available, or an error occured
-  useEffect(() => {
-    if (onboardingStatus === 'complete') {
-      // TODO: add matomo
-      if (locationServiceActive) {
-        Initializers[Initializer.LocationService]();
-      }
-      if (pushNotificationsActive) {
-        Initializers[Initializer.PushNotifications]();
-      }
-    }
-  }, [onboardingStatus]);
+  useInitializeAfterOnboarding(onboardingStatus === 'complete');
 
   // render null while onboarding status is loading from AsyncStorage
   if (onboardingStatus === 'loading') {
@@ -78,6 +93,5 @@ export const OnboardingManager = ({ children }: { children: React.ReactNode }) =
     return <AppIntroScreen setOnboardingComplete={setOnboardingComplete} />;
   }
 
-  // TODO: move matomo provider  here
-  return <>{children}</>;
+  return <CustomMatomoProvider>{children}</CustomMatomoProvider>;
 };
--- a/src/OnboardingManager.tsx
+++ b/src/OnboardingManager.tsx
@@ -1,3 +1,4 @@
+import * as ScreenOrientation from 'expo-screen-orientation';
 import * as SplashScreen from 'expo-splash-screen';
 import React, { useContext, useEffect, useState } from 'react';
 
@@ -34,6 +35,8 @@ const useInitializeAfterOnboarding = (onboardingComplete: boolean) => {
       if (pushNotificationsActive) {
         Initializers[Initializer.PushNotifications]();
       }
+
+      ScreenOrientation.lockAsync(ScreenOrientation.OrientationLock.DEFAULT);
     }
   }, [onboardingComplete]);
 };
--- a/src/OnboardingManager.tsx
+++ b/src/OnboardingManager.tsx
@@ -1,0 +1,74 @@
+import * as SplashScreen from 'expo-splash-screen';
+import React, { useContext, useEffect, useState } from 'react';
+
+import { addToStore, readFromStore } from './helpers';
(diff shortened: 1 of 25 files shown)


digorath:
@donni106 i will take care of the todos soon. those will only be minor changes i think, so it is still worthwhile to review already

digorath:
location services initialization needs to be fixed.
currently it is not properly disabled, if globalsettings do not have it enabled.

src/helpers/storageHelper.js
```diff
@@ -10,6 +10,7 @@ export const storageHelper = {
   globalSettings: () => readFromStore('globalSettings'),
   setGlobalSettings: (globalSettings) => addToStore('globalSettings', globalSettings),
   locationSettings: () => readFromStore('locationSettings'),
+  onboardingSettings: () => readFromStore('onboardingSettings'),
```
donni106, line 13: can you move this one line down please to keep `locationSettings` and `setLocationSettings` next to each other like als the "getter and setter"

src/index.js
```diff
@@ -178,9 +178,7 @@ const MainAppWithApolloProvider = () => {
     }
 
     // if there are no locationSettings yet, set the defaults as fallback
-    const locationSettings = globalSettings.settings.locationService
-      ? (await storageHelper.locationSettings()) || { locationService: true }
-      : { locationService: false };
+    const locationSettings = (await storageHelper.locationSettings()) || {};
```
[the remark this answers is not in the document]
digorath: please check my changes here  a9a92b77c797f7a654775d42ab5eefa0082cb164

src/screens/AppIntroScreen.tsx
```diff
@@ -0,0 +1,109 @@
...
+      <AppIntroSlider<AppIntroSlide>
+        renderItem={renderSlide}
+        data={slides}
+        keyExtractor={keyExtractor}
+        onSlideChange={(_, index) => {
+          slides[index]?.onLeaveSlide?.(true);
+        }}
+        onDone={setOnboardingComplete}
+        renderDoneButton={() => <SliderButton label="Weiter" />}
+        renderNextButton={() => <SliderButton label="Weiter" />}
```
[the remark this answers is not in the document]
digorath: ✅  3e4e21b9a2dcb6a4200d056c34301fea1826ab29

src/OnboardingManager.tsx
```diff
@@ -0,0 +1,100 @@
...
+      } finally {
+        await SplashScreen.hideAsync();
+      }
+    };
+
+    if (onboardingActive) {
+      loadAndSetOnboardingStatus();
+    } else {
+      setOnboardingStatus('complete');
+      SplashScreen.hideAsync();
```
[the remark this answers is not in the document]
digorath: in l. 76 we are inside of an async function. my expected default behaviour of an async function is, that all its async calls have resolved by the time the promise returned by the function is resolved. (unless there are specific reasons not to).

in l. 84, we cannot await as it is inside of the sync effect block

src/screens/AppIntroScreen.tsx
```diff
@@ -0,0 +1,109 @@
...
+        dotStyle={styles.inactiveDot}
+        activeDotStyle={styles.activeDot}
+        scrollEnabled={false}
```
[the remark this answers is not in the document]
digorath: correct, we should disable dot click.

src/jsonValidation/appIntroValidator.ts
```diff
@@ -0,0 +1,34 @@
+import { isArray, isObjectLike, isString } from 'lodash';
+
+import { Initializer, Initializers } from '../helpers/initializationHelper';
```
donni106, line 3: do you mind putting initializationHelper in index.js for helpers to be able to import from "../helpers" at all places?

src/screens/AppIntroScreen.tsx
```diff
@@ -0,0 +1,109 @@
...
+};
+
+type Props = {
+  setOnboardingComplete: () => void;
+};
+
+const renderSlide: ListRenderItem<AppIntroSlide> = ({ item }) => {
+  return (
+    <ScrollView>
+      <Image
```
donni106, line 28: what about

Bildschirmfoto 2021-11-24 um 12 48 46

because of mixing js and ts?

digorath: yes.
it occured when we stopped destructuring `source` from `props` in `image`.

we did not notice it back then, and i did not find the right place to change it yet.

src/index.js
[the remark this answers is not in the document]
donni106: i do not get it. all changes are regarding onboarding and the new screen. what do i miss?

src/screens/SettingsScreen.js
```diff
@@ -105,7 +111,7 @@ export const SettingsScreen = () => {
 
       // settings should sometimes contain matomo analytics next, depending on server settings
       if (settings.matomo) {
-        const { consent: matomoValue = false } = await storageHelper.matomoSettings();
+        const { consent: matomoValue } = await matomoSettings();
```
donni106, line 114: nice. we can get rid of default value because of

```
// if there are no matomo settings yet, init without consent
if (!settings) {
  settings = {
    consent: false
  };
}
```

right?

src/OnboardingManager.tsx
```diff
@@ -34,6 +35,8 @@ const useInitializeAfterOnboarding = (onboardingComplete: boolean) => {
       if (pushNotificationsActive) {
         Initializers[Initializer.PushNotifications]();
       }
+
+      ScreenOrientation.lockAsync(ScreenOrientation.OrientationLock.DEFAULT);
```
donni106, line 39: you removed the comment here, which could somewhen be helpful, as the method is not very self explanatory.
otherwise one need to know what the lock does and what DEFAULT stands for.

> // set orientation to "default", to allow both portrait and landscape

src/jsonValidation/appIntroValidator.ts
```diff
@@ -0,0 +1,34 @@
...
+};
+
+const isValidAppIntroSlideData = (json: unknown): json is AppIntroSlideData => {
+  if (!isObjectLike(json)) {
+    return false;
+  }
+
+  const { image, title, text, onLeaveSlide } = json as AppIntroSlideData;
+
+  const isInitializer = !onLeaveSlide || Object.values(Initializer).includes(onLeaveSlide);
```
donni106, line 20: could this line get a comment please.
trying to get the context from code is not that easy.
Bildschirmfoto 2021-11-24 um 15 19 14
Bildschirmfoto 2021-11-24 um 15 19 21

i guess the values of `Initializer` are:
- 'locationService'
- 'matomoTracking'
- 'pushNotifications

so we check here if a `onLeaveSlide` is on of them?

src/screens/HomeScreen.js
```diff
@@ -74,6 +74,8 @@ export const HomeScreen = ({ navigation, route }) => {
     undefined,
     globalSettings?.settings?.pushNotifications
   );
+
+  useTrackAppStart();
```
donni106, line 78: is this the first possible spot to run this?
when entering the home screen, we track that separately already.
the app start should be earlier in my understanding.

src/screens/AppIntroScreen.tsx
```diff
@@ -0,0 +1,110 @@
...
+    }
+  }, [error, loading, slides?.length]);
+
+  if (error || loading || !slides?.length) {
+    return null;
+  }
+
+  return (
+    <SafeAreaViewFlex>
+      <AppIntroSlider<AppIntroSlide>
```
donni106, line 65: there is an unexpected behavior when clicking "next" very fast. it somehow jumps around. tested with my older and newer ios. this does not happen with my android! please try out if it happens to you as well.

src/index.js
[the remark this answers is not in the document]
digorath: seems i failed when copying the commit hash..  58b671324099184ef5334ebfc9910734876f858c

src/helpers/storageHelper.js
digorath: no i can't. because i removed it instead (was unused)

src/screens/AppIntroScreen.tsx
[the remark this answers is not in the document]
digorath: ✅

src/jsonValidation/appIntroValidator.ts
digorath: Screenshot 2021-11-25 at 11 19 17

no, as that creates require cycles.

src/screens/SettingsScreen.js
digorath: correct. we always initialize it as part of the helper function

src/OnboardingManager.tsx
digorath: added it again

src/jsonValidation/appIntroValidator.ts
digorath: yes. the purpose is to filter out slides that have invalid `onLeaveSlide`s in them. that way slides corresponding to features of newer app versions will not be shown.

src/screens/HomeScreen.js
digorath: i changed it to be called immediately upon instance creation.

the hook is then unused so i removed it

src/screens/AppIntroScreen.tsx
digorath: it happens for me as well, and not on android either.

unfortunately we do not provide the whole button component, but only the visual layer to the app intro slider.

what do you think is the best course of action here?

src/jsonValidation/appIntroValidator.ts
donni106: > could this line get a comment please.

do you mind adding your sentence as comment there?

src/screens/HomeScreen.js
donni106: thanks, makes more sense to me. i changed `info` to `userInfo` per rebase. so update you local branch first in terms of updating something else.

src/jsonValidation/appIntroValidator.ts
donni106: i do it

donni106: did it

src/screens/AppIntroScreen.tsx
donni106: all right. please check if you can pass an own button component to avoid "their" problems.
otherwise you can check issues of the repo, if this is public already. otherwise create one and provide a PR with some possible fix?

src/screens/AppIntroScreen.tsx
donni106: ideas, what you would do and how long it could take?

src/screens/HomeScreen.js
digorath: ✅

src/screens/AppIntroScreen.tsx
digorath: so. it is possible to provide a custom `renderPagination` method, and i tried a bit around.

1. it already is a bit of work to have it look and behave exactly the same (using as much of the already implemented functionality as you can) while creating a custom `onPress` for the next button. (with that i mean having it be its own touchable that uses the `goToSlide` of the `ref` for the app intro slider)
2. after trying around for a while i did not come to a clean solution. everything i tried seems to revolve around the internal state for `activeIndex` not updating fast enough. i tried several workarounds but none solved the problem of the `goToSlide` function being called multiple times with the same values (we need the `activeIndex` to calculate the args for the call).

for now i do not see a way for us to **easily** fix it. neither by using a fork, nor by creating a custom `renderPagination`.

it can probably be done by altering the state management of the `AppIntroSlider` in a more fundamental and complex way, though.

src/screens/AppIntroScreen.tsx
digorath: it is probably a matter of finding a decent name for the destructured `source.props`. we tried already and failed, that is why we ended up with the solution of not destructuring it. which then causes type inference to malfunction.

changing it to be a typescript file would also be a solution, but i would need to check how many issues that brings with it, due to other things not being typed yet

src/screens/AppIntroScreen.tsx
digorath: https://github.com/Jacse/react-native-app-intro-slider/issues/254

src/screens/AppIntroScreen.tsx
donni106: ok #410

src/screens/AppIntroScreen.tsx
donni106: let's see when and how they will answer. i also subscribed to the issue. thanks

src/screens/AppIntroScreen.tsx
```diff
@@ -106,6 +107,6 @@ const styles = StyleSheet.create({
   sliderButton: {
     borderBottomColor: colors.darkText,
     borderBottomWidth: 1,
-    marginTop: 14 // no normalization here as the dots position does not use normalization either
+    marginVertical: 16 // no normalization here as the dots position does not use normalization either
```
donni106, line 110: 12 makes it more aligned with the dots in my test than 16?

digorath: hmm. when testing it made it look better with 16 and both top and bottom, instead of only top.

that is also the margin that is used for the dots component internally. 🤔

donni106: i should have attached screenshots of my test yesterday. will contact you on this one later

donni106: |16|14|12|
|---|---|---|
|IMG_0019|IMG_0020|IMG_0021|

donni106: screenshots taken with iPhone 6s real device.
comparing them (full screen after clicking the image), i think the 12 aligns best? 

Bildschirmfoto 2021-12-01 um 19 42 55

digorath: ✅

PR #403, opened 2021-11-23
Survey: hide comment section when disabled

digorath:
- updated `Survey` type to include `canComment` flag
- hide `CommentSection` if `canComment` flag is not equal to `true`

- [x] update query

---

SVA-363

### DIFF
--- a/src/screens/SurveyDetailScreen.tsx
+++ b/src/screens/SurveyDetailScreen.tsx
@@ -187,13 +187,15 @@ export const SurveyDetailScreen = ({ route }: Props) => {
               />
             )}
           </WrapperWithOrientation>
-          <CommentSection
-            archived={archived}
-            comments={survey.comments}
-            isMultilingual={survey.isMultilingual}
-            scrollViewRef={scrollViewRef}
-            surveyId={surveyId}
-          />
+          {!!survey.canComment && (
+            <CommentSection
+              archived={archived}
+              comments={survey.comments}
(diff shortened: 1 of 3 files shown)


src/screens/SurveyDetailScreen.tsx
```diff
@@ -149,12 +149,14 @@ export const SurveyDetailScreen = ({ route }: Props) => {
               />
             )}
           </WrapperWithOrientation>
-          <CommentSection
-            archived={archived}
-            comments={survey.comments}
-            scrollViewRef={scrollViewRef}
-            surveyId={surveyId}
-          />
+          {!!survey.canComment && (
```
donni106, line 152: could we avoid `!!` as `canComment` is typed as boolean?

donni106 (approved):

src/screens/SurveyDetailScreen.tsx
donni106: maybe better to keep them for some `null` or `undefined` cases?

digorath: i typed the `Survey` type to have it as an optional value, because i did not know if we will get `null` from the server under any circumstances. (e.g. it does not get set when it is off at creation, or the db migration for backward compatability was not completed in time/successfully)

donni106: yes, was my second thought as well even when i think this cannot happen. let's be more safe here.

PR #405, opened 2021-11-23
WebViews: add props for extended usage

### LINKED ISSUE #300: Faster scrolling in web views
https://github.com/ikuseiGmbH/smart-village-app-app/blob/a6cf9cdc34dc1fcb3c4440661d4e4406a7149a72/src/screens/WebScreen.js#L29-L38

we should add `decelerationRate="normal"` to improve scroll behavior in web views.

(https://github.com/react-native-webview/react-native-webview/blob/19980d888d66554875f3ac64b3e8a35bd7ad998b/src/WebViewTypes.ts#L378-L389)

donni106:
feat: optimize webview scroll behavior

- added `decelerationRate="normal"` to WebScreen to increase speed of scrolling like user are used to from the browser

Closes #300

feat: add custom useragent string to identify app requests on web servers

- auto organized imports
- used scheme string to pass as application name for user agent that can be used on web servers
  - first use case is bbnavi, where the user agent will be used

BBNAV-61

### DIFF
--- a/src/screens/WebScreen.js
+++ b/src/screens/WebScreen.js
@@ -35,6 +35,8 @@ export const WebScreen = ({ route }) => {
               <ActivityIndicator color={colors.accent} />
             </LoadingContainer>
           )}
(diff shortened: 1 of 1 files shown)


digorath (approved):
🚀

PR #407, opened 2021-11-25
Feature: nested info screen

digorath:
- this screen handles a new json format that allows for an easy tree structure
  - each screen may include html content at the top
  - each screen includes a list of textlistitems at the bottom

- added skip option to useStaticContent

---

SVA-314

## How to test:

- a new link using this screen was added to the drawer and the about section

## Screenshots:

Simulator Screen Shot - iPhone 12 Pro - 2021-11-25 at 10 20 20


src/screens/NestedInfoScreen.tsx
```diff
@@ -0,0 +1,118 @@
...
+  return (
+    <Wrapper>
+      {/* @ts-expect-error HtmlView uses memo in js, which is not inferred correctly */}
+      <HtmlView html={html} />
+    </Wrapper>
+  );
+};
+
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+export const NestedInfoScreen = ({ navigation, route }: StackScreenProps<any>) => {
```
donni106, line 49: was this screen based on another one? are lines copied from another screen?

src/screens/NestedInfoScreen.tsx
```diff
@@ -0,0 +1,118 @@
...
+  const sectionData = [
+    {
+      title: data.title,
+      data: data.children?.map((child) => {
+        return {
+          ...child,
+          params: {
+            ...child.params,
+            rootRouteName: child.params?.rootRouteName ?? rootRouteName,
+            title: child.params?.title ?? navigationTitle
```
donni106, line 90: please check against the navigation title für the screen entering per "loop" from home screen. i have an empty one and could not make it appear. i have added a `title` inside child params, but this just applies for the second time navigated to "loop".

donni106 (changes requested):

src/screens/NestedInfoScreen.tsx
```diff
@@ -0,0 +1,118 @@
...
+          }
+        };
+      })
+    }
+  ];
+
+  return (
+    <SafeAreaViewFlex>
+      <WrapperWithOrientation>
```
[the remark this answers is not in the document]
donni106: 66b44d0

src/screens/NestedInfoScreen.tsx
digorath: the title is set (as with other about links as well) through the about home static content. because of that there is no initial title set.
there is an initial title, when you open it through the drawer (because it is set as part of the drawer content).

does that explain/solve this?

src/screens/NestedInfoScreen.tsx
digorath: i don't remember. is there anything wrong with this line?

donni106: no, i thought of how i review, if i compare to another screen or not

src/screens/NestedInfoScreen.tsx
donni106: yes, thanks, works, knot untied

src/screens/NestedInfoScreen.tsx
```diff
@@ -0,0 +1,118 @@
...
+  LoadingContainer,
+  LoadingSpinner,
+  SafeAreaViewFlex,
+  SectionHeader,
+  Wrapper,
+  WrapperWithOrientation
+} from '../components';
+import { colors } from '../config';
+import { useStaticContent } from '../hooks';
+import { useRenderItem } from '../hooks/listHooks';
```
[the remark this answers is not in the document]
digorath: with rebasing another branch that commit was lost there. so i reintroduced it here via cherry pick 🍒 

no further adjustments were needed. the imports were already all through hooks index

src/screens/NestedInfoScreen.tsx
digorath: it probably still is helpful to do that

digorath:
@donni106 i decided to go the extra step and use the `useStaticContent` hook here as well now.

src/components/screens/About.js
```diff
@@ -1,37 +1,25 @@
...
 import { SettingsContext } from '../../SettingsProvider';
 import { LoadingContainer } from '../LoadingContainer';
 import { SectionHeader } from '../SectionHeader';
 import { VersionNumber } from '../VersionNumber';
 
 export const About = ({ navigation, withHomeRefresh, withSettings }) => {
-  const { isConnected, isMainserverUp } = useContext(NetworkContext);
+  const { data: aboutData, loading, refetch } = useStaticContent({
+    name: 'homeAbout',
+    type: 'json'
```
donni106, line 17: should there be a `refreshTimeKey` with `'publicJsonFile-homeAbout'` here, because otherwise the `name` will be used and could lead to different behaviors than before?

digorath: this would lead to users possibly needing to fetch one extra time after updating the app. i changed it to use the `refreshTimeKey` as suggested.

donni106: which would be fine one extra time.
but not necessary to do "something else than before".
you know how we like it :)

donni106 (approved):

PR #404, opened 2021-11-23
Survey: multilanguage optional

digorath:
- added new flag `isMultiLanguage` to `Survey` type
- when used no polish texts are displayed
- adjusted some string constructions
- adjusted `useLanguages` hook to accept `isMultiLanguage`
- slight adjustments to styling of answers where needed

- [x] adjust query

---

SVA-362

## Screenshots:

Simulator Screen Shot - iPhone 12 Pro - 2021-11-23 at 16 24 46
Simulator Screen Shot - iPhone 12 Pro - 2021-11-23 at 16 24 20
Simulator Screen Shot - iPhone 12 Pro - 2021-11-23 at 16 25 58

PR #408, opened 2021-11-25
Survey: allow selection of multiple answers for surveys that have it enabled

digorath:
- updated all relevant places to use string arrays instead of a single string for selections of answers
- added extra logic to ensure old stored submissions are handled correctly
- removed redundant condition from submission function
  - it does not cause real issues if the condition is not met
  - the condition is already checked on the detail screen itself as part of the button disabled condition
- added a hint that multiple options can be selected

---

SVA-361

## How to test:

- The survey with title `Test mit HTML` is multi select

## Screenshots:

Simulator Screen Shot - iPhone 12 Pro - 2021-11-25 at 17 37 23

### DIFF
--- a/src/screens/SurveyDetailScreen.tsx
+++ b/src/screens/SurveyDetailScreen.tsx
@@ -146,6 +146,18 @@ export const SurveyDetailScreen = ({ route }: Props) => {
             {!archived && (
               <>
                 <Wrapper>
+                  {!!survey.questionAllowMultipleResponses && (
+                    <>
+                      <RegularText center error>
+                        {texts.survey.multiSelectPossible.de}
+                      </RegularText>
+                      {/* TODO: update with new optional multilanguage logic */}
+                      <RegularText center error italic>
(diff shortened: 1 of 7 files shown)


src/screens/SurveyDetailScreen.tsx
```diff
@@ -111,26 +119,37 @@ export const SurveyDetailScreen = ({ route }: Props) => {
...
             ))}
             {!archived && (
               <>
                 <Wrapper>
+                  {!!survey.questionAllowMultipleResponses && (
+                    <>
+                      <RegularText center error>
+                        {texts.survey.multiSelectPossible.de}
+                      </RegularText>
+                      {/* TODO: update with new optional multilanguage logic */}
```
digorath, line 139: this needs to be updated after #404 is merged

donni106 (approved):

src/screens/SurveyDetailScreen.tsx
digorath: ✅

PR #414, opened 2021-12-03
Chore: unify top filters

digorath:
- removed `MapSwitchHeader`
- updated index screen logic to use the `IndexFilterWrapperAndList`
- reused SectionHeader in SettingsScreen

---

SVA-382

## Screenshots:

|Before|After|
|---|---|
|Simulator Screen Shot - iPhone 12 Pro - 2021-12-03 at 12 09 34|Simulator Screen Shot - iPhone 12 Pro - 2021-12-03 at 12 01 14|


src/components/IndexFilterWrapperAndList.js
```diff
@@ -2,10 +2,9 @@ import PropTypes from 'prop-types';
 import React from 'react';
 import { FlatList } from 'react-native';
 
-import { Touchable } from '../Touchable';
-import { WrapperHorizontal } from '../Wrapper';
-import { RegularText } from '../Text';
-
+import { Touchable } from './Touchable';
+import { WrapperHorizontal } from './Wrapper';
+import { RegularText } from './Text';
```
donni106, line 7: `IndexFilterElement.js` was also missing adjusted imports as well as `BB-BUS/IndexFilter.js`

Bildschirmfoto 2021-12-03 um 16 22 40

i amended the changes and force pushed

src/screens/IndexScreen.js
```diff
@@ -29,20 +29,34 @@ import { usePosition, useTrackScreenViewAsync } from '../hooks';
 
 const { MATOMO_TRACKING } = consts;
 
+const TopFilter = {
+  list: 'list',
+  map: 'listTypes'
```
donni106, line 34: why `listTypes` for `map`? in what way is this wording connected?

src/screens/IndexScreen.js
```diff
@@ -29,20 +29,34 @@ import { usePosition, useTrackScreenViewAsync } from '../hooks';
 
 const { MATOMO_TRACKING } = consts;
 
+const TopFilter = {
```
donni106, line 32: like `INITIAL_TOP_FILTER` this could be `TOP_FILTER`, because this is used as holding constant? as well as `LIST` and `MAP`? and then this would be fit very well placed in our `consts.js` file, where we already import `MATOMO_TRACKING` from, right?

src/screens/SettingsScreen.js
```diff
@@ -24,15 +22,7 @@ const { MATOMO_TRACKING } = consts;
...
-const renderSectionHeader = ({ section: { title } }) =>
-  !!title && (
-    <View>
-      <TitleContainer>
-        <Title accessibilityLabel={`(${title}) ${consts.a11yLabel.heading}`}>{title}</Title>
-      </TitleContainer>
-      {device.platform === 'ios' && <TitleShadow />}
-    </View>
-  );
+const renderSectionHeader = ({ section: { title } }) => !!title && <SectionHeader title={title} />;
```
donni106, line 25: how is this related?

donni106 (changes requested):

src/screens/IndexScreen.js
digorath: this should have been `map` already. i somehow missed the rename here.

src/components/IndexFilterWrapperAndList.js
digorath: good catch. i have no idea why it did not cause issues while i tested it on my simulator. i must have been in an unsaved state.

src/screens/IndexScreen.js
digorath: I capitalized the names, but kept them in this file. they are specific to this screen and there is no need to extract them from here imo.

src/screens/SettingsScreen.js
digorath: this is directly related. i stumbled across this while looking at this screen for reference for the filter.

src/screens/SettingsScreen.js
```diff
@@ -42,14 +42,14 @@ const onDeactivatePushNotifications = (revert) => {
   setInAppPermission(false).then((success) => !success && revert());
 };
 
-const Filter = {
+const FILTER = {
```
donni106, line 45: following the changes, this should be also capitalized entirely?

```
const FILTER = {
  GENERAL: 'general',
  LIST_TYPES: 'listTypes'
};
```

as well as renamed to `LIST_FILTER` as we have now the separation across code with `TOP_FILTER` and this one here?

donni106 (approved):
thanks, added one further request

src/screens/SettingsScreen.js
digorath: ✅

PR #419, opened 2021-12-17
Feature: replace directus by new bb bus api

digorath:
## What has changed

- the graphql queries to directus have all been replaced by queries to the new endpoint
- minimal changes to namings were made to make the variable names be more in line with the names provided by the api
- removed all `un-nesting` steps that were required for data provided by directus
- removed no longer used `deep-rename-keys` dependency

## How to test
- testing this requires new secrets, which include the url to the new endpoint
- for the internal development server the new top10 ids have been added. other servers do not provide the required data yet


digorath:
I rebased it onto the current master and removed the no longer used dependency `deep-rename-keys`

digorath:
@donni106 Codeclimate url gives me a 404. Can you confirm?

donni106:
> @donni106 Codeclimate url gives me a 404. Can you confirm?

nope. and it is working for you again as well, right?

src/screens/BB-BUS/IndexScreen.js
```diff
@@ -98,35 +92,21 @@ export const IndexScreen = ({ navigation }) => {
     (data, areaId) => [(data[0].object || data[0]).id, data.length, areaId].join('-')
   );
 
-  const getCommunities = (data) => {
-    const snake_caseData = data?.directus?.community?.data;
+  const getAreas = (data) => {
+    const areas = data?.area;
```
donni106, line 96: i found out that some areas show up twice

IMG_814225044D8C-1

src/queries/BB-BUS/getService.ts
```diff
@@ -0,0 +1,73 @@
+import gql from 'graphql-tag';
+
+export const GET_SERVICE = gql`
+  query getService($areaId: ID!, $id: ID!) {
```
donni106, line 4: i do not know if this is the best place to put it, but i have some unwanted behavior.
i searched for a different city "brandenburg" and selected "brandenburg an der havel". then i searched for the pst "gewerbe abmeldung". when i opened the details, i reached the details of this pst but from "bad belzig" (initial area). i can see this in logs or in app when viewing "ansprechpartner".
logs at that point:
```
"organisationalUnits": Array [
    Object {
      "__typename": "OrganisationalUnit",
      "addresses": Array [
        Object {
          "__typename": "Address",
          "city": "Bad Belzig",
          "houseNumber": "6",
          "street": "Wiesenburger Straße",
          "zipcode": "14806",
        },
      ],
```

src/screens/BB-BUS/DetailScreen.js
```diff
@@ -136,13 +110,15 @@ export const DetailScreen = ({ route }) => {
 
   const fetchPolicy = graphqlFetchPolicy({ isConnected, isMainserverUp, refreshTime });
 
-  const { data, loading, refetch } = useQuery(GET_DIRECTUS, {
-    variables: GET_SERVICE(id),
+  const areaId = secrets[namespace]?.busBb?.areaId;
```
donni106, line 113: ah i think this is the reason for https://github.com/ikuseiGmbH/smart-village-app-app/pull/419/files#r784222846.
here the area id from secrets is used instead of a other possibly selected one, right?

src/screens/BB-BUS/IndexScreen.js
digorath: these are duplicates due to the name occurring with two different ids on the side of the external endpoint:

```
{
        "externalId": "100008239",
        "name": "Niemegk"
},
{
        "externalId": "100025122",
        "name": "Niemegk"
},
```

src/queries/BB-BUS/getService.ts
digorath: this should be fixed now.

src/screens/BB-BUS/DetailScreen.js
digorath: this should be fixed now.

src/screens/BB-BUS/DetailScreen.js
```diff
@@ -100,6 +100,7 @@ export const DetailScreen = ({ route }) => {
   const rootRouteName = route.params?.rootRouteName ?? '';
   const headerTitle = route.params?.title ?? '';
   const details = route?.params?.data ?? '';
+  const areaId = route.params?.areaId ?? secrets[namespace]?.busBb?.areaId;
```
donni106, line 103: we have `secrets[namespace]?.busBb?.areaId?.toString()` in screens/BB-BUS/IndexScreen.js. should we remove the `toString` there to be in place with here?

src/screens/BB-BUS/DetailScreen.js
```diff
@@ -107,20 +107,16 @@ export const DetailScreen = ({ route }) => {
 
   const openWebScreen = useOpenWebScreen(headerTitle, undefined, rootRouteName);
 
-  const refreshTime = useRefreshTime(`BBBUS-service-${id}`, consts.REFRESH_INTERVALS.BB_BUS);
```
donni106, line 135: > removed caching for details page to avoid problems due to pst id not being a unique key without areaId

what about using `BBBUS-service-${areaId}-${id}` here? and reverting `BB_BUS: NEVER` to `BB_BUS: ONCE_A_DAY` again in `REFRESH_INTERVALS` constants? would this help to be able to change less code and keep a bit of caching?

src/screens/BB-BUS/DetailScreen.js
digorath: the api returns `areaIds` as `strings`, so for prop type validation we should treat `areaId` as a string. Instead of removing it there, I added it to the missing places

src/screens/BB-BUS/DetailScreen.js
digorath: the problem is the following, if cache is used
1. open details of service `S` for area `A`
2. open details of service `S` for area `B`
-> cache for service `S` for area `A` is overwritten by service `S` for area `B`
3. open details of service `S` for area `A` again
-> cache is used but contains details for area `B`

this happens because the cache only works based on ids of entries, but the id of service `S` is the same for area `A` and area `B`

regardless of that, I changed the caching back from `NEVER` to `ONCE_A_DAY`. for the index screen bb bus queries

donni106: thanks, this helped me to understand it more clearly. so the query with `variables: { id, areaId }` is just using the `id` for caching? i somehow thought that a combination of variables will be used.

digorath: it is not using query variables at all for caching. just the id of the response

donni106: jo, i did some more research.
we can come back to it at some later point: https://www.apollographql.com/docs/react/v2/caching/cache-configuration/#assigning-unique-identifiers

src/screens/BB-BUS/DetailScreen.js
```diff
@@ -40,24 +41,22 @@ const TEXT_BLOCKS_SORTER = {
...
 };
 
-const FormButton = ({ link, rootRouteName }) => {
+const FormButton = ({ link, name }) => {
   const { url } = link;
 
-  const openWebScreen = useOpenWebScreen(name, url, rootRouteName);
-
-  return <Button title={url} onPress={openWebScreen} invert />;
+  return <Button title={name} onPress={() => openURL(url)} invert />;
```
donni106, line 47: i think we should use `openLink` from our `linkHelper`, that wrapps `openURL` from `expo-linking` with some additional stuff

digorath: ✅

donni106 (approved):

PR #426, opened 2022-01-10
setup: sentry

donni106:
- followed instructions on https://docs.expo.dev/guides/using-sentry and the project getting started guide of Sentry
- the secrets needs to be extended with:
```
module.exports.sentryApi = {
  config: {
    authToken: '',
    organization: '',
    project: ''
  },
  dsn: ''
};
```
- added an additional JavaScript Expo config to be able to extend with a post publish hook for Sentry
  - JavaScript config is used to be able to import credentials from secrets
    > By default, Expo CLI will read the app.json first and send the normalized results to the app.config.js
    
    https://docs.expo.dev/workflow/configuration/#dynamic-configuration-with-appconfigjs
- added check for presence of `dsn` secret key in order to not init Sentry if it does not exist to prevent app crash
- updated secrets template
  - updated nesting of `secrets` and added missing keys
  - added `sentryApi` keys

SVA-404

---

connection tested successfully for iOS (first) and Android (second):

Bildschirmfoto 2022-01-10 um 13 11 01


App.js
```diff
@@ -1,9 +1,17 @@
...
+import * as Sentry from '@sentry/react-native';
 
 import { MainApp } from './src';
-import { fontConfig } from './src/config';
+import { fontConfig, sentryApi } from './src/config';
+
+Sentry.init({
+  dsn: sentryApi.dsn,
+  enableNative: false, // Note: Native crash reporting is not available with the classic build system (expo build:[ios|android]), but is available via EAS Build.
+  enableInExpoDevelopment: true,
```
okwast, line 12: Hey @donni106, I guess we want to turn this off as all errors while development will go to sentry if we leave this on.

okwast:
I'm not that deep in the processes here, how is the process of adding the sentry entry to the secrets?
As the app crashes when the entry in the secrets is missing I would suggest we add it to the secrets before merging.

Another question: Might it be useful to be able to turn Sentry on and off for individual clients?

okwast:
@donni106 Did you leave out the [step 3](https://docs.expo.dev/guides/using-sentry/#step-3-app-config) of the setup instructions on purpose or did you just miss it?
I verified that it already works, however we only get a stack trace from the main.js bundle which is quite cumbersome to work with. When we upload the source maps we get the stack trace based on the TS code which is definitely easier to work with.

okwast (changes requested):

donni106:
> @donni106 Did you leave out the [step 3](https://docs.expo.dev/guides/using-sentry/#step-3-app-config) of the setup instructions on purpose or did you just miss it? I verified that it already works, however we only get a stack trace from the main.js bundle which is quite cumbersome to work with. When we upload the source maps we get the stack trace based on the TS code which is definitely easier to work with.

yes. i have a separate commit for this as this should not be public in my understanding. i get in touch with you on slack on that.

App.js
donni106: true. i will comment it out, so that devs can quickly enable it if necessary.

donni106:
> I'm not that deep in the processes here, how is the process of adding the sentry entry to the secrets? As the app crashes when the entry in the secrets is missing I would suggest we add it to the secrets before merging.
> 
> Another question: Might it be useful to be able to turn Sentry on and off for individual clients?

good point. i will add a check for presence on the entry in the secrets. this way, we could also toggle it for different clients. even if this is not the best practice for individual toggling in long term.

App.js
donni106: done in e3046032453d60f92be801e71f824b42c134df45

donni106:
> i will add a check for presence on the entry in the secrets.

done in e304603

app.config.js
```diff
@@ -0,0 +1,12 @@
...
+export default {
+  hooks: {
+    postPublish: [
+      sentryApi?.config && {
+        file: 'sentry-expo/upload-sourcemaps',
+        config: sentryApi.config
+      }
+    ]
+  }
+};
```
donni106, line 12: with the new `app.config.js` my Expo Go app lists this project in "recently opened" now as `mobile-app` and not `Smart Village App` anymore???!??!???

donni106, line 12: i wasn't quite sure, if this extends or overwrites. in my understanding it extends simmply this way.
otherwise we would need to do:
```
export default ({ config }) => ({
  ...config,
  hooks: {
    ...
  }
});
```

okwast (approved):
Great job, looking good to me now. I could see errors of Android and iOS in the sentry console.

I fixed the app.config.js to be a function now. If you are happy with the changes we are good to merge this 🚀

app.config.js
donni106: nice, i wanted to hear this

donni106: great, thanks

PR #427, opened 2022-01-10
feat(categories): filter by ids

digorath:
- added ids as query param to categories query

---

SVA-366

---

## Screenshots:
Simulator Screen Shot - iPhone 12 Pro - 2022-01-10 at 17 13 52


okwast (approved):
Looking good to me. But are there no types for this?

digorath:
> Looking good to me. But are there no types for this?

unfortunately there are no types for the queries/queryVariables

PR #429, opened 2022-01-14
Fix: photo permission bug

### LINKED ISSUE #410: Type `Image`
it is probably a matter of finding a decent name for the destructured `source.props`. we tried already and failed, that is why we ended up with the solution of not destructuring it. which then causes type inference to malfunction.

changing it to be a typescript file would also be a solution, but i would need to check how many issues that brings with it, due to other things not being typed yet

_Originally posted by @digorath in https://github.com/ikuseiGmbH/smart-village-app-app/pull/398#discussion_r758585863_

digorath:
- trimmed the uri if present
- destructured source from props again to get rid of type inference issue

---

SVA-391
closes #410 

---

## How to test:

the home image on the internal development currently has whitespace in the url. with this fix the permission request should no longer trigger

### DIFF
--- a/src/components/Image.js
+++ b/src/components/Image.js
@@ -31,7 +31,7 @@ export const Image = ({
   resizeMode,
   borderRadius,
   refreshInterval,
-  ...props
+  source: sourceProp
 }) => {
   const [source, setSource] = useState(null);
   const { globalSettings } = useContext(SettingsContext);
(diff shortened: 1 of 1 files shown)


okwast (approved):
Looking good to me. Not triggering the media library request for me anymore. Even not having endless cycles. Good job. Ready to merge 🚀

donni106 (approved):

PR #438, opened 2022-01-26
fix(sentry): moved sentry config to another position in the secrets t…

okwast:
…o allow sentry config per instance

SVA-404

Moved sentry config in the secrets to be set per instance to allow turning it off, on and setting it to another sentry instance for every app instance.

### DIFF
--- a/App.js
+++ b/App.js
@@ -4,8 +4,9 @@ import React, { useEffect, useState } from 'react';
 import * as Sentry from 'sentry-expo';
 
 import { MainApp } from './src';
-import { fontConfig, sentryApi } from './src/config';
+import { fontConfig, namespace, secrets } from './src/config';
 
+let sentryApi = secrets[namespace].sentryApi;
 if (sentryApi?.dsn) {
   Sentry.init({
     dsn: sentryApi.dsn,
--- a/App.js
+++ b/App.js
@@ -6,7 +6,7 @@ import * as Sentry from 'sentry-expo';
 import { MainApp } from './src';
 import { fontConfig, namespace, secrets } from './src/config';
 
-let sentryApi = secrets[namespace].sentryApi;
+const sentryApi = secrets[namespace].sentryApi;
 if (sentryApi?.dsn) {
   Sentry.init({
     dsn: sentryApi.dsn,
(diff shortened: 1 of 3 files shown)


App.js
```diff
@@ -4,8 +4,9 @@ import React, { useEffect, useState } from 'react';
 import * as Sentry from 'sentry-expo';
 
 import { MainApp } from './src';
-import { fontConfig, sentryApi } from './src/config';
+import { fontConfig, namespace, secrets } from './src/config';
 
+let sentryApi = secrets[namespace].sentryApi;
```
donni106, line 9: `let`? `const` here, right?

app.config.js
```diff
@@ -1,13 +1,17 @@
-import { sentryApi } from './src/config/secrets';
+import { secrets } from './src/config/secrets';
+
+let hooks = secrets.sentryApi?.config
```
donni106, line 3: `let`? `const` here as well, right? or what was your thought?

donni106: `namespace` of `secrets` missing here?

donni106 (changes requested):

App.js
okwast: Yes definitely. I probably was still in Swift-mode 😅

app.config.js
okwast: Great catch 👏 

Fixed it to also take the namespace into account now and double checked that it is working as expected.

okwast:
@donni106 great review. Updated the PR now. Back to you 🔙

app.config.js
```diff
@@ -1,13 +1,20 @@
...
+export default ({ config }) => {
+  const sentryConfig = secrets[config.slug]?.sentryApi?.config;
+  const hooks = sentryConfig
+    ? {
+        postPublish: [
+          {
+            file: 'sentry-expo/upload-sourcemaps',
+            config: sentryConfig
+          }
+        ]
```
donni106, line 12: when reading this and the part in App.js i think the "same" stuff is done way too differently. it could be easier to read and capture if it is more in common, what do you think about:

```
const namespace = config.slug;
const sentryApi = secrets[namespace].sentryApi;
const hooks = sentryApi?.config
  ? {
      postPublish: [
        {
          file: 'sentry-expo/upload-sourcemaps',
          config: sentryApi.config
        }
      ]
```

okwast: I don't really mind in that case. I updated to your version and pushed it. Still works the same way 🙂

donni106 (approved):

PR #441, opened 2022-02-01
fix(leaflet-crash): switch to another fork

okwast:
- added a new commit to the fork of @digorath to disable hardware accelerration on android to prevent android devices from crashing when navigating to a screen containing a map

SVA-404

This PR updates the dependency to react-native-webview-leaflet to another custom fork which contains a change to disable hardware acceleration on android devices as this causes crashes on some Android devices when showing the map

---

SVA-404

---

## How to test:
* [X] Android portrait mode
* [ ] iOS portrait mode
* [ ] Android landscape mode
* [ ] iOS landscape mode

Open the app, open the drawer menu, open the app settings.
The app should crash at this point.

*NOTE:* The crash only happens sometimes and only on some devices. We could trigger the crash on two Pixel devices and a Samsung Galaxy S8

### DIFF
--- a/package.json
+++ b/package.json
@@ -89,7 +89,7 @@
     "react-native-snap-carousel": "~3.9.0",
     "react-native-svg": "12.1.1",
     "react-native-webview": "11.13.0",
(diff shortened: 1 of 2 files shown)


digorath (approved):
looks good to me. 💪

PR #446, opened 2022-02-03
fix: public json contents

donni106:
- removed `JSON.parse`s around public json contents because the
  `publicJsonFile` query now returns proper json instead of a string
  - therefore we also do not need the try catch bocks for possible
    errors with `JSON.parse` anymore

---

ℹ️ i will bring this to int/development demo app directly per `expo publish`

### DIFF
--- a/src/hooks/staticContent.ts
+++ b/src/hooks/staticContent.ts
@@ -78,7 +78,7 @@ export const useStaticContent = <T>({
 
     try {
       if (data) {
-        const json = JSON.parse(data?.publicJsonFile?.content);
+        const json = data?.publicJsonFile?.content;
 
         return parseFromJson ? parseFromJson(json) : json;
       }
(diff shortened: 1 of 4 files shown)


src/hooks/staticContent.ts
```diff
@@ -72,7 +72,7 @@ export const useStaticContent = <T>({
 
     try {
```
okwast, line 73: You said we don't need the try catches any more, but we want to keep it here, for the case the `parseFromJson` function throws an error?

okwast (approved):
@donni106 looking good to me and app works great again.

I personally prefer setting default values what you don't do here all over but as the next lines just return if it is false, it is only personal taste, so I'm fine with that.

I only found one hint. Feel free to ignore it though.

src/hooks/staticContent.ts
donni106: true, i also thought about it and decided to leave it as is due to uncertainty on my part in this point

PR #447, opened 2022-02-03
feature(feedback): Added feedback to detail screens

okwast:
* Created feedback footer
* Added feedback footer to detail screen
* Added route info to send feedback info

SVA-327

## Screenshots:

Simulator Screen Shot - iPhone 13 mini - 2022-02-03 at 15 18 53


digorath (approved):
Code looks good!
Works great!

src/components/Text.js
```diff
@@ -56,59 +56,65 @@ export const RegularText = styled(Text)`
...
+    props.lineThrough &&
+    css`
+      text-decoration: line-through;
+    `};
+
+  ${(props) =>
+    props.underline &&
+    css`
+      text-decoration: underline;
+    `};
```
donni106, line 69: any reason for moving them up here?

src/config/navigation/defaultStackConfig.tsx
```diff
@@ -153,7 +153,13 @@ export const defaultStackConfig = ({
     },
     {
       routeName: ScreenName.Form,
-      screenComponent: FormScreen
+      screenComponent: FeedbackScreen,
+      screenOptions: { title: texts.screenTitles.feedback }
```
donni106, line 157: this is kept for backwards compatibility, for cases where "FORM" is used to navigate?

src/components/Text.js
digorath: they need to be defined before the change of their color (which was added in this pr as well)

src/config/navigation/defaultStackConfig.tsx
digorath: yes

src/screens/FeedbackScreen.js
```diff
@@ -47,7 +50,8 @@ export const FormScreen = () => {
         email,
         phone,
         message,
-        consent
+        consent,
+        appInfo: appInfo
```
donni106, line 54: this could be just `appInfo`, right?

src/components/Text.js
donni106: good to know!

src/screens/FeedbackScreen.js
okwast: Yes, I renamed it and it seems like I missed that.

src/components/FeedbackFooter.tsx
```diff
@@ -0,0 +1,31 @@
...
+import { BoldText } from '.';
+
+export const FeedbackFooter: FC = () => {
+  const navigation = useNavigation();
+
+  return (
+    <View style={styles.container}>
+      <Pressable onPress={() => navigation.navigate('Feedback')} style={styles.button}>
+        <BoldText underline placeholder>
+          FEEDBACK
```
donni106, line 14: it would have been nicer if this had been written in the `texts`

src/components/FeedbackFooter.tsx
```diff
@@ -0,0 +1,31 @@
...
+import { Pressable, StyleSheet, View } from 'react-native';
+
+import { BoldText } from '.';
+
+export const FeedbackFooter: FC = () => {
+  const navigation = useNavigation();
+
+  return (
+    <View style={styles.container}>
+      <Pressable onPress={() => navigation.navigate('Feedback')} style={styles.button}>
```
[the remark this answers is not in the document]
donni106: #463

src/components/FeedbackFooter.tsx
donni106: #463

PR #448, opened 2022-02-04
Update base components

donni106:
feat: remove `big` prop for `BoldText`

- the `big` definition is already existing for `RegularText`
  thus was a duplication

feat: add smarter button style

- some apps already have a custom button style set
  which should be possible right from master, therefore
  a `buttonStyle` is added
- for `invert` buttons both styles should be applied,
  therefore the condition on `invert` is removed and
  the styles passed per array
- sorted styles alphabetically per key for better overview

### DIFF
--- a/src/components/Button.js
+++ b/src/components/Button.js
@@ -25,7 +25,7 @@ export const Button = ({ title, onPress, invert, disabled }) => {
       ]}
       disabledStyle={styles.buttonStyleDisabled}
       disabledTitleStyle={styles.titleStyle}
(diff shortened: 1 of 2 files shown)


digorath (approved):
for now this is ok, but if new releases incorporate a custom button style, the style should be moved to config, just like all the other configurable parts of the app

donni106:
> for now this is ok, but if new releases incorporate a custom button style, the style should be moved to config, just like all the other configurable parts of the app

good point. we can face that at herzberg

PR #450, opened 2022-02-09
fix(app-intro): fix app intro content below status bar

okwast:
Solved the issue that the app intro content is under the status bar. The status bar component in the app intro screen is not translucent any more.

SVA-305

---

## How to test:
* [x] Android portrait mode
* [ ] iOS portrait mode
* [ ] Android landscape mode
* [ ] iOS landscape mode


donni106:
branch name should be `feature/SVA-305-app-intro-status-bar-fix` for consistency, the `SVA` is missing.
@okwast i have also set you as assignee and added missing labels. please try to think of it next time.

donni106 (approved):

PR #451, opened 2022-02-10
fix(location overview): filter by location

digorath:
- added missing query variable
- changed the way the query variables get passed to the location overview to avoid future oversights

--- 

SVA-367

### DIFF
--- a/src/screens/IndexScreen.js
+++ b/src/screens/IndexScreen.js
@@ -210,9 +210,8 @@ export const IndexScreen = ({ navigation, route }) => {
           route={route}
           position={position}
           queryVariables={{
-            category: queryVariables.category,
-            categoryId: queryVariables.categoryId,
-            dataProvider: queryVariables.dataProvider
+            ...queryVariables,
+            limit: undefined
           }}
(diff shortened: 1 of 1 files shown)


src/screens/IndexScreen.js
```diff
@@ -210,9 +210,8 @@ export const IndexScreen = ({ navigation, route }) => {
           route={route}
           position={position}
           queryVariables={{
-            category: queryVariables.category,
-            categoryId: queryVariables.categoryId,
-            dataProvider: queryVariables.dataProvider
+            ...queryVariables,
+            limit: undefined
```
donni106, line 214: is `limit` already included in `queryVariables` and needs to be overwritten?

digorath: correct. that was the initial reason for selecting specific query variables there.
back then there were less query variables, so selecting seemed reasonable.

it turned out that even though new ones got added, the only one we want to keep omitting is the limit. so i changed it to overwrite the limit instead

donni106 (approved):

PR #452, opened 2022-02-11
fix: respect `item.bottomDivider` in list hooks if present

donni106:
a `bottomDivider` could be passed, which should be used in first case.


okwast (approved):
@donni106 looks good to me and works fine 👍 

Only one question: The `CardListItem` does not have a `bottomDivider` and therefore did not get the new option?

donni106:
> Only one question: The `CardListItem` does not have a `bottomDivider` and therefore did not get the new option?

correct, `bottomDivider` does not need any consideration with `CardListItem`

PR #456, opened 2022-02-11
fix(theming): Set the text color for the date picker of waste reminder

okwast:
* The settings of the waste calendar have a time picker to select when to be notified. The time within the picker was invisible when the iPhone was on DarkMode as it was light text on white background. The text color is now set to be textDark to be visible again

discovered while working on SVA-327

---

## How to test:
* [x] Android portrait mode
* [x] iOS portrait mode
* [x] Android landscape mode
* [x] iOS landscape mode

| Before | After |
| --- | --- |
| IMG_A39F21564079-1 | IMG_E8D6326B6C00-1 |

### DIFF
--- a/src/components/wasteCalendar/WasteReminderSettings.tsx
+++ b/src/components/wasteCalendar/WasteReminderSettings.tsx
@@ -347,6 +347,7 @@ export const WasteReminderSettings = ({
                           mode="time"
                           onChange={onDatePickerChange}
                           value={localSelectedTime || new Date()}
+                          textColor={colors.darkText}
                         />
                       </SafeAreaView>
                     </View>


src/components/wasteCalendar/WasteReminderSettings.tsx
```diff
@@ -28,6 +28,7 @@ import { ReminderSettings, WasteTypeData } from '../../types';
 import { Button } from '../Button';
 import { areValidReminderSettings, parseReminderSettings } from '../../jsonValidation';
 import { SectionHeader } from '../SectionHeader';
+import { FeedbackFooter } from '../FeedbackFooter';
```
donni106, line 31: hello sneaky import, how is it going?

donni106 (approved):

PR #459, opened 2022-02-14
fix(feedback footer): require cycle

digorath:
the import from `'.'` caused a require cycle, so i removed it.


okwast (approved):
Looking great. Hunt down all the require cycles!

PR #463, opened 2022-02-17
refactor(feedback): string to variables for consistency

donni106:


okwast (approved):
Nice improvement. Looking good and works fine.

PR #469, opened 2022-02-24
fix(static content): check for content emptiness

donni106:
- if a json static content is not found there could be
  a `content: {}` which we assume as error now
  - without it this crashed the app

### DIFF
--- a/src/hooks/staticContent.ts
+++ b/src/hooks/staticContent.ts
@@ -1,3 +1,4 @@
+import _isEmpty from 'lodash/isEmpty';
 import { useCallback, useContext, useMemo, useState } from 'react';
 import { useQuery } from 'react-apollo';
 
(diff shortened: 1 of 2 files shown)


okwast (approved):
Good job. Works as expected.

Ready to 🚀

src/hooks/staticContent.ts
```diff
@@ -77,10 +78,12 @@ export const useStaticContent = <T>({
     }
 
     try {
-      if (data) {
+      if (!_isEmpty(data?.publicJsonFile?.content)) {
         const json = data?.publicJsonFile?.content;
 
         return parseFromJson ? parseFromJson(json) : json;
+      } else {
+        setError(true);
```
donni106, line 86: this is now always fired for the time, data is not fetched entirely before isEmpty can check the real data.
it needs a check for loading state or that the response has finished with some data.

donni106: #479

PR #472, opened 2022-02-28
fix(waste calendar): Fixed the missing time zone

okwast:
The time send to the API only contained the hours and minutes so that the server assumed it is UTC while the user sees the time in their local time zone

SVA-467


donni106:
additional adjustments on main server for sending push notifications at correct time with https://github.com/ikuseiGmbH/smart-village-app-mainserver/pull/300

PR #474, opened 2022-03-02
feat(image): check status of image uri before using it

donni106:
- if an image uri responses with problematic state like 404
  we had an infinite loading spinner
- added a `fetch` for the image uri to check if it is available
  (status 200) before using it
- added a `NO_IMAGE` fallback source to identify problematic
  uris and render nothing for that cases

|before|after|
|---|---|
|Simulator Screen Shot - iPhone 11 Pro - 2022-03-02 at 15 33 10|Simulator Screen Shot - iPhone 11 Pro - 2022-03-02 at 15 33 20|

### DIFF
--- a/src/components/Image.js
+++ b/src/components/Image.js
@@ -92,7 +92,7 @@ export const Image = ({
     return () => (mounted = false);
   }, [timestamp, refreshInterval, sourceProp, setSource]);
 
-  if (source?.uri === 'NO_IMAGE') return null;
+  if (source?.uri === NO_IMAGE.uri) return null;
 
   return (
     <View>


src/components/Image.js
```diff
@@ -70,26 +72,41 @@ export const Image = ({
...
-                sourceProp.uri,
-                err
-              );
-              mounted && setSource(sourceProp);
-            })
-        : mounted && setSource(sourceProp);
+      if (sourceProp.uri) {
+        fetch(sourceProp.uri)
+          .then((response) => {
+            if (response?.status == 200) {
```
donni106, line 78: could this result in cases, where we have no image now compared with before? do we need to see more status as ok and fine here?

src/components/Image.js
```diff
@@ -70,26 +72,41 @@ export const Image = ({
...
+                .catch((err) => {
+                  console.warn(
+                    'An error occurred with cache management for an image',
+                    sourceProp.uri,
+                    err
+                  );
+                  mounted && setSource(NO_IMAGE);
+                });
+            } else {
+              mounted && setSource(NO_IMAGE);
```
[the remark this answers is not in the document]
okwast: Sounds reasonable.

src/components/Image.js
donni106: nice that you have checked the code and what you found out. love it, let's do it the way you suggested.

okwast (approved):
I added the changes I suggested now. Feel free to have a look and merge if you think this is OK.

src/components/Image.js
```diff
@@ -74,22 +76,24 @@ export const Image = ({
...
             })
-        : mounted && setSource(sourceProp);
+        : mounted && setSource(NO_IMAGE);
     }
 
     return () => (mounted = false);
   }, [timestamp, refreshInterval, sourceProp, setSource]);
 
+  if (source?.uri === 'NO_IMAGE') return null;
```
donni106, line 95: we should better check for `NO_IMAGE.uri` here instead of hardcoded string, right?

digorath: ✅

PR #478, opened 2022-03-03
fix(local time): parsing moments from utc strings

digorath:
- added new helper to use where needed

SVA-496

---

We should check in the future, which places might be affected as well.

### DIFF
--- a/src/helpers/momentHelper.js
+++ b/src/helpers/momentHelper.js
@@ -17,12 +17,12 @@ export const momentFormat = (
 ) => moment(dateTime, dateTimeFormat).format(returnFormat);
 
 /**
- * @param {string} date Date as utc string
+ * @param {string} dateTime Date as utc string
  * @param {string} [returnFormat] format to return in
  * @returns formated date to given format
  */
-export const momentFormatUtcToLocal = (date, returnFormat = 'DD.MM.YYYY') =>
-  moment.utc(date, 'YYYY-MM-DD HH:mm:ss Z').local().format(returnFormat);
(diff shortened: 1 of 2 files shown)


src/helpers/momentHelper.js
```diff
@@ -16,6 +16,14 @@ export const momentFormat = (
   dateTimeFormat = 'YYYY-MM-DD HH:mm:ss Z'
 ) => moment(dateTime, dateTimeFormat).format(returnFormat);
 
+/**
+ * @param {string} date Date as utc string
+ * @param {string} [returnFormat] format to return in
+ * @returns formated date to given format
+ */
+export const momentFormatUtcToLocal = (date, returnFormat = 'DD.MM.YYYY') =>
```
donni106, line 24: `date`? isn't it `dateTime` as well as for the existing `momentFormat` helper. espacially if the format is fix to `'YYYY-MM-DD HH:mm:ss Z'`

src/helpers/momentHelper.js
```diff
@@ -16,6 +16,14 @@ export const momentFormat = (
...
+  moment.utc(date, 'YYYY-MM-DD HH:mm:ss Z').local().format(returnFormat);
```
donni106, line 25: `date`? isn't it `dateTime` as well as for the existing `momentFormat` helper. espacially if the format is fix to `'YYYY-MM-DD HH:mm:ss Z'`

digorath: ✅

donni106 (approved):

PR #479, opened 2022-03-04
fix(static content): add condition for setting error

donni106:
- follow up for #469
- set error true if there is bad data without `publicJsonFile.content`
  or some `queryError`, otherwise it would be always true for the time
  data is not yet fetched entirely

### DIFF
--- a/src/hooks/staticContent.ts
+++ b/src/hooks/staticContent.ts
@@ -68,23 +68,25 @@ export const useStaticContent = <T>({
   );
 
   const refetchCallback = useCallback(async () => {
-    setError(false);
     return await refetch?.();
   }, [refetch]);
 
   const publicFileData = useMemo(() => {
+    setError(false);
+
     if (type === 'html') {
       return data?.publicHtmlFile?.content;
     }
 
     try {
-      if (!_isEmpty(data?.publicJsonFile?.content)) {
-        const json = data?.publicJsonFile?.content;
+      const json = data?.publicJsonFile?.content;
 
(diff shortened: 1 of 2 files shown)


src/hooks/staticContent.ts
```diff
@@ -82,7 +82,8 @@ export const useStaticContent = <T>({
         const json = data?.publicJsonFile?.content;
 
         return parseFromJson ? parseFromJson(json) : json;
-      } else {
+      } else if (data || queryError) {
+        // set error true if there is bad data without `publicJsonFile.content` or some `queryError`
```
okwast, line 86: Good point. It really makes sense to have some check here. Didn't see that in the review of the first PR.

However as the original issue was to check if there is data, I'm a little unsure if it is smart to check for data.
Wouldn't it make more sense to check for `loading` instead? Like `!loading || queryError`?

digorath:
with the current way of handling the error, the multi button screen does never show anything because there will always be an error!

src/hooks/staticContent.ts
donni106: at this particular point, we know that `data` is just correct, if it has `publicJsonFile.content`, otherwise it wouldn't be a correct static content response from the server. this means, if there is other `data` with some random information, we can handle it as error.
but i guess considering `loading` here in addition makes sense, so that we would not have an truthy error state if loading - i will change to: `!loading && data` to not have `error: true, loading: true` possible at the same time in the hook return. in addition i will remove `|| queryError` as this is already considered in the hook return for `error`.

PR #433, opened 2022-01-21
feat(empty screen): new empty message component

ardasnturk:
"No content found" text added by default on a page without content.

The EmptyPage.js file has been created in the Components folder. Updated IndexScreen.js and HtmlScreen.js pages. 

home screen events widget | no events for the current day | home screen about item at the bottom | no content
-- | -- | -- | --
 HomeScreen |  Simulator Screen Shot - iPhone 13 - 2022-01-21 at 13 33 27 | HomeScreen  |  Simulator Screen Shot - iPhone 13 - 2022-01-21 at 13 33 37 
HomeScreen | IndexScreen | HomeScreen | HtmlScreen

Fixes #49 

SVA-493

PR #481, opened 2022-03-09
feat(category tree): children categories display on screen

ardasnturk:
- A new Query has been added to IndexScreen.js,
- Filtering was done in the query,
- Children data has been added in the category query,
- Children data has been added to the Param information in the category parser.

## How to test:
* [x] iOS portrait mode
* [x] iOS landscape mode
* [ ] Android landscape mode
* [ ] Android portrait mode

## Screenshots:

Before | After |
|--|--
| Simulator Screen Shot - iPhone 13 - 2022-03-08 at 16 29 58 | Simulator Screen Shot - iPhone 13 - 2022-03-08 at 16 29 14 | 
IndexScreen.js | IndexScreen.js |

SVA-28


ardasnturk:
Update Details:

- Extra query removed,
- Navigation in CategoryListItem.js has been changed with push,
- Updated param parameter in ListItemParser.
- Added 2 props to CategoryList.js:
scrollEnabled
noSectionHeader

ardasnturk:
Update details:

- Sub category sorted alphabetically,
- Removed sub categories from the Main Category list,
- Fixed the title error on Sub category list and Sub category detail page,
- Sub Category list has been added under the header.

|Category List|SubCategory List|SubCategory Detail Screen|
|---|---|---|
Simulator Screen Shot - iPhone 13 - 2022-03-28 at 15 12 24|Simulator Screen Shot - iPhone 13 - 2022-03-28 at 15 12 07|Simulator Screen Shot - iPhone 13 - 2022-03-28 at 15 12 13|

SVA-28

src/helpers/parser/listItemParser.js
```diff
@@ -180,6 +180,7 @@ const parseCategories = (data, skipLastDivider, routeName = 'Category') => {
     pointsOfInterestCount: category.pointsOfInterestCount,
     toursCount: category.toursCount,
     routeName,
+    parent: category.parent ? true : false,
```
donni106, line 183: is this especially needed, or would `parent: category.parent` already work? the value of parent could be checked elsewhere as well, instead of having it `true` or `false` i guess.
on the other hand, if we would need to access data of the parent, these data were gone and exchanged with the `true`, right?

src/screens/IndexScreen.js
```diff
@@ -279,6 +270,13 @@ export const IndexScreen = ({ navigation, route }) => {
                 ListHeaderComponent={
                   showFilter ? (
                     <DropdownHeader {...{ query, queryVariables, data, updateListData }} />
+                  ) : subCategories?.length ? (
+                    <CategoryList
+                      navigation={navigation}
+                      data={subCategories}
+                      horizontal={false}
+                      noSectionHeader={true}
+                    />
```
donni106, line 279: currently this depends on `showFilter`, which could result in a missing category list, ih a filter is shown. both needs to be possible at the same time.

donni106, line 279: currently this depends on `showFilter`, which could result in a missing category list, ih a filter is shown. both needs to be possible at the same time.

src/components/CategoryList.js
```diff
@@ -27,24 +27,27 @@ export class CategoryList extends React.PureComponent {
...
+        data: _filter(data, (category) => category.toursCount > 0 && !category.parent)
       }
     ];
 
     return (
       <SectionList
         keyExtractor={this.keyExtractor}
         sections={sectionedData}
-        initialNumToRender={data.length}
+        initialNumToRender={data.length > 1 ? data.length : 2}
```
donni106, line 50: why is this condition necessary here? could you add a comment to explain right there at the code?

src/components/CategoryList.js
```diff
@@ -54,7 +57,7 @@ export class CategoryList extends React.PureComponent {
             section={section}
           />
         )}
-        renderSectionHeader={this.renderSectionHeader}
+        renderSectionHeader={noSectionHeader ? null : this.renderSectionHeader}
```
donni106, line 60: let's put the condition inside of `renderSectionHeader` to have the render part more readable here and the logic separated.

example for update of the function:

```
const { noSectionHeader } = this.props;

if (!data?.length || noSectionHeader) return null;
```

src/screens/IndexScreen.js
```diff
@@ -268,6 +270,13 @@ export const IndexScreen = ({ navigation, route }) => {
                 ListHeaderComponent={
                   showFilter ? (
                     <DropdownHeader {...{ query, queryVariables, data, updateListData }} />
+                  ) : subCategories?.length ? (
+                    <CategoryList
+                      navigation={navigation}
+                      data={subCategories}
+                      horizontal={false}
+                      noSectionHeader={true}
```
donni106, line 278: in most cases we want a section header. i think it is easier to think and to explain with a prop `hasSectionHeader` here, that default to `true` and is disabled with `false` here in this on place.

src/helpers/parser/listItemParser.js
```diff
@@ -173,15 +173,17 @@ const parseTours = (data, skipLastDivider) => {
...
     id: category.id,
     title: category.name,
     pointsOfInterestCount: category.pointsOfInterestCount,
     toursCount: category.toursCount,
-    routeName: 'Category',
+    routeName,
+    parent: category.parent ? true : false,
     params: {
       title: category.name,
+      categories: parseCategories(category.children, skipLastDivider, 'Index'),
```
donni106, line 186: instead of hardcoding the route string here, `ScreenName.Index` could be used.

src/helpers/parser/listItemParser.js
```diff
@@ -173,15 +173,17 @@ const parseTours = (data, skipLastDivider) => {
   }));
 };
 
-const parseCategories = (data, skipLastDivider) => {
+const parseCategories = (data, skipLastDivider, routeName = 'Category') => {
```
donni106, line 176: instead of hardcoding the route string here, `ScreenName.Category` could be used.

src/queries/categories.js
```diff
@@ -7,6 +7,15 @@ export const GET_CATEGORIES = gql`
       name
       pointsOfInterestCount
       toursCount
+      children {
+        id
+        name
+        pointsOfInterestCount
+        toursCount
+      }
```
donni106, line 15: in order to have multiple nested data in this request, we need more layers here:

```
children {
  id
  name
  pointsOfInterestCount
  toursCount
  children {
    id
    name
    pointsOfInterestCount
    toursCount
    children {
      id
      name
      pointsOfInterestCount
      toursCount
      children {
        id
        name
        pointsOfInterestCount
        toursCount
        children {
          id
          name
          pointsOfInterestCount
          toursCount
        }
      }
    }
  }
}
```

src/components/CategoryList.js
ardasnturk: If data.length is 1, it is not displayed on the screen. Section list needs more than 1 data.

src/components/CategoryList.js
ardasnturk: Yes I did and it worked fine 👍

src/helpers/parser/listItemParser.js
ardasnturk: change has been made

src/helpers/parser/listItemParser.js
ardasnturk: change has been made

src/screens/IndexScreen.js
ardasnturk: I edited the code. It's currently going through two different filters and showing whichever is needed.

src/screens/IndexScreen.js
ardasnturk: Yes that's true. I edited the code.

src/queries/categories.js
donni106: for now we use the multi-nested query structure

donni106 (approved):
i think this can land in 2.5.3 now

ardasnturk:
looks good

src/screens/IndexScreen.js
```diff
@@ -236,7 +236,7 @@ export const IndexScreen = ({ navigation, route }) => {
               withDate: false
             });
 
-            if (!listItems) return null;
+            if (!listItems?.length && !categories?.length) return null;
```
donni106, line 239: this could be changed to the new `EmptyMessage` from #433 after 2.5.3 merge

donni106: done with 3197128 and updated empty message text for lists with categories:

Simulator Screen Shot - iPhone 11 Pro - 2022-04-25 at 14 34 25

PR #492, opened 2022-03-29
Fix(location): Reduced accuracy for location access on iOS

okwast:
* The location access was to slow for older iOS devices resulting in lists depending on the location being either slow or not ordered by location
* Accuracy was highest which is <<10m accuracy and is now set to balanced which is <100m accuracy

SVA-473


digorath (approved):
looks good to me.
in my test the accuracy was not only <100 meters but more around 10-15 meters 💪

PR #506, opened 2022-04-20
feat(bb bus): migrate to api v4

digorath:
## What has changed

* this PR introduces the necessary changes to work with the new version of the bb bus server that connects to a newer version of the bus api

## Note

* when this gets merged, the area ids of the communities need to be updated accordingly, as the ids used by the API have changed.
* the url pointing to the bb bus server needs to be adjusted as well

---

SVA-423


donni106 (approved):

donni106:
⚠️  **NEW SECRETS NECESSARY** ⚠️

PR #516, opened 2022-04-28
style(consul): comment input

ardasnturk:
- added shadow to comment input (example taken from LinkedIn)
- relocated the submit button
- used `TouchableOpacity` is instead of `Button`
- changed placeholders of reply and comment inputs
- added send icon
---

## Screenshots:

|||
|--|--|
Simulator Screen Shot - iPhone 13 - 2022-04-28 at 16 31 11 | IMG_9463
|New Comment Input | LinkedIn Comment Input|

SVA-403

### DIFF
--- a/src/components/consul/detail/ConsulCommentListItem.js
+++ b/src/components/consul/detail/ConsulCommentListItem.js
@@ -2,7 +2,7 @@ import PropTypes from 'prop-types';
 import React, { useState } from 'react';
 import { useMutation } from 'react-apollo';
 import { useForm } from 'react-hook-form';
-import { Alert, StyleSheet, View } from 'react-native';
+import { Alert, StyleSheet, TouchableOpacity, View } from 'react-native';
 
 import { colors, Icon, normalize, texts } from '../../../config';
 import { ConsulClient } from '../../../ConsulClient';
@@ -12,11 +12,10 @@ import {
   DELETE_COMMENT
 } from '../../../queries/consul';
 import { ScreenName } from '../../../types';
-import { Button } from '../../Button';
 import { Input } from '../../form';
 import { RegularText } from '../../Text';
 import { Touchable } from '../../Touchable';
-import { WrapperRow, WrapperVertical } from '../../Wrapper';
+import { WrapperRow } from '../../Wrapper';
 
 import { ConsulPublicAuthor } from './ConsulPublicAuthor';
 
@@ -213,26 +212,25 @@ export const ConsulCommentListItem = ({ commentItem, onRefresh, replyList, navig
         : null}
 
       {showReply ? (
-        <>
+        <WrapperRow style={{ marginTop: normalize(10) }}>
           <Input
             multiline
             minHeight={50}
             name="comment"
             label={texts.consul.commentLabel}
-            placeholder={texts.consul.comment}
+            placeholder={texts.consul.answer}
             autoCapitalize="none"
             control={control}
+            chat
           />
-          <WrapperVertical>
-            <Button
-              onPress={handleSubmit(onSubmit)}
-              title={
-                isLoading ? texts.consul.submittingCommentButton : texts.consul.commentAnswerButton
-              }
-              disabled={isLoading}
-            />
-          </WrapperVertical>
-        </>
+          <TouchableOpacity
+            onPress={handleSubmit(onSubmit)}
+            style={styles.button}
+            disabled={isLoading}
+          >
+            <Icon.Send color={colors.primary} size={normalize(16)} />
+          </TouchableOpacity>
+        </WrapperRow>
       ) : null}
     </View>
   );
@@ -289,6 +287,12 @@ const styles = StyleSheet.create({
     flexDirection: 'row',
     justifyContent: 'flex-start',
     marginTop: normalize(10)
+  },
+  button: {
+    alignSelf: 'flex-end',
+    alignItems: 'flex-end',
+    marginBottom: normalize(10),
+    width: '10%'
   },
   deleteButton: {
     alignItems: 'center',
--- a/src/components/consul/detail/ConsulCommentListItem.js
+++ b/src/components/consul/detail/ConsulCommentListItem.js
@@ -2,7 +2,7 @@ import PropTypes from 'prop-types';
 import React, { useState } from 'react';
 import { useMutation } from 'react-apollo';
 import { useForm } from 'react-hook-form';
-import { Alert, StyleSheet, TouchableOpacity, View } from 'react-native';
+import { Alert, StyleSheet, TouchableOpacity, View, Keyboard } from 'react-native';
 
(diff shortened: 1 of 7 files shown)


src/components/screens/consul/Debates/DebateDetail.js
```diff
@@ -150,25 +150,26 @@ export const DebateDetail = ({ data, onRefresh, route, navigation }) => {
...
+              multiline
+              minHeight={50}
+              name="comment"
+              label={texts.consul.commentLabel}
+              placeholder={texts.consul.comment}
+              autoCapitalize="none"
+              control={control}
+              chat
+            />
+            <TouchableOpacity
```
donni106, line 165: the new comment input adjustment is missing for proposals

src/components/screens/consul/Debates/DebateDetail.js
```diff
@@ -150,25 +150,26 @@ export const DebateDetail = ({ data, onRefresh, route, navigation }) => {
...
               onPress={handleSubmit(onSubmit)}
-              title={
-                loading ? texts.consul.submittingCommentButton : texts.consul.commentAnswerButton
-              }
+              style={styles.button}
```
donni106, line 167: what do you think about hiding the keyboard after submission of comments? i felt a bit surprised, that it was still open after sending a test comment.

src/components/consul/detail/ConsulCommentListItem.js
```diff
@@ -213,26 +212,25 @@ export const ConsulCommentListItem = ({ commentItem, onRefresh, replyList, navig
...
-              onPress={handleSubmit(onSubmit)}
-              title={
-                isLoading ? texts.consul.submittingCommentButton : texts.consul.commentAnswerButton
-              }
-              disabled={isLoading}
-            />
-          </WrapperVertical>
-        </>
+          <TouchableOpacity
+            onPress={handleSubmit(onSubmit)}
```
donni106, line 227: i cannot click on the send icon for answers, if the keyboard is open. there is something missing on the view, that its pressable while keyboard is open i guess.

donni106 (changes requested):

src/config/Icon.tsx
```diff
@@ -147,6 +147,7 @@ export const Icon = {
   Settings: (props: IconProps) => (
     <NamedIcon name={device.platform === 'ios' ? 'ios-settings' : 'md-settings'} {...props} />
   ),
+  Send: (props: IconProps) => <NamedIcon name={'send'} {...props} />,
```
donni106, line 150: does `send` work for ios and android? if yes, i would suggest to format it `<NamedIcon name="send" {...props} />` because there is no javascript needed for the plain string.

src/components/screens/consul/Debates/DebateDetail.js
```diff
@@ -78,101 +77,100 @@ export const DebateDetail = ({ data, onRefresh, route, navigation }) => {
...
       reset();
+      Keyboard.dismiss();
     } catch (err) {
       console.error(err);
     }
   };
 
   return (
-    <SafeAreaViewFlex>
-      <WrapperWithOrientation>
```
donni106, line 88: i found in the commit message:

> - removed from detail pages in `ConsulDetailScreen`
>  because the main component is wrapped with
>  `WrapperWithOrientation` and `SafeAreaViewFlex`.

i was asking myself "what was removed"? i needed to take a look in the code to find out. could have been mentioned, that these components were removed, because already wrapped on a higher level.

src/components/screens/consul/Polls/PollDetail.js
```diff
@@ -122,19 +122,18 @@ export const PollDetail = ({ data, onRefresh, route, navigation }) => {
...
-        </Wrapper>
-      </WrapperWithOrientation>
-    </SafeAreaViewFlex>
+          <TouchableOpacity
+            onPress={handleSubmit(onSubmit)}
+            style={styles.button}
+            disabled={loading}
+          >
+            <Icon.Send color={colors.primary} size={normalize(16)} />
+          </TouchableOpacity>
```
donni106, line 133: while testing i wondered, if expired polls are ok to be commented, or if this should be turned off as well? how does consul handle that on their website? are comments possible forever?

src/components/consul/detail/ConsulCommentListItem.js
ardasnturk: i will look into this.

ardasnturk: adding `keyboardShouldPersistTaps` props to ListItem fixed the problem.

6d09d940ed07b93248d82864cc5fa3cd0d68ac79

src/config/Icon.tsx
donni106: checking 2752a877c0fc28a4677f432b87e92ca1689237ee is saying: 
> This commit does not belong to any branch on this repository, and may belong to a fork outside of the repository.

you mean 6d09d940ed07b93248d82864cc5fa3cd0d68ac79?

src/components/consul/detail/ConsulCommentListItem.js
donni106: why have you used always there? what would other options be that react native provides us? have you checked the existing code base?

src/components/screens/consul/Debates/DebateDetail.js
donni106: ok thanks.

did you understand what i mean? you wrote `removed from detail pages...`. 
but, what was removed?

src/config/Icon.tsx
ardasnturk: Yes. https://github.com/ikuseiGmbH/smart-village-app-app/commit/6d09d940ed07b93248d82864cc5fa3cd0d68ac79

src/components/consul/detail/ConsulCommentListItem.js
ardasnturk: yes, it doesn't need to be `always`. i will update it to `handled`.

ardasnturk: b725a4b6a93fddb8f9dde638db29c9c5fe0f8591

donni106: do you know why? what is the difference in our case?

ardasnturk: i read the documentation and understood how it works.

- 'never', tapping outside of the focused text input when the keyboard is up dismisses the keyboard. When this happens, children won't receive the tap.
- 'always', the keyboard will not dismiss automatically, and the scroll view will not catch taps, but children of the scroll view can catch taps.
- 'handled', the keyboard will not dismiss automatically when the tap was handled by children of the scroll view (or captured by an ancestor).
- false, deprecated, use 'never' instead
- true, deprecated, use 'always' instead

donni106 (approved):

PR #521, opened 2022-04-29
refactor(consul): code clearing

ardasnturk:
- simplified `RefreshControl` with `usePullToRefetch`
  to avoid code duplication.
- added "is" to the beginning of `useState(boolean)`
  expressions to understand that the variable
  is boolean.
- sorted imports and variables alphabetically
  to improve readability
- added `defaultValues` props to input to
  eliminate warn error
- deleted props names written twice to
  simplify code
- replaced `val` in 'onSubmit(val)' with a
  more understandable name
- updated propTypes to `navigation.object`
  as all necessary information is available in `navigation.object`
- removed component statement to improve
  readability of file directory
- moved normalize import from `react-native-
  elements` to `config` due to style bug
- renamed `onRefresh` props to `refetch` to
  eliminate name confusion

SVA-403


src/components/consul/detail/ConsulCommentListItem.js
```diff
@@ -36,14 +36,14 @@ const deleteCommentAlert = (onDelete) =>
...
-  const [showReply, setShowReply] = useState(false);
+export const ConsulCommentListItem = ({ commentItem, refetch, replyList, navigation }) => {
   const [isLoading, setIsLoading] = useState(false);
+  const [isShowReply, setIsShowReply] = useState(false);
+  const [isShowResponse, setIsShowResponse] = useState(false);
 
   const { control, handleSubmit, reset } = useForm({
     defaultValues: {
-      comment: ''
+      comment: null
```
donni106, line 46: why was this changed to null?

src/components/consul/detail/ConsulCommentListItem.js
```diff
@@ -36,14 +36,14 @@ const deleteCommentAlert = (onDelete) =>
 
 /* eslint-disable complexity */
 /* NOTE: we need to check a lot for presence, so this is that complex */
-export const ConsulCommentListItem = ({ commentItem, onRefresh, replyList, navigation }) => {
-  const [showResponse, setShowResponse] = useState(false);
-  const [showReply, setShowReply] = useState(false);
+export const ConsulCommentListItem = ({ commentItem, refetch, replyList, navigation }) => {
   const [isLoading, setIsLoading] = useState(false);
+  const [isShowReply, setIsShowReply] = useState(false);
+  const [isShowResponse, setIsShowResponse] = useState(false);
```
donni106, line 42: `show` would be enough to indicate a boolean. it does not have to be `is` all the time. see it as natural language when you ask for something.

1. is something loading? -> `isLoading`?
2. should replies be shown? -> `showReply`?
3. should responses be shown? -> `showResponses`?

questions for conditions could be understand very fine i guess. with `is` and `show` at the same time, it seems duplicated for me and i have difficulties to make a natural sentence. you know what i mean?

src/components/consul/detail/ConsulQuestionsList.js
```diff
@@ -1,35 +1,33 @@
...
-import { ConsulQuestionsListItem } from './ConsulQuestionsListItem';
 import { ConsulQuestionsDescriptionListItem } from './ConsulQuestionsDescriptionListItem';
+import { ConsulQuestionsListItem } from './ConsulQuestionsListItem';
 
-const text = texts.consul;
-
-export const ConsulQuestionsList = ({ data, onRefresh, token, disabled }) => {
-  const [userAnswer, setAnswerUser] = useState(false);
+export const ConsulQuestionsList = ({ data, refetch, token, disabled }) => {
+  const [isUserAnswer, setIsAnswerUser] = useState(false);
```
donni106, line 13: `isUserAnswer` does not match `setIsAnswerUser` as expected

src/components/consul/detail/ConsulVideo.js
```diff
@@ -24,14 +24,14 @@ export const ConsulVideoComponent = ({ videoUrl }) => {
...
+ConsulVideo.propTypes = {
   videoUrl: PropTypes.string
 };
 
 const styles = StyleSheet.create({
   videoView: {
     flex: 1,
-    width: '100%',
-    height: 250
+    height: 250,
```
donni106, line 34: nothing new in code here, but with seeing it i wonder if `250` always works for different devices and orientations? can you have a look in different situations please.

src/components/consul/detail/ConsulVoting.js
```diff
@@ -80,18 +80,18 @@ export const ConsulVotingComponent = ({ votesData, onRefresh, id }) => {
...
-  onRefresh: PropTypes.func
-};
-
 const styles = StyleSheet.create({
+  icon: {
+    paddingHorizontal: 10
+  },
   iconButton: {
     alignItems: 'center',
     marginHorizontal: 10
```
donni106, line 89: do we need normalization here? or is it ok not having it in this particular place.

src/components/screens/consul/Debates/DebateDetail.js
```diff
@@ -53,14 +53,14 @@ export const DebateDetail = ({ data, onRefresh, route, navigation }) => {
...
-      if (val) return setUserId(JSON.parse(val).id);
+    getConsulUser().then((userInfo) => {
+      if (userInfo) return setUserId(JSON.parse(userInfo).id);
     });
   }, []);
 
   const { control, handleSubmit, reset } = useForm({
     defaultValues: {
-      comment: ''
+      comment: null
```
donni106, line 63: same here as asked before, why changed to `null`?

src/components/screens/consul/Debates/NewDebate.js
```diff
@@ -42,15 +42,15 @@ export const NewDebate = ({ navigation, data, query }) => {
...
       id: data?.id,
       attributes: {
         translationsAttributes: {
-          title: val.title,
-          description: val.description
+          title: newDebateData.title,
+          description: newDebateData.description
         },
-        tagList: val.tagList,
+        tagList: newDebateData.tagList,
```
donni106, line 53: these should be secured with conditional chaining i guess to avoid crashes if `newDebateData` is undefined.

src/components/screens/consul/Polls/PollDetail.js
```diff
@@ -53,7 +53,7 @@ export const PollDetail = ({ data, onRefresh, route, navigation }) => {
 
   const { control, handleSubmit, reset } = useForm({
     defaultValues: {
-      comment: ''
+      comment: null
```
donni106, line 56: again this as twice before

src/components/screens/consul/Proposals/NewProposal.js
```diff
@@ -65,11 +72,11 @@ export const NewProposal = ({ navigation, data, query }) => {
...
-      title: data?.title,
-      description: data?.description,
-      tagList: data?.tagList?.toString(),
-      summary: data?.summary,
-      videoUrl: data?.videoUrl
+      title: data?.title ? data?.title : null,
+      description: data?.description ? data?.description : null,
+      tagList: data?.tagList ? data?.tagList?.toString() : null,
+      summary: data?.summary ? data?.summary : null,
+      videoUrl: data?.videoUrl ? data?.videoUrl : null
```
donni106, line 79: again, why have you extended the conditions here to return `null`s instead of `undefined`s?

src/components/screens/consul/Proposals/ProposalDetail.js
```diff
@@ -69,14 +69,14 @@ export const ProposalDetail = ({ data, onRefresh, route, navigation }) => {
...
-      if (val) return setUserId(JSON.parse(val).id);
+    getConsulUser().then((userInfo) => {
+      if (userInfo) return setUserId(JSON.parse(userInfo).id);
     });
   }, []);
 
   const { control, handleSubmit, reset } = useForm({
     defaultValues: {
-      comment: ''
+      comment: null
```
donni106, line 79: again this :)

src/components/screens/consul/Proposals/ProposalDetail.js
```diff
@@ -246,15 +246,13 @@ export const ProposalDetail = ({ data, onRefresh, route, navigation }) => {
 };
 /* eslint-enable complexity */
 
+const styles = StyleSheet.create({
+  imageContainerStyle: { alignSelf: 'center' }
```
donni106, line 250: i think style sheets are never one lined across the whole code base. could you please adjust it here as well?

src/screens/consul/ConsulIndexScreen.js
```diff
@@ -53,13 +53,13 @@ const showRegistrationFailAlert = (navigation) =>
...
-  const [refreshing, setRefreshing] = useState(false);
-  const [sortingType, setSortingType] = useState(INITIAL_TOP_SORTING);
   const [filterType, setFilterType] = useState(INITIAL_TOP_FILTERING_FOR_POLLS);
   const [queryVariables, setQueryVariables] = useState(route.params?.queryVariables ?? {});
+  const [sortingType, setSortingType] = useState(INITIAL_TOP_SORTING);
   const bookmarkable = route.params?.bookmarkable;
-  const query = route.params?.query ?? '';
   const extraQuery = route.params?.extraQuery ?? '';
+  const query = route.params?.query ?? '';
+  let type = null;
```
donni106, line 62: this `type` is used for two purposes, right? is there a specific reason using it globally? it is very not clear when reading `type = null`.
i saw the first for sorting list data, where `type` is used to store a selected one. couldn't it be a local `const selectedSortingType` for example?
and second usage is for filtering, where `type` is also used to store a selected one. couldn't it be a local `const selectedFilterType` for example?
this way there would no `type` variable, which has not much meaning and is used for multiple purposes.

src/screens/consul/LoginRegister/ConsulLoginScreen.js
```diff
@@ -34,42 +35,51 @@ const showLoginFailAlert = () =>
...
 
   const {
     control,
     formState: { errors },
     handleSubmit
-  } = useForm();
+  } = useForm({
+    defaultValues: {
+      email: null,
+      password: null
```
donni106, line 48: does react hook form like `null` more than `''`?

donni106 (changes requested):
i like this pull request a lot! with a few answers and adjustments this will be a big improvement for the code base.

src/components/consul/detail/ConsulCommentListItem.js
ardasnturk: i thought `null` was more accurate both in terms of readability and in terms of disk space.

src/components/consul/detail/ConsulCommentListItem.js
ardasnturk: Yes, I understand better. Thank you.

src/components/consul/detail/ConsulCommentListItem.js
donni106: does the react-hook-form docs say something about that?

src/components/consul/detail/ConsulVoting.js
ardasnturk: I have tested. Better to use. It is not understandable unless you are very careful with devices of different sizes. But still there is a slight difference.

src/components/consul/detail/ConsulVideo.js
ardasnturk: I tested. There is no big difference in devices of different sizes. How do you think I can do this?

src/components/screens/consul/Debates/DebateDetail.js
ardasnturk: i thought null was more accurate both in terms of readability and in terms of disk space.

src/components/screens/consul/Debates/NewDebate.js
```diff
@@ -21,17 +21,17 @@ const graphqlErr = (err) => Alert.alert('Hinweis', err);
...
     formState: { errors },
     handleSubmit
   } = useForm({
     defaultValues: {
-      title: data?.title,
-      description: data?.description,
-      tagList: data?.tagList?.toString()
+      title: data?.title ? data?.title : null,
+      description: data?.description ? data?.description : null,
+      tagList: data?.tagList ? data?.tagList?.toString() : null
```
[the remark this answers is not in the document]
ardasnturk: React-Hook-Form was giving warn error for `defaultValues`. To clear this error.

src/screens/consul/LoginRegister/ConsulLoginScreen.js
ardasnturk: The documentation states that you can use null or empty strings.

`defaultValues` Note: It is encouraged that you set a defaultValue for all inputs to non-undefined such as the empty string or null.

[React-Hook-Form v6](https://react-hook-form.com/v6/api/)

donni106: ah ok. good.
what do you think about using empty string for string fields and null for others?

src/components/consul/detail/ConsulVideo.js
donni106: i thought for landscape for example the video will be very wide but still 250 in height. but maybe our orientation wrapper already saves us for having such "weird" renderings.

src/screens/consul/LoginRegister/ConsulLoginScreen.js
ardasnturk: All code needs to stay in the same structure. I will update my code with empty string.

src/screens/consul/ConsulIndexScreen.js
ardasnturk: Good suggestion. I wanted to do 2 different jobs with a single variable.

src/screens/consul/LoginRegister/ConsulLoginScreen.js
donni106: do you mean updating empty strings just for string values or for everything? i think we should not have default values with empty strings, if the values at the end are numbers for examples.

src/components/consul/detail/ConsulVideo.js
ardasnturk: Most probably. Should I change this?

src/components/consul/detail/ConsulCommentListItem.js
ardasnturk: https://github.com/ikuseiGmbH/smart-village-app-app/pull/521#discussion_r862853106

src/components/consul/detail/ConsulVideo.js
donni106: did you encounter problems now that need to be changed? which one?

src/screens/consul/ConsulIndexScreen.js
ardasnturk: 8d1121aeab4ff33f5e37bc5f2568faec55c0b11f

src/components/consul/detail/ConsulVoting.js
ardasnturk: 8d1121aeab4ff33f5e37bc5f2568faec55c0b11f

src/components/consul/detail/ConsulVideo.js
ardasnturk: i didn't see any problem.

src/components/screens/consul/Proposals/NewProposal.js
```diff
@@ -120,13 +120,13 @@ export const NewProposal = ({ navigation, data, query }) => {
...
-          title: newProposalData.title,
-          summary: newProposalData.summary,
-          description: newProposalData.description
+          title: newProposalData?.title,
+          summary: newProposalData?.summary,
+          description: newProposalData?.description
         },
-        tagList: newProposalData.tagList,
+        tagList: newProposalData?.tagList,
         termsOfService: termsOfService,
```
donni106, line 128: `termsOfService: termsOfService` -> `termsOfService` ;)

donni106: and this is also a boolean, i just see: `const [termsOfService, setTermsOfService] = useState(data?.termsOfService ?? false);`

so please change it to `hasAcceptedTermsOfService` or something like that more understandable

src/screens/consul/ConsulIndexScreen.js
```diff
@@ -59,7 +59,8 @@ export const ConsulIndexScreen = ({ navigation, route }) => {
   const bookmarkable = route.params?.bookmarkable;
   const extraQuery = route.params?.extraQuery ?? '';
   const query = route.params?.query ?? '';
-  let type = null;
+  const selectedSortingType = sortingType.find((data) => data.selected);
+  const selectedFilterType = filterType.find((data) => data.selected);
```
donni106, line 63: please place them in the corresponding methods. we do not need them globally. it is only needed in the local methods (list data callback for sorting and effect for filtering)

src/components/consul/detail/ConsulVideo.js
donni106: ok, then we do not need to change anything.

src/screens/consul/ConsulIndexScreen.js
ardasnturk: 7a3a9dacaabe27e7b1291911444f7d2b05906983

src/components/screens/consul/Proposals/NewProposal.js
ardasnturk: 7a3a9dacaabe27e7b1291911444f7d2b05906983

donni106 (approved):

PR #522, opened 2022-04-30
fix(consul): android crash fix

ardasnturk:
- fixed the error caused by `TouchableNativeFeedback` used for android
   because more than one component is not
   supported.
- replaced with `TouchableOpacity` as there
   is no `Touchable` style support in
  `ConsulQuestionsListItem`.

## How to test:
* [x] Android portrait mode
* [ ] iOS portrait mode
* [x] Android landscape mode
* [ ] iOS landscape mode

## Screenshots:

|||
|--|--|
Screenshot_20220430-181004 | Screenshot_20220430-180824 | 
|old question item|new question item|

SVA-403

### DIFF
--- a/src/components/consul/detail/ConsulCommentListItem.js
+++ b/src/components/consul/detail/ConsulCommentListItem.js
@@ -145,11 +145,13 @@ export const ConsulCommentListItem = ({ commentItem, onRefresh, replyList, navig
 
           {commentUserId === userId && (
             <>
-              <Touchable onPress={() => deleteCommentAlert(onDelete)} style={styles.deleteButton}>
-                <Icon.Trash size={normalize(12)} color={colors.error} />
-                <RegularText error smallest>
-                  {` ${texts.consul.delete}`}
-                </RegularText>
+              <Touchable onPress={() => deleteCommentAlert(onDelete)}>
+                <View style={styles.deleteButton}>
+                  <Icon.Trash size={normalize(12)} color={colors.error} />
+                  <RegularText error smallest>
+                    {` ${texts.consul.delete}`}
+                  </RegularText>
+                </View>
               </Touchable>
 
               <Space />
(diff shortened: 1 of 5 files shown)


src/components/consul/detail/ConsulQuestionsListItem.js
```diff
@@ -58,7 +57,7 @@ export const ConsulQuestionsListItem = ({ questionItem, onRefresh, token, disabl
           ]}
         >
           <RegularText placeholder={!disabled}>{item.title}</RegularText>
-        </Touchable>
+        </TouchableOpacity>
```
donni106, line 60: there is also a `<Touchable` including `<RegularText`... in `ConsulCommentListItem`. Does this need adjustments as well?

src/components/consul/detail/ConsulCommentListItem.js
```diff
@@ -241,13 +243,15 @@ export const ConsulCommentListItem = ({ commentItem, onRefresh, replyList, navig
...
-      <RegularText smallest>{like ? cachedVotesUp : cachedVotesDown}</RegularText>
+    <Touchable onPress={onPress}>
+      <View style={styles.iconButton}>
+        <Icon.Like
+          color={color}
+          style={[styles.icon, { transform: disslike && [{ rotateX: '180deg' }] }]}
+          size={normalize(16)}
+        />
+        <RegularText smallest>{like ? cachedVotesUp : cachedVotesDown}</RegularText>
+      </View>
```
donni106, line 254: according to this, does the following from `ConsulDocumentListItem `also need a change?

```
<Touchable
  style={[styles.button, styles.buttonClose]}
  onPress={() => setModalVisible(!modalVisible)}
>
  <Icon.Close color={colors.lightestText} size={normalize(16)} />
</Touchable>
```

ardasnturk: Yes, I missed this. I'm updating now.

ardasnturk: 92dbcae56ec8279ee67d9b39e7fd53ac74c1a527

src/screens/consul/ConsulIndexScreen.js
```diff
@@ -113,7 +113,7 @@ export const ConsulIndexScreen = ({ navigation, route }) => {
 
   if (isError) showRegistrationFailAlert(navigation);
 
-  if (!Component) return null;
+  if (!Component || !listItems) return null;
```
donni106, line 116: what about empty screen component here?

ardasnturk: i thought about it. we have `ListEmptyComponent` inside each of our lists. we show it here. in the current situation, an alert appears on the screen and when the button is pressed, it returns to the previous page.

donni106: ok

PR #525, opened 2022-05-04
feat(intro): repeat app intro

ardasnturk:
- added an option to the `SettingsScreen` so that
  users can see the intro of the application again
- added new texts to the `texts` file for the intro
  option in the settings

---

## How to test:
* [ ] Android portrait mode
* [x] iOS portrait mode
* [ ] Android landscape mode
* [x] iOS landscape mode

## Screenshots:

|SettingsScreen.js|SettingsScreen.js|SettingsScreen.js|AppIntroScreen.js|
|--|--|--|--|
Simulator Screen Shot - iPhone 13 - 2022-05-04 at 10 36 26 | Simulator Screen Shot - iPhone 13 - 2022-05-04 at 10 36 30 |  Simulator Screen Shot - iPhone 13 - 2022-05-04 at 10 36 34 | Simulator Screen Shot - iPhone 13 - 2022-05-04 at 10 36 45
|Settings Screen|If the user turns on this option, they will see an alert|If the user turns this option off, they will see an alert|If the application is restarted with the option turned on, the intro will appear on the screen|

SVA-458

### DIFF
--- a/src/config/texts.js
+++ b/src/config/texts.js
@@ -476,7 +476,7 @@ export const texts = {
     onboarding: {
       onActivate: 'Beim nächsten Start wird die App-Einführung angezeigt.',
       onDeactivate: 'Die App-Einführung wird beim nächsten Start nicht angezeigt.',
-      ok: 'OK'
+      ok: 'Ok'
     },
     permanentFilter: {
       sectionHeader: 'Datenquellen'
--- a/src/config/texts.js
+++ b/src/config/texts.js
@@ -473,6 +473,11 @@ export const texts = {
       save: 'Speichern',
       sectionHeader: 'Standort'
     },
+    onboarding: {
+      onActivate: 'Beim nächsten Start wird die App-Einführung angezeigt.',
+      onDeactivate: 'Die App-Einführung wird beim nächsten Start nicht angezeigt.',
+      ok: 'OK'
+    },
     permanentFilter: {
       sectionHeader: 'Datenquellen'
     }
@@ -492,6 +497,7 @@ export const texts = {
       textList: 'Textliste'
     },
     locationService: 'Ortungsdienste',
+    onboarding: 'App-Einführung',
     pushNotifications: 'Push-Benachrichtigungen',
     tabs: {
       general: 'Allgemein',
--- a/src/screens/SettingsScreen.js
+++ b/src/screens/SettingsScreen.js
@@ -23,6 +23,7 @@ import {
 import { useMatomoTrackScreenView } from '../hooks';
 import { PushNotificationStorageKeys, setInAppPermission } from '../pushNotifications';
 import { SettingsContext } from '../SettingsProvider';
+import { ONBOARDING_STORE_KEY } from '../OnboardingManager';
 
 const { MATOMO_TRACKING } = consts;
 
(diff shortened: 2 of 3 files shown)


src/screens/SettingsScreen.js
```diff
@@ -52,6 +58,8 @@ const INITIAL_FILTER = [
   { id: TOP_FILTER.LIST_TYPES, title: texts.settingsTitles.tabs.listTypes, selected: false }
 ];
 
+const ONBOARDING_STORE_KEY = 'ONBOARDING_STORE_KEY';
```
donni106, line 61: this key shouldn't be duplicated. it is already existing in `OnboardingManager`, where we should export it to use it here.

src/screens/SettingsScreen.js
```diff
@@ -132,6 +140,42 @@ export const SettingsScreen = () => {
...
 
+      if (settings.onboarding) {
+        const onboarding = await readFromStore(ONBOARDING_STORE_KEY);
+
+        additionalSectionedData.push({
+          data: [
+            {
+              title: texts.settingsTitles.onboarding,
+              topDivider: false,
+              value: onboarding === 'complete' ? false : true,
```
donni106, line 151: `onboarding === 'complete'` this expression already returns `true` or `false`, which is not needed to do explicitly.
i would suggest `value: onboarding === 'complete'` here

src/screens/SettingsScreen.js
```diff
@@ -132,6 +140,42 @@ export const SettingsScreen = () => {
...
       }
 
+      if (settings.onboarding) {
+        const onboarding = await readFromStore(ONBOARDING_STORE_KEY);
+
+        additionalSectionedData.push({
+          data: [
+            {
+              title: texts.settingsTitles.onboarding,
+              topDivider: false,
```
donni106, line 150: as this is the third entry, a top divider is necessary to separate from matomo setting.

src/config/texts.js
```diff
@@ -473,6 +473,11 @@ export const texts = {
       save: 'Speichern',
       sectionHeader: 'Standort'
     },
+    onboarding: {
+      onActivate: 'Beim nächsten Start wird die App-Einführung angezeigt.',
+      onDeactivate: 'Die App-Einführung wird beim nächsten Start nicht angezeigt.',
+      ok: 'OK'
```
donni106, line 479: for the `dateTimePicker` around line 65 we have `ok: 'Ok'`, so i would suggest to do it the same way here for consistency.

donni106 (changes requested):
nice PR description and well done. just a few things to make it perfect.

src/config/texts.js
ardasnturk: I wanted to add it again for code readability. I will use this.

ardasnturk: should it stay like this for now? the same is true for matomo setting.

src/screens/SettingsScreen.js
ardasnturk: a7587b4e5e398ac83c0cd01a398797166b4d72b2

src/screens/SettingsScreen.js
ardasnturk: a7587b4e5e398ac83c0cd01a398797166b4d72b2

src/screens/SettingsScreen.js
ardasnturk: a7587b4e5e398ac83c0cd01a398797166b4d72b2

src/config/texts.js
ardasnturk: a7587b4e5e398ac83c0cd01a398797166b4d72b2

donni106 (approved):

PR #528, opened 2022-05-09
fix(volunteer): select multiple dropdown

ardasnturk:
although the index changed when searching via dropdown, the data the user wanted to select was being matched with the data before the search was made
the array shown on the screen and the selected array are not the same

### DIFF
--- a/src/components/DropdownSelect.js
+++ b/src/components/DropdownSelect.js
@@ -89,7 +89,13 @@ export const DropdownSelect = ({
           let updatedData = [...data];
 
           if (multipleSelect) {
-            updatedData[index].selected = !updatedData[index].selected;
+            updatedData = updatedData.map((entry) => {
+              if (entry.value === value) {
+                entry.selected = !entry.selected;
+              }
+
+              return entry;
+            });
           } else {
             // only trigger onPress if a new selection is made
             if (selectedValue === value) return;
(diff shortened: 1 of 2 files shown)


src/components/DropdownSelect.js
```diff
@@ -89,7 +89,9 @@ export const DropdownSelect = ({
           let updatedData = [...data];
 
           if (multipleSelect) {
-            updatedData[index].selected = !updatedData[index].selected;
+            updatedData.map((entry) => {
+              if (entry.value === value) entry.selected = !entry.selected;
+            });
```
donni106, line 94: nice, so this is the same as for not multiple, right?

```
updatedData = updatedData.map((entry) => ({
  ...entry,
  selected: entry.value === value
}));
```

i will test it and then reduce duplication before merge.

ardasnturk: in theory it looks the same. but i tried both. it works better that way. it will be much better if you find a more optimized method.

donni106: no not at all. it needs to be different.

donni106: but i missed the returned value updating `updatedData`, which we pass in `setData(updatedData);` a few lines later. so i added that part. it seemed that mutating `updatedData` implicitly also worked, but isn't that readable and understandable.

donni106 (approved):

PR #529, opened 2022-05-12
feature: images upload

ardasnturk:
- added a new type named `PICKER` for the
  photo upload component to the `INPUTS` JSON
- added a new button to the render function of
  the `Controller` component to select an image
  using the image picker
- added 2 different error messages because the
  server only supports jpg format and the size of
  the photo can be up to 1 megabyte.
- added `imageAttributes` to variables object
  for image to be used in proposal mutation
- used `useSelectImage()` hooks to select
  image from library
- deleted `resourceId` props from `upload.js`
  as we don't need it
- changed the name of the alert function that
  appears if the user has not approved the
  TermsOfService to be more clear
- added defaultValue for image to remove warn
- added new `image` props to proposal edit
  button to show image on edit page
- added new texts to `texts` file to show in
  photo upload component
- added new props to `useSelectImage`
  hook to adjust image quality and reduce
  size
- added `if` check in `useEffect` in `Controller`
  to check if there is image information in `value`
- added new props to `ImagePicker.ts` as we
  need images of different sizes and kept the
  default props in `launchImageLibraryAsync`

⚠️ image upload function does not work for editing proposal for now! ⚠️
In order to use the editing method, the `cachedImagePath`
information must be in the `query`. The server does not accept 
`absoluteUrl` for photo upload.

---

## How to test:
* [x] Android portrait mode
* [x] iOS portrait mode
* [x] Android landscape mode
* [x] iOS landscape mode

## Screenshots:

||||
|--|--|--|
Simulator Screen Shot - iPhone 13 - 2022-05-12 at 14 33 47 | Simulator Screen Shot - iPhone 13 - 2022-05-12 at 14 34 03 | Simulator Screen Shot - iPhone 13 - 2022-05-12 at 14 34 07
Simulator Screen Shot - iPhone 13 - 2022-05-17 at 15 10 35 | Simulator Screen Shot - iPhone 13 - 2022-05-17 at 15 10 28 | Simulator Screen Shot - iPhone 13 - 2022-05-17 at 15 10 22

SVA-596

### DIFF
--- a/src/components/screens/consul/Proposals/NewProposal.js
+++ b/src/components/screens/consul/Proposals/NewProposal.js
@@ -61,12 +61,6 @@ export const NewProposal = ({ navigation, data, query }) => {
   );
   const [isLoading, setIsLoading] = useState(false);
   const [tags, setTags] = useState([]);
-  const { selectImage } = useSelectImage(
-    undefined, // onChange
-    false, // allowsEditing,
-    undefined, // aspect,
-    undefined // quality
-  );
 
   const {
     control,
@@ -283,7 +277,7 @@ export const NewProposal = ({ navigation, data, query }) => {
             <Controller
               name={item.name}
               control={control}
-              render={(field) => <ImageSelector {...{ control, field, item, selectImage }} />}
+              render={(field) => <ImageSelector {...{ control, field, item }} />}
             />
           )}
         </Wrapper>
@@ -320,12 +314,19 @@ export const NewProposal = ({ navigation, data, query }) => {
   );
 };
 
-const ImageSelector = ({ control, field, item, selectImage }) => {
+const ImageSelector = ({ control, field, item }) => {
   const [errorMessage, setErrorMessage] = useState('');
   const [imageInfoText, setImageInfoText] = useState('');
 
   const { buttonTitle, infoText } = item;
   const { name, onChange, value } = field;
+
+  const { selectImage } = useSelectImage(
+    undefined, // onChange
+    false, // allowsEditing,
+    undefined, // aspect,
+    undefined // quality
+  );
 
   return (
     <>
--- a/src/components/screens/consul/Proposals/NewProposal.js
+++ b/src/components/screens/consul/Proposals/NewProposal.js
@@ -48,26 +48,11 @@ const ITEM_TYPES = {
   PICKER: 'picker'
 };
 
-const imageProps = {
-  onChange: null,
-  allowsEditing: false,
-  aspect: null,
-  quality: 0.2
-};
-
 const showPrivacyCheckRequireAlert = () =>
   Alert.alert(texts.consul.privacyCheckRequireTitle, texts.consul.privacyCheckRequireBody);
 const graphqlErr = (err) => Alert.alert(texts.consul.privacyCheckRequireTitle, err);
-const showImageUploadSizeError = () =>
-  Alert.alert(
-    texts.consul.privacyCheckRequireTitle,
-    texts.consul.startNew.newProposalImageUploadSizeErrorAlertTitle
-  );
-const showImageUploadFormatError = () =>
-  Alert.alert(
-    texts.consul.privacyCheckRequireTitle,
-    texts.consul.startNew.newProposalImageUploadFormatErrorAlertTitle
-  );
+const showImageUploadError = (errorText) =>
+  Alert.alert(texts.consul.privacyCheckRequireTitle, errorText);
 
 export const NewProposal = ({ navigation, data, query }) => {
   const [hasAcceptedTermsOfService, setHasAcceptedTermsOfService] = useState(
@@ -76,10 +61,10 @@ export const NewProposal = ({ navigation, data, query }) => {
   const [isLoading, setIsLoading] = useState(false);
   const [tags, setTags] = useState([]);
   const { imageUri, selectImage } = useSelectImage(
-    imageProps.onChange,
-    imageProps.allowsEditing,
-    imageProps.aspect,
-    imageProps.quality
+    undefined, // onChange
+    false, // allowsEditing,
+    undefined, // aspect,
+    0.2 // quality
   );
 
   const {
@@ -133,8 +118,31 @@ export const NewProposal = ({ navigation, data, query }) => {
     setValue('tagList', filterData.toString());
   }, [tags]);
 
+  const uploadImage = async (imageData) => {
+    let imageAttributes = null;
+    const imageUploadData = await uploadAttachment(imageData, 'image');
+
+    if (imageUploadData.status === 200) {
+      const { filename, cached_attachment } = JSON.parse(imageUploadData.body);
+
+      imageAttributes = {
+        title: filename,
+        cachedAttachment: cached_attachment
+      };
+    } else if (imageUploadData.status === 422) {
+      const errors = JSON.parse(imageUploadData.body).errors.toLowerCase().split(' ').join('-');
+
+      showImageUploadError(
+        texts.consul.startNew[errors] ?? texts.consul.startNew.generalPhotoUploadError
+      );
+
+      return;
+    }
+
+    return imageAttributes;
+  };
+
   const onSubmit = async (newProposalData) => {
-    let imageData = null;
     let variables = {
       id: data?.id,
       attributes: {
@@ -161,25 +169,12 @@ export const NewProposal = ({ navigation, data, query }) => {
     if (!hasAcceptedTermsOfService) return showPrivacyCheckRequireAlert();
 
     if (newProposalData.image) {
-      const imageUploadData = await uploadAttachment(newProposalData.image, 'image', 'Proposal');
+      const imageAttributes = await uploadImage(newProposalData.image);
 
-      if (imageUploadData.status === 200) {
-        imageData = JSON.parse(imageUploadData.body);
-
-        let imageAttributes = {
-          title: imageData?.filename,
-          cachedAttachment: imageData?.cached_attachment
-        };
-
-        variables.attributes = { ...variables.attributes, imageAttributes };
-      } else if (imageUploadData.status === 422) {
-        if (JSON.parse(imageUploadData.body).errors.length < 100) {
-          showImageUploadSizeError();
-        } else {
-          showImageUploadFormatError();
-        }
-
+      if (!imageAttributes) {
         return;
+      } else {
+        variables.attributes = { ...variables.attributes, imageAttributes };
       }
     }
 
@@ -292,42 +287,11 @@ export const NewProposal = ({ navigation, data, query }) => {
 
           {item.type === ITEM_TYPES.PICKER && (
             <Controller
-              key={item.name}
               name={item.name}
               control={control}
-              render={({ onChange, value }) => {
-                useEffect(() => {
-                  if (!value) {
-                    onChange(imageUri);
-                  }
-                }, [imageUri]);
-
-                return (
-                  <>
-                    <Input
-                      hidden
-                      label={item.label}
-                      control={control}
-                      name={item.name}
-                      value={value}
-                    />
-                    <RegularText smallest placeholder>
-                      {item.infoText}
-                    </RegularText>
-
-                    {value ? (
-                      <WrapperRow center spaceBetween>
-                        <Image source={{ uri: value }} style={styles.image} />
-                        <Touchable onPress={() => onChange(null)}>
-                          <Icon.Trash color={colors.error} size={normalize(16)} />
-                        </Touchable>
-                      </WrapperRow>
-                    ) : (
-                      <Button title={item.buttonTitle} invert onPress={selectImage} />
-                    )}
-                  </>
-                );
-              }}
+              render={({ onChange, value }) => (
+                <ImageSelector {...{ control, imageUri, item, onChange, selectImage, value }} />
+              )}
             />
           )}
         </Wrapper>
@@ -361,6 +325,38 @@ export const NewProposal = ({ navigation, data, query }) => {
   );
 };
 
+const ImageSelector = ({ control, imageUri, item, onChange, selectImage, value }) => {
+  const { buttonTitle, infoText, label, name } = item;
+
+  useEffect(() => {
+    // the `!value` control is used for editing the proposal
+    // the proposal contains an image, it is for showing the preview to the users
+    if (!value) {
+      onChange(imageUri);
+    }
+  }, [imageUri]);
+
+  return (
+    <>
+      <Input hidden label={label} control={control} name={name} value={value} />
+      <RegularText smallest placeholder>
+        {infoText}
+      </RegularText>
+
+      {value ? (
+        <WrapperRow center spaceBetween>
+          <Image source={{ uri: value }} style={styles.image} />
+          <Touchable onPress={() => onChange(null)}>
+            <Icon.Trash color={colors.error} size={normalize(16)} />
+          </Touchable>
+        </WrapperRow>
+      ) : (
+        <Button title={buttonTitle} invert onPress={selectImage} />
+      )}
+    </>
+  );
+};
+
 const styles = StyleSheet.create({
   image: {
     height: imageHeight(imageWidth() * 0.6),
@@ -383,6 +379,15 @@ NewProposal.propTypes = {
   data: PropTypes.object,
   navigation: PropTypes.object.isRequired,
   query: PropTypes.string
+};
+
+ImageSelector.propTypes = {
+  item: PropTypes.object,
+  control: PropTypes.object,
+  value: PropTypes.string,
+  onChange: PropTypes.func,
+  selectImage: PropTypes.func,
+  imageUri: PropTypes.string
 };
 
 const INPUTS = [
@@ -447,12 +452,12 @@ const INPUTS = [
     title: texts.consul.startNew.proposalVideoUrlInfo
   },
   {
-    buttonTitle: texts.consul.startNew.newProposalImageAddButtonTitle,
-    infoText: texts.consul.startNew.newProposalImageAddInfoText,
-    label: texts.consul.startNew.newProposalImageAddTitle,
+    type: ITEM_TYPES.PICKER,
     name: 'image',
+    label: texts.consul.startNew.newProposalImageAddTitle,
     rules: { required: false },
-    type: ITEM_TYPES.PICKER
+    buttonTitle: texts.consul.startNew.newProposalImageAddButtonTitle,
+    infoText: texts.consul.startNew.newProposalImageAddInfoText
   },
   {
     type: ITEM_TYPES.TITLE,
--- a/src/components/screens/consul/Proposals/NewProposal.js
+++ b/src/components/screens/consul/Proposals/NewProposal.js
@@ -1,3 +1,4 @@
+import * as FileSystem from 'expo-file-system';
 import PropTypes from 'prop-types';
 import React, { useEffect, useState } from 'react';
 import { useMutation } from 'react-apollo';
@@ -17,9 +18,7 @@ import { Checkbox } from '../../../Checkbox';
 import { Input } from '../../../form';
 import { Image } from '../../../Image';
 import { Label } from '../../../Label';
-import { LoadingSpinner } from '../../../LoadingSpinner';
 import { RegularText } from '../../../Text';
-import { Touchable } from '../../../Touchable';
 import { Wrapper, WrapperHorizontal, WrapperRow } from '../../../Wrapper';
 
 const TAG_CATEGORIES = [
@@ -60,11 +59,11 @@ export const NewProposal = ({ navigation, data, query }) => {
   );
   const [isLoading, setIsLoading] = useState(false);
   const [tags, setTags] = useState([]);
-  const { imageUri, selectImage } = useSelectImage(
+  const { selectImage } = useSelectImage(
     undefined, // onChange
     false, // allowsEditing,
     undefined, // aspect,
-    0.2 // quality
+    undefined // quality
   );
 
   const {
@@ -119,24 +118,22 @@ export const NewProposal = ({ navigation, data, query }) => {
   }, [tags]);
 
   const uploadImage = async (imageData) => {
-    let imageAttributes = null;
+    let imageAttributes;
     const imageUploadData = await uploadAttachment(imageData, 'image');
 
     if (imageUploadData.status === 200) {
-      const { filename, cached_attachment } = JSON.parse(imageUploadData.body);
+      const { cached_attachment: cachedAttachment, filename: title } = JSON.parse(
+        imageUploadData.body
+      );
 
       imageAttributes = {
-        title: filename,
-        cachedAttachment: cached_attachment
+        title,
+        cachedAttachment
       };
     } else if (imageUploadData.status === 422) {
       const errors = JSON.parse(imageUploadData.body).errors.toLowerCase().split(' ').join('-');
 
-      showImageUploadError(
-        texts.consul.startNew[errors] ?? texts.consul.startNew.generalPhotoUploadError
-      );
-
-      return;
+      throw errors;
     }
 
     return imageAttributes;
@@ -168,23 +165,26 @@ export const NewProposal = ({ navigation, data, query }) => {
 
     if (!hasAcceptedTermsOfService) return showPrivacyCheckRequireAlert();
 
-    if (newProposalData.image) {
-      const imageAttributes = await uploadImage(newProposalData.image);
+    setIsLoading(true);
 
-      if (!imageAttributes) {
-        return;
-      } else {
+    if (newProposalData.image) {
+      try {
+        const imageAttributes = await uploadImage(newProposalData.image);
+
         variables.attributes = { ...variables.attributes, imageAttributes };
+      } catch (error) {
+        setIsLoading(false);
+
+        return showImageUploadError(
+          texts.consul.startNew[error] ?? texts.consul.startNew.generalPhotoUploadError
+        );
       }
     }
 
     switch (query) {
       case QUERY_TYPES.CONSUL.START_PROPOSAL:
-        setIsLoading(true);
-
         try {
           await submitProposal({ variables });
-          setIsLoading(false);
 
           navigation.navigate(ScreenName.ConsulIndexScreen, {
             title: texts.consul.homeScreen.proposals,
@@ -203,11 +203,8 @@ export const NewProposal = ({ navigation, data, query }) => {
         }
         break;
       case QUERY_TYPES.CONSUL.UPDATE_PROPOSAL:
-        setIsLoading(true);
         try {
           let data = await updateProposal({ variables });
-
-          setIsLoading(false);
 
           navigation.navigate(ScreenName.ConsulDetailScreen, {
             query: QUERY_TYPES.CONSUL.PROPOSAL,
@@ -223,8 +220,6 @@ export const NewProposal = ({ navigation, data, query }) => {
         break;
     }
   };
-
-  if (isLoading) return <LoadingSpinner loading />;
 
   return (
     <>
@@ -289,9 +284,7 @@ export const NewProposal = ({ navigation, data, query }) => {
             <Controller
               name={item.name}
               control={control}
-              render={({ onChange, value }) => (
-                <ImageSelector {...{ control, imageUri, item, onChange, selectImage, value }} />
-              )}
+              render={(field) => <ImageSelector {...{ control, field, item, selectImage }} />}
             />
           )}
         </Wrapper>
@@ -312,9 +305,12 @@ export const NewProposal = ({ navigation, data, query }) => {
 
         <Wrapper>
           <Button
+            disabled={isLoading}
             onPress={handleSubmit(onSubmit)}
             title={
-              query === QUERY_TYPES.CONSUL.START_PROPOSAL
+              isLoading
+                ? texts.consul.startNew.updateButtonDisabledLabel
+                : query === QUERY_TYPES.CONSUL.START_PROPOSAL
                 ? texts.consul.startNew.newProposalStartButtonLabel
                 : texts.consul.startNew.updateButtonLabel
             }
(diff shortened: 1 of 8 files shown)


src/components/screens/consul/Proposals/NewProposal.js
```diff
@@ -39,33 +44,43 @@ const ITEM_TYPES = {
...
-// 	  }
-//  ],
-
-const showRegistrationFailAlert = () =>
+const imageProps = {
+  onChange: null,
+  allowsEditing: false,
+  aspect: null,
+  quality: 0.2
+};
```
donni106, line 56: why have you decided to create a `const` for `imageProps`. they are only used once and could be passed to `useSelectImage` directly, or am i wrong?
alphabetical order would be nice.
why have you chosen `null` over `undefined`, is there any reason?
why have you chosen `quality: 0.2`? wouldn't `undefined` work here as well to take images as they are?

src/queries/consul/uploads/upload.js
```diff
@@ -0,0 +1,49 @@
+import * as FileSystem from 'expo-file-system';
+
+import { namespace, secrets } from '../../../config';
+import { getConsulAuthToken } from '../../../helpers';
+
+// https://docs.expo.io/versions/latest/sdk/filesystem/#filesystemuploadasyncurl-fileuri-options
+export const uploadAttachment = async (uri, resourceRelation, resourceType) => {
```
donni106, line 7: we could do `(uri, resourceRelation, resourceType = 'Proposal')` and avoid passing it with calling `await uploadAttachment(newProposalData.image, 'image');`, because it will be also `'Proposal'` for documents. if we somewhen need something different, we could pass a third param.

src/queries/consul/uploads/upload.js
```diff
@@ -0,0 +1,49 @@
...
+  const token = await getConsulAuthToken();
+
+  const fetchObj = {
+    httpMethod: 'POST',
```
donni106, line 11: shouldn't this be `method`? how come `httpMethod`? i think this happened as we called together. in other files we always have `method` so we should here and for volunteer, right?

src/components/screens/consul/Proposals/NewProposal.js
```diff
@@ -129,10 +146,42 @@ export const NewProposal = ({ navigation, data, query }) => {
...
+        // 	  }
+        //  ],
       }
     };
 
-    if (!hasAcceptedTermsOfService) return showRegistrationFailAlert();
+    if (!hasAcceptedTermsOfService) return showPrivacyCheckRequireAlert();
+
+    if (newProposalData.image) {
+      const imageUploadData = await uploadAttachment(newProposalData.image, 'image', 'Proposal');
```
donni106, line 164: can `item.name` (with `name: 'image'`) from the fields config be used here instead of the hardcoded `'image'`?

src/components/screens/consul/Proposals/NewProposal.js
```diff
@@ -129,10 +146,42 @@ export const NewProposal = ({ navigation, data, query }) => {
...
+        imageData = JSON.parse(imageUploadData.body);
+
+        let imageAttributes = {
+          title: imageData?.filename,
+          cachedAttachment: imageData?.cached_attachment
+        };
+
+        variables.attributes = { ...variables.attributes, imageAttributes };
+      } else if (imageUploadData.status === 422) {
+        if (JSON.parse(imageUploadData.body).errors.length < 100) {
```
donni106, line 176: what does the `100` stand for? how does the errors look like in that case?

src/components/screens/consul/Proposals/NewProposal.js
```diff
@@ -239,6 +289,47 @@ export const NewProposal = ({ navigation, data, query }) => {
               </ScrollView>
             </>
           )}
+
+          {item.type === ITEM_TYPES.PICKER && (
+            <Controller
+              key={item.name}
```
donni106, line 295: do we need a `key` here as we have no array that we loop for rendering several components?

src/components/screens/consul/Proposals/NewProposal.js
```diff
@@ -239,6 +289,47 @@ export const NewProposal = ({ navigation, data, query }) => {
...
+              name={item.name}
+              control={control}
+              render={({ onChange, value }) => {
+                useEffect(() => {
+                  if (!value) {
+                    onChange(imageUri);
+                  }
+                }, [imageUri]);
```
donni106, line 303: this needs to be placed in the main component and not inside the render prop here. this seems bad practice to me.

src/components/screens/consul/Proposals/NewProposal.js
donni106: what is used now at the end, when you set it to `undefined`?
and still a question, why decided to create a `const` for `imageProps`. they are only used once and could be passed to `useSelectImage` directly, or am i wrong?

src/components/screens/consul/Proposals/NewProposal.js
```diff
@@ -351,6 +458,14 @@ const INPUTS = [
     type: ITEM_TYPES.INFO_TEXT,
     title: texts.consul.startNew.proposalVideoUrlInfo
   },
+  {
+    buttonTitle: texts.consul.startNew.newProposalImageAddButtonTitle,
+    infoText: texts.consul.startNew.newProposalImageAddInfoText,
+    label: texts.consul.startNew.newProposalImageAddTitle,
+    name: 'image',
+    rules: { required: false },
+    type: ITEM_TYPES.PICKER
```
donni106, line 467: `type` is always the first key in this structure but now its alphabetical - just for the new entry. i think it helped to have `type` first to quickly differentiate between the entries. what do you think could be the best structure to keep the json easily readable?

src/components/screens/consul/Proposals/NewProposal.js
donni106: alright, let's keep it simple here.

src/components/screens/consul/Proposals/NewProposal.js
donni106: i see. this seems not very long lasting for me.
if there are errors, we should use them for translation.
one possibility would be to lowercase and hyphenate the string:

```
const errors = "Choose image content type image/png does not match any of accepted content types jpg";

errors.toLowerCase().split(" ").join("-")
=> "choose-image-content-type-image/png-does-not-match-any-of-accepted-content-types-jpg"
```

with such string we could list it in `texts` and use it as follows:

```
'choose-image-content-type-image/png-does-not-match-any-of-accepted-content-types-jpg': 'Bildtyp image/png entspricht nicht dem akzeptierten Typen jpg'
```

src/components/screens/consul/Proposals/NewProposal.js
```diff
@@ -361,6 +334,36 @@ export const NewProposal = ({ navigation, data, query }) => {
...
 };
 
+const ImageSelector = ({ control, imageUri, item, onChange, selectImage, value }) => {
+  const { buttonTitle, infoText, label, name } = item;
+
+  useEffect(() => {
+    if (!value) {
+      onChange(imageUri);
+    }
+  }, [imageUri]);
```
donni106, line 344: i do not understand this part when reading.
why is the dependency `imageUri` but the logic is also depending on the value of `value`. why is there a check for `!value`? what if `value` exists? what is the effect used for?

src/components/screens/consul/Proposals/NewProposal.js
ardasnturk: Yes, this is a logical method. 👍

src/components/screens/consul/Proposals/NewProposal.js
ardasnturk: You say right. In terms of readability, it would be great if they were all in the same structure. I will fix this.

src/components/screens/consul/Proposals/NewProposal.js
ardasnturk: According to `imageUri`, I need to run the `onChange` function again, because after I click the button, I do not receive `imageUri` data. After selecting the photo, I can access this data.

The reason I check `!value` is that if I want to edit a proposal with a photo, it automatically sends the value `undefined` to the `onChange` function and the user cannot see the preview of the picture. This check is just a check for editing the proposal.

src/components/screens/consul/Proposals/NewProposal.js
ardasnturk: You are right. The reason I created `const` is to see their names when sending props to `useSelectImage`. Otherwise I will just have to send data and we will not be able to see which data is 'quality' and which data is 'allowsEditing'. I could not create props as object in `useSelectImage`. It already had the `onChange` prop in it. I just added to it.

src/components/screens/consul/Proposals/NewProposal.js
donni106: alright thanks, can you please write something down as comment in the code for our future selfs

src/components/screens/consul/Proposals/NewProposal.js
donni106: > to see their names when sending props to `useSelectImage`

i understand. on the other hand, `useSelectImage` is typed and the props are clear. one could easily check whats param 1, 2, 3 or 4. what about:

```
const { imageUri, selectImage } = useSelectImage(
  undefined, // onChange,
  false, // allowsEditing,
  undefined, // aspect,
  undefined // quality
);
```

so the additional variable could be avoided.

have you checked, what is used now at the end, when you set the values to `undefined`?

ardasnturk: Since there are default values in `imagePicker.ts`, if I enter the `quality` value as undefined, it automatically becomes `quality: 1`. So I can't give undefined value for `quality`. According to my tests, when the highest `quality` value is 0.2, photo uploading takes place. For other props, `null` or `undefined` value works the same.

donni106: correct, so we must passt a value. and since we do not know better, let's take the 0.2 according to your tests.
what about `aspect`? what can be passed to overwrite the default value with "nothing"?

ardasnturk: The aspect value can remain `undefined`. To use aspect, the value of `allowsEditing` must be `true`. But we are currently sending it as `false`.

src/components/screens/consul/Proposals/NewProposal.js
ardasnturk: 4b6da24309c37c24328a4e90c8d96a9212c89326

src/components/screens/consul/Proposals/NewProposal.js
ardasnturk: 8130952b6dccca99da959e64754c14698f3b10bf

src/components/screens/consul/Proposals/NewProposal.js
ardasnturk: 8130952b6dccca99da959e64754c14698f3b10bf

src/components/screens/consul/Proposals/NewProposal.js
```diff
@@ -129,10 +154,29 @@ export const NewProposal = ({ navigation, data, query }) => {
...
-    if (!hasAcceptedTermsOfService) return showRegistrationFailAlert();
+    if (!hasAcceptedTermsOfService) return showPrivacyCheckRequireAlert();
+
+    if (newProposalData.image) {
+      const imageAttributes = await uploadImage(newProposalData.image);
+
+      if (!imageAttributes) {
+        return;
+      } else {
+        variables.attributes = { ...variables.attributes, imageAttributes };
```
donni106, line 177: wouldn't this work in any case? why does this need to be an if-else with a return case?

src/components/screens/consul/Proposals/NewProposal.js
```diff
@@ -270,7 +325,43 @@ export const NewProposal = ({ navigation, data, query }) => {
...
+  return (
+    <>
+      <Input hidden label={label} control={control} name={name} value={value} />
+      <RegularText smallest placeholder>
+        {infoText}
+      </RegularText>
+
+      {value ? (
+        <WrapperRow center spaceBetween>
+          <Image source={{ uri: value }} style={styles.image} />
```
donni106, line 348: why is `value` used here and not `imageUri`?

src/components/screens/consul/Proposals/NewProposal.js
```diff
@@ -270,7 +325,43 @@ export const NewProposal = ({ navigation, data, query }) => {
...
+          <Touchable onPress={() => onChange(null)}>
```
donni106, line 349: would a `reset({ image: '' })` do the same here, which would be more readable than some `onChange(null)` without context.

src/components/screens/consul/Proposals/NewProposal.js
```diff
@@ -117,6 +118,30 @@ export const NewProposal = ({ navigation, data, query }) => {
...
+    let imageAttributes = null;
+    const imageUploadData = await uploadAttachment(imageData, 'image');
+
+    if (imageUploadData.status === 200) {
+      const { filename, cached_attachment } = JSON.parse(imageUploadData.body);
+
+      imageAttributes = {
+        title: filename,
+        cachedAttachment: cached_attachment
+      };
```
donni106, line 131: what do you say about:

```
const { filename: title, cached_attachment: cachedAttachment } = JSON.parse(imageUploadData.body);

imageAttributes = {
  title,
  cachedAttachment
};
```

src/components/screens/consul/Proposals/NewProposal.js
ardasnturk: If the image upload fails, the proposal opens without the image. I had to check for this. If the image upload is unsuccessful, it should notify the user and then change this image.

src/components/screens/consul/Proposals/NewProposal.js
```diff
@@ -239,6 +284,16 @@ export const NewProposal = ({ navigation, data, query }) => {
               </ScrollView>
             </>
           )}
+
+          {item.type === ITEM_TYPES.PICKER && (
+            <Controller
+              name={item.name}
+              control={control}
+              render={({ onChange, value }) => (
+                <ImageSelector {...{ control, imageUri, item, onChange, selectImage, value }} />
```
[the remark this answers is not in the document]
ardasnturk: If I use it like this, `value` will not get any empty value after the image is selected once. The user will not be able to change the image as it will be filled with `imageUri`. I need to think of something else for this.

I need `value`. I need to fetch as `image: "xxx"` in `onSubmit` function. I can only achieve this with `onChange`. Am I wrong?

src/components/screens/consul/Proposals/NewProposal.js
ardasnturk: `imageUri` does not return an empty value after selecting an image. It just changes itself. If I use `imageUri` here, the user will not be able to delete the image and select another image after selecting it. Or I'll have to delete the value of `onChange("")` plus the `imageUri`.

src/components/screens/consul/Proposals/NewProposal.js
```diff
@@ -129,10 +154,29 @@ export const NewProposal = ({ navigation, data, query }) => {
...
     };
 
-    if (!hasAcceptedTermsOfService) return showRegistrationFailAlert();
+    if (!hasAcceptedTermsOfService) return showPrivacyCheckRequireAlert();
+
+    if (newProposalData.image) {
+      const imageAttributes = await uploadImage(newProposalData.image);
+
+      if (!imageAttributes) {
+        return;
```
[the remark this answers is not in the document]
ardasnturk: 0bc389056155a1878daf189d621c7353bcc70657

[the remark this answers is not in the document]
ardasnturk: 0bc389056155a1878daf189d621c7353bcc70657

src/components/screens/consul/Proposals/NewProposal.js
ardasnturk: 0bc389056155a1878daf189d621c7353bcc70657

src/components/screens/consul/Proposals/NewProposal.js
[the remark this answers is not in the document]
ardasnturk: 0bc389056155a1878daf189d621c7353bcc70657

src/components/screens/consul/Proposals/NewProposal.js
ardasnturk: 0bc389056155a1878daf189d621c7353bcc70657

src/components/screens/consul/Proposals/NewProposal.js
ardasnturk: 0bc389056155a1878daf189d621c7353bcc70657

src/components/screens/consul/Proposals/NewProposal.js
```diff
@@ -117,6 +118,30 @@ export const NewProposal = ({ navigation, data, query }) => {
...
+        cachedAttachment: cached_attachment
+      };
+    } else if (imageUploadData.status === 422) {
+      const errors = JSON.parse(imageUploadData.body).errors.toLowerCase().split(' ').join('-');
+
+      showImageUploadError(
+        texts.consul.startNew[errors] ?? texts.consul.startNew.generalPhotoUploadError
+      );
+
+      return;
```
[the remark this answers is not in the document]
ardasnturk: I tried a different method. I hope this is better.

0bc389056155a1878daf189d621c7353bcc70657

src/components/screens/consul/Proposals/NewProposal.js
```diff
@@ -117,6 +117,28 @@ export const NewProposal = ({ navigation, data, query }) => {
...
+        cachedAttachment
+      };
+    } else if (imageUploadData.status === 422) {
+      const errors = JSON.parse(imageUploadData.body).errors.toLowerCase().split(' ').join('-');
+
+      throw errors;
+    }
+
+    return imageAttributes;
+  };
```
donni106, line 140: not bad, thanks. i like the approach.
do we still need `let imageAttributes;` now?

if status is 200 we could

```
return {
  title,
  cachedAttachment
}
```

and then have another if status is 422 to throw an error.

this should be the same but less and more straight forward code i guess? what do you think?

src/components/screens/consul/Proposals/NewProposal.js
```diff
@@ -325,33 +321,78 @@ export const NewProposal = ({ navigation, data, query }) => {
...
   return (
     <>
-      <Input hidden label={label} control={control} name={name} value={value} />
+      <Input
+        {...item}
+        control={control}
+        errorMessage={errorMessage}
+        hidden
+        name={name}
+        validate={true}
```
donni106, line 338: we like it short, so `validate` for passing `true` is enoufg, like `hidden` :)

src/components/screens/consul/Proposals/NewProposal.js
```diff
@@ -325,33 +321,78 @@ export const NewProposal = ({ navigation, data, query }) => {
...
+            // `1048576` = the byte equivalent of 1MB
+            if (size > 1048576) {
+              setErrorMessage(
+                texts.consul.startNew['choose-image-must-be-in-between-0-bytes-and-1-mb']
+              );
+            }
+
+            // if the photo that wants to be uploaded is not a jpg,
+            // we show this error message on the screen
+            if (!uri.includes('.jpg')) {
```
donni106, line 375: i am curios if `jpeg` works for consul?

src/components/screens/consul/Proposals/NewProposal.js
```diff
@@ -325,33 +321,78 @@ export const NewProposal = ({ navigation, data, query }) => {
...
+              setErrorMessage(
+                texts.consul.startNew[
+                  'choose-image-content-type-image/png-does-not-match-any-of-accepted-content-types-jpg'
+                ]
+              );
+            }
+
+            // if both requests are not met, a message with all
+            // errors is displayed on the screen
+            if (!uri.includes('.jpg') && size > 1048576) {
```
donni106, line 385: the checks for `.jpg` and `1048576` are duplicated here. this could be smarter. any ideas?

src/hooks/imagePicker.ts
```diff
@@ -38,6 +38,7 @@ export const useSelectImage = (
 
     if (!result.cancelled) {
       onChange ? onChange(setImageUri)(result.uri) : setImageUri(result.uri);
+      return result;
```
donni106, line 41: what is `result` in this case? does it still satisfy typescript and match the other return somehow?

src/components/screens/consul/Proposals/NewProposal.js
```diff
@@ -325,33 +321,78 @@ export const NewProposal = ({ navigation, data, query }) => {
   );
 };
 
-const ImageSelector = ({ control, imageUri, item, onChange, selectImage, value }) => {
-  const { buttonTitle, infoText, label, name } = item;
+const ImageSelector = ({ control, field, item, selectImage }) => {
+  const [errorMessage, setErrorMessage] = useState();
```
donni106, line 325: we see `errorMessage` always as string, right? event if this is no typescript file, i guess having a string as initial value is more clear `useState('');`

src/components/screens/consul/Proposals/NewProposal.js
donni106: ok, let's keep it

```
onChange('');
setErrorMessage('');
```

this seams clear to me

src/components/screens/consul/Proposals/NewProposal.js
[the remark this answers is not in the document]
donni106: follow up in: https://github.com/ikuseiGmbH/smart-village-app-app/pull/529#discussion_r874642441

src/components/screens/consul/Proposals/NewProposal.js
ardasnturk: I just tested for it. Unfortunately it does.

src/components/screens/consul/Proposals/NewProposal.js
ardasnturk: 18e756843bc8ae851b92b31cf86be6945d2450ab

src/components/screens/consul/Proposals/NewProposal.js
ardasnturk: 18e756843bc8ae851b92b31cf86be6945d2450ab

src/components/screens/consul/Proposals/NewProposal.js
ardasnturk: 18e756843bc8ae851b92b31cf86be6945d2450ab

src/components/screens/consul/Proposals/NewProposal.js
```diff
@@ -270,7 +318,92 @@ export const NewProposal = ({ navigation, data, query }) => {
...
+          onPress={async () => {
+            const { uri, type } = await selectImage();
+            const { size } = await FileSystem.getInfoAsync(uri);
+            const errorMessage = imageErrorMessageGenerator({
+              size,
+              uri: uri.substr(uri.length - 4)
+            });
+
+            setErrorMessage(texts.consul.startNew[errorMessage]);
+            setImageInfoText(`(${type}/${uri.substr(uri.length - 3)}, ${bytesToSize(size)})`);
```
donni106, line 372: what if the filetype ending is not only 3 chars? like jpeg? this is very unflexible.
doesn't `FileSystem.getInfoAsync(uri)` return some filetype stuff? i would think so and we would not need to build the string on our own.

src/components/screens/consul/Proposals/NewProposal.js
```diff
@@ -270,7 +318,92 @@ export const NewProposal = ({ navigation, data, query }) => {
...
+
+const imageErrorMessageGenerator = ({ size, uri }) => {
+  const errorMessage =
+    uri !== '.jpg' && uri !== 'jpeg' && size > 1048576
+      ? 'choose-image-content-type-image/png-does-not-match-any-of-accepted-content-types-jpg,-choose-image-must-be-in-between-0-bytes-and-1-mb'
+      : uri !== '.jpg' && uri !== 'jpeg'
+      ? 'choose-image-content-type-image/png-does-not-match-any-of-accepted-content-types-jpg'
+      : size > 1048576
+      ? 'choose-image-must-be-in-between-0-bytes-and-1-mb'
+      : '';
```
donni106, line 397: why `.jpg` but `jpeg` without `.`? ;)

please use consts for `.jpg`, `.jpeg` and `1048576`  inside of this function to not duplicate these values.

example:

```
const isJPG = uri === '.jpg' || uri === '.jpeg';
const isGreater1MB = size > 1048576;
```

then you can use `errorMessage = !isJPG && !isGreater1MB ? ...` and so on, which is more readable.

the `uri` is always just the file ending/file type and not the whole path?

src/components/screens/consul/Proposals/NewProposal.js
```diff
@@ -334,62 +332,44 @@ const ImageSelector = ({ control, field, item, selectImage }) => {
...
-          </TouchableOpacity>
-        </WrapperRow>
+        <>
+          <WrapperRow center spaceBetween>
+            <Image source={{ uri: value }} style={styles.image} />
+
+            <TouchableOpacity
+              onPress={() => {
+                onChange('');
+                setErrorMessage('');
```
donni106, line 351: so we also need to reset the `setImageInfoText` to `''` here?

src/components/screens/consul/Proposals/NewProposal.js
```diff
@@ -334,62 +332,44 @@ const ImageSelector = ({ control, field, item, selectImage }) => {
...
+              }}
+            >
+              <Icon.Trash color={colors.error} size={normalize(16)} />
+            </TouchableOpacity>
+          </WrapperRow>
+          <RegularText smallest>{imageInfoText}</RegularText>
```
donni106, line 357: shouldn't this be rendered only if set? so wrapping it with `{!!imageInfoText && ...}`?

src/components/screens/consul/Proposals/NewProposal.js
ardasnturk: sorry, i can't find the type of the image
type comes only as `image` or `video`.

`selectImage()` :
```
Object {
  "cancelled": false,
  "height": 800,
  "type": "image",
  "uri": "file:///Use......0EE9A6.png",
  "width": 800,
}
```

`FileSystem.getInfoAsync(uri)` :
```
Object {
  "exists": true,
  "isDirectory": false,
  "modificationTime": 1652857258.6079366,
  "size": 239169,
  "uri": "file:///Use......2C660.jpg",
}
```

src/components/screens/consul/Proposals/NewProposal.js
ardasnturk: Yes, I didn't think about it that way. Thank you. 

While sending the Uri data to the function, it is shown as type in the function because I clip it.

src/components/screens/consul/Proposals/NewProposal.js
ardasnturk: After the picture was deleted, I did not delete it because it was not displayed on the screen. But it can be deleted.

src/components/screens/consul/Proposals/NewProposal.js
ardasnturk: 4f4769453427750e74db1b209268861c890b1939

src/components/screens/consul/Proposals/NewProposal.js
ardasnturk: 4f4769453427750e74db1b209268861c890b1939

src/components/screens/consul/Proposals/NewProposal.js
ardasnturk: 4f4769453427750e74db1b209268861c890b1939

src/components/screens/consul/Proposals/NewProposal.js
donni106: ok, so when its clipped, it is not a `uri` anymore. this needs to be named as something like `imageType` please

src/components/screens/consul/Proposals/NewProposal.js
```diff
@@ -270,7 +318,96 @@ export const NewProposal = ({ navigation, data, query }) => {
...
+      ) : (
+        <Button
+          title={buttonTitle}
+          invert
+          onPress={async () => {
+            const { uri, type } = await selectImage();
+            const { size } = await FileSystem.getInfoAsync(uri);
+            const errorMessage = imageErrorMessageGenerator({
+              size,
+              imageType: uri.substr(uri.length - 4)
```
donni106, line 369: on a different place it is `- 3` and here `- 4`? as said there, we need a more safe regular expression to determine the image type. and both places must use the same.

src/components/screens/consul/Proposals/NewProposal.js
```diff
@@ -270,7 +318,96 @@ export const NewProposal = ({ navigation, data, query }) => {
...
+const bytesToSize = (bytes) => {
+  const sizes = ['Bytes', 'KB', 'MB', 'GB', 'TB'];
+  if (bytes == 0) return '0 Byte';
+  let i = parseInt(Math.floor(Math.log(bytes) / Math.log(1024)));
+
+  return Math.round(bytes / Math.pow(1024, i), 2) + ' ' + sizes[i];
+};
+
+const imageErrorMessageGenerator = ({ size, imageType }) => {
+  const isJPG = imageType === '.jpg' ?? 'jpeg';
```
donni106, line 391: `.` is still missing in `jpeg`, right?

src/components/screens/consul/Proposals/NewProposal.js
ardasnturk: I just take the last 3 letters of the uri and determine the type of the picture. In another way, the picture type does not reach me.

src/components/screens/consul/Proposals/NewProposal.js
ardasnturk: If I use -3, I will get the `peg` string for `jpeg` images.

src/components/screens/consul/Proposals/NewProposal.js
ardasnturk: `.` is not missing. I have the last 4 letters of the uri data.

jpg = `.jpg`
jpeg = `jpeg`

src/hooks/imagePicker.ts
ardasnturk: I am using the `uri` and `type` data from the `result` object. There is `imageUri` returned in `useSelectImage`. But this information is returned to me when I use `useSelectImage`. I need the return in `selectImage`.

src/components/screens/consul/Proposals/NewProposal.js
donni106: i understand. it is related to the `- 4`. this is too specific. let's think of some RegEx that will help us in all situations.

ardasnturk: 96320ff2f0d88d127a9a420ef721f9dc1b50613c

src/components/screens/consul/Proposals/NewProposal.js
```diff
@@ -47,6 +47,8 @@ const ITEM_TYPES = {
   PICKER: 'picker'
 };
 
+const IMAGE_TYPE_REGEX = /\.(gif|jpe?g|tiff?|png|webp|bmp)$/i;
```
donni106, line 50: seems to work fine

Bildschirmfoto 2022-05-18 um 13 52 53

src/hooks/imagePicker.ts
donni106: ok we have called 1 on 1 and discussed the situation. all fine now.

donni106 (approved):
let's do it :)

PR #535, opened 2022-05-18
feat(register): in-app implementation

donni106:
- unified login and registration components, texts
  and email regex between volunteer and consul
  - removed `refreshControl` because it setup in consul
    but without any functionality
  - removed state for `isLoading` because `isLoading` can be
    used from `useMutation` for consul
  - refactored duplicated `rightIcon` function to a
    `InputSecureTextIcon` component for reusability
- changed registration endpoint to the new v2 url
  to use the new functionality
- added firstname and lastname to the registration
  because the new endpoints requires them
- moved login types to user and added type for registration
  to have them at one place

HDVT-39

⚠️  there is a bug which causes data not to be returned because of missing approval of privacy policies.
i already opened an issue for this: https://github.com/ikuseiGmbH/smart-village-app-humhub-ext/issues/22


src/screens/volunteer/VolunteerRegistrationScreen.tsx
```diff
@@ -79,13 +110,12 @@ export const VolunteerRegistrationScreen = ({ navigation }: StackScreenProps<any
                 textContentType="username"
                 autoCapitalize="none"
                 validate
-                rules={{ required: true }}
-                errorMessage={
-                  errors.username && `${texts.volunteer.username} muss ausgefüllt werden`
-                }
+                rules={{ required: texts.volunteer.usernameError }}
```
ardasnturk, line 113: I think the {minLenth} rule needs to be set here. There is an example of this on line 182.

ardasnturk (commented):
- after logging out of my account, I am not allowed to log in again. I have to close and open the app. 
- also, I can't create a new account, it automatically sends me to `VolunteerRegisteredScreen`.
- after registering to the system, the confirmation e-mail did not come. Is this normal?
- after registering, I can log in directly without checking whether the user is real or not. Is it supposed to be like this?
- shouldn't there be a confirmation like `Terms of service` when registering?

src/screens/volunteer/VolunteerRegistrationScreen.tsx
```diff
@@ -17,43 +18,72 @@ import {
...
+    isLoading: isLoadingMe,
+    isError: isErrorMe,
+    isSuccess: isSuccessMe,
+    data: dataMe
+  } = useQuery(QUERY_TYPES.VOLUNTEER.ME, me, {
+    enabled: !!data?.auth_token // the query will not execute until the auth token exists
+  });
+
+  const onSubmit = (registerData: VolunteerRegistration) => mutateRegister(registerData);
+  console.warn(data);
```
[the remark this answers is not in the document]
donni106: good catch!

src/screens/volunteer/VolunteerRegistrationScreen.tsx
donni106: true!

> Username should contain at least 4 characters.

src/config/consts.js
```diff
@@ -53,6 +53,9 @@ export const consts = {
     website: '(Webseite)',
     webViewHint: '(Öffnet Webseite in der aktuellen App)'
   },
+
+  EMAIL_REGEX: /^[a-zA-Z0-9.!#$%&’*+/=?^_`{|}~-]+@[a-zA-Z0-9-]+(?:\.[a-zA-Z0-9-]+)*$/,
```
donni106, line 57: this seems not to be enough:

Bildschirmfoto 2022-05-20 um 19 08 27

`.de` is missing but it gets valid

ardasnturk: then the same problem exists in Consul

donni106: true. with this PR, both will use the same regex. i will investigate further on monday.

donni106: 98e8b40bbee458039eb8e6a3cb373a7fc08f28f0

donni106:
> * after logging out of my account, I am not allowed to log in again. I have to close and open the app.
> * also, I can't create a new account, it automatically sends me to `VolunteerRegisteredScreen`.

this should behave totally better now. please do some tests again.

> * after registering to the system, the confirmation e-mail did not come. Is this normal?
> * after registering, I can log in directly without checking whether the user is real or not. Is it supposed to be like this?

yes, there is a direct login and no e-mail to keep things simple for users.

> * shouldn't there be a confirmation like `Terms of service` when registering?

added it, thanks. there was the server part missing until now.

ardasnturk (approved):
looks nice 🎉

PR #538, opened 2022-05-27
feat(consul): filter draft proposal

ardasnturk:
- added a new `published` field to the `proposals` and
  `user` query to see if the `Proposal` is drafted
- added the published field to listItemParser to
  be able to use it in list item
- added if condition to `listItemParser` to add
  `draft` text to `title` according to `Public Proposals`
  query and according to `published` data
- add a new if condition to `listHooks` to not show it on
  proposal data listing page in draft state
- edited subtitle in `listItemParser` to improve
  readability
- added draft text to `texts.js`

SVA-617

## How to test:
* [ ] Android portrait mode
* [x] iOS portrait mode
* [ ] Android landscape mode
* [x] iOS landscape mode

## Screenshots:

|Before List|After List|My Content|
|--|--|--|
Simulator Screen Shot - iPhone 13 - 2022-05-27 at 10 39 53 | Simulator Screen Shot - iPhone 13 - 2022-05-27 at 10 39 23 |  Simulator Screen Shot - iPhone 13 - 2022-05-27 at 10 39 18 | 
|Proposals.js|Proposals.js||

### DIFF
--- a/src/helpers/parser/listItemParser.js
+++ b/src/helpers/parser/listItemParser.js
@@ -270,7 +270,7 @@ const parseConsulData = (data, query, skipLastDivider) => {
     }
 
     if (query === QUERY_TYPES.CONSUL.PUBLIC_PROPOSALS && !consulData.published) {
-      title = `${texts.consul.draft} - ${consulData.title ?? consulData.body}`;
+      title = `${texts.consul.draft} - ${title}`;
     }
 
     return {
--- a/src/helpers/parser/listItemParser.js
+++ b/src/helpers/parser/listItemParser.js
@@ -257,9 +257,8 @@ const querySwitcherForDetail = (query) => {
 
 const parseConsulData = (data, query, skipLastDivider) => {
   return data?.nodes?.map((consulData, index) => {
-    let subtitle = momentFormatUtcToLocal(
-      consulData.publicCreatedAt ? consulData.publicCreatedAt : consulData.createdAt
-    );
+    let subtitle = momentFormatUtcToLocal(consulData.publicCreatedAt ?? consulData.createdAt);
+    let title = consulData.title ?? consulData.body;
 
     if (query === QUERY_TYPES.CONSUL.PUBLIC_COMMENTS) {
       subtitle = consulData.commentableTitle;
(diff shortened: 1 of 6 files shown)


src/helpers/parser/listItemParser.js
```diff
@@ -270,13 +269,18 @@ const parseConsulData = (data, query, skipLastDivider) => {
         momentFormatUtcToLocal(consulData.endsAt);
     }
 
+    if (query === QUERY_TYPES.CONSUL.PUBLIC_PROPOSALS && !consulData.published) {
+      title = `${texts.consul.draft} - ${consulData.title ?? consulData.body}`;
```
donni106, line 273: as `title` is already set to `consulData.title ?? consulData.body` you can use the `title` here: 

```js
title = `${texts.consul.draft} - ${title}`
```

src/hooks/listHooks.js
```diff
@@ -62,6 +62,8 @@ export const useRenderItem = (query, navigation, options = {}) => {
     }
     default: {
       renderItem = ({ item, index, section }) => {
+        if (query === QUERY_TYPES.CONSUL.PROPOSALS && !item.published) return;
```
donni106, line 65: good idea. but i think the list data should be cleaned up instead of rendering nothing for some entries here. there should be only list data in the array passed to the list component that we want to render on the screen. i guess the `listItems` in the `ConsulIndexScreen` need to be modified base on `item.published` and all entries could be rejected, that are draft. the list item rendering would not need to be touched. the logics are already very heavy here deciding which list item component to use. so we should not add another complexity with checking for specific data to do something event more different. what do you think?

donni106 (changes requested):

src/helpers/parser/listItemParser.js
ardasnturk: I don't understand how I missed this. 🤦‍♂️

src/hooks/listHooks.js
ardasnturk: I guess I can add a filter for `Proposals`. If the `item.pulished` value is `true`, it will be displayed in the listing process on the screen.

src/helpers/parser/listItemParser.js
donni106: :) happens

src/hooks/listHooks.js
donni106: nice, very good solution!

donni106 (approved):

PR #539, opened 2022-05-27
feature(consul): new query integration

ardasnturk:
- added new field prepared for image to
  `proposals` query
- updated the `imageUrlMedium` data on
  the `ProposalDetail` page as `image`
  because it is no longer in the query
- added `image?.imageUrlLarge` to the Image
  component as `imageUrlMedium` is no longer
  available
- updated image props in edit button to
  `image?.imageUrlLarge`
- added `imageId` props that we need to delete
  the image in the edit button
- added an if condition before the `uploadData`
  function as the server does not accept an image
  data containing an absolute url
- added a new regex to `consts.js` to check if we
  have absolute url data
- deleted the data that has the url in the document
  array because the server does not accept absolute url

## How to test:
* [x] Android portrait mode
* [x] iOS portrait mode
* [x] Android landscape mode
* [x] iOS landscape mode

SVA-611

### DIFF
--- a/src/components/screens/consul/Proposals/NewProposal.js
+++ b/src/components/screens/consul/Proposals/NewProposal.js
@@ -4,7 +4,7 @@ import { useMutation } from 'react-apollo';
 import { Controller, useForm } from 'react-hook-form';
 import { Alert, ScrollView, StyleSheet, TouchableOpacity } from 'react-native';
 
-import { colors, namespace, secrets, texts } from '../../../../config';
+import { colors, consts, namespace, secrets, texts } from '../../../../config';
 import { ConsulClient } from '../../../../ConsulClient';
 import { documentErrorMessageGenerator, imageErrorMessageGenerator } from '../../../../helpers';
 import { QUERY_TYPES } from '../../../../queries';
@@ -18,6 +18,8 @@ import { Input } from '../../../form';
 import { Label } from '../../../Label';
 import { RegularText } from '../../../Text';
 import { Wrapper, WrapperHorizontal } from '../../../Wrapper';
+
+const { URL_REGEX } = consts;
 
 const TAG_CATEGORIES = [
   { name: 'Associations', id: 0, selected: false },
@@ -149,6 +151,8 @@ export const NewProposal = ({ navigation, data, query }) => {
     setIsLoading(true);
 
     if (newProposalData.image) {
+      if (URL_REGEX.test(newProposalData.image)) return;
+
       try {
         const imageAttributes = await uploadData(newProposalData.image, 'image');
 
--- a/src/components/screens/consul/Proposals/NewProposal.js
+++ b/src/components/screens/consul/Proposals/NewProposal.js
@@ -170,10 +170,18 @@ export const NewProposal = ({ navigation, data, query }) => {
 
     if (newProposalData.documents) {
       let documents = JSON.parse(newProposalData.documents);
+      let documentsLength = documents.length;
+
       variables.attributes = {
         ...variables.attributes,
         documentsAttributes: []
       };
+
+      while (documentsLength--) {
+        if (documents[documentsLength].url) {
+          documents.splice(documentsLength, 1);
+        }
+      }
 
       for (let i = 0; i < documents.length; i++) {
         const { cachedAttachment, title } = documents[i];
--- a/src/components/screens/consul/Proposals/ProposalDetail.js
+++ b/src/components/screens/consul/Proposals/ProposalDetail.js
@@ -47,7 +47,7 @@ export const ProposalDetail = ({ data, refetch, route, navigation }) => {
     description,
     documents,
     id,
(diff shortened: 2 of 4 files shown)


src/components/screens/consul/Proposals/NewProposal.js
```diff
@@ -166,11 +170,19 @@ export const NewProposal = ({ navigation, data, query }) => {
...
       variables.attributes = {
         ...variables.attributes,
         documentsAttributes: []
       };
 
+      while (documentsLength--) {
+        if (documents[documentsLength].url) {
+          documents.splice(documentsLength, 1);
+        }
+      }
```
donni106, line 184: what is this exactly for?

src/components/screens/consul/Proposals/NewProposal.js
```diff
@@ -149,6 +151,8 @@ export const NewProposal = ({ navigation, data, query }) => {
     setIsLoading(true);
 
     if (newProposalData.image) {
+      if (URL_REGEX.test(newProposalData.image)) return;
```
donni106, line 154: in which cases are there absolute urls for images, that we need to check here?

src/components/screens/consul/Proposals/ProposalDetail.js
```diff
@@ -149,7 +149,8 @@ export const ProposalDetail = ({ data, refetch, route, navigation }) => {
                   description,
                   documents,
                   id,
-                  image: imageUrlMedium,
+                  image: image?.imageUrlLarge,
+                  imageId: image?.id,
```
donni106, line 153: what is `imageId` used for? i can only find `image: data?.image || '',` in `NewProposal`.

src/components/screens/consul/Proposals/NewProposal.js
ardasnturk: If a proposal with a PDF inside is desired to be edited, these PDFs have an absolute url. If we do not delete these urls from the array, an error message will come from the server. I thought I needed a countdown while loop to prevent this.

src/components/screens/consul/Proposals/NewProposal.js
ardasnturk: We need this control if it contains a propsal image and the user wants to edit it. The user may want to edit only the text without deleting the picture. If such a situation occurs, we should not send the absolute url value to the server.

src/components/screens/consul/Proposals/ProposalDetail.js
ardasnturk: The `imageId` information is required to delete the image. But yes you are right, this information is also in the `image` object.

src/components/screens/consul/Proposals/NewProposal.js
donni106: could the same result be achieved with a `filter` method rejecting the entries with `url`? i guess it would be more readable that way.

ardasnturk: [11cd880](https://github.com/ikuseiGmbH/smart-village-app-app/pull/539/commits/11cd880fd4a1818be68e2554dd6598e4211f65e7)

src/components/screens/consul/Proposals/ProposalDetail.js
donni106: ok right and we need it concerning deletions in the other PR, i see

src/components/screens/consul/Proposals/NewProposal.js
ardasnturk: [e74baff](https://github.com/ikuseiGmbH/smart-village-app-app/pull/539/commits/e74baffa719ec66f2d89cb7518205b6a0c11dd46)

src/components/screens/consul/Proposals/NewProposal.js
```diff
@@ -148,7 +150,10 @@ export const NewProposal = ({ navigation, data, query }) => {
 
     setIsLoading(true);
 
-    if (newProposalData.image) {
+    // if the image is an absolute url, we are editing and do not want to upload a new image
+    const isImageToUpload = !URL_REGEX.test(newProposalData.image);
+
+    if (newProposalData.image && isImageToUpload) {
```
donni106, line 156: i guess we should go even further in refactoring and do

```
const isImageToUpload = newProposalData.image && !URL_REGEX.test(newProposalData.image);

if (isImageToUpload) {
```

otherwise `isImageToUpload` could be `true` but there is no image, just because the regex test returns `false`. this would contradict the variable name.

ardasnturk: [2d81526](https://github.com/ikuseiGmbH/smart-village-app-app/pull/539/commits/2d81526eb0d4553070cace89fba45903e64c0360)

donni106 (approved):

PR #540, opened 2022-05-27
feature(consul): add delete mutation

ardasnturk:
- added new `deleteImage` and `deleteDocument` 
  mutation
- added `deleteImage` mutation to `ImageSelector`
  component to delete image when editing proposal
  containing image
- added `imageId` prop to `ImageSelector` because
  id is required when deleting the image
- added `deleteDocument` mutation to `DocumentSelector`
  component to delete the document while editing the
  proposal containing the document
- added a new alert to prevent users from accidentally
  deleting image and document
- added `onDeleteImage` and `onDeleteDocument` function 
  that is triggered according to the result of Alert for the 
  user who wants to delete a image and document
- added texts for new alert

## How to test:
* [x] Android portrait mode
* [x] iOS portrait mode
* [x] Android landscape mode
* [x] iOS landscape mode

SVA-612

### DIFF
--- a/src/components/consul/selectors/DocumentSelector.js
+++ b/src/components/consul/selectors/DocumentSelector.js
@@ -13,19 +13,18 @@ import { Input } from '../../form';
 import { RegularText } from '../../Text';
 import { WrapperRow } from '../../Wrapper';
 
-const deleteDocumentAlert = ({ onDeleteDocument, documentId, index }) =>
+const deleteDocumentAlert = (onPress) =>
   Alert.alert(
     texts.consul.startNew.deleteAttributesAlertTitle,
     texts.consul.startNew.documentDeleteAlertBody,
     [
       {
         text: texts.consul.abort,
-        onPress: () => null,
         style: 'cancel'
       },
       {
         text: texts.consul.startNew.deleteAttributesButtonText,
-        onPress: () => onDeleteDocument(documentId, index),
+        onPress,
         style: 'destructive'
       }
     ]
@@ -75,9 +74,7 @@ export const DocumentSelector = ({ control, field, item }) => {
                 <RegularText>{item.title}</RegularText>
 
                 <TouchableOpacity
-                  onPress={() =>
-                    deleteDocumentAlert({ onDeleteDocument, documentId: item.id, index })
-                  }
+                  onPress={() => deleteDocumentAlert(() => onDeleteDocument(item.id, index))}
                 >
                   <Icon.Trash color={colors.error} size={normalize(16)} />
                 </TouchableOpacity>
--- a/src/components/consul/selectors/DocumentSelector.js
+++ b/src/components/consul/selectors/DocumentSelector.js
@@ -1,7 +1,7 @@
 import PropTypes from 'prop-types';
 import React, { useEffect, useState } from 'react';
 import { useMutation } from 'react-apollo';
-import { TouchableOpacity, View } from 'react-native';
+import { Alert, TouchableOpacity, View } from 'react-native';
 
 import { colors, Icon, normalize, texts } from '../../../config';
 import { ConsulClient } from '../../../ConsulClient';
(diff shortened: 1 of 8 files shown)


donni106:
i received `GraphQL error: Field 'imageUrlMedium' doesn't exist on type 'Proposal'`. so this cannot be tested independently because a change on the server is to blame that changes from another PR is necessary first #539.

src/components/consul/selectors/DocumentSelector.js
```diff
@@ -1,28 +1,66 @@
...
 import { WrapperRow } from '../../Wrapper';
 
+const deleteDocumentAlert = ({ onDeleteDocument, documentId, index }) =>
+  Alert.alert(
+    texts.consul.startNew.deleteAttributesAlertTitle,
+    texts.consul.startNew.documentDeleteAlertBody,
+    [
+      {
+        text: texts.consul.abort,
+        onPress: () => null,
```
donni106, line 23: is passing a method returning `null` necessary here for doing "nothing" on press? what if `onPress` would not be given?

src/components/consul/selectors/DocumentSelector.js
```diff
@@ -37,10 +75,9 @@ export const DocumentSelector = ({ control, field, item }) => {
                 <RegularText>{item.title}</RegularText>
 
                 <TouchableOpacity
-                  onPress={() => {
-                    setDocumentsAttributes(deleteArrayItem(documentsAttributes, index));
-                    setInfoAndErrorText(deleteArrayItem(infoAndErrorText, index));
-                  }}
+                  onPress={() =>
+                    deleteDocumentAlert({ onDeleteDocument, documentId: item.id, index })
```
donni106, line 79: would it be possible to pass the entire method with params directly here, as `documentId` and `index` are only used to call the method in the alert. would `deleteDocumentAlert(() => onDeleteDocument(item.id, index))` work, when changing to the following in the alert definition:

```js
const deleteDocumentAlert = (onPress) {

  ...

  text: texts.consul.startNew.deleteAttributesButtonText,
  onPress,
  style: 'destructive'
```

src/components/consul/selectors/ImageSelector.js
```diff
@@ -14,19 +17,54 @@ import { WrapperRow } from '../../Wrapper';
...
+    texts.consul.startNew.imageDeleteAlertBody,
+    [
+      {
+        text: texts.consul.abort,
+        onPress: () => null,
+        style: 'cancel'
+      },
+      {
+        text: texts.consul.startNew.deleteAttributesButtonText,
+        onPress: () => onDeleteImage(),
```
donni106, line 32: if the param for `deleteImageAlert` would be `onPress` instead of `onDeleteImage` this could be refactored to the following, as we pass no further params to the callback method:

```js
text: texts.consul.startNew.deleteAttributesButtonText,
onPress,
```

src/components/screens/consul/Proposals/NewProposal.js
```diff
@@ -298,7 +298,7 @@ export const NewProposal = ({ navigation, data, query }) => {
               control={control}
               render={(field) =>
                 item.name === 'image' ? (
-                  <ImageSelector {...{ control, field, item }} />
+                  <ImageSelector {...{ control, field, item, imageId: data?.imageId }} />
```
donni106, line 301: aaahhh this is the `imageId` i was questioning in the other PR, right? https://github.com/ikuseiGmbH/smart-village-app-app/pull/539#discussion_r884953915

donni106 (changes requested):

src/components/consul/selectors/ImageSelector.js
ardasnturk: [b2165fe](https://github.com/ikuseiGmbH/smart-village-app-app/pull/540/commits/b2165fe121ef9f8a52809c0490e7fae9b0b7046a)

src/components/consul/selectors/DocumentSelector.js
ardasnturk: [b2165fe](https://github.com/ikuseiGmbH/smart-village-app-app/pull/540/commits/b2165fe121ef9f8a52809c0490e7fae9b0b7046a)

donni106 (approved):

PR #542, opened 2022-05-30
feat: change app store language

ardasnturk:
- fixed an issue where app language was showing as
  English in App Store

> ⚠️ **We can't see if the language problem of the application in the App Store is resolved or not, until we get the build and upload it to the App Store.** ⚠️

SVA-621

### DIFF
--- a/app.json
+++ b/app.json
@@ -28,7 +28,9 @@
       "supportsTablet": true,
       "infoPlist": {
         "NSCameraUsageDescription": "Diese App nutzt die Kamera um QR-Codes zu scannen.",
-        "NSPhotoLibraryUsageDescription": "Diese App nutzt die Medienbibliothek für die Auswahl eines Profilbildes."
+        "NSPhotoLibraryUsageDescription": "Diese App nutzt die Medienbibliothek für die Auswahl eines Profilbildes.",
+        "CFBundleLocalizations": ["de"],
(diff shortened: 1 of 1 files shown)


donni106 (approved):

app.json
```diff
@@ -28,7 +28,9 @@
       "supportsTablet": true,
       "infoPlist": {
         "NSCameraUsageDescription": "Diese App nutzt die Kamera um QR-Codes zu scannen.",
-        "NSPhotoLibraryUsageDescription": "Diese App nutzt die Medienbibliothek für die Auswahl eines Profilbildes."
+        "NSPhotoLibraryUsageDescription": "Diese App nutzt die Medienbibliothek für die Auswahl eines Profilbildes.",
+        "CFBundleLocalizations": ["de"],
+        "CFBundleDevelopmentRegion": "de_DE"
```
donni106, line 33: i found the following:

> CFBundleLocalizations (Info.plist array specifying all supported localizations)
> CFBundleDevelopmentRegion (Info.plist value specifying the default language for the app)

https://docs.microsoft.com/en-us/xamarin/ios/app-fundamentals/localization/

default for `CFBundleDevelopmentRegion` seems to be "en_US". i think this key would be enough for our case, but let us keep the localizations as well.

PR #543, opened 2022-05-31
feat(bus bb): remove teaser and fix order typo

donni106:
- adjusted title for "Kosten..." to match the api response
  in order to be sorted correctly
- added `TEASER` as block to be removed from rendering

SVA-576

|before|after|
|---|---|
|IMG_0068|IMG_0067|


ardasnturk (approved):
Look good! 👌

PR #544, opened 2022-05-31
fix(categories): avoid intro text on sub category lists

donni106:
- added check for `hasSectionHeader` in order to render the
  intro text, because only the first category list screen has sections
  and sub category list screen receives always `false`
  - in order to test this, `texts.categoryList.intro` needs some test string

SVA-615

|first category list screen|sub category list before|sub category list after|
|---|---|---|
|IMG_0070|IMG_0069|IMG_F56A90110032-1|

### DIFF
--- a/src/components/CategoryList.js
+++ b/src/components/CategoryList.js
@@ -29,7 +29,7 @@ export class CategoryList extends React.PureComponent {
   };
 
   render() {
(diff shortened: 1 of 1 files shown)


ardasnturk (approved):
Looks good!

PR #545, opened 2022-05-31
fix(home screen): render button for all points of interest and tours

donni106:
- adjusted condition to set `showButton` depending on correct `query`
  in order to check the received data length
- reordered `fetchPolicy` for consistency

SVA-614

|before|after|
|---|---|
|IMG_0072|IMG_0073|

### DIFF
--- a/src/components/HomeSection.tsx
+++ b/src/components/HomeSection.tsx
@@ -3,7 +3,7 @@ import React from 'react';
 import { useQuery } from 'react-apollo';
 
 import { useHomeRefresh } from '../hooks';
-import { getQuery } from '../queries';
+import { getQuery, QUERY_TYPES } from '../queries';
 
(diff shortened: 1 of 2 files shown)


ardasnturk (commented):
Please add a post explaining how I can test this feature.
In v2.6.0 I can see this button on the screen.
I can confirm after running the test. 👍

ardasnturk:
> > Please add a post explaining how I can test this feature. In v2.6.0 I can see this button on the screen. I can confirm after running the test. 👍
> 
> you can see the button in `release/next/hdvt-consul`? the currently (pre) released 2.6.0 apps are missing the button, that's why i made a fix.

You are right. I can not see. I just asked because I couldn't test it in the same branch. Thank you for the information. 👍

ardasnturk (approved):
Looks good!

PR #551, opened 2022-06-02
setup eas build

donni106:
setup(eas build): initial configuration

  - ran `eas build:configure` after installing the cli tool per `npm install -g eas-cli`

setup(eas build): individual configuration

  - adjusted the configs to our build tools and versions from `.nvmrc` and `.yvmrc`, see https://docs.expo.dev/build/eas-json/#configuring-your-build-tools and https://docs.expo.dev/build/eas-json/#build-profiles
    - `production` builds will be submitted to app stores, for release to the general public or as part of a store-facilitated testing process
    - `preview` builds don't include developer tools, they are intended to be installed by your team and other stakeholders, to test out the app in production-like circumstances
    - `development` builds include developer tools, and they are never submitted to an app store
      - `development-simulator` alternatively prefer the development build to run in an simulator
        - note: for android there is no such special key needed
  - changed (yarn) scripts to new eas commands and sorted them accordingly

setup(eas build): create main `index.js`

  - created main entry point file as described here: https://docs.expo.dev/build-reference/migrating/#custom--main--entry-point-in

setup(eas build): specify expo `runtimeVersion`

  - specified a `runtimeVersion` to ensure that future updates are delivered only to compatible builds, see https://docs.expo.dev/distribution/runtime-versions
    - this way we avoid that live apps receives updates too early, which was the case with `expo build`/`expo publish`
  - the value must be one of https://docs.expo.dev/eas-update/runtime-versions
    - `nativeVersion` sounds as best fit for us right now
  - this `runtimeVersion` should be updated at least whenever native modules or JS–native interfaces changes
  - the `runtimeVersion` could theoretically differ for iOS and Android

SVA-444

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,16 +1,17 @@
 {
   "main": "node_modules/expo/AppEntry.js",
   "scripts": {
-    "android": "expo start --android",
-    "eject": "expo eject",
-    "ios": "expo start --ios",
-    "lint": "./node_modules/.bin/eslint __tests__ src --ext .js,.jsx,.ts,.tsx",
     "start": "expo start",
+    "start:android": "expo start --android",
+    "start:ios": "expo start --ios",
+    "build:android": "eas build --platform android",
+    "build:ios": "eas build --platform ios",
+    "submit:android": "eas submit --platform android",
+    "submit:ios": "eas submit --platform ios",
+    "lint": "./node_modules/.bin/eslint __tests__ src --ext .js,.jsx,.ts,.tsx",
     "test": "./node_modules/.bin/jest",
     "test:debug": "node --inspect-brk node_modules/jest/bin/jest.js --runInBand",
-    "test:coverage": "node_modules/.bin/jest --coverage",
-    "build:android": "expo build:android",
-    "build:ios": "expo build:ios"
+    "test:coverage": "node_modules/.bin/jest --coverage"
   },
   "jest": {
     "preset": "jest-expo",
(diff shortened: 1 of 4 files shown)


ardasnturk (approved):
Looks good! 🎉

package.json
```diff
@@ -1,16 +1,17 @@
...
-    "eject": "expo eject",
-    "ios": "expo start --ios",
-    "lint": "./node_modules/.bin/eslint __tests__ src --ext .js,.jsx,.ts,.tsx",
     "start": "expo start",
+    "start:android": "expo start --android",
+    "start:ios": "expo start --ios",
+    "build:android": "eas build --platform android",
+    "build:ios": "eas build --platform ios",
+    "submit:android": "eas submit --platform android",
+    "submit:ios": "eas submit -platform ios",
```
ardasnturk, line 10: should be `--platform`

ardasnturk, line 10: maybe we should combine build and submit here, because submit always requires a prior build. or we always do `yarn build:ios && yarn submit:ios` to combine them.

donni106: i think we will need multiple commands anyways, so let's do not have a combined one.

donni106: fixed 80965f875c77457b1d8bf5d875167c5c8b8dea4d

donni106: there is an `--auto-submit` flag for the `build` command, which seems to combine build and submit. This could be interesting for our `yarn submit:ios` as we always need to build before submission.

donni106: on the other hand regarding CI functionalities in a future release step, a build cannot happen on the server due to necessary cli prompts. so we still need a submit only here i guess.

PR #552, opened 2022-06-03
setup development build with `expo-dev-client`

donni106:
- installed `expo-dev-client` as documented here: https://docs.expo.dev/development/getting-started/#installing--expo-dev-client--in-your-project
- adjusted eas.json as there was an error saying: `Too long chain of profile extensions, make sure "extends" keys do not make a cycle`
- added `ios.bundleIdentifier` and `android.package` as local run scripts expected them
- automatically changed things after running `yarn ios` like adjustment of package.json: `Removed "main": "node_modules/expo/AppEntry.js" from package.json because we recommend using index.js as main instead` or adding `react-dom` and `react-native-web` to the dependencies
- prepared custom expo client with `yarn android` locally -> worked
  - was running automatically on an android emulator
- prepared custom expo client with `yarn ios` locally -> failed
  - this needs a proper setup Xcode, which seems to fail for me as my is outdated because of an old macOS
  - in addition i needed to remove `"EXCLUDED_ARCHS[sdk=iphonesimulator*]" = "arm64 ";` in `project.pbxproj` files, to make everything compile sucessfully except for @react-native-community/datetimepicker
    - this package seems to failed because of outdated macOS & Xcode
- building with `yarn build:ios-simulator` was successful on expo servers but therefore it was needed to temporarily comment out the gitignorance of `/src/config/secrets.js` as it would be missing on the server processes otherwise
  - you can find the different development builds in our expo account on expo.dev like previous `expo build`s
- after downloading the custom client follow these instructions: https://docs.expo.dev/build-reference/simulators/#installing-your-build-on-the-simulator
  - when running on the simulator everything should work 99% as known from the Expo Go app
- to start developing run `yarn dev` which will create a local server for the custom client

SVA-444

|server|app|
|---|---|
|Bildschirmfoto 2022-06-03 um 12 43 02|Simulator Screen Shot - iPhone 11 Pro - 2022-06-03 at 12 41 17-side|


donni106:
the custom dev client for this branch can be downloaded here: https://expo.dev/accounts/ikusei/projects/smart-village-app/builds/70aa8e58-2cb1-4132-aa2b-ccaebd1b3f66

ardasnturk (approved):

PR #553, opened 2022-06-14
fix(matomo): setup with condition

donni106:
- added check for existence of mandatory matomo values for initialization of an instance
- there was a warning otherwise which caused an annoying warning message on the screen with every app reload


ardasnturk (approved):
Looks good 👍

ardasnturk (approved):

PR #557, opened 2022-06-22
fix(global settings): initializations and accesses

donni106:
- adjusted conditional chaining and initial empty objects everywhere to avoid errors while accessing settings objects
  - removed `matomo: false` fallback in settings screen as we now always can access the settings object and if there is no matomo it resolves to false
  - we cannot add fallback initial values for every possible settings
- added fallback to `initialGlobalSettings` on setting up the global settings to ensure a valid value also if there was nothing stored locally before or accessing failed somehow
- fixed error: `[Unhandled promise rejection: TypeError: null is not an object (evaluating 'globalSettings.settings')]`

### DIFF
--- a/src/SettingsProvider.js
+++ b/src/SettingsProvider.js
@@ -5,7 +5,8 @@ export const SettingsContext = createContext({
   globalSettings: {
     filter: {},
     sections: {},
-    settings: {}
+    settings: {},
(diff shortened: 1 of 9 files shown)


ardasnturk (approved):
Looks good! 🎉

PR #569, opened 2022-07-08
fix(storage name): change directory

ardasnturk:
- changed the name of the storage to the
  name of the file to resolve the error of
  objects not displaying properly
  -  a note has been added to the file so
      that it can work at the desired performance.
      will be updated later

SVA-633

---

each texture file has a special name. viro uses these special names. 
we need to save them in the file directory with these names. the problem 
here is that the texture names of different objects may be the same. if 
this happens, when downloading the file, it saves it in place of the 
previous downloaded one and the object cannot be loaded properly. 
we need a different directory for each object. the file system does not 
allow us to create new directories. we have to find a different way!

---

Expo-FileSystem Document: https://docs.expo.dev/versions/latest/sdk/filesystem/#filesystemdocumentdirectory

---

if we want it to run at full performance, we need a 
different directory for each file. When creating 
these directories we get the following error. 
It works fine for testing but not for release.

Bildschirmfoto 2022-07-08 um 15 27 01


src/helpers/augmentedReality/downloadObject.js
```diff
@@ -13,16 +13,27 @@ export const downloadObject = async ({ index, data, setData }) => {
   for (const objectItem of dataItem?.downloadableUris) {
     const { uri, title, type, id } = objectItem;
 
-    const storageName = storageNameCreator({ dataItem, objectItem });
+    const { directoryName, folderName, storageName } = storageNameCreator({ dataItem, objectItem });
 
     const downloadResumable = FileSystem.createDownloadResumable(
       uri,
-      FileSystem.cacheDirectory + storageName,
+      directoryName,
```
donni106, line 20: what if we would use different `storageName`s instead of several directories? would it be possible to prefix `storageName`s with a timestamp or unique id to be always different? this way we could avoid the directory logics i think?

src/helpers/augmentedReality/storageNameCreator.js
```diff
@@ -1,6 +1,13 @@
...
-  const { title, type, id } = objectItem;
+  const { id: objectId, title: objectTitle } = dataItem;
+  const { title, type } = objectItem;
+  const replacedObjectTitle = objectTitle.replace(/\s+/g, '');
 
-  return `${objectId}-${title}${id}.${type}`;
+  return {
+    directoryName: documentDirectory + `${replacedObjectTitle}_${objectId}/${title}.${type}`,
+    folderName: documentDirectory + `${replacedObjectTitle}_${objectId}`,
+    storageName: `${replacedObjectTitle}_${objectId}_${title}.${type}`
```
donni106, line 11: why did the filename divider become underscore instead of hyphens?

src/helpers/augmentedReality/downloadObject.js
ardasnturk: it's hard to escape the logic of the index. Each texture file has a specific name. Viro needs this name to display a model. If we change the name of the texture file, the 3D object cannot be displayed properly.

src/helpers/augmentedReality/storageNameCreator.js
ardasnturk: No particular reason 😂🤷‍♂️

donni106 (approved):

src/helpers/augmentedReality/storageNameCreator.js
donni106, line 11: please create variables to reduce duplications,example:

```js
1 = `${replacedObjectTitle}_${objectId}`
2 = `${title}.${type}`
...
directoryName: documentDirectory + `${1}/${2}`,
folderName: documentDirectory + 1
storageName: `${1}_${2}`
```

something like this?

src/helpers/augmentedReality/storageNameCreator.js
```diff
@@ -1,6 +1,13 @@
+import { documentDirectory } from 'expo-file-system';
+
 export const storageNameCreator = ({ dataItem, objectItem }) => {
-  const { id: objectId } = dataItem;
-  const { title, type, id } = objectItem;
+  const { id: objectId, title: objectTitle } = dataItem;
+  const { title, type } = objectItem;
+  const replacedObjectTitle = objectTitle.replace(/\s+/g, '');
```
donni106, line 6: what does `replacedObjectTitle` tell us? could this be something more meaningful? like `titleWithoutSpaces`?

src/helpers/augmentedReality/storageNameCreator.js
```diff
@@ -1,6 +1,13 @@
+import { documentDirectory } from 'expo-file-system';
+
 export const storageNameCreator = ({ dataItem, objectItem }) => {
-  const { id: objectId } = dataItem;
-  const { title, type, id } = objectItem;
+  const { id: objectId, title: objectTitle } = dataItem;
+  const { title, type } = objectItem;
```
donni106, line 5: this confuses me when reading.
why does the `dataItem.title` be renamed to `objectTitle` and the `objectItem.title` becomes `title`?

donni106: what about not destructing at all and use `dataItem.id`, `dataItem.title`, `objectItem.title` and `objectItem.type` where necessary?

ardasnturk: https://github.com/ikuseiGmbH/smart-village-app-app/pull/569/commits/1dbcc5eb3ef7c049a142104e92b97fe6a1ba2bda

src/helpers/augmentedReality/storageNameCreator.js
ardasnturk: Yeah, it's better now.

https://github.com/ikuseiGmbH/smart-village-app-app/pull/569/commits/1dbcc5eb3ef7c049a142104e92b97fe6a1ba2bda

ardasnturk: Yeah, it's better now.

https://github.com/ikuseiGmbH/smart-village-app-app/pull/569/commits/881f40de6376a8c53355e44d9c6fe9d8dc3749f8

src/helpers/augmentedReality/storageNameCreator.js
ardasnturk: Yeah, it's better now.

https://github.com/ikuseiGmbH/smart-village-app-app/pull/569/commits/881f40de6376a8c53355e44d9c6fe9d8dc3749f8

PR #572, opened 2022-07-12
chore(expo sdk): upgrade to 44

donni106:
This PR upgrades Expo SDK to 44. There were a few breaking changes which were hard to tackle. The setup of Reanimated2 was difficult and it was not obvious to include `import 'react-native-reanimated';` in `index.js`. On the other hand the `style={{ flex: 1 }}` was required on the new `GestureHandlerRootView` wrapper to not have a white screen.

SVA-649


ardasnturk (approved):
Looks good 🎉

PR #573, opened 2022-07-12
fix(ar): rn elements crash

ardasnturk:
- changed `accesibilityRole` to `button` because
  the application with viro crashed because of
  `accessibilityRole`


donni106:
this branch is based on `epic/ar-modul` but is setup in this PR to be merged into `feature/SVA-633-show-ar-object`. i think this is not intended as there are a lot of files which i would have guessed unchanged in the context of this PR. is there a rebase necessary?

donni106 (commented):
this branch is based on `epic/ar-modul` but is setup in this PR to be merged into `feature/SVA-633-show-ar-object`. i think this is not intended as there are a lot of files which i would have guessed unchanged in the context of this PR. is there a rebase necessary?

PR #575, opened 2022-07-12
chore(expo sdk): upgrade to 45

donni106:
This PR upgrades Expo SDK to 45. It was hard to find out which caused the problem with babel stuff. After deleting yarn.lock and node_modules to reinstall everything from scratch the problem was solved. So maybe you also need to do these steps.

SVA-650


ardasnturk (approved):
Looks good 🎉

PR #578, opened 2022-07-12
feat(ar): start animation

ardasnturk:
- added the ability to start and stop
  animation to 3D object
- added a new if to the parser to
  distinguish if there is an animated
  3D object
- added animation prop to `Viro3DObject`
  to start and stop animation

SVA-633

## How to test:
* [x] Android portrait mode
* [x] iOS portrait mode
* [ ] Android landscape mode
* [ ] iOS landscape mode

## Screenshots:

|Animation Start Button|
|--|
IMG_0201 3

### DIFF
--- a/src/screens/augmentedReality/ARShowScreen.js
+++ b/src/screens/augmentedReality/ARShowScreen.js
@@ -12,6 +12,7 @@ import { LoadingSpinner } from '../../components';
 import { colors, Icon, normalize } from '../../config';
 
 export const ARShowScreen = ({ navigation, route }) => {
+  const [startAnimation, setStartAnimation] = useState(false);
   const [isLoading, setIsLoading] = useState(true);
   const [isVideoRecording, setIsVideoRecording] = useState(false);
   const data = route?.params?.data ?? [];
@@ -69,7 +70,7 @@ export const ARShowScreen = ({ navigation, route }) => {
         initialScene={{
           scene: AugmentedRealityView
         }}
-        viroAppProps={{ object }}
+        viroAppProps={{ object, startAnimation }}
         style={styles.arSceneNavigator}
       />
 
@@ -105,6 +106,16 @@ export const ARShowScreen = ({ navigation, route }) => {
         <Icon.NamedIcon name="camera" color={colors.darkText} size={normalize(30)} />
       </TouchableOpacity>
 
+      <TouchableOpacity
+        style={[styles.animationButton, styles.generalButtonStyle]}
+        onPress={() => setStartAnimation(!startAnimation)}
+      >
+        {startAnimation ? (
+          <Icon.NamedIcon name="stop" color={colors.primary} size={normalize(40)} />
+        ) : (
+          <Icon.NamedIcon name="play" color={colors.primary} size={normalize(40)} />
+        )}
+      </TouchableOpacity>
     </>
   );
 };
@@ -113,7 +124,8 @@ const AugmentedRealityView = ({ sceneNavigator }) => {
   const [rotation, setRotation] = useState([0, 0, 0]);
   const [position, setPosition] = useState([0, -1, -5]);
 
-  const { object } = sceneNavigator.viroAppProps;
+  const { object, startAnimation } = sceneNavigator.viroAppProps;
+
 
   const moveObject = (newPosition) => {
     setPosition(newPosition);
@@ -155,6 +167,11 @@ const AugmentedRealityView = ({ sceneNavigator }) => {
         onLoadStart={handleLoadStart}
         onLoadEnd={handleLoadEnd}
         onError={(event) => handleError(event)}
+        animation={{
+          loop: true,
+          name: object.animationName,
+          run: startAnimation
+        }}
       />
     </ViroARScene>
   );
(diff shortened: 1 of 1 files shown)


src/screens/augmentedReality/ARShowScreen.js
```diff
@@ -105,6 +106,16 @@ export const ARShowScreen = ({ navigation, route }) => {
...
       </TouchableOpacity>
 
+      <TouchableOpacity
+        style={[styles.animationButton, styles.generalButtonStyle]}
+        onPress={() => setStartAnimation(!startAnimation)}
+      >
+        {startAnimation ? (
+          <Icon.NamedIcon name="stop" color={colors.primary} size={normalize(40)} />
+        ) : (
+          <Icon.NamedIcon name="play" color={colors.primary} size={normalize(40)} />
```
donni106, line 116: why are these buttons bigger? should be 30 as well, right?

src/screens/augmentedReality/ARShowScreen.js
```diff
@@ -155,6 +167,11 @@ const AugmentedRealityView = ({ sceneNavigator }) => {
         onLoadStart={handleLoadStart}
         onLoadEnd={handleLoadEnd}
         onError={(event) => handleError(event)}
+        animation={{
+          loop: true,
+          name: object.animationName,
+          run: startAnimation
+        }}
```
donni106, line 174: could there be a problem if we have no `animationName` here?

src/screens/augmentedReality/ARShowScreen.js
```diff
@@ -12,6 +12,7 @@ import { LoadingSpinner } from '../../components';
 import { colors, Icon, normalize } from '../../config';
 
 export const ARShowScreen = ({ navigation, route }) => {
+  const [startAnimation, setStartAnimation] = useState(false);
```
donni106, line 15: i think it would be better to start the animation instantly, why should users want to not have something animated?

src/screens/augmentedReality/ARShowScreen.js
```diff
@@ -105,6 +106,16 @@ export const ARShowScreen = ({ navigation, route }) => {
         <Icon.NamedIcon name="camera" color={colors.darkText} size={normalize(30)} />
       </TouchableOpacity>
 
+      <TouchableOpacity
+        style={[styles.animationButton, styles.generalButtonStyle]}
+        onPress={() => setStartAnimation(!startAnimation)}
+      >
+        {startAnimation ? (
+          <Icon.NamedIcon name="stop" color={colors.primary} size={normalize(40)} />
```
donni106, line 114: this button behave more like a "pause" button, right? animations get paused instead of completely stoped to the beginning. in this case let's use a "pause" icon here.

src/screens/augmentedReality/ARShowScreen.js
```diff
@@ -193,6 +214,11 @@ const objectParser = async ({ item, setObject, onPress }) => {
 };
 
 var styles = StyleSheet.create({
+  animationButton: {
+    alignSelf: 'center',
+    bottom: 100,
+    padding: 12
```
donni106, line 220: normalize for some cases? can this conflict somehow with safeareas of particular devices?

src/screens/augmentedReality/ARShowScreen.js
ardasnturk: I can do this, but I think the animation will work in conjunction with the sound. I think it will be a bit interesting if users start listening to the sound before they realise what is happening.

src/screens/augmentedReality/ARShowScreen.js
ardasnturk: I need to explore this with a more specific animation. The animation of the sample file we have is not very clear. I think this button will stop the animation. It doesn't make it pause.

src/screens/augmentedReality/ARShowScreen.js
ardasnturk: I'll try it on different devices.

src/screens/augmentedReality/ARShowScreen.js
ardasnturk: If there is no animation name, the 3D object just stays as it is loaded and no error occurs.

src/screens/augmentedReality/ARShowScreen.js
ardasnturk: https://github.com/ikuseiGmbH/smart-village-app-app/pull/578/commits/b6f83169a8c3647c64aa371c9815f19973fbfb81

src/screens/augmentedReality/ARShowScreen.js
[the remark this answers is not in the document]
ardasnturk: https://github.com/ikuseiGmbH/smart-village-app-app/pull/578/commits/b6f83169a8c3647c64aa371c9815f19973fbfb81

src/screens/augmentedReality/ARShowScreen.js
ardasnturk: Yes it is different, I looked more carefully and the animation is just paused. That's why I changed the icon.

https://github.com/ikuseiGmbH/smart-village-app-app/pull/578/commits/b6f83169a8c3647c64aa371c9815f19973fbfb81

PR #579, opened 2022-07-15
feat(ar): add tour id

ardasnturk:
- added `tourID` prop to `AugmentedReality` component
  to show 3D model list only on a detail page of
  `Internationaler Kunstwanderweg`
- added manually the `tourID` prop to be able to view the
  list in the settings page
- added `id` to the Tour page to add the id of the tour as
  a prop to the `AugmentedReality` component

SVA-633

## How to test:
* [x] Android portrait mode
* [x] iOS portrait mode
* [ ] Android landscape mode
* [ ] iOS landscape mode


src/components/augmentedReality/AugmentedReality.js
```diff
@@ -39,6 +39,10 @@ export const AugmentedReality = ({ navigation, onSettingsScreen }) => {
     setIsLoading(false);
   };
 
+  if (error) return null;
```
donni106, line 42: what about using our `EmptyMessage` component here for users instead of a white screen?

src/components/augmentedReality/AugmentedReality.js
```diff
@@ -14,16 +15,15 @@ import { Wrapper, WrapperRow } from '../Wrapper';
...
-    name: 'arDownloadableDataList',
+export const AugmentedReality = ({ navigation, onSettingsScreen, tourID }) => {
+  const { data: staticData, error, loading, refetch } = useStaticContent({
+    name: `arDownloadableDataList-${tourID}`,
     type: 'json'
   });
 
   const [data, setData] = useState([]);
   const [isLoading, setIsLoading] = useState(loading);
+  const [isModalVisible, setIsModalVisible] = useState(false);
```
donni106, line 26: how come the modal state here?

src/components/screens/Tour.js
```diff
@@ -97,7 +98,7 @@ export const Tour = ({ data, navigation, route }) => {
           </View>
         )}
 
-        {!!augmentedReality && <AugmentedReality navigation={navigation} />}
+        {/* {!!augmentedReality && <AugmentedReality navigation={navigation} />} */}
```
donni106, line 101: this was for local testing? i think it should be down here uncommented?

src/screens/SettingsScreen.js
```diff
@@ -225,7 +225,9 @@ export const SettingsScreen = () => {
         />
       )}
       {selectedFilterId === TOP_FILTER.LIST_TYPES && <ListSettings />}
-      {selectedFilterId === TOP_FILTER.AR_DOWNLOAD_LIST && <AugmentedReality onSettingsScreen />}
+      {selectedFilterId === TOP_FILTER.AR_DOWNLOAD_LIST && (
+        <AugmentedReality onSettingsScreen tourID="579" />
```
donni106, line 229: nice, so this works until somehow id changes of the data on the server.
we need to check if the id is the same for int/development and bad belzig.

src/components/screens/Tour.js
ardasnturk: https://github.com/ikuseiGmbH/smart-village-app-app/pull/579/commits/799cbe77e5eca37c9114f903dd0c1d5d388c0e1a

src/components/augmentedReality/AugmentedReality.js
ardasnturk: It's just been moved. I thought it would be better to keep all the states together.

src/components/augmentedReality/AugmentedReality.js
ardasnturk: what about using our EmptyMessage component here for users instead of a white screen?

src/screens/SettingsScreen.js
ardasnturk: Yes, the tour id for Bad Belzig is different.

src/components/augmentedReality/AugmentedReality.js
donni106: i see

src/screens/SettingsScreen.js
donni106: so we need to change this in releases accordingly. please keep the int/development one here as the branches on Github belongs to int/development.

src/components/augmentedReality/AugmentedReality.js
ardasnturk: a blank page is not displayed here. Only the Augmented section is not displayed.

donni106: ah jo, good!

src/screens/SettingsScreen.js
ardasnturk: https://github.com/ikuseiGmbH/smart-village-app-app/pull/579/commits/f32b86fdbf21f3987b336340446d8ca3ef0ac25e

donni106 (approved):

PR #583, opened 2022-07-15
fix: solve empty section error

ardasnturk:
- fixed the error of displaying empty sections on
  `CrossData` page
- added `navigateButton` prop to `CrossData` to
  make `Alle anzeigen` button appear in sections
- added if to `DataListSection` to prevent rendering
  if `listData` is empty
- added to show placeholder render on screen if
  `listData` comes empty
- added null return if placeholder is empty

## How to test:
* [ ] Android portrait mode
* [x] iOS portrait mode
* [ ] Android landscape mode
* [ ] iOS landscape mode

## Screenshots:
|Before|After|
|--|--|
 IMG_0252| IMG_089E275CD3FD-1

### DIFF
--- a/src/components/DataListSection.tsx
+++ b/src/components/DataListSection.tsx
@@ -58,9 +58,8 @@ export const DataListSection = ({
   if (loading) {
     return (
       <View>
-        {!!sectionTitle && (
-          <SectionHeader onPress={navigate} title={sectionTitle ?? getTitleForQuery(query)} />
-        )}
+        <SectionHeader onPress={navigate} title={sectionTitle ?? getTitleForQuery(query)} />
+
         <LoadingContainer>
           <ActivityIndicator color={colors.accent} />
         </LoadingContainer>
@@ -80,20 +79,14 @@ export const DataListSection = ({
   if (listData?.length) {
     return (
       <View>
-        {!!sectionTitle && (
-          <SectionHeader onPress={navigate} title={sectionTitle ?? getTitleForQuery(query)} />
-        )}
-        {!!limit &&
-          (listData?.length ? (
-            <ListComponent
-              data={isRandom ? _shuffle(listData).slice(0, limit) : listData.slice(0, limit)}
-              horizontal={horizontal}
-              navigation={navigation}
-              query={query}
-            />
-          ) : (
-            !!placeholder && placeholder
-          ))}
+        <SectionHeader onPress={navigate} title={sectionTitle ?? getTitleForQuery(query)} />
+
+        <ListComponent
+          data={isRandom ? _shuffle(listData).slice(0, limit) : listData.slice(0, limit)}
+          horizontal={horizontal}
+          navigation={navigation}
+          query={query}
+        />
         {!!linkTitle && !!navigateLink && showLink && (
           <Wrapper>
             <Touchable onPress={navigateLink}>
--- a/src/components/DataListSection.tsx
+++ b/src/components/DataListSection.tsx
@@ -77,36 +77,49 @@ export const DataListSection = ({
     skipLastDivider: true
   });
 
+  if (listData?.length) {
+    return (
+      <View>
+        {!!sectionTitle && (
+          <SectionHeader onPress={navigate} title={sectionTitle ?? getTitleForQuery(query)} />
+        )}
+        {!!limit &&
+          (listData?.length ? (
+            <ListComponent
+              data={isRandom ? _shuffle(listData).slice(0, limit) : listData.slice(0, limit)}
+              horizontal={horizontal}
+              navigation={navigation}
+              query={query}
+            />
+          ) : (
+            !!placeholder && placeholder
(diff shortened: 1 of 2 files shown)


src/components/DataListSection.tsx
```diff
@@ -77,36 +77,49 @@ export const DataListSection = ({
...
-            </BoldText>
-          </Touchable>
-        </Wrapper>
-      )}
-      {!!buttonTitle && !!navigateButton && showButton && (
-        <Wrapper>
-          <Button title={buttonTitle} onPress={navigateButton} />
-        </Wrapper>
-      )}
+      <SectionHeader onPress={navigate} title={sectionTitle ?? getTitleForQuery(query)} />
```
donni106, line 121: In line 83 there is a `!!sectionTitle` before rendering the section header. Do we need it also here?
I think we can probably remove it in line 83 as well, right, as we have `sectionTitle ?? getTitleForQuery(query)`, so there is some fallback implemented for the prop?

src/components/DataListSection.tsx
```diff
@@ -77,36 +77,49 @@ export const DataListSection = ({
...
   });
 
+  if (listData?.length) {
+    return (
+      <View>
+        {!!sectionTitle && (
+          <SectionHeader onPress={navigate} title={sectionTitle ?? getTitleForQuery(query)} />
+        )}
+        {!!limit &&
+          (listData?.length ? (
```
donni106, line 87: if we check for `listData?.length` in line 80 now, we do not need to do it here again, right?

src/components/DataListSection.tsx
```diff
@@ -77,36 +77,49 @@ export const DataListSection = ({
...
+            <ListComponent
+              data={isRandom ? _shuffle(listData).slice(0, limit) : listData.slice(0, limit)}
+              horizontal={horizontal}
+              navigation={navigation}
+              query={query}
+            />
+          ) : (
+            !!placeholder && placeholder
+          ))}
```
donni106, line 96: this `placeholder` part can be removed now because we always have list data for this whole render part here, right?

donni106 (changes requested):
good catches! thanks. few questions about the refactoring:

src/components/DataListSection.tsx
ardasnturk: This js is used in more than one place. All pages have different props. And I tried to keep the latest updated version. But I can change it. 👍

src/components/DataListSection.tsx
ardasnturk: Yes, this check is unnecessary.

donni106: can be removed at line 61 as well, right?

src/components/DataListSection.tsx
ardasnturk: https://github.com/ikuseiGmbH/smart-village-app-app/pull/583/commits/c572a17d9f6c37319dec493afd2b8a6a9c43d631

src/components/DataListSection.tsx
ardasnturk: https://github.com/ikuseiGmbH/smart-village-app-app/pull/583/commits/c572a17d9f6c37319dec493afd2b8a6a9c43d631

src/components/DataListSection.tsx
ardasnturk: https://github.com/ikuseiGmbH/smart-village-app-app/pull/583/commits/c572a17d9f6c37319dec493afd2b8a6a9c43d631

src/components/DataListSection.tsx
```diff
@@ -80,20 +80,15 @@ export const DataListSection = ({
...
-        {!!sectionTitle && (
-          <SectionHeader onPress={navigate} title={sectionTitle ?? getTitleForQuery(query)} />
+        <SectionHeader onPress={navigate} title={sectionTitle ?? getTitleForQuery(query)} />
+        {!!limit && (
+          <ListComponent
+            data={isRandom ? _shuffle(listData).slice(0, limit) : listData.slice(0, limit)}
+            horizontal={horizontal}
+            navigation={navigation}
+            query={query}
+          />
```
donni106, line 90: do we need to check the `limit` as well already in line 80? what if there is no limit here? `data` could go wrong where `limit` is used, right? so it seems, that `limit` is mandatory.

ok as i checked now, `limit = 3` is set per default. so we do not need any check for `limit` anymore, right?

src/components/DataListSection.tsx
donni106: what about 61?

```
{!!sectionTitle && (
    <SectionHeader onPress={navigate} title={sectionTitle ?? getTitleForQuery(query)} />
  )}
```

ardasnturk: I realised after I shared the code. I'll fix it.

donni106 (approved):

PR #584, opened 2022-07-15
refactor(ar): delete object move and rotate functions

ardasnturk:
- deleted object `moveObject` and `rotateObject` functions
  as they are not necessary
- replace the `position` and `rotation` props with default
  data since the object  `moveObject` and `rotateObject`
  functions are deleted

SVA-633

### DIFF
--- a/src/screens/augmentedReality/ARShowScreen.js
+++ b/src/screens/augmentedReality/ARShowScreen.js
@@ -153,24 +153,7 @@ export const ARShowScreen = ({ navigation, route }) => {
 };
 
 const AugmentedRealityView = ({ sceneNavigator }) => {
-  const [rotation, setRotation] = useState([0, 0, 0]);
-  const [position, setPosition] = useState([0, -1, -5]);
-
   const { object, isStartAnimation, setIsObjectLoading } = sceneNavigator.viroAppProps;
-
-  const moveObject = (newPosition) => {
-    setPosition(newPosition);
-  };
-
-  const rotateObject = (rotateState, rotationFactor) => {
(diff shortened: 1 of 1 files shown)


src/screens/augmentedReality/ARShowScreen.js
```diff
@@ -180,11 +163,9 @@ const AugmentedRealityView = ({ sceneNavigator }) => {
         source={{ uri: object.vrx }}
         resources={[{ uri: object.png }]}
         type="VRX"
-        position={position}
+        position={[0, -1, -5]}
         scale={[0.02, 0.02, 0.02]}
-        rotation={rotation}
-        onDrag={moveObject}
-        onRotate={rotateObject}
+        rotation={[0, 0, 0]}
```
donni106, line 168: please keep these two as variables outside of the rendering, even if they are no state. we know that we will probably make these values coming from other data depending on the ar model:

```
const position = [0, -1, -5];
const rotation = [0, 0, 0];

...

position={position}
rotation={rotation}
```

and i think we should do it with the `scale` as well:

```
const scale = [0.02, 0.02, 0.02];

...

scale={scale}
```

donni106 (changes requested):

src/screens/augmentedReality/ARShowScreen.js
ardasnturk: https://github.com/ikuseiGmbH/smart-village-app-app/pull/584/commits/c66b97c24afe2ee31d4e1c6d02332677c239d827

donni106: nice

donni106 (approved):

PR #585, opened 2022-07-17
Include publiccode.yml

pwilimzig:
publiccode.yml is a metadata standard for repositories containing software developed or acquired by the Public Administration, aimed at making them easily discoverable and thus reusable by other entities. > https://yml.publiccode.tools/

This file should be included into our repo and must be maintained in the future!

### DIFF
--- a/publiccode.yml
+++ b/publiccode.yml
@@ -1,0 +1,120 @@
+publiccodeYmlVersion: '0.3'
+categories:
+  - online-community
+  - application-development
+  - appointment-scheduling
+  - communications
+  - customer-service-and-support
+  - digital-citizenship
+description:
+  en:
+    features:
+      - News
+      - Events
+      - Points of interest
+      - Service links
+      - Push notifications
+      - Weather
+      - many more
+    longDescription: >-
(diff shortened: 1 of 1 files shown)


publiccode.yml
```diff
@@ -0,0 +1,120 @@
...
+softwareType: standalone/mobile
+softwareVersion: 2.6.0
+url: 'https://github.com/ikuseiGmbH/smart-village-app-app'
+usedBy:
+  - Stadt Bad Belzig
+  - Stadt Herzberg/Elster
+  - Stadt Frankfurt (Oder)
+  - Landkreis Uckermark
+  - Amt Gransee und Gemeinden
+  - Gemeinde Schönefeld
```
donni106, line 120: `releaseDate`, `softwareVersion` and `usedBy` seems the main parts for more continuous updates, right? Other parts like `features` could be updated with `softwareVersion` and the other keys are more long-term, right?

donni106 (approved):

publiccode.yml
pwilimzig: Right

PR #586, opened 2022-07-18
feat(ar): add viro sound

ardasnturk:
- integrated viro sound to allow users to
  listen to audio
- added pause and start to a single state
  to synchronise the animation with the audio
  file
- updated `isStartAnimation` state name to
  `isStartAnimationAndSound` for clarity
- added control according to the mp3 data of
  the 3D object so that the application does
  not crash
- added `isObjectLoading` control to run the
  sound file when the object is loaded

SVA-633

## How to test:
* [x] Android portrait mode
* [x] iOS portrait mode
* [ ] Android landscape mode
* [ ] iOS landscape mode

### DIFF
--- a/src/screens/augmentedReality/ARShowScreen.js
+++ b/src/screens/augmentedReality/ARShowScreen.js
@@ -5,14 +5,15 @@ import {
   Viro3DObject,
   ViroAmbientLight,
   ViroARScene,
-  ViroARSceneNavigator
+  ViroARSceneNavigator,
+  ViroSound
 } from '@viro-community/react-viro';
 
 import { LoadingSpinner } from '../../components';
 import { colors, Icon, normalize, texts } from '../../config';
 
 export const ARShowScreen = ({ navigation, route }) => {
-  const [isStartAnimation, setIsStartAnimation] = useState(true);
+  const [isStartAnimationAndSound, setIsStartAnimationAndSound] = useState(true);
   const [isLoading, setIsLoading] = useState(true);
   const [isObjectLoading, setIsObjectLoading] = useState(true);
   const [isVideoRecording, setIsVideoRecording] = useState(false);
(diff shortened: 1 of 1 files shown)


src/screens/augmentedReality/ARShowScreen.js
```diff
@@ -153,12 +160,26 @@ export const ARShowScreen = ({ navigation, route }) => {
...
+    isStartAnimationAndSound,
+    setIsStartAnimationAndSound,
+    object
+  } = sceneNavigator.viroAppProps;
 
   return (
     <ViroARScene dragType="FixedToWorld">
       <ViroAmbientLight color={'#fff'} />
 
+      {object.mp3 && !isObjectLoading && (
```
donni106, line 175: `mp3` is not boolean. do we need `!!object.mp3` here?

src/screens/augmentedReality/ARShowScreen.js
```diff
@@ -153,12 +160,26 @@ export const ARShowScreen = ({ navigation, route }) => {
...
+        <ViroSound
+          source={{ uri: object.mp3 }}
+          paused={!isStartAnimationAndSound}
+          onFinish={() => setIsStartAnimationAndSound(false)}
```
donni106, line 179: we have the animation set with `loop: true`, can this be a problem with sound, when we set the flag to `false` here on finish? will the sound stop while the animation runs multiple times?

ardasnturk: If the sound ends, the animation will stop automatically. This is because the animation state is set to `false` in the `onFinish` function of `ViroSound`. But if we want a loop then I can delete `onFinish` and set `loop: true` instead.

donni106: äh, loop is already true:

```
animation={{
  loop: true,
```

that was why i asked that question.

ardasnturk: I can also make `loop: true` for `ViroSound`. That's what I meant.

donni106: you have to decide what's working, i don't know.
animation and sound needs to be in sync all the time.

ardasnturk: The animation and the sound file run at the same time. 👍

donni106 (approved):

PR #589, opened 2022-07-20
refactor(consul): save user id

ardasnturk:
- updated to store only the user's id information
  to avoid storing unused information about the
  user in memory
- updated with the user's id information directly
  from `asyncStorage` instead of `JSON.parse(...).id`
  for places where user id is needed

### DIFF
--- a/src/helpers/consul/index.js
+++ b/src/helpers/consul/index.js
@@ -13,9 +13,9 @@ export const setConsulAuthToken = async (authToken) => {
 
 export const getConsulAuthToken = async () => await SecureStore.getItemAsync(CONSUL_AUTH_TOKEN);
 
-export const setConsulUser = async (user) => {
-  if (user) {
-    await SecureStore.setItemAsync(CONSUL_USER, user.id);
+export const setConsulUser = async (userId) => {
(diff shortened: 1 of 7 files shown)


src/helpers/consul/index.js
```diff
@@ -15,7 +15,7 @@ export const getConsulAuthToken = async () => await SecureStore.getItemAsync(CON
 
 export const setConsulUser = async (user) => {
   if (user) {
-    await SecureStore.setItemAsync(CONSUL_USER, JSON.stringify(user));
+    await SecureStore.setItemAsync(CONSUL_USER, user.id);
```
donni106, line 18: could we pass the `userId` directly to this function instead of the whole `user`? this way we always know to receive a number instead of some object with values we cannot know 100%

ardasnturk: Yes, that's right. We only need the id information anyway.


https://github.com/ikuseiGmbH/smart-village-app-app/pull/589/commits/75e5cac47ab9ecaa53e670a89cb4c2bbd96025be

donni106 (approved):

PR #592, opened 2022-07-20
feat(ar): recognise image and show model

ardasnturk:
- added image recognition feature to show
  3D model
- wrapped augmented reality features with
  the `ViroARImageMarker` component to
  make the image recognition feature work
- added `targetImage` in `createTargets`
  object from `ViroARTrackingTargets` to
  add image information
- update the scale because the model looks
  too big
- added animation of the model and sound file
  when the image is recognised

SVA-633

## How to test:
* [ ] Android portrait mode
* [x] iOS portrait mode
* [ ] Android landscape mode
* [ ] iOS landscape mode

In order to test this feature, you must first delete any downloaded model data on your device. Then refresh the model list and download it again (54.9 MB). Because the target image must be added to the server for the image recognition feature.

After the download is complete, go to `ARShowScreen` and show the image shared below to the camera. 

augmentedReality

### DIFF
--- a/src/components/augmentedReality/AugmentedRealityView.js
+++ b/src/components/augmentedReality/AugmentedRealityView.js
@@ -1,0 +1,122 @@
+import PropTypes from 'prop-types';
+import React from 'react';
+import {
+  Viro3DObject,
+  ViroAmbientLight,
+  ViroARImageMarker,
+  ViroARScene,
+  ViroARTrackingTargets,
+  ViroSound
+} from '@viro-community/react-viro';
+
+import { texts } from '../../config';
+
+export const AugmentedRealityView = ({ sceneNavigator }) => {
+  const {
+    isObjectLoading,
+    setIsObjectLoading,
+    isStartAnimationAndSound,
+    setIsStartAnimationAndSound,
+    object
+  } = sceneNavigator.viroAppProps;
+
+  // TODO: these data can be updated according to the data coming from the server when the
+  //       real 3D models arrive
+  const position = [0, 0, 0];
+  const rotation = [0, 0, 0];
+  const scale = [0.002, 0.002, 0.002];
+
+  ViroARTrackingTargets.createTargets({
+    targetImage: {
+      physicalWidth: 0.2, // real world width in meters
+      source: { uri: object.target },
+      type: 'image'
+    }
+  });
+
+  return (
+    <ViroARScene dragType="FixedToWorld">
+      <ViroAmbientLight color={'#fff'} />
+
+      {object.target ? (
+        <ViroARImageMarker
+          onAnchorFound={() => setIsStartAnimationAndSound(true)} // animation and sound file are started after the image is recognised
+          pauseUpdates // prevents the model from continuous jumping
+          target={'targetImage'}
+        >
+          <ViroSoundAnd3DObject
+            {...{
+              isObjectLoading,
+              isStartAnimationAndSound,
+              object,
+              position,
+              rotation,
+              scale,
+              setIsObjectLoading,
+              setIsStartAnimationAndSound
+            }}
+          />
+        </ViroARImageMarker>
+      ) : (
+        <ViroSoundAnd3DObject
+          {...{
+            isObjectLoading,
+            isStartAnimationAndSound,
+            object,
+            position,
+            rotation,
+            scale,
+            setIsObjectLoading,
+            setIsStartAnimationAndSound
+          }}
+        />
+      )}
+    </ViroARScene>
+  );
+};
+
+const ViroSoundAnd3DObject = (item) => {
+  const {
+    isObjectLoading,
+    isStartAnimationAndSound,
+    object,
+    position,
+    rotation,
+    scale,
+    setIsObjectLoading,
+    setIsStartAnimationAndSound
+  } = item;
+
+  return (
+    <>
+      {!!object.mp3 && !isObjectLoading && (
+        <ViroSound
+          source={{ uri: object.mp3 }}
+          paused={!isStartAnimationAndSound}
+          onFinish={() => setIsStartAnimationAndSound(false)}
+        />
+      )}
+
+      <Viro3DObject
+        source={{ uri: object.vrx }}
+        resources={[{ uri: object.png }]}
+        type="VRX"
+        position={position}
+        rotation={rotation}
+        scale={scale}
+        onLoadStart={() => setIsObjectLoading(true)}
+        onLoadEnd={() => setIsObjectLoading(false)}
+        onError={() => alert(texts.augmentedReality.arShowScreen.objectLoadErrorAlert)}
+        animation={{
+          loop: true,
+          name: object.animationName,
+          run: isStartAnimationAndSound
+        }}
+      />
+    </>
+  );
+};
+
+AugmentedRealityView.propTypes = {
+  sceneNavigator: PropTypes.object
+};
--- a/src/components/augmentedReality/AugmentedRealityView.js
+++ b/src/components/augmentedReality/AugmentedRealityView.js
@@ -53,8 +53,11 @@ export const AugmentedRealityView = ({ sceneNavigator }) => {
     />
   );
 
+  /* see `onAnchorFound` function to auto-start animation and sound file 
+     when showing a 3D model that is not `TARGET` in the future time */
+
   return (
-    <ViroARScene dragType="FixedToWorld">
+    <ViroARScene>
       <ViroAmbientLight color={colors.surface} />
 
       {object?.target ? (
--- a/src/components/augmentedReality/AugmentedRealityView.js
+++ b/src/components/augmentedReality/AugmentedRealityView.js
@@ -1,5 +1,3 @@
-import PropTypes from 'prop-types';
-import React from 'react';
 import {
   Viro3DObject,
   ViroAmbientLight,
@@ -8,8 +6,10 @@ import {
   ViroARTrackingTargets,
   ViroSound
 } from '@viro-community/react-viro';
+import PropTypes from 'prop-types';
+import React, { useEffect } from 'react';
 
-import { texts } from '../../config';
+import { colors, texts } from '../../config';
 
 export const AugmentedRealityView = ({ sceneNavigator }) => {
   const {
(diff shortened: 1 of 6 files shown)


src/screens/augmentedReality/ARShowScreen.js
```diff
@@ -174,38 +176,52 @@ const AugmentedRealityView = ({ sceneNavigator }) => {
...
   //       real 3D models arrive
-  const position = [0, -1, -5];
+  const position = [0, 0, 0];
   const rotation = [0, 0, 0];
-  const scale = [0.02, 0.02, 0.02];
+  const scale = [0.002, 0.002, 0.002];
+
+  ViroARTrackingTargets.createTargets({
+    targetImage: {
+      physicalWidth: 0.2, // real world width in meters
```
donni106, line 185: is this also something that could change depending on the ar model or given trigger image thus it needs to be defined per server json?

src/screens/augmentedReality/ARShowScreen.js
```diff
@@ -174,38 +176,52 @@ const AugmentedRealityView = ({ sceneNavigator }) => {
...
+      source: { uri: object.target },
```
donni106, line 186: what happens if `object.target` would be undefined?

src/screens/augmentedReality/ARShowScreen.js
```diff
@@ -174,38 +176,52 @@ const AugmentedRealityView = ({ sceneNavigator }) => {
...
-      {!!object.mp3 && !isObjectLoading && (
-        <ViroSound
-          source={{ uri: object.mp3 }}
-          paused={!isStartAnimationAndSound}
-          onFinish={() => setIsStartAnimationAndSound(false)}
-        />
-      )}
+      <ViroARImageMarker
+        onAnchorFound={() => setIsStartAnimationAndSound(true)} // animation and sound file are started after the image is recognised
+        pauseUpdates={true} // prevents the model from continuous jumping
```
donni106, line 197: `pauseUpdates` would be enough to pass `true` here as prop, right?

src/screens/augmentedReality/ARShowScreen.js
```diff
@@ -174,38 +176,52 @@ const AugmentedRealityView = ({ sceneNavigator }) => {
...
-        onError={() => alert(texts.augmentedReality.arShowScreen.objectLoadErrorAlert)}
-        animation={{
-          loop: true,
-          name: object.animationName,
-          run: isStartAnimationAndSound
-        }}
-      />
+        <Viro3DObject
+          source={{ uri: object.vrx }}
+          resources={[{ uri: object.png }]}
```
donni106, line 210: what happens if `object.vrc` or `object.png` would be undefined? is this checked somewhere before?

src/screens/augmentedReality/ARShowScreen.js
ardasnturk: If undefined, nothing happens. But the object cannot be displayed.

src/screens/augmentedReality/ARShowScreen.js
ardasnturk: Yes, you're right.

src/screens/augmentedReality/ARShowScreen.js
ardasnturk: I will need to add a control in the parser for these. If undefined, the application crashes!

src/screens/augmentedReality/ARShowScreen.js
ardasnturk: I guess we don't need to add information about it in the json.

src/screens/augmentedReality/ARShowScreen.js
ardasnturk: Added a control and prevented the application from crashing.

https://github.com/ikuseiGmbH/smart-village-app-app/pull/592/commits/cb2f8b030d0be12176bb12acbbc265177dbd7e43

src/screens/augmentedReality/ARShowScreen.js
ardasnturk: updated the code to show the object directly in case there is no target image

https://github.com/ikuseiGmbH/smart-village-app-app/pull/592/commits/cb2f8b030d0be12176bb12acbbc265177dbd7e43

src/components/augmentedReality/AugmentedRealityView.js
```diff
@@ -0,0 +1,122 @@
...
+    isStartAnimationAndSound,
+    setIsStartAnimationAndSound,
+    object
+  } = sceneNavigator.viroAppProps;
+
+  // TODO: these data can be updated according to the data coming from the server when the
+  //       real 3D models arrive
+  const position = [0, 0, 0];
+  const rotation = [0, 0, 0];
+  const scale = [0.002, 0.002, 0.002];
```
donni106, line 27: why did the scale change from 0.02 to 0.002? isn't it nearly 0 now and 0 would work the same?

ardasnturk: how can we see if we make the size of the object 0? Yes, this is close to 0, but not 0. 0.02 was updated to 0.002 because it was too big. If I give it 0, it will not appear on the screen. If you delete scale, you will see how big the default size of the object is.

donni106: ah ja ok i see. so we would need to give the artists some information about the size of the arts in order to not have to set a scale here? or to just have a scale here for being set in order it is still too big.
is there some value we can give the artists to not provide too big arts?

ardasnturk: Unfortunately, I don't know how this value is determined.

donni106: can you find an object where no scale is needed? than we could send this to artists as template?

ardasnturk: I guess I'll have to try a lot of objects. I'm not sure I can find this one.

donni106: ok

src/components/augmentedReality/AugmentedRealityView.js
```diff
@@ -0,0 +1,122 @@
...
+
+  ViroARTrackingTargets.createTargets({
+    targetImage: {
+      physicalWidth: 0.2, // real world width in meters
+      source: { uri: object.target },
+      type: 'image'
+    }
+  });
```
donni106, line 35: do we need this once on mount thus put it in a use effect with `[]`? it is somehow "flying around" here out of anything?

src/components/augmentedReality/AugmentedRealityView.js
```diff
@@ -0,0 +1,122 @@
...
+
+  return (
+    <ViroARScene dragType="FixedToWorld">
```
donni106, line 38: what does `FixedToWorld` mean?

src/components/augmentedReality/AugmentedRealityView.js
```diff
@@ -0,0 +1,122 @@
...
+      <ViroAmbientLight color={'#fff'} />
```
donni106, line 39: if we use just a string, we should use it as `color="#fff"`. but i suggest to use our colors with `colors.surface` here.

src/components/augmentedReality/AugmentedRealityView.js
```diff
@@ -0,0 +1,122 @@
...
+            isObjectLoading,
+            isStartAnimationAndSound,
+            object,
+            position,
+            rotation,
+            scale,
+            setIsObjectLoading,
+            setIsStartAnimationAndSound
+          }}
+        />
```
donni106, line 72: we could reduce duplication here with having a const for the component with its props that we use inside the rendering.

src/components/augmentedReality/AugmentedRealityView.js
```diff
@@ -0,0 +1,122 @@
...
+
+  return (
+    <ViroARScene dragType="FixedToWorld">
+      <ViroAmbientLight color={'#fff'} />
+
+      {object.target ? (
+        <ViroARImageMarker
+          onAnchorFound={() => setIsStartAnimationAndSound(true)} // animation and sound file are started after the image is recognised
+          pauseUpdates // prevents the model from continuous jumping
+          target={'targetImage'}
```
donni106, line 45: i will create a constant `TARGET_IMAGE` that we can use here and in the `ViroARTrackingTargets.createTargets`

donni106: b697dd1498c5744680244368c0a1076d20cf7d3f

src/components/augmentedReality/AugmentedRealityView.js
donni106: b697dd1498c5744680244368c0a1076d20cf7d3f

src/components/augmentedReality/AugmentedRealityView.js
donni106: b697dd1498c5744680244368c0a1076d20cf7d3f

src/components/augmentedReality/AugmentedRealityView.js
donni106: b697dd1498c5744680244368c0a1076d20cf7d3f

src/components/augmentedReality/AugmentedRealityView.js
donni106: https://viro-community.readme.io/docs/viroarscene#dragtype

src/components/augmentedReality/AugmentedRealityView.js
```diff
@@ -0,0 +1,119 @@
...
+      {object?.target ? (
+        <ViroARImageMarker
+          onAnchorFound={() => setIsStartAnimationAndSound(true)} // animation and sound file are started after the image is recognised
+          pauseUpdates // prevents the model from continuous jumping
+          target={TARGET}
+        >
+          {ViroContent}
+        </ViroARImageMarker>
+      ) : (
+        ViroContent
```
donni106, line 69: in this case, sound and animation did not start automatically. do we need to change that? for now i guess no, because we will have targets in any case i think.

src/components/augmentedReality/AugmentedRealityView.js
ardasnturk: Yes, since we've removed the `onDrag` feature, we can delete it.

src/components/augmentedReality/AugmentedRealityView.js
ardasnturk: Yes, I don't think it's necessary for now as we have a target, but a note can be written for the future.

donni106 (approved):

PR #594, opened 2022-07-21
refactor(ar): add native folder to gitignore

ardasnturk:
- deleted the `iOS` and `Android` folders
  as we always create them locally
- added these folders to `gitignore` to
  avoid uploading native code folders on every
  commit

SVA-633


.gitignore
```diff
@@ -257,6 +260,11 @@ buck-out/
 # Bundle artifacts
 *.jsbundle
 
+# iOS ######
+
+# Native iOS Folder
+./ios
+
```
donni106, line 267: this seems to not work, as the folders are in git anyways. normally it needs two commits, one to delete the folders and another to add them to gitignore.
As you can see `/ios/Pods/` in here i guess `/android` and `/ios` would be the paths we need?

ardasnturk: https://github.com/ikuseiGmbH/smart-village-app-app/pull/594/commits/54ceb1a60dcf6dbdcf1e6ec06d3853a8fffe6877

donni106 (approved):

PR #598, opened 2022-07-22
fix(ar): solve android crash for recognise image

ardasnturk:
- added `orientation` prop to `createTargets`
  because image recognition feature is crash
  in android
  -  orientation prop: determines the orientation
      of the source image

[viro document for orientation](https://viro-community.readme.io/docs/viroartrackingtargets#orientation)

SVA-633

### DIFF
--- a/src/components/augmentedReality/AugmentedRealityView.js
+++ b/src/components/augmentedReality/AugmentedRealityView.js
@@ -31,6 +31,7 @@ export const AugmentedRealityView = ({ sceneNavigator }) => {
   useEffect(() => {
     ViroARTrackingTargets.createTargets({
       [TARGET]: {
+        orientation: 'Up',
         physicalWidth: 0.2, // real world width in meters
         source: { uri: object.target },
         type: 'image'


src/components/augmentedReality/AugmentedRealityView.js
```diff
@@ -31,6 +31,7 @@ export const AugmentedRealityView = ({ sceneNavigator }) => {
   useEffect(() => {
     ViroARTrackingTargets.createTargets({
       [TARGET]: {
+        orientation: 'Up',
```
donni106, line 34: sure we need this? 'Up' seems to be the default: https://github.com/ViroCommunity/viro/blob/7fa83641c807f3065291d0d3a29bfef4cacabaa9/ios/ViroReact/AR/Modules/VRTARTrackingTargetsModule.mm#L284

donni106: if this is the correct file, i cannot find a default for android: https://github.com/ViroCommunity/viro/blob/7fa83641c807f3065291d0d3a29bfef4cacabaa9/android/viro_bridge/src/main/java/com/viromedia/bridge/module/ARTrackingTargetsModule.java

donni106 (approved):

src/components/augmentedReality/AugmentedRealityView.js
ardasnturk: When I go to the augmented reality page on Android, it says that the `orientation` prop is missing and crash the application.

ardasnturk: I think it is assigned as default data only for iOS. We can write this on discord.

donni106: yes please share it with the community, it is something weird having a crash in this situation.

PR #599, opened 2022-07-22
feat: add filter for daily events on index screen

donni106:
- added a new switch on event index screens to toggle daily events which are default when coming from the event widget
- needed to make this `refetch` magic like we already did in `updateListData` to re-render the index screen
- re-organized imports in index screen and refactored `... ?... : null` to a shorter `&&` for the opening times filter

refactor(lists ): optimize behavior when filtering

- created a callback that handles the update of the switch state as well as the update of the query variables
  - this way we get rid of one re-rendering (first switch statte, second query variables state) which seems to affect the list rendering in a weird way
- refactored loading screen to refresh control to not have a full full white screen with the loading spinner when filtering but a loading spinner in the body of the list implemented per `ListEmptyComponent` using the `loading` flag
  - to position it vertically centered, lists needed a new `contentContainerStyle`
- refactored building of list items in a callback `buildListItems` to remove it from the render section and to benefit a bit from the memoization
- refactored line lengths of comments to be within 100 chars
- re-organized imports of touched files to be consistent in code base

HDVT-73

---

|22.07.2022 on|22.07.2022 off|
|---|---|
|Simulator Screen Shot - iPhone 8 Plus - 2022-07-22 at 18 45 10|Simulator Screen Shot - iPhone 8 Plus - 2022-07-22 at 18 45 17|

### DIFF
--- a/src/components/widgets/EventWidget.tsx
+++ b/src/components/widgets/EventWidget.tsx
@@ -12,6 +12,8 @@ import { WidgetProps } from '../../types';
 
 import { DefaultWidget } from './DefaultWidget';
 
+const currentDate = moment().format('YYYY-MM-DD');
+
 export const EventWidget = ({ text }: WidgetProps) => {
   const navigation = useNavigation();
   const refreshTime = useRefreshTime('event-widget', consts.REFRESH_INTERVALS.ONCE_A_DAY);
@@ -19,7 +21,6 @@ export const EventWidget = ({ text }: WidgetProps) => {
 
   const fetchPolicy = graphqlFetchPolicy({ isConnected, isMainserverUp, refreshTime });
 
-  const currentDate = moment().format('YYYY-MM-DD');
   const queryVariables = {
     dateRange: [currentDate, currentDate],
     order: 'listDate_ASC'
@@ -36,7 +37,8 @@ export const EventWidget = ({ text }: WidgetProps) => {
       title: text ?? texts.homeTitles.events,
       query: QUERY_TYPES.EVENT_RECORDS,
       queryVariables,
-      rootRouteName: 'EventRecords'
+      rootRouteName: 'EventRecords',
+      filterByDailyEvents: true
     });
   }, [navigation, text, queryVariables]);
 
--- a/src/screens/IndexScreen.js
+++ b/src/screens/IndexScreen.js
@@ -49,6 +49,7 @@ const keyForSelectedValueByQuery = {
   [QUERY_TYPES.EVENT_RECORDS]: 'categoryId',
   [QUERY_TYPES.NEWS_ITEMS]: 'dataProvider'
 };
+
 const getAdditionalQueryVariables = (query, selectedValue, excludeDataProviderIds) => {
   const keyForSelectedValue = keyForSelectedValueByQuery[query];
   const additionalQueryVariables = {};
@@ -136,17 +137,14 @@ export const IndexScreen = ({ navigation, route }) => {
 
   const updateListDataByDropdown = useCallback(
     (selectedValue) => {
-      const additionalQueryVariables = getAdditionalQueryVariables(
-        query,
-        selectedValue,
-        excludeDataProviderIds
-      );
-
       if (selectedValue) {
         // remove a refetch key if present, which was necessary for the "- Alle -" selection
         delete queryVariables.refetch;
 
-        setQueryVariables({ ...queryVariables, ...additionalQueryVariables });
+        setQueryVariables({
+          ...queryVariables,
+          ...getAdditionalQueryVariables(query, selectedValue, excludeDataProviderIds)
+        });
       } else {
         setQueryVariables((prevQueryVariables) => {
           // remove the filter key for the specific query, when selecting "- Alle -"
@@ -158,7 +156,7 @@ export const IndexScreen = ({ navigation, route }) => {
         });
       }
     },
-    [excludeDataProviderIds, setQueryVariables, query, queryVariables]
+    [excludeDataProviderIds, query, queryVariables]
   );
 
   const updateListDataByDailySwitch = useCallback(() => {
@@ -186,7 +184,7 @@ export const IndexScreen = ({ navigation, route }) => {
 
       return !oldSwitchValue;
     });
-  }, [filterByDailyEvents, setQueryVariables, queryVariables]);
+  }, [filterByDailyEvents, queryVariables]);
 
   // if we show the map or want to sort by distance, we need to fetch all the entries at once
   // this is not a big issue if we want to sort by distance, because getting the location usually
@@ -205,7 +203,7 @@ export const IndexScreen = ({ navigation, route }) => {
     // we want to ensure when changing from one index screen to another, for example from
     // news to events, that the query variables are taken freshly. otherwise the mounted screen can
     // have query variables from the previous screen, that does not work. this can result in an
-    // empty screen because the query is not retuning anything.
+    // empty screen because the query is not returning anything.
     const variables = {
       ...(route.params?.queryVariables ?? {}),
       ...getAdditionalQueryVariables(query, undefined, excludeDataProviderIds)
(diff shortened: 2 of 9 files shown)


src/components/widgets/EventWidget.tsx
```diff
@@ -12,14 +12,15 @@ import { WidgetProps } from '../../types';
 
 import { DefaultWidget } from './DefaultWidget';
 
+const currentDate = moment().format('YYYY-MM-DD');
+
```
ardasnturk, line 16: How did you benefit from moving it here?

src/screens/IndexScreen.js
```diff
@@ -159,6 +165,28 @@ export const IndexScreen = ({ navigation, route }) => {
...
+        // remove the filter key for the specific query, when unselecting daily events
+        delete queryVariables.dateRange;
+        // need to spread the `prevQueryVariables` into a new object with additional refetch key
+        // to force the Query component to update the data, otherwise it is not fired somehow
+        // because the state variable wouldn't change
+        return { ...prevQueryVariables, refetch: true };
+      });
+    }
+  }, [filterByDailyEvents]);
+
```
ardasnturk, line 189: when I switch the filter off and on, the events load late. This may be due to internet connection speed, I'm not sure. But it doesn't look good when the list is refreshed late.

https://user-images.githubusercontent.com/11755668/180493941-d70c5f86-5713-4842-a384-5e635aa70c7d.mov

src/screens/IndexScreen.js
```diff
@@ -195,18 +223,17 @@ export const IndexScreen = ({ navigation, route }) => {
 
   return (
     <SafeAreaViewFlex>
-      {query === QUERY_TYPES.POINTS_OF_INTEREST ? (
+      {query === QUERY_TYPES.POINTS_OF_INTEREST && (
```
ardasnturk, line 226: Why did you replace the `?` operator with the `&&` operator?

src/screens/IndexScreen.js
```diff
@@ -271,7 +298,18 @@ export const IndexScreen = ({ navigation, route }) => {
...
                   <>
                     {!!showFilter && (
-                      <DropdownHeader {...{ query, queryVariables, data, updateListData }} />
+                      <>
+                        <DropdownHeader {...{ query, queryVariables, data, updateListData }} />
+                        {query === QUERY_TYPES.EVENT_RECORDS && (
+                          <View>
+                            <OptionToggle
+                              label={texts.eventRecord.filterByDailyEvents}
+                              onToggle={() => setFilterByDailyEvents((value) => !value)}
```
ardasnturk, line 307: Instead of `(value) => !value`, how about `!filterByDailyEvents`?

src/components/widgets/EventWidget.tsx
donni106: i thought to move it outside of the component as it has nothing to do with the component and its lifecycle, it does not depend on anything. it is like a constant value for each day, which do not need to be recalculated on every mount or re-renderings. the variable can be created and "stored" more globally in the code base for its usage in the EventWidget component.

src/screens/IndexScreen.js
donni106: interesting, you mean the behavior at the end, right, where there is no loading spinner when toggling off. did this happen on different devices? at my tests while development the loading spinner took place in both cases, when switching on and also off.

src/screens/IndexScreen.js
donni106: I replaced it because of our more consistent way of rendering "something or nothing". I felt it like a interruption of our code flow and way of reading it. why do we need a short-hand if-else here with explicitly rendering `null` if the query is not for POIs? when reading the code, we want to do something if we have a query for POIs and nothing else. it is like rendering other stuff only if some condition is true. imagine our `DataListSection` component with `... ? ... : null` all the time:

Bildschirmfoto 2022-07-25 um 09 49 30

this would become a lot more unreadable and confusing in my opinion. that's why i adjusted it here as well when i saw it.

src/screens/IndexScreen.js
donni106: I implemented it in the same way as for the existing toggle

```
<OptionToggle
  label={texts.pointOfInterest.filterByOpeningTime}
  onToggle={() => setFilterByOpeningTimes((value) => !value)}
  value={filterByOpeningTimes}
/>
```

i also asked myself to do it this way or setting `!filterByDailyEvents` and update the other part with `!filterByOpeningTimes`. but i don't wanted to test these special changes but be sure that everything works the same as before. i also would need to read about react best practices about when to use pre-value passed as param and when to use the state variable directly. sure, in cases where we have the setter method used somewhere without having its state variable, one would use the passed pre-value param. but in cases where we use it in the same component/file? could you make a short research please if you find out more about the best use?

ardasnturk: Yes, after a little research, I understand better. Thank you.

src/screens/IndexScreen.js
ardasnturk: Was there any activity event while you were doing this test? Because I got this error after setting a new event for a specific day.

src/screens/IndexScreen.js
ardasnturk: Yes, it certainly reads better.

src/screens/IndexScreen.js
donni106: And what is the outcome of the research? Can you tell me please in which situation which usage is preferred?

src/screens/IndexScreen.js
donni106: > after setting a new event for a specific day

what does this mean? i did nothing special, just opened the app and toggled the switch with the outcomes of my screenshots and the white screen with loading indicator in between.

ardasnturk: Today `25.07.2022`. When we activate this filter, it only shows today's events. If there is no event today, it works fine. But if there is an event today, this problem occurs.

src/screens/IndexScreen.js
ardasnturk: Since it's easy to get confused with all the callbacks and timing, and to avoid any unexpected side-effects, it's best to use the functional updater state setter argument. We can use this method especially when the state before updating the value is unknown.

src/screens/IndexScreen.js
donni106: Ok thanks, I understand and will check more on this.

donni106: please check again now. there are several changes according behavior and rendering. there is no more "white screen with loading" when filtering but a loading spinner inside the view under the filters.
with refactoring stuff i hope the problem of weird list item rendering on toggling for daily events is gone.

src/screens/IndexScreen.js
```diff
@@ -107,37 +114,93 @@ export const IndexScreen = ({ navigation, route }) => {
...
+          return { ...prevQueryVariables, refetchDate: true };
         });
       }
-    },
-    [excludeDataProviderIds, setQueryVariables, query, queryVariables]
-  );
+
+      return !oldSwitchValue;
+    });
+  }, [filterByDailyEvents, queryVariables]);
```
ardasnturk, line 197: When I go to the "Veranstaltungen" page from the main page, today's events are displayed. That's because the filter is switched on. However, if I go directly to the "Veranstaltungen" page from the "Meldungen" or "Nachrichten" page, all events are displayed, even though the filter is activated.

donni106: thanks for this very good review catch! fixed with bb5a256

ardasnturk (commented):
When I switch off and on the filter on the "Activity" page, the list refreshes itself late, as shown in the video. What could be the reason for this?

https://user-images.githubusercontent.com/11755668/182142019-654c609f-03bc-4ad5-844a-f04bef5d55df.MP4

donni106:
> When I switch off and on the filter on the "Activity" page, the list refreshes itself late, as shown in the video. What could be the reason for this?
> 
>  RPReplay_Final1659354629.MP4

:o i have no idea right now. it did not happen in my simulator. can you reproduce this in different devices?
first thing i could think of is that some state is changed before another, so the old list data is rendered before `data` was fully updated, which should be handled by the `loading` value.

ardasnturk:
> > When I switch off and on the filter on the "Activity" page, the list refreshes itself late, as shown in the video. What could be the reason for this?
> > RPReplay_Final1659354629.MP4
> 
> :o i have no idea right now. it did not happen in my simulator. can you reproduce this in different devices? first thing i could think of is that some state is changed before another, so the old list data is rendered before `data` was fully updated, which should be handled by the `loading` value.

I tried it on 2 other devices. iPhone 13 Pro Max and Pixel 4 XL. The result on both is as in the video.

PR #605, opened 2022-07-25
feat(ar): add map view to augmented reality

ardasnturk:
- added `IndexFilterWrapperAndList` component to
  Augmented Reality section for map view and list view
- added `mapToMapMarkers` function to make the
  location data from the server suitable for the map
  component
- added new texts to `texts` file to show in filter section
- added redirection to `artworkDetailScreen` with
  markers on the map
- added `showUserLocation` as a prop and set true
  by default
- added full screen map view
- added a new button for full screen
  navigation at the bottom right of the
  small map view
- added `onMaximizeButtonPress` prop to
  `AugmentedReality` component to switch
  to large map view
- added new map navigation to
  `defaultStackConfig`
- added artwork information window shown below
  the map when clicking on the marker on the full
  page map
- used `navigationToArtworksDetailScreen` helper
  to navigate to `artworkDetailScreen` from the
  opened detail window

## How to test:
* [x] Android portrait mode
* [x] iOS portrait mode
* [ ] Android landscape mode
* [ ] iOS landscape mode

## Screenshots:

|MapView on AR Section|ListView on AR Section|
|--|--|
IMG_0029 | IMG_0030

|Full Screen Map|Full Screen Map with Art details|
|--|--|
IMG_0031 |  IMG_0032

### DIFF
--- a/src/components/map/Map.tsx
+++ b/src/components/map/Map.tsx
@@ -9,11 +9,13 @@ import { imageHeight, imageWidth } from '../../helpers';
 import { MapMarker } from '../../types';
 
 type Props = {
+  isMaximizeButtonVisible?: boolean;
   locations?: MapMarker[];
   mapCenterPosition?: { lat: number; lng: number };
   mapStyle?: StyleProp<ViewStyle>;
   onMapPress?: () => void;
   onMarkerPress?: (arg0?: string) => void;
+  onMaximizeButtonPress?: () => void;
   showsUserLocation?: boolean;
   style?: StyleProp<ViewStyle>;
   zoom?: number;
@@ -25,11 +27,13 @@ const LATITUDE_DELTA = 0.0922;
 const LONGITUDE_DELTA = LATITUDE_DELTA * (device.width / (device.height / 2));
 
 export const Map = ({
+  isMaximizeButtonVisible,
   locations,
   mapCenterPosition,
   mapStyle,
   onMapPress,
   onMarkerPress,
+  onMaximizeButtonPress,
   showsUserLocation = true,
   style,
   zoom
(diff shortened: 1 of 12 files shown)


src/components/map/Map.tsx
```diff
@@ -94,6 +99,11 @@ export const Map = ({
           </Marker>
         ))}
       </MapView>
+      {isMaximizeButtonVisible && (
+        <TouchableOpacity style={stylesForMap().maximizeMapButton} onPress={onMaximizeButtonPress}>
+          <Icon.NamedIcon name="expand" color={colors.primary} size={normalize(18)} />
```
donni106, line 104: i will move this to configs/Icon.tsx, to have icons in one place. we need to adjust `name`s across different cities according to their selected icon font set, which should take place in one file at best.

donni106: this will become `<Icon.ExpandMap size={normalize(18)} />` here

src/screens/MapViewScreen.js
```diff
@@ -0,0 +1,86 @@
...
+                  data,
+                  isNavigation: true,
+                  modelId,
+                  navigation,
+                  refetch
+                })
+              }
+            >
+              <Icon.ArrowRight size={normalize(30)} color={colors.primary} />
+            </TouchableOpacity>
```
donni106, line 65: the whole row needs to be pressable as always, not just the button. and with a touchable (os dependent) not just opacity, as elsewhere. i will adjust that.
the primary color is default, so i will remove that as well.

src/screens/MapViewScreen.js
```diff
@@ -0,0 +1,86 @@
...
+      {isAugmentedReality && modelData && (
+        <Wrapper>
+          <WrapperRow spaceBetween>
+            <View style={styles.augmentedRealityInfoContainer}>
+              <RegularText big>{modelData.title}</RegularText>
+              <RegularText small>{modelData.locationInfo}</RegularText>
+              <WrapperVertical>
+                <RegularText placeholder numberOfLines={2}>
+                  {modelData.description}
+                </RegularText>
```
donni106, line 49: the description is html, so i will replace the text rendering with a html view

PR #607, opened 2022-08-05
fix(linking): error android linking not opening

ardasnturk:
- fixed `canOpenURL` bug that prevented links
  from opening on android

SVA-689

## How to test:
* [x] Android portrait mode
* [x] iOS portrait mode
* [x] Android landscape mode
* [x] iOS landscape mode


donni106 (approved):
https://support.google.com/googleplay/android-developer/answer/10158779?hl=en

PR #608, opened 2022-08-05
feat(ar): add `tourStops` query

ardasnturk:
- added `tourStops` query to `tour` query
- updated all helpers accordingly as the model
  information is included in the `payload` object
- changed from `png` to `texture` in the resource
  section to support different file types
- removed `refetch` because the data came with
  tour data
- updated `DOWNLOAD_TYPE` to `downloadType`
- added `imageType` to `storageNameCreator`
  and updated `objectName` to capture the actual
  name and extension of the texture file

SVA-633

### DIFF
--- a/src/components/augmentedReality/AugmentedReality.js
+++ b/src/components/augmentedReality/AugmentedReality.js
@@ -15,23 +15,12 @@ import { ARModal } from './ARModal';
 import { ARObjectList } from './ARObjectList';
 import { WhatIsARButton } from './WhatIsARButton';
 
-const OnSettingsScreen = ({ id, isLoading, setData }) => {
-  const { data } = useQuery(getQuery(QUERY_TYPES.TOUR_STOPS), { variables: { id } });
-
-  return (
-    <ARObjectList
-      data={data?.tour?.tourStops}
-      setData={setData}
-      isLoading={isLoading}
-      showDeleteAllButton
-      showDownloadAllButton
-      showFreeSpace
-      showTitle
-    />
-  );
-};
-
 export const AugmentedReality = ({ id, navigation, onSettingsScreen, tourStops }) => {
+  const { data: tourData } = useQuery(getQuery(QUERY_TYPES.TOUR_STOPS), {
+    variables: { id },
+    skip: !onSettingsScreen
+  });
+
   const [isARSupported, setIsARSupported] = useState(false);
   const [data, setData] = useState([]);
   const [isLoading, setIsLoading] = useState(false);
@@ -61,7 +50,17 @@ export const AugmentedReality = ({ id, navigation, onSettingsScreen, tourStops }
   if (!isARSupported) return null;
 
   if (onSettingsScreen) {
-    return <OnSettingsScreen {...{ id, isLoading, setData }} />;
+    return (
+      <ARObjectList
+        data={tourData?.tour?.tourStops}
+        setData={setData}
+        isLoading={isLoading}
+        showDeleteAllButton
+        showDownloadAllButton
+        showFreeSpace
+        showTitle
+      />
+    );
   }
 
   if (isLoading) return <LoadingSpinner loading />;
@@ -112,10 +111,4 @@ AugmentedReality.propTypes = {
   navigation: PropTypes.object,
   onSettingsScreen: PropTypes.bool,
   tourStops: PropTypes.array
-};
-
-OnSettingsScreen.propTypes = {
-  id: PropTypes.string,
-  isLoading: PropTypes.bool,
-  setData: PropTypes.func
 };
--- a/src/components/augmentedReality/AugmentedReality.js
+++ b/src/components/augmentedReality/AugmentedReality.js
@@ -1,6 +1,7 @@
 import PropTypes from 'prop-types';
 import React, { useEffect, useState } from 'react';
 import { isARSupportedOnDevice } from '@viro-community/react-viro';
+import { useQuery } from 'react-apollo';
 
 import { consts, device, texts } from '../../config';
 import { checkDownloadedData } from '../../helpers';
@@ -8,15 +9,24 @@ import { Button } from '../Button';
 import { LoadingSpinner } from '../LoadingSpinner';
 import { Title, TitleContainer, TitleShadow } from '../Title';
 import { Wrapper } from '../Wrapper';
+import { GET_TOUR_STOPS } from '../../queries/tours';
 
 import { ARModal } from './ARModal';
 import { ARObjectList } from './ARObjectList';
 import { WhatIsARButton } from './WhatIsARButton';
+import { getQuery, QUERY_TYPES } from '../../queries';
 
-export const AugmentedReality = ({ tourStops, navigation, onSettingsScreen }) => {
+export const AugmentedReality = ({ id, navigation, onSettingsScreen }) => {
+  const {
+    data: {
+      tour: { tourStops }
+    },
+    loading
+  } = useQuery(getQuery(QUERY_TYPES.TOUR_STOPS), { variables: { id } });
+
   const [isARSupported, setIsARSupported] = useState(false);
   const [data, setData] = useState([]);
-  const [isLoading, setIsLoading] = useState(false);
+  const [isLoading, setIsLoading] = useState(loading);
   const [isModalVisible, setIsModalVisible] = useState(false);
 
   useEffect(() => {
@@ -42,7 +52,7 @@ export const AugmentedReality = ({ tourStops, navigation, onSettingsScreen }) =>
 
   if (!isARSupported) return null;
 
-  if (isLoading || !tourStops) return <LoadingSpinner loading />;
+  if (isLoading || !tourStops?.length) return <LoadingSpinner loading />;
 
   const a11yText = consts.a11yLabel;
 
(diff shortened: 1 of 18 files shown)


src/screens/SettingsScreen.js
```diff
@@ -240,6 +240,7 @@ export const SettingsScreen = () => {
         />
       )}
       {selectedFilterId === TOP_FILTER.LIST_TYPES && <ListSettings />}
+      {/* TODO: this will crash now as the props were changed in `AugmentedReality`, right? */}
       {selectedFilterId === TOP_FILTER.AR_DOWNLOAD_LIST && (
         <AugmentedReality onSettingsScreen tourID="579" />
       )}
```
donni106, line 246: this will crash now as the props were changed in `AugmentedReality`, right?

src/screens/augmentedReality/ARShowScreen.js
```diff
@@ -163,17 +163,21 @@ export const ARShowScreen = ({ navigation, route }) => {
...
+  item?.payload?.localUris?.forEach((item) => {
+    if (item.type === 'texture') {
+      parsedObject[item.type].push({ uri: item.uri });
+    } else {
+      parsedObject[item.type] = item.uri;
+    }
   });
 
-  if (!parsedObject.png || !parsedObject.vrx) {
+  if (!parsedObject?.texture?.length || !parsedObject?.vrx) {
```
donni106, line 180: with creating `const parsedObject = { texture: [] };` the check of `!parsedObject.texture` would be always false, so I added length check for array

donni106 (changes requested):
I did some refactorings and added a TODO, please check the comments and if everything still works 😄

src/screens/SettingsScreen.js
ardasnturk: There will be no application crash here. Only the list is not loaded. But I think I solved this problem in [984398](https://github.com/ikuseiGmbH/smart-village-app-app/pull/608/commits/a2eab2d01e81cd949d8827aeb854f772ad59f304).

src/screens/augmentedReality/ARShowScreen.js
ardasnturk: Thank you 🙏

src/helpers/augmentedReality/storageNameCreator.js
```diff
@@ -1,9 +1,14 @@
...
+import { consts } from '../../config';
+
+const { IMAGE_TYPE_REGEX } = consts;
+
 export const storageNameCreator = ({ dataItem, objectItem }) => {
+  const imageType = IMAGE_TYPE_REGEX.exec(objectItem.uri);
   const objectItemTitleWithoutSpaces = dataItem.title.replace(/\s+/g, '');
   const dataDirectoryName = `${objectItemTitleWithoutSpaces}_${dataItem.id}`;
-  const objectName = `${objectItem.title}.${objectItem.type}`;
+  const objectName = `${objectItem.title}.${imageType ? imageType[1] : objectItem.type}`;
```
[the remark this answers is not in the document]
ardasnturk: This only varies according to the texture types. For different types we can use the information from the server. For example, I am writing a few different file types:

Ch17_1001.**texture** = Ch17_1001.**png**
Ch17_1002.**texture** = Ch17_1002.**jpg**
augmentedReality.**target** = augmentedReality.**png**
Musik.**mp3** = Musik.**mp3**
test.**vrx** = test.**vrx**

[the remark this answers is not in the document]
donni106: ok, this is not very clear to me when reading the line. can you refactor this to a function with a certain name and a comment including these example so that we still know about it in the future?

src/components/augmentedReality/AugmentedReality.js
```diff
@@ -1,29 +1,28 @@
...
-    data: staticData,
-    error,
-    loading,
-    refetch
-  } = useStaticContent({
-    name: `arDownloadableDataList-${tourID}`,
-    type: 'json'
-  });
+    data: {
+      tour: { tourStops }
```
donni106, line 22: is this safe? i think it could be a problem destructing for `tourStops` if there is no `tour` in `data`. how can you be sure, that this does not happen?

src/components/augmentedReality/AugmentedReality.js
```diff
@@ -38,22 +37,22 @@ export const AugmentedReality = ({ navigation, onSettingsScreen, tourID }) => {
...
     setIsLoading(true);
     await checkDownloadedData({ data, setData });
     setIsLoading(false);
   };
 
-  if (error || !isARSupported) return null;
+  if (!isARSupported) return null;
 
-  if (isLoading || !staticData) return <LoadingSpinner loading />;
+  if (isLoading || !tourStops?.length) return <LoadingSpinner loading />;
```
donni106, line 55: when i look at this now, is it always correct to show a loading spinner if there are no tour stops?

src/components/screens/Tour.js
```diff
@@ -98,7 +96,7 @@ export const Tour = ({ data, navigation, route }) => {
           </View>
         )}
 
-        {!!augmentedReality && <AugmentedReality navigation={navigation} tourID={id} />}
+        {!!tourStops?.length && <AugmentedReality {...{ id, navigation }} />}
```
donni106, line 99: now we will have tour stops data here through the tour query and make the same query for the data again in `AugmentedReality`? i suggest not doing it this way. if there is already data for tour stops we can pass and use them like before.
only for settingsscreen we can make a new query. therefore we should skip the useQuery hook if not `onSettingsScreen`. you will find such `skip` already in the code for other places i guess.

src/helpers/augmentedReality/storageNameCreator.js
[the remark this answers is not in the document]
ardasnturk: I will do it. 👍

src/components/augmentedReality/AugmentedReality.js
ardasnturk: Hmm. I'll have to think of something for that.

src/components/augmentedReality/AugmentedReality.js
ardasnturk: In fact, this component is not displayed when `tourStops is not present. This place can be organised.

src/components/screens/Tour.js
ardasnturk: Yeah, I'll fix that.

ardasnturk: https://github.com/ikuseiGmbH/smart-village-app-app/pull/608/commits/afa7f49f3c7659b48f0bc217d6e549966e443c24

src/components/augmentedReality/AugmentedReality.js
ardasnturk: https://github.com/ikuseiGmbH/smart-village-app-app/pull/608/commits/afa7f49f3c7659b48f0bc217d6e549966e443c24

src/components/augmentedReality/AugmentedReality.js
ardasnturk: https://github.com/ikuseiGmbH/smart-village-app-app/pull/608/commits/afa7f49f3c7659b48f0bc217d6e549966e443c24

src/helpers/augmentedReality/storageNameCreator.js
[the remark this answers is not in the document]
ardasnturk: https://github.com/ikuseiGmbH/smart-village-app-app/pull/608/commits/afa7f49f3c7659b48f0bc217d6e549966e443c24

src/components/augmentedReality/AugmentedReality.js
```diff
@@ -14,20 +15,26 @@ import { ARModal } from './ARModal';
...
-  const {
-    data: staticData,
-    error,
-    loading,
-    refetch
-  } = useStaticContent({
-    name: `arDownloadableDataList-${tourID}`,
-    type: 'json'
-  });
+const OnSettingsScreen = ({ id, isLoading, setData }) => {
```
donni106, line 18: we name components with `...Screen` if the present a screen, like index, detail ...
calling `return <OnSettingsScreen {...{ id, isLoading, setData }} />;` now suggests some screen component being rendered instead of a small component like this.
i suggest a rollback and working with `skip` in the `useQuery` to only call it on the settings screen.
feel free to ask again, if something is not clear.

src/helpers/augmentedReality/storageNameCreator.js
```diff
@@ -1,9 +1,26 @@
...
+		Ch17_1002.texture       = Ch17_1002.jpg
+		augmentedReality.target = augmentedReality.png
+		Musik.mp3               = Musik.mp3
+		test.vrx                = test.vrx
+ 	*/
+  const textureType = IMAGE_TYPE_REGEX.exec(objectItem.uri);
   const objectItemTitleWithoutSpaces = dataItem.title.replace(/\s+/g, '');
   const dataDirectoryName = `${objectItemTitleWithoutSpaces}_${dataItem.id}`;
-  const objectName = `${objectItem.title}.${objectItem.type}`;
+  const objectName = `${objectItem.title}.${textureType ? textureType[1] : objectItem.type}`;
```
donni106, line 23: please create a separate function that can be commented and reused instead of having "it all" inside here.

src/components/augmentedReality/AugmentedReality.js
ardasnturk: You're absolutely right. I think it's much clearer now.
https://github.com/ikuseiGmbH/smart-village-app-app/pull/608/commits/39ed66226c960aa10e02b3e40ebcb1e5c5432701

PR #611, opened 2022-08-09
feat(volunteer): add gifted chat

ardasnturk:
- added `gifted-chat` package
- added new chat component
- added `UserAvatar` prepared for use in chat component
- added `alwaysShowSend` prop to show the send button
  on the screen continuously
- added `onFocus` control to `messagesContainerStyle`
  so that the messages are not under the input when the
  keyboard appears on the screen
- added `onScrollToBottom` prop to quickly navigate to the
  newest incoming message
- added `renderAvatar` prop to see avatar next to messages
  from another user
- added `renderBubble` prop to create our own design in
  message bubbles
- added `renderComposer` prop to render text input in the
  design we want
- return `renderDay` prop as `null` because the date information
  of the message is shown inside the message bubble in the
  created design and add date information to `renderTime` prop
  with `RegularText` component
- changed the design of the message sending button with the
  `renderSend` prop to match the design
- added `renderInputToolbar` prop to create the design of the
  area behind `textInput` and send message buttons
- removed some wrappers and `VolunteerMessageTextField` 
  from `VolunteerDetailScreen` because they are no longer 
  used
- integrated `Chat` component into `VolunteerMessage` to 
  use gifted chat component
- created expressions with `forEach` inside `useEffect` to 
  create an array of objects with the structure `giftedChat` 
  wants
- used mutateAsync function to send message
- reversed the array with the `reverse()` function so that 
  the newest message is displayed below the screen and the 
  oldest message is displayed above the screen

HDVT-85

⚠️ `react-native-gifted-chat` supports version `^4.2.4` of `react-native-safe-area-context`.
you need to update `react-native-safe-area-context` for this ⚠️

## How to test:
* [ ] Android portrait mode
* [x] iOS portrait mode
* [ ] Android landscape mode
* [x] iOS landscape mode

## Screenshots:

|Chat Screen|Chat Screen (with Keyboard)|Chat Screen (with New Message)|
|--|--|--|
Simulator Screen Shot - iPhone 13 - 2022-08-09 at 12 06 22 | Simulator Screen Shot - iPhone 13 - 2022-08-09 at 12 06 41 | Simulator Screen Shot - iPhone 13 - 2022-08-09 at 12 06 46

|Landscape mode|
|--|
Simulator Screen Shot - iPhone 13 - 2022-08-09 at 12 48 38 |

### DIFF
--- a/package.json
+++ b/package.json
@@ -83,7 +83,7 @@
     "react-native-expo-image-cache": "4.1.0",
     "react-native-gesture-handler": "~1.10.2",
     "react-native-get-random-values": "~1.7.0",
-    "react-native-gifted-chat": "^1.0.4",
+    "react-native-gifted-chat": "0.16.3",
     "react-native-markdown-display": "^7.0.0-alpha.2",
     "react-native-modal-dropdown": "https://github.com/donni106/react-native-modal-dropdown#feature/search-options",
     "react-native-qrcode-svg": "^6.1.1",
--- a/package.json
+++ b/package.json
@@ -83,6 +83,7 @@
     "react-native-expo-image-cache": "4.1.0",
     "react-native-gesture-handler": "~1.10.2",
     "react-native-get-random-values": "~1.7.0",
+    "react-native-gifted-chat": "^1.0.4",
     "react-native-markdown-display": "^7.0.0-alpha.2",
     "react-native-modal-dropdown": "https://github.com/donni106/react-native-modal-dropdown#feature/search-options",
     "react-native-qrcode-svg": "^6.1.1",
--- a/src/components/Chat.js
+++ b/src/components/Chat.js
@@ -1,0 +1,222 @@
+import 'moment/locale/de';
+import PropTypes from 'prop-types';
+import React, { useCallback, useEffect, useState } from 'react';
+import { StyleSheet, View } from 'react-native';
+import { Avatar } from 'react-native-elements';
+import {
+  Bubble,
+  Composer,
+  GiftedChat,
+  InputToolbar,
+  MessageText,
+  Send
+} from 'react-native-gifted-chat';
+
+import { colors, Icon, normalize } from '../config';
+import { momentFormat } from '../helpers';
+
+import { RegularText } from './Text';
+
+/* eslint-disable react-native/no-inline-styles */
+/* eslint-disable react/prop-types */
+
+const UserAvatar = ({ uri, title }) => (
+  <Avatar
+    containerStyle={styles.spacing}
+    overlayContainerStyle={[styles.overlayContainerStyle, !uri && styles.border]}
+    placeholderStyle={styles.placeholderStyle}
+    rounded
+    source={uri ? { uri } : undefined}
+    renderPlaceholderContent={
+      <Avatar
+        containerStyle={[styles.containerStyle]}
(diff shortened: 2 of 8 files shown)


package.json
```diff
@@ -1,134 +1,135 @@
...
+		"jest-expo": "^43.0.0",
+		"jest-styled-components": "^7.0.3",
+		"mixin-deep": "^1.3.2",
+		"prettier": "^2.2.1",
+		"react-test-renderer": "^17.0.2",
+		"set-value": "^4.1.0",
+		"typescript": "~4.3.5"
+	},
+	"resolutions": {
+		"markdown-it": "^12.3.2"
```
donni106, line 133: this file is linted wrongly again, i will fix this before merge to see the changes

src/components/Chat.js
```diff
@@ -0,0 +1,214 @@
+import 'moment/locale/de';
```
donni106, line 1: i think we do not need this here, as the momentFormat helper is setting the locale

src/components/Chat.js
```diff
@@ -0,0 +1,214 @@
...
+  Send
+} from 'react-native-gifted-chat';
+
+import { colors, Icon, normalize } from '../config';
+import { momentFormat } from '../helpers';
+
+import { RegularText } from './Text';
+
+/* eslint-disable react-native/no-inline-styles */
+/* eslint-disable react/prop-types */
```
donni106, line 21: i will remove these and fix the nline style with adding a style and the prop types with adding the missing prop type for `currentMessage`

src/components/Chat.js
```diff
@@ -0,0 +1,214 @@
...
+        <Bubble
+          {...props}
+          wrapperStyle={{
+            left: bubbleWrapperStyleLeft || {
+              backgroundColor: colors.gray20
+            },
+            right: bubbleWrapperStyleRight || {
+              // TODO: added manually because there is no similar colour in the colours file.
+              //       can be edited later!
+              backgroundColor: '#E8F1E9'
```
donni106, line 102: ok

src/components/Chat.js
```diff
@@ -0,0 +1,214 @@
...
+  const onSend = useCallback((messages) => {
+    onSendButton(messages[0].text);
+  }, []);
+
+  return (
+    <GiftedChat
+      alwaysShowSend
+      messages={messages}
+      minInputToolbarHeight={normalize(96)}
+      onSend={(messages) => onSend(messages)}
```
donni106, line 82: can be shortened to `onSend={onSend}`

src/components/Chat.js
```diff
@@ -0,0 +1,214 @@
...
+        />
+      )}
+      renderComposer={(props) => (
+        <Composer
+          {...props}
+          textInputStyle={styles.textInputStyle}
+          textInputProps={textInputProps}
+        />
+      )}
+      renderDay={() => null}
```
donni106, line 114: wow i just saw this and took it.
so we can shorten the messages to only render time, because date is there as kind of title

image

src/components/Chat.js
donni106: or even different with

```
onSend={(messages) => {
  onSendButton(messages[0].text);
}}
```

why did you decide to have a `useCallback` without dependency? i guess we cannot profit from that.

PR #612, opened 2022-08-09
feat(volunteer): update route for user requests with id

donni106:
- in order to request user data without being an admin we needed a new endpoint in our v2 api


ardasnturk (approved):
looks good 🎉

PR #613, opened 2022-08-09
feat(volunteer): hdvt events between main server events

donni106:
This integrates volunteer events in main server events if a global setting is activated (`hdvt.events = true`).

HDVT-43

---

|home section|index|index today|detail|
|---|---|---|---|
|Simulator Screen Shot - iPhone 11 - 2022-08-09 at 18 39 28|Simulator Screen Shot - iPhone 11 - 2022-08-09 at 17 57 25|Simulator Screen Shot - iPhone 11 - 2022-08-09 at 17 57 31|Simulator Screen Shot - iPhone 11 - 2022-08-09 at 18 39 45|

### DIFF
--- a/src/components/volunteer/VolunteerEventRecord.tsx
+++ b/src/components/volunteer/VolunteerEventRecord.tsx
@@ -108,6 +108,10 @@ export const VolunteerEventRecord = ({
           />
         )}
 
+        <Wrapper>
+          <InfoCard category={category} webUrls={webUrls} openWebScreen={openWebScreen} />
+        </Wrapper>
+
         {!!appointments?.length && (
           <View>
             <TitleContainer>
@@ -151,10 +155,6 @@ export const VolunteerEventRecord = ({
             </Wrapper>
           </View>
         )}
-
-        <Wrapper>
-          <InfoCard category={category} webUrls={webUrls} openWebScreen={openWebScreen} />
-        </Wrapper>
 
         {isAttendingEvent !== undefined && (
           <Wrapper>
--- a/src/components/volunteer/VolunteerEventRecord.tsx
+++ b/src/components/volunteer/VolunteerEventRecord.tsx
@@ -62,12 +62,12 @@ export const VolunteerEventRecord = ({
     }
   ];
 
-  const [isAttendingEvent, setIsAttendingEvent] = useState(false);
+  const [isAttendingEvent, setIsAttendingEvent] = useState<boolean>();
 
   const checkIfAttending = useCallback(async () => {
     const { currentUserId } = await volunteerUserData();
 
-    setIsAttendingEvent(isAttending(currentUserId, attending));
+    !!currentUserId && setIsAttendingEvent(isAttending(currentUserId, attending));
   }, [participants]);
 
   useEffect(() => {
@@ -99,7 +99,7 @@ export const VolunteerEventRecord = ({
         )}
         {device.platform === 'ios' && <TitleShadow />}
 
-        {!!attending?.length && (
+        {isAttendingEvent !== undefined && !!attending?.length && (
           <VolunteerEventAttending
             calendarEntryId={id}
             data={attending}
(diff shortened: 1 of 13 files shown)


src/components/volunteer/VolunteerEventRecord.tsx
```diff
@@ -99,7 +99,7 @@ export const VolunteerEventRecord = ({
         )}
         {device.platform === 'ios' && <TitleShadow />}
 
-        {!!attending?.length && (
+        {isAttendingEvent !== undefined && !!attending?.length && (
```
ardasnturk, line 102: why did you choose `isAttendingEvent !== undefined` here? can we get the same result with `!!isAttendingEvent`?

src/components/volunteer/VolunteerEventRecord.tsx
```diff
@@ -152,18 +156,16 @@ export const VolunteerEventRecord = ({
...
-
-        <Wrapper>
-          {!isAttendingEvent && <RegularText small>{texts.volunteer.attendInfo}</RegularText>}
-          <Button
-            title={isAttendingEvent ? texts.volunteer.notAttend : texts.volunteer.attend}
-            invert={isAttendingEvent}
-            onPress={attend}
-          />
-        </Wrapper>
+        {isAttendingEvent !== undefined && (
```
ardasnturk, line 159: why did you choose `isAttendingEvent !== undefined` here? can we get the same result with `!!isAttendingEvent`?

src/components/widgets/EventWidget.tsx
```diff
@@ -42,9 +53,21 @@ export const EventWidget = ({ text }: WidgetProps) => {
...
+  const todaysDataVolunteerEvents = dataVolunteerEvents?.filter(
+    ({ listDate }: { listDate: string }) => listDate === currentDate
+  );
+
+  let eventCount = data?.eventRecords?.length || 0;
 
-  const eventCount = data?.eventRecords?.length;
+  if (showVolunteerEvents && todaysDataVolunteerEvents?.length) {
+    eventCount += todaysDataVolunteerEvents.length;
+  }
```
ardasnturk, line 70: why did you want to increase the `eventCount` here?

src/screens/IndexScreen.js
```diff
@@ -322,6 +348,14 @@ export const IndexScreen = ({ navigation, route }) => {
               ]) &&
               !Object.prototype.hasOwnProperty.call(queryVariables, 'refetch');
 
+            // apply additional data if volunteer events should be presented and
+            // no category selection is made, because the category has nothing to do with
+            // volunteer data
+            const additionalData =
+              isCalendarWithVolunteerEvents && !hasCategoryFilterSelection
+                ? dataVolunteerEvents
+                : undefined;
```
ardasnturk, line 357: what happens if we use the `&&` operator instead of `?` here?

ardasnturk (commented):
When I want to refresh the events page, it stutters the same way.

https://user-images.githubusercontent.com/11755668/183879698-9a0af67d-9d10-4965-9916-95d8d013a9b1.mp4

There is a strange stutter when going from another page to the events page. Do you know what's causing it?

https://user-images.githubusercontent.com/11755668/183879811-64a554f7-a803-4dc2-990f-7ad6f8f39445.mp4

src/components/volunteer/VolunteerEventRecord.tsx
donni106: we need to check if `isAttendingEvent` is not set, which means its neither `true` or `false`. if it is `true` or `false` their was a successful validation of the attending status, otherwise we do not have a logged in user, so we will not show any buttons.

src/components/volunteer/VolunteerEventRecord.tsx
donni106: we need to check if `isAttendingEvent` is not set, which means its neither `true` or `false`. if it is `true` or `false` their was a successful validation of the attending status, otherwise we do not have a logged in user, so we will not show any buttons.

this is what i meant with `- added condition to not show event attending button if not logged in` in the corresponding commit.

src/components/widgets/EventWidget.tsx
donni106: to add the number of events from volunteer module. otherwise the widget shows a smaller number then the index when clicking on it.

this is what i meant with `add volunteer events count to event widget count` in the commit.

src/screens/IndexScreen.js
donni106: `additionalData` would be boolean or array instead of undefined or array. but we types with `additional?: unknown[]` which corresponds to undefined or array.

ardasnturk: Thanks for the information. 👍

ardasnturk (approved):

PR #617, opened 2022-08-10
fix: solve cycle warn

ardasnturk:
- solved cycle error caused by imports

### DIFF
--- a/src/components/volunteer/VolunteerFormCalendar.tsx
+++ b/src/components/volunteer/VolunteerFormCalendar.tsx
@@ -15,7 +15,7 @@ import { CREATE_EVENT_RECORDS } from '../../queries/eventRecords';
 import { calendarNew, calendarUpload, groups } from '../../queries/volunteer';
 import { VolunteerCalendar, VolunteerGroup } from '../../types';
 import { Button } from '../Button';
-import { DocumentSelector, ImageSelector } from '../consul';
+import { DocumentSelector, ImageSelector } from '../consul/selectors';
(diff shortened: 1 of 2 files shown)


src/components/volunteer/index.ts
```diff
@@ -16,6 +17,7 @@ export * from './VolunteerHomeSection';
 export * from './VolunteerMessage';
 export * from './VolunteerMessageTextField';
 export * from './VolunteerPostListItem';
+export * from './VolunteerPostListItem';
```
donni106, line 20: new duplications here?

src/components/volunteer/index.ts
```diff
@@ -1,4 +1,5 @@
 export * from './VolunteerApplicantListItem';
+export * from './VolunteerApplicantListItem';
```
donni106, line 2: new duplication here?

ardasnturk: https://github.com/ikuseiGmbH/smart-village-app-app/pull/617/commits/6ea15dda586b5bd947e8b7cd4a2facb746f89026

src/components/volunteer/index.ts
ardasnturk: https://github.com/ikuseiGmbH/smart-village-app-app/pull/617/commits/6ea15dda586b5bd947e8b7cd4a2facb746f89026

donni106 (approved):

src/components/volunteer/index.ts
donni106: updating the import in the other file was the correct solution

PR #618, opened 2022-08-11
feat(maps): `react-native-maps` instead of `react-native-webview-leaflet`

donni106:
This integrates `react-native-maps` instead of `react-native-webview-leaflet` in every component we use maps.

SVA-634


app.json
```diff
@@ -39,7 +40,12 @@
         "READ_EXTERNAL_STORAGE",
         "WRITE_EXTERNAL_STORAGE"
       ],
-      "jsEngine": "hermes"
+      "jsEngine": "hermes",
+      "config": {
+        "googleMaps": {
+          "apiKey": "abcdef"
```
donni106, line 46: this needs to be changed individually in releases with a secret google maps api key

package.json
```diff
@@ -105,12 +106,12 @@
     "react-native-screens": "~3.11.1",
     "react-native-snap-carousel": "~3.9.0",
     "react-native-svg": "12.3.0",
+    "react-native-web": "0.17.7",
     "react-native-webview": "11.18.1",
     "react-native-webview-leaflet": "https://github.com/donni106/react-native-webview-leaflet#28dc2b06f957bec3e64a439f1fac62e1390da624",
```
donni106, line 111: TODO: remove the old package

donni106: 507703ffdd9919d30d5dac39e87e43d7982ce9eb

ardasnturk (approved):

ardasnturk (approved):

PR #619, opened 2022-08-12
feat(maps): add polyline to map

ardasnturk:
- added `Polyline` to `Map` component to show
  the route on the map
  -  to see the route we need location information
      of type `LatLng[]`
- added `geometryTourData` to `Tour` query for
  route information
- added the necessary props to display the route
  on the fullscreen map page
- added control to prevent an error if the
  `augmentedReality` object does not have location
  information

SVA-634

## How to test:
* [ ] Android portrait mode
* [x] iOS portrait mode
* [ ] Android landscape mode
* [ ] iOS landscape mode

## Screenshots:

|Polyline without loading the map|Polyline wit loading the map|Full screen Polyline on the map|
|--|--|--|
IMG_0059 | IMG_0061 | IMG_0060

### DIFF
--- a/src/components/map/Map.tsx
+++ b/src/components/map/Map.tsx
@@ -1,7 +1,7 @@
 import React, { useRef } from 'react';
 import { StyleProp, StyleSheet, TouchableOpacity, View, ViewStyle } from 'react-native';
 import { normalize } from 'react-native-elements';
-import MapView, { MAP_TYPES, Marker, UrlTile } from 'react-native-maps';
+import MapView, { LatLng, MAP_TYPES, Marker, Polyline, UrlTile } from 'react-native-maps';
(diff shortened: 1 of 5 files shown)


src/components/map/Map.tsx
```diff
@@ -80,6 +82,9 @@ export const Map = ({
           left: 0
         }}
       >
+        {!!geometryTourData?.length && (
+          <Polyline coordinates={geometryTourData} strokeWidth={2} strokeColor={colors.primary} />
```
donni106, line 86: is there some z index or layering stuff? i have the polyline under the map tiles for some reason.

donni106: when i change `<UrlTile` and `<Polyline` it works

PR #621, opened 2022-08-16
feature(event): not showing count in event widget

ardasnturk:
- added `noCount` bool expression in
  `additionalProps` to catch changes
  on the server
- added `noFilterByDailyEvents` bool
  expression in `additionalProps` to
  catch changes on the server
- added not showing count on widget if
  `additionalProps.noCount` is set to
  `true` in `globalSettings`
- added control to `filterByDailyEvents`
  param to turn off the daily event filter
  when opening the index page
- converted `styles` to function and 
  added `count` prop to set styles 
  according to the number of widgets
- deleted the blank line rendered to 
  the screen in case there is no count 
  to centre the icon
- changed `count` to a boolean expression 
  by adding `!` at the beginning to 
  eliminate the problem of `false` if 
  `count` is `0`

SVA-664

⚠️ please add the necessary objects to `globalSettings` to try it out! 
you can find information about this in `Confluence` ⚠️

## How to test:
* [x] Android portrait mode
* [x] iOS portrait mode
* [x] Android landscape mode
* [x] iOS landscape mode

## Screenshots:
|event widget with count|event widget without count|
|--|--|
Simulator Screen Shot - iPhone 13 - 2022-08-16 at 11 55 59 | Simulator Screen Shot - iPhone 13 - 2022-09-07 at 16 20 18

### DIFF
--- a/src/components/widgets/DefaultWidget.tsx
+++ b/src/components/widgets/DefaultWidget.tsx
@@ -16,11 +16,11 @@ type Props = {
 export const DefaultWidget = ({ Icon, count, onPress, text }: Props) => {
   return (
     <TouchableOpacity onPress={onPress}>
-      <WrapperVertical style={styles().container}>
+      <WrapperVertical style={styles.container}>
         <WrapperRow center>
-          <Icon style={styles(!!count?.toString()).icon} />
+          <Icon style={[styles.iconWithoutCount, !!count?.toString() && styles.iconWithCount]} />
           <BoldText primary big>
-            {count}
+            {count ?? ''}
           </BoldText>
         </WrapperRow>
         <RegularText primary small>
@@ -33,14 +33,15 @@ export const DefaultWidget = ({ Icon, count, onPress, text }: Props) => {
 
 /* eslint-disable react-native/no-unused-styles */
 /* this works properly, we do not want that warning */
-const styles = (count?: boolean) =>
-  StyleSheet.create({
-    container: {
-      alignItems: 'center'
-    },
-    icon: {
-      paddingRight: count ? normalize(8) : 0,
-      paddingBottom: normalize(3)
-    }
-  });
+const styles = StyleSheet.create({
+  container: {
+    alignItems: 'center'
+  },
+  iconWithCount: {
+    paddingRight: normalize(8)
+  },
+  iconWithoutCount: {
+    paddingBottom: normalize(3)
+  }
+});
 /* eslint-enable react-native/no-unused-styles */
--- a/src/components/widgets/DefaultWidget.tsx
+++ b/src/components/widgets/DefaultWidget.tsx
@@ -31,8 +31,6 @@ export const DefaultWidget = ({ Icon, count, onPress, text }: Props) => {
   );
 };
 
-/* eslint-disable react-native/no-unused-styles */
-/* this works properly, we do not want that warning */
 const styles = StyleSheet.create({
   container: {
     alignItems: 'center'
@@ -44,4 +42,3 @@ const styles = StyleSheet.create({
     paddingBottom: normalize(3)
   }
 });
-/* eslint-enable react-native/no-unused-styles */
--- a/src/components/widgets/DefaultWidget.tsx
+++ b/src/components/widgets/DefaultWidget.tsx
@@ -16,11 +16,11 @@ type Props = {
 export const DefaultWidget = ({ Icon, count, onPress, text }: Props) => {
   return (
     <TouchableOpacity onPress={onPress}>
-      <WrapperVertical style={styles.container}>
+      <WrapperVertical style={styles().container}>
         <WrapperRow center>
-          <Icon style={styles.icon} />
+          <Icon style={styles(!!count?.toString()).icon} />
           <BoldText primary big>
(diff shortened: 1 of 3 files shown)


src/components/widgets/EventWidget.tsx
```diff
@@ -71,7 +71,7 @@ export const EventWidget = ({ text }: WidgetProps) => {
 
   return (
     <DefaultWidget
-      count={eventCount}
+      count={additionalProps?.noCount ? null : eventCount}
```
donni106, line 74: can we make the icon horizontally centered if the count is `null`? currently it looks a bit off 
image

ardasnturk: I thought about doing that. But if there is no number, we add a blank line. If there is no blank line, it looks like this. Which is better?
Simulator Screen Shot - iPhone 13 - 2022-09-07 at 12 10 41

donni106: this looks even more off. i meant that the icon gets horizontally centered above the text, you know?

ardasnturk: i think it looks better now. see the updated picture in the PR message.

donni106: image

yes! that's it!

src/components/widgets/DefaultWidget.tsx
```diff
@@ -16,11 +16,11 @@ type Props = {
 export const DefaultWidget = ({ Icon, count, onPress, text }: Props) => {
   return (
     <TouchableOpacity onPress={onPress}>
-      <WrapperVertical style={styles.container}>
+      <WrapperVertical style={styles().container}>
         <WrapperRow center>
-          <Icon style={styles.icon} />
+          <Icon style={styles(!!count?.toString()).icon} />
```
donni106, line 21: why are you doing `toString` to pass the value?

i also think that no function for styles could be a better solution. if there is no count, you do not need to apply the icon style with padding right but always padding bottom, correct?

what about:

```
<WrapperVertical style={styles.container}>
        <WrapperRow center>
          <Icon style={!!count ? styles.iconWithCount : styles.iconWithoutCount} />
```

src/components/widgets/DefaultWidget.tsx
```diff
@@ -16,11 +16,11 @@ type Props = {
...
           <BoldText primary big>
-            {count ?? ' '}
+            {count}
```
donni106, line 23: can you think of a case where `count` is something that results in rendering a `null` or `undefined` or something on the screen? do we need to be more safe here somehow?

ardasnturk: if the `count` value is `null` or `undefined`, it will not be displayed. i don't think the application will crash because of this.

but I can still add such a control.
image

src/components/widgets/DefaultWidget.tsx
ardasnturk: This is a different solution. Both have the same end result. But in the future I will use what you say to make it more understandable. 👍

ardasnturk: If `count` is 0, this will always be false. So I had to convert it to string first. Otherwise, if count is 0, `paddingRight: 0` becomes

src/components/widgets/DefaultWidget.tsx
ardasnturk: https://github.com/ikuseiGmbH/smart-village-app-app/pull/621/commits/095c971020462743627d19c5c2fd946ea322aa22

src/components/widgets/DefaultWidget.tsx
ardasnturk: https://github.com/ikuseiGmbH/smart-village-app-app/pull/621/commits/095c971020462743627d19c5c2fd946ea322aa22

donni106: the rendered result is the same but the processed code is not. making the `StyleSheet` functional has a few downsides. linting is affected, you have to call the styles in a functional way even if you do not have to pass anything and the `StyleSheet` const is created all the time you call the function. this should be only a work around for larger problems.
in our case in this PR we want to have styles for two cases. this is not a big problem we need to solve with functional styles. please read about `StyleSheet` in a function, maybe you can find out other reasons you can share.

src/components/widgets/DefaultWidget.tsx
```diff
@@ -31,11 +31,17 @@ export const DefaultWidget = ({ Icon, count, onPress, text }: Props) => {
   );
 };
 
+/* eslint-disable react-native/no-unused-styles */
+/* this works properly, we do not want that warning */
```
donni106, line 35: we do not need this anymore without function, right?

ardasnturk: Yes, you're right, we don't need it. 👍

src/components/widgets/DefaultWidget.tsx
ardasnturk: Yes you are right, the code has been updated. 👍

donni106 (approved):

PR #631, opened 2022-09-09
feat(widget): add new water temperature widget

ardasnturk:
- added `WaterTemperatureWidget` to show
  water temperature widget on screen
- added if check to not show the widget on the 
  screen if there is no response from the server
- added `iconName` and `staticContentName`
  fields to `additionalProps` for remote control
  of widget
- added to `Widgets` to show the new widget
  in the widget area
- added `useWaterTemperature` function to
  ensure that the water temperature value comes
  from the server
- added `getWaterTemperature` request to get
  the data from the server
- added new type for `WaterTemperature`
- added `iconName` and `staticContentName`
  types to `WidgetProps` to customise widgets
  and control them remotely

SVA-658

note: to customise the widget, new fields need to be created for the widget in `globalSettings`
you can check the documentation page for this.

## How to test:
* [x] Android portrait mode
* [x] iOS portrait mode
* [x] Android landscape mode
* [x] iOS landscape mode

## Screenshots:

|widget's appearance on the home screen|`StaticContent` page to go to when clicking the widget|view of the widget on a horizontal screen|
|--|--|--|
Simulator Screen Shot - iPhone 13 - 2022-09-09 at 14 19 05 | Simulator Screen Shot - iPhone 13 - 2022-09-09 at 14 19 08 | Simulator Screen Shot - iPhone 13 - 2022-09-09 at 14 20 34

### DIFF
--- a/src/components/widgets/WaterTemperatureWidget.tsx
+++ b/src/components/widgets/WaterTemperatureWidget.tsx
@@ -1,0 +1,38 @@
+import { useNavigation } from '@react-navigation/core';
+import React, { useCallback } from 'react';
+
+import { Icon, normalize, texts } from '../../config';
+import { useHomeRefresh } from '../../hooks';
+import { useWaterTemperature } from '../../hooks/waterTemperature';
+import { QUERY_TYPES } from '../../queries';
+import { ScreenName, WidgetProps } from '../../types';
+
+import { DefaultWidget } from './DefaultWidget';
+
+export const WaterTemperatureWidget = ({ text, additionalProps }: WidgetProps) => {
+  const navigation = useNavigation();
+  const { temperature, refresh } = useWaterTemperature();
+
+  const onPress = useCallback(() => {
+    navigation.navigate(ScreenName.Html, {
+      title: text ?? texts.waterTemperature.headerTitle,
+      query: QUERY_TYPES.WATER_TEMPERATURE,
+      queryVariables: { name: additionalProps?.staticContentName ?? 'water-temperature' }
+    });
+  }, [navigation, text]);
+
+  useHomeRefresh(refresh);
+
+  if (!temperature) return null;
+
+  return (
+    <DefaultWidget
+      count={(temperature ?? 0) + '°C'}
+      Icon={() => (
+        <Icon.NamedIcon name={additionalProps?.iconName ?? 'water'} size={normalize(22)} />
+      )}
+      onPress={onPress}
+      text={text ?? texts.widgets.water}
+    />
+  );
+};
--- a/src/components/widgets/WaterTemperatureWidget.tsx
+++ b/src/components/widgets/WaterTemperatureWidget.tsx
@@ -15,7 +15,7 @@ export const WaterTemperatureWidget = ({ text, additionalProps }: WidgetProps) =
 
   const onPress = useCallback(() => {
     navigation.navigate(ScreenName.Html, {
-      title: text ?? texts.waterTemperature.headerTitle,
+      title: additionalProps?.staticContentTitle ?? texts.waterTemperature.headerTitle,
       query: QUERY_TYPES.WATER_TEMPERATURE,
       queryVariables: { name: additionalProps?.staticContentName ?? 'water-temperature' }
     });
@@ -23,11 +23,9 @@ export const WaterTemperatureWidget = ({ text, additionalProps }: WidgetProps) =
 
   useHomeRefresh(refresh);
 
-  if (!temperature) return null;
-
   return (
     <DefaultWidget
-      count={(temperature ?? 0) + '°C'}
+      count={(temperature ?? '—') + '°C'}
       Icon={() => (
         <Icon.NamedIcon name={additionalProps?.iconName ?? 'water'} size={normalize(22)} />
       )}
--- a/src/components/widgets/WaterTemperatureWidget.tsx
+++ b/src/components/widgets/WaterTemperatureWidget.tsx
@@ -17,7 +17,7 @@ export const WaterTemperatureWidget = ({ text, additionalProps }: WidgetProps) =
     navigation.navigate(ScreenName.Html, {
       title: additionalProps?.staticContentTitle ?? texts.waterTemperature.headerTitle,
       query: QUERY_TYPES.WATER_TEMPERATURE,
-      queryVariables: { name: additionalProps?.staticContentName ?? 'water-temperature' }
+      queryVariables: { name: additionalProps?.staticContentName }
     });
   }, [navigation, text]);
 
@@ -26,9 +26,7 @@ export const WaterTemperatureWidget = ({ text, additionalProps }: WidgetProps) =
   return (
     <DefaultWidget
       count={(temperature ?? '—') + '°C'}
-      Icon={() => (
-        <Icon.NamedIcon name={additionalProps?.iconName ?? 'water'} size={normalize(22)} />
-      )}
+      Icon={() => <Icon.NamedIcon name={additionalProps?.iconName} size={normalize(22)} />}
       onPress={onPress}
       text={text ?? texts.widgets.water}
     />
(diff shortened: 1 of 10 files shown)


src/queries/waterTemperature.js
```diff
@@ -0,0 +1,12 @@
+import { waterTemperatureEndpoint } from '../config';
+
+export const getWaterTemperature = async () => {
+  const fetchObj = {
+    method: 'GET',
+    headers: { Authorization: 'readonly:LjjhVGU46EBcKND9j3cX' }
```
donni106, line 6: what is this auth? is this publicly allowed? this should go to the secrets or am i wrong?

src/queries/waterTemperature.js
```diff
@@ -0,0 +1,12 @@
+import { waterTemperatureEndpoint } from '../config';
```
donni106, line 1: it seems like the export is missing for this?

src/queries/waterTemperature.js
```diff
@@ -0,0 +1,12 @@
...
+
+export const getWaterTemperature = async () => {
+  const fetchObj = {
+    method: 'GET',
+    headers: { Authorization: 'readonly:LjjhVGU46EBcKND9j3cX' }
+  };
+
+  return await fetch(waterTemperatureEndpoint, fetchObj)
+    .then((response) => response.json())
+    .catch((err) => console.error(err));
```
donni106, line 11: isn't `useQuery` handling the error cases? for volunteer and consul we also have no `then...catch`

src/types/WaterTemperature.ts
```diff
@@ -0,0 +1,3 @@
+export type WaterTemperature = {
+  temperature: { temperature: string } | undefined;
```
donni106, line 2: is this allowed to be undefined? in the `useWaterTemperature` hook it is not marked as optional per `?`. 
and i guess that this type should be `temperature?: string` or what am i missing concerning this nesting of `temperature`? when we return in the hook, it is not nested, right?

src/components/widgets/WaterTemperatureWidget.tsx
```diff
@@ -0,0 +1,38 @@
...
+
+export const WaterTemperatureWidget = ({ text, additionalProps }: WidgetProps) => {
+  const navigation = useNavigation();
+  const { temperature, refresh } = useWaterTemperature();
+
+  const onPress = useCallback(() => {
+    navigation.navigate(ScreenName.Html, {
+      title: text ?? texts.waterTemperature.headerTitle,
+      query: QUERY_TYPES.WATER_TEMPERATURE,
+      queryVariables: { name: additionalProps?.staticContentName ?? 'water-temperature' }
```
donni106, line 20: i am thinking about removing the fallback. if someone forgets to addd a static content name, then there is something missing. we cannot achieve a fallback static content with `water-temperature` for every city.

src/components/widgets/WaterTemperatureWidget.tsx
```diff
@@ -0,0 +1,38 @@
...
+    });
+  }, [navigation, text]);
+
+  useHomeRefresh(refresh);
+
+  if (!temperature) return null;
+
+  return (
+    <DefaultWidget
+      count={(temperature ?? 0) + '°C'}
```
donni106, line 30: if we return `null` earlier, there will be always a `temperature` here, so we do not need to `?? 0`, right?!

src/queries/waterTemperature.js
donni106: i cannot see any widget. maybe this is the reason?

donni106 (changes requested):

src/components/widgets/WaterTemperatureWidget.tsx
```diff
@@ -0,0 +1,38 @@
...
+
+import { DefaultWidget } from './DefaultWidget';
+
+export const WaterTemperatureWidget = ({ text, additionalProps }: WidgetProps) => {
+  const navigation = useNavigation();
+  const { temperature, refresh } = useWaterTemperature();
+
+  const onPress = useCallback(() => {
+    navigation.navigate(ScreenName.Html, {
+      title: text ?? texts.waterTemperature.headerTitle,
```
donni106, line 18: so the title will be always the `text` right, because this is always given. if there would be no `text`, there would be no widget title on the home screen. in this case, we will never have the longer page title, right? maybe we want to add a new additional prop for the `staticContentTitle`?

src/hooks/waterTemperature.ts
```diff
@@ -0,0 +1,33 @@
...
+  );
+
+  const refresh = useCallback(async () => {
+    setRefreshing(true);
+    await refetch?.();
+    setRefreshing(false);
+  }, [refetch]);
+
+  return {
+    temperature: temperature?.temperature,
```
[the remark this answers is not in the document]
ardasnturk: Yeah, you're right, I'll update.

src/queries/waterTemperature.js
ardasnturk: Yeah, you're right, I'll update. It's my fault.

src/queries/waterTemperature.js
ardasnturk: Yeah, I'll send this to you.

src/queries/waterTemperature.js
ardasnturk: Yeah, you're right, I'll update.

src/types/WaterTemperature.ts
ardasnturk: Wow. I'm gonna do it. 👍

src/components/widgets/WaterTemperatureWidget.tsx
```diff
@@ -0,0 +1,38 @@
...
+      query: QUERY_TYPES.WATER_TEMPERATURE,
+      queryVariables: { name: additionalProps?.staticContentName ?? 'water-temperature' }
+    });
+  }, [navigation, text]);
+
+  useHomeRefresh(refresh);
+
+  if (!temperature) return null;
```
donni106, line 26: i checked locally and it is confusing, that this widget is the only one loading later then the others. let's not `null` it here but use a fallback `'—'` like for the weather widget

image

```
count={(temperature ?? '—') + '°C'}
```

src/components/widgets/WaterTemperatureWidget.tsx
ardasnturk: I can add a control for that. If `staticContentName` is not added, no redirection is done.

src/components/widgets/WaterTemperatureWidget.tsx
ardasnturk: Yeah, you're right, I'll update.

src/components/widgets/WaterTemperatureWidget.tsx
ardasnturk: Yeah, that'll be better.

src/components/widgets/WaterTemperatureWidget.tsx
ardasnturk: I can change this. 👍

src/components/widgets/WaterTemperatureWidget.tsx
donni106: we do not need that. if user setup a widget, they need to setup all necessary parts. this is not created automatically or with something dynamical.

src/queries/waterTemperature.js
```diff
@@ -0,0 +1,12 @@
+import { waterTemperatureEndpoint } from '../config';
+
+export const getWaterTemperature = async () => {
+  const { serverUrl, authSecret } = waterTemperatureEndpoint;
+
+  const fetchObj = {
+    method: 'GET',
+    headers: { Authorization: `readonly:${authSecret}` }
```
donni106, line 8: i will add the whole `Authorization` as `authSecret`. There is no good reason to keep the `readonly` here and the code there.

src/types/WaterTemperature.ts
```diff
@@ -0,0 +1,3 @@
+export type WaterTemperature = {
+  temperature?: string;
+};
```
donni106, line 3: this can be removed after correction of the used types

donni106 (approved):

PR #634, opened 2022-09-19
feat(ar): add new server requests, spatial sound, video and image components

ardasnturk:
- added new data requested from the server
  -  chromaKeyFilteredVideo
  -  maxDistance
  -  minDistance
  -  physicalWidth
  -  position
  -  rolloffModel
  -  rotation
  -  scale
  -  isSpatialSound
- added new data to the array of downloaded objects `(localUris)` inside `downloadObject`
- added `?` to provide data control
- deleted undefined data with `JSON.parse` because the data coming from the server may be undefined
- deleted previously used data as `position`, `rotation` and `scale` information comes from the server
- added `chromaKeyFilteredVideo` to remove unwanted background colour in videos
- added `ViroSpatialSound` component so that the sound file can have a Spatial property
- added `ViroVideo` component to show video on stage
- Added `ViroImage` component to show an image on stage

SVA-633

### DIFF
--- a/src/components/augmentedReality/AugmentedRealityView.js
+++ b/src/components/augmentedReality/AugmentedRealityView.js
@@ -49,7 +49,7 @@ export const AugmentedRealityView = ({ sceneNavigator }) => {
     />
   );
 
-  /* see `onAnchorFound` function to auto-start animation and sound file 
+  /* see `onAnchorFound` function to auto-start animation and sound file
      when showing a 3D model that is not `TARGET` in the future time */
 
   return (
@@ -58,7 +58,8 @@ export const AugmentedRealityView = ({ sceneNavigator }) => {
 
       {object?.target ? (
         <ViroARImageMarker
-          onAnchorFound={() => setIsStartAnimationAndSound(true)} // animation and sound file are started after the image is recognised
+          // animation and sound file are started after the image is recognized
+          onAnchorFound={() => setIsStartAnimationAndSound(true)}
           pauseUpdates // prevents the model from continuous jumping
           target={TARGET}
         >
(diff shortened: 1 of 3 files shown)


src/screens/augmentedReality/ARShowScreen.js
```diff
@@ -122,9 +122,20 @@ const objectParser = async ({ item, setObject, setIsLoading, onPress }) => {
 
   item?.payload?.localUris?.forEach((item) => {
     if (item.type === 'texture') {
-      parsedObject[item.type].push({ uri: item.uri });
+      parsedObject[item?.type].push({ uri: item?.uri });
     } else {
-      parsedObject[item.type] = item.uri;
+      parsedObject[item?.type] = {
```
donni106, line 127: why did you add the `?` here for the object keys `parsedObject[item?.type]`?
think about not having an `item`, this will become undefined. so you have `parsedObject[undefined]` which makes no sense. if you want to be sure, that `item` exists in order to access `type` you need to check that explicitly before pushing or assigning.

src/helpers/augmentedReality/downloadObject.js
```diff
@@ -39,7 +39,22 @@ export const downloadObject = async ({ index, data, setData }) => {
 
       downloadedData[index].payload.downloadType = DOWNLOAD_TYPE.DOWNLOADED;
       downloadedData[index].payload.size += size;
-      downloadedData[index].payload.localUris.push({ uri, id, size, title, type });
+      downloadedData[index].payload.localUris.push({
```
donni106, line 42: does `downloadedData[index].payload.localUris` always exist to be able to push something to it? maybe an additional check before pushing would help?

src/components/augmentedReality/AugmentedRealityView.js
```diff
@@ -76,41 +71,78 @@ export const AugmentedRealityView = ({ sceneNavigator }) => {
...
+          loop
+          materials={['chromaKeyFilteredVideo']}
+          position={object?.mp4?.position}
+          rotation={object?.mp4?.rotation}
+          scale={object?.mp4?.scale}
+          source={{ uri: object?.mp4?.uri }}
+        />
+      )}
+
+      {!!object?.image && !isObjectLoading && (
```
donni106, line 124: you are checking `!isObjectLoading` every time. i would suggest to check that once before all the several objects.

src/components/augmentedReality/AugmentedRealityView.js
```diff
@@ -76,41 +71,78 @@ export const AugmentedRealityView = ({ sceneNavigator }) => {
...
     setIsObjectLoading,
     setIsStartAnimationAndSound
   } = item;
 
+  // if the `chromaKeyFilteredVideo` prop is undefined, the default colour of GreenScreen is defined
+  ViroMaterials.createMaterials({
+    chromaKeyFilteredVideo: {
+      chromaKeyFilteringColor: object?.mp4?.chromaKeyFilteredVideo || '#00FF00'
+    }
+  });
```
donni106, line 89: is this ok to be here without a condition? is this necessary in any case? or would it make sense to have a condition to add that materials just in case we have video stuff?

ardasnturk: Yes, you're right. 
https://github.com/ikuseiGmbH/smart-village-app-app/pull/634/commits/c69f385584050e19f46a35da65fddf329cc59a5c

src/components/augmentedReality/AugmentedRealityView.js
ardasnturk: Yes, you're right. 
https://github.com/ikuseiGmbH/smart-village-app-app/pull/634/commits/c69f385584050e19f46a35da65fddf329cc59a5c

src/helpers/augmentedReality/downloadObject.js
ardasnturk: Yes, you're right. 
https://github.com/ikuseiGmbH/smart-village-app-app/pull/634/commits/689faf5322a8c44097e267c3ca81f3b154d94ccd

src/screens/augmentedReality/ARShowScreen.js
ardasnturk: Yes, you're right. 
https://github.com/ikuseiGmbH/smart-village-app-app/pull/634/commits/689faf5322a8c44097e267c3ca81f3b154d94ccd

src/components/augmentedReality/AugmentedRealityView.js
```diff
@@ -76,41 +71,83 @@ export const AugmentedRealityView = ({ sceneNavigator }) => {
...
+
   return (
     <>
-      {!!object.mp3 && !isObjectLoading && (
-        <ViroSound
-          source={{ uri: object.mp3 }}
-          paused={!isStartAnimationAndSound}
-          onFinish={() => setIsStartAnimationAndSound(false)}
-        />
+      {isObjectLoading && (
```
donni106, line 95: should be `!isObjectLoading` as before, right?

donni106: you also wrote that in the commit message: `added a main control with `!isObjectLoading` because...` :)

src/helpers/augmentedReality/downloadObject.js
```diff
@@ -39,7 +39,22 @@ export const downloadObject = async ({ index, data, setData }) => {
 
       downloadedData[index].payload.downloadType = DOWNLOAD_TYPE.DOWNLOADED;
       downloadedData[index].payload.size += size;
-      downloadedData[index].payload.localUris.push({ uri, id, size, title, type });
+      downloadedData[index]?.payload?.localUris?.push({
```
donni106, line 42: there is something set for `downloadedData[index].payload.size` in line 41 (and before)without checks and then in line 42 `downloadedData[index]?.payload?.localUris?` is checked for `push`. 
this is a bit confusing because if there is no `downloadedData[index].payload` in line 41 we already have an error and checks in line 42 do not help. i will refactor this with a surrounding check i guess, before setting any new values.

src/helpers/augmentedReality/downloadObject.js
```diff
@@ -39,7 +39,22 @@ export const downloadObject = async ({ index, data, setData }) => {
...
+        type,
+        chromaKeyFilteredVideo: objectItem?.chromaKeyFilteredVideo, // HEX Color Code
+        maxDistance: parseFloat(objectItem?.maxDistance),
+        minDistance: parseFloat(objectItem?.minDistance),
+        physicalWidth: parseFloat(objectItem?.physicalWidth),
+        position: objectItem?.position, // Array [x,y,z]
+        rolloffModel: objectItem?.rolloffModel, // String (none, linear, or logarithmic)
+        rotation: objectItem?.rotation, // Array [x,y,z]
+        scale: objectItem?.scale, // Array [x,y,z]
+        isSpatialSound: objectItem?.isSpatialSound // Boolean
```
donni106, line 56: why aren't these destructed from `objectItem` as `uri`, `id`, ...?
i will change that to be just like that.

donni106: d72ecb6

src/screens/augmentedReality/ARShowScreen.js
```diff
@@ -122,9 +122,20 @@ const objectParser = async ({ item, setObject, setIsLoading, onPress }) => {
 
   item?.payload?.localUris?.forEach((item) => {
     if (item.type === 'texture') {
-      parsedObject[item.type].push({ uri: item.uri });
+      parsedObject[item.type].push({ uri: item?.uri });
```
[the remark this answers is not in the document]
donni106: d72ecb6

PR #640, opened 2022-09-23
fix(map): fix map on settings screen

ardasnturk:
- added `latitude` and `longitude` information for the application to read since the location information from `globalSettings` is `lat` and `lng`

SVA-634

### DIFF
--- a/src/components/settings/LocationSettings.js
+++ b/src/components/settings/LocationSettings.js
@@ -32,6 +32,8 @@ const useSystemPermission = () => {
 const getLocationMarker = (locationObject) => ({
   ...baseLocationMarker,
   position: {
+    latitude: locationObject?.coords?.lat,
+    longitude: locationObject?.coords?.lng,
     ...locationObject.coords
   }
 });


donni106 (approved):

src/components/settings/LocationSettings.js
```diff
@@ -32,6 +32,8 @@ const useSystemPermission = () => {
 const getLocationMarker = (locationObject) => ({
   ...baseLocationMarker,
   position: {
+    latitude: locationObject?.coords?.lat,
+    longitude: locationObject?.coords?.lng,
     ...locationObject.coords
```
donni106, line 37: this can be removed now, right?

ardasnturk: I recommend it not to be removed. There may be other features inside the object.

donni106: ok  than please change the order and put the spreading before the hardcoded keys. otherwise `latitude` and `longitude` would be overridden with values in `coords` if existing

PR #655, opened 2022-10-04
fix(volunteer): solve `useNativeDrive` warn

ardasnturk:
- added `lightboxProps` to the `MessageImage` component to remove the `useNativeDrive` warning when an image is opened in the `GiftedChat` component

SVA-767

### DIFF
--- a/src/components/Chat.js
+++ b/src/components/Chat.js
@@ -208,7 +208,12 @@ export const Chat = ({
       )}
       renderMessageImage={(props) =>
         props?.currentMessage?.image?.map(({ uri }, index) => (
-          <MessageImage {...props} key={`image-${index}`} currentMessage={{ image: uri }} />
+          <MessageImage
+            {...props}
+            key={`image-${index}`}
+            currentMessage={{ image: uri }}
+            lightboxProps={{ springConfig: { useNativeDriver: false } }}
+          />
         ))
       }
       renderMessageVideo={(props) =>


src/components/Chat.js
```diff
@@ -208,7 +208,12 @@ export const Chat = ({
       )}
       renderMessageImage={(props) =>
         props?.currentMessage?.image?.map(({ uri }, index) => (
-          <MessageImage {...props} key={`image-${index}`} currentMessage={{ image: uri }} />
+          <MessageImage
+            {...props}
+            key={`image-${index}`}
+            currentMessage={{ image: uri }}
+            lightboxProps={{ springConfig: { useNativeDriver: false } }}
```
donni106, line 215: thanks. i wonder what the spring config means. what is the difference between setting `useNativeDriver` true or false?

ardasnturk: The package where this problem occurs is called `Lightbox`. `Lightbox` uses the `springConfig` prop for animation. Actually, `tension` and `friction` properties should be added to this prop. But since it is an object, it has the ability to add more props related to animations.

If set to true, it means that we can send everything about the animation to native before the animation starts and let the native code perform the animation in the UI thread without having to bridge every frame. Once the animation has started, the JS thread can be blocked and the animation will continue to run smoothly.

donni106: And with having it set to false, it is completely done by JS? Is this the better case for us? Have you tried both variations?

ardasnturk: Yes, I tried both versions. If I set it to true, it causes a crash in the application. The animation that `Lightbox` uses probably doesn't support `useNativeDrive`. 

`Incurred error: Style property 'left' is not supported by native animated module`

donni106 (approved):

PR #661, opened 2022-10-07
feat(ar): add `tourId` to ar object in `globalSettings`

ardasnturk:
- updated id to send tourId added to `globalSettings` to `AugmentedReality` component in settings page
- changed the ar variable in `globalSettings` from a boolean expression to an object, so this value is changed to a boolean expression in `SettingsScreen` and `AugmentedReality` component
- updated settings object in `globalSettings` to be accessible from everywhere since it is used in multiple places in `SettingsScreen`

SVA-633

### DIFF
--- a/src/screens/SettingsScreen.js
+++ b/src/screens/SettingsScreen.js
@@ -64,6 +64,7 @@ const INITIAL_FILTER = [
 
 export const SettingsScreen = () => {
   const { globalSettings } = useContext(SettingsContext);
+  const { settings = {} } = globalSettings;
   const [sectionedData, setSectionedData] = useState([]);
   const [filter, setFilter] = useState(INITIAL_FILTER);
   const selectedFilterId = filter.find((entry) => entry.selected)?.id;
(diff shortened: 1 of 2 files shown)


src/screens/SettingsScreen.js
```diff
@@ -244,7 +241,7 @@ export const SettingsScreen = () => {
       )}
       {selectedFilterId === TOP_FILTER.LIST_TYPES && <ListSettings />}
       {selectedFilterId === TOP_FILTER.AR_DOWNLOAD_LIST && (
-        <AugmentedReality id="579" onSettingsScreen />
+        <AugmentedReality id={settings?.ar?.tourId} onSettingsScreen />
```
donni106, line 244: if there is `!!settings.ar` this means that `settings.ar` is accessible, so you do not need the optional chaining here. please access `settings.ar.tourId` directly

ardasnturk: Yes! You're right! 👍
https://github.com/ikuseiGmbH/smart-village-app-app/pull/661/commits/639962767920160946ddf04cb47003983ba368e2

donni106: or am i wrong? the check for `!!settings.ar` is not related to the rendering part here, right? but the selected filter is only visible, if `!!settings.ar` was true before, right? so we can be sure, that `settings.ar` is accessible here?

donni106 (approved):

donni106 (approved):

PR #669, opened 2022-10-24
feat(ar): add description to artwork detail page

ardasnturk:
- added `description` to `HtmlView` in `ArtworkDetailScreen` to show the description of the artwork
- deleted html static content because we no longer need it

SVA-565

### DIFF
--- a/src/screens/augmentedReality/ArtworkDetailScreen.js
+++ b/src/screens/augmentedReality/ArtworkDetailScreen.js
@@ -23,7 +23,7 @@ export const ArtworkDetailScreen = ({ route, navigation }) => {
   const [data, setData] = useState(route?.params?.data ?? []);
   const [isLoading, setIsLoading] = useState(true);
   const index = route?.params?.index;
-  const description = data?.[index]?.description || undefined;
+  const description = data[index]?.description;
   const { downloadType } = data[index]?.payload;
 
   useEffect(() => {
--- a/src/screens/augmentedReality/ArtworkDetailScreen.js
+++ b/src/screens/augmentedReality/ArtworkDetailScreen.js
@@ -16,27 +16,15 @@ import {
 } from '../../components';
 import { texts } from '../../config';
 import { checkDownloadedData, downloadObject, DOWNLOAD_TYPE } from '../../helpers';
-import { usePullToRefetch, useStaticContent } from '../../hooks';
 import { ScreenName } from '../../types';
 
 export const ArtworkDetailScreen = ({ route, navigation }) => {
-  const {
-    data: artworkDetail = '',
-    error,
-    loading,
-    refetch
-  } = useStaticContent({
-    type: 'html',
-    name: 'artworkDetail'
-  });
-
   const [isModalVisible, setIsModalVisible] = useState(false);
   const [data, setData] = useState(route?.params?.data ?? []);
-  const [isLoading, setIsLoading] = useState(loading);
+  const [isLoading, setIsLoading] = useState(true);
   const index = route?.params?.index;
+  const description = data?.[index]?.description || undefined;
   const { downloadType } = data[index]?.payload;
-
-  const RefreshControl = usePullToRefetch(refetch);
 
   useEffect(() => {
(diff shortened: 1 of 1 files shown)


src/screens/augmentedReality/ArtworkDetailScreen.js
```diff
@@ -16,28 +16,16 @@ import {
...
-    type: 'html',
-    name: 'artworkDetail'
-  });
-
   const [isModalVisible, setIsModalVisible] = useState(false);
   const [data, setData] = useState(route?.params?.data ?? []);
-  const [isLoading, setIsLoading] = useState(loading);
+  const [isLoading, setIsLoading] = useState(true);
   const index = route?.params?.index;
+  const description = data?.[index]?.description || undefined;
```
donni106, line 26: why have you added `|| undefined`. what do you think is the benefit?

src/screens/augmentedReality/ArtworkDetailScreen.js
```diff
@@ -71,17 +59,19 @@ export const ArtworkDetailScreen = ({ route, navigation }) => {
     return <LoadingSpinner loading />;
   }
 
-  if (error || !data.length) {
+  if (!data.length) {
```
donni106, line 62: can `data` be empty at this point? do we still need this?

donni106 (changes requested):

src/screens/augmentedReality/ArtworkDetailScreen.js
ardasnturk: We don't really need it, but isn't security always good? 😂

src/screens/augmentedReality/ArtworkDetailScreen.js
ardasnturk: I think it can be deleted.

ardasnturk: https://github.com/ikuseiGmbH/smart-village-app-app/pull/669/commits/fd0b64e1f35b55330d35672b966b746520c1c8e1

src/screens/augmentedReality/ArtworkDetailScreen.js
donni106: we try to do things we know and not hoping that something is ok to have or keep.

we are initializing data with `const [data, setData] = useState(route?.params?.data ?? []);`. so if there is no data from params it would be set to an empty array. this means checking for length makes sense.

it would be different, if we would not have the fallback to an empty array. in that case we would need to check what values data can have from all the possible params that can occur.

src/screens/augmentedReality/ArtworkDetailScreen.js
donni106: and why?

src/screens/augmentedReality/ArtworkDetailScreen.js
```diff
@@ -16,28 +16,16 @@ import {
...
-    type: 'html',
-    name: 'artworkDetail'
-  });
-
   const [isModalVisible, setIsModalVisible] = useState(false);
   const [data, setData] = useState(route?.params?.data ?? []);
-  const [isLoading, setIsLoading] = useState(loading);
+  const [isLoading, setIsLoading] = useState(true);
   const index = route?.params?.index;
+  const description = data?.[index]?.description;
```
donni106, line 26:
```
const description = data?.[index]?.description;
const { downloadType } = data[index]?.payload;
```

why are you checking the presence of data for description now? for payload this is not the case.

src/screens/augmentedReality/ArtworkDetailScreen.js
ardasnturk: I added it to avoid an error in case there is no description.

donni106: `data?.[index]?.description` this results in `undefined` if there is no description, so this would be `undefined || undefiend`, you know. that's why we do not need `|| undefined`.

donni106 (approved):

PR #672, opened 2022-10-24
feat(ar): remove description on map page

ardasnturk:
- removed the description text when shown on the map page because it was shown full screen
- changes have been made to display the `locationInfo` data carried inside the payload object on the screen

SVA-565

### DIFF
--- a/src/screens/MapViewScreen.js
+++ b/src/screens/MapViewScreen.js
@@ -2,7 +2,7 @@ import PropTypes from 'prop-types';
 import React, { useEffect, useState } from 'react';
 import { StyleSheet, View } from 'react-native';
 
-import { HtmlView, Map, RegularText, Touchable, Wrapper, WrapperRow } from '../components';
+import { Map, RegularText, Touchable, Wrapper, WrapperRow } from '../components';
 import { colors, Icon, normalize } from '../config';
 import { navigationToArtworksDetailScreen } from '../helpers';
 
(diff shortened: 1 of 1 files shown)


src/screens/MapViewScreen.js
```diff
@@ -53,14 +53,9 @@ export const MapViewScreen = ({ navigation, route }) => {
...
-                {!!modelData.locationInfo && (
-                  <RegularText small>{modelData.locationInfo}</RegularText>
-                )}
-                {!!modelData.description && (
-                  <View style={styles.marginTop}>
-                    <HtmlView html={modelData.description} tagsStyles={{ p: styles.p }} />
-                  </View>
+                {!!modelData?.title && <RegularText big>{modelData?.title}</RegularText>}
+                {!!modelData?.payload?.locationInfo && (
+                  <RegularText small>{modelData?.payload?.locationInfo}</RegularText>
```
donni106, line 58: if you check for `!!modelData?.title` and `!!modelData?.payload?.locationInfo` why did you add the `?` in the rendering part `modelData?.title` and `modelData?.payload?.locationInfo`? what do you think is the benefit?

`modelData` is already checked at line 42, so it is safe to access `title` and `locationInfo` in these lines. the changes should be reverted in my opinion, or am i wrong?

removing the description rendering should have been the only change in this commit i suggest.

donni106 (changes requested):

src/screens/MapViewScreen.js
ardasnturk: Yes! You're right!!! 👍
But I'm still adding the undefined control of the payload

https://github.com/ikuseiGmbH/smart-village-app-app/pull/672/commits/07885128b69631df11a7404fd72a2c215a9f2e37

donni106 (approved):

PR #673, opened 2022-10-26
feat(volunteer): send messages to group admins

ardasnturk:
- added `fetchGroupMembers` request to add the IDs of the group's administrators as param to the send message button
- added filter method in `fetchGroupMembers` function to collect group admins into an array
- added a button to the `VolunteerForm` screen to send a message to all admins at the same time
- added `contactGroupOwner` text to texts to show the name of the button

HDVT-88

## How to test:
* [x] Android portrait mode
* [x] iOS portrait mode
* [x] Android landscape mode
* [x] iOS landscape mode

## Screenshots:

|user not a member of the group|user a member of the group|screen to be redirected after pressing the button|
|--|--|--|
Simulator Screen Shot - iPhone 8 Plus - 2022-10-26 at 12 33 09 | Simulator Screen Shot - iPhone 8 Plus - 2022-10-26 at 12 33 44 | Simulator Screen Shot - iPhone 8 Plus - 2022-10-26 at 12 35 29


src/components/volunteer/VolunteerGroup.tsx
```diff
@@ -82,6 +88,18 @@ export const VolunteerGroup = ({
...
 
+  useEffect(() => {
+    fetchGroupMembers();
+  }, []);
+
+  // added this request to add the IDs of the group's admins as param to the send message button
+  const fetchGroupMembers = async () => {
+    const { results } = await groupMembership({ id });
+    setGroupAdmins(
+      results?.filter(({ role }: any) => role === 'admin').map(({ user: { id } }: any) => id)
```
donni106, line 99: which roles can exist? let's have a const enum object like for join policy, visibility and member status types.

src/components/volunteer/VolunteerGroup.tsx
```diff
@@ -68,6 +73,7 @@ export const VolunteerGroup = ({
   const [isGroupMember, setIsGroupMember] = useState<boolean | undefined>();
   const [isGroupOwner, setIsGroupOwner] = useState(false);
   const [isGroupApplicant, setIsGroupApplicant] = useState(false);
+  const [groupAdmins, setGroupAdmins] = useState<Array<string>>([]);
```
donni106, line 76: `string`? there will be numbers in the array, right? and the default value could be the `id` of the `owner` prop.

src/components/volunteer/VolunteerGroup.tsx
```diff
@@ -152,6 +170,22 @@ export const VolunteerGroup = ({
...
+        {(isGroupOwner || !isGroupMember) && (
+          <Wrapper>
+            <Button
+              title={texts.volunteer.contactGroupOwner}
+              onPress={() =>
+                navigation.push(ScreenName.VolunteerForm, {
+                  title: texts.volunteer.conversationAllStart,
+                  query: QUERY_TYPES.VOLUNTEER.CONVERSATION,
+                  rootRouteName: ROOT_ROUTE_NAMES.VOLUNTEER,
+                  selectedUserIds: groupAdmins
```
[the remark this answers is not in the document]
ardasnturk: If the array is empty, there will be no selected person in the dropdown list on the redirected page. I can add a control for this.

src/components/volunteer/VolunteerGroup.tsx
ardasnturk: Oh, yeah. I'll fix that. Why wouldn't the default value be an empty array?

src/components/volunteer/VolunteerGroup.tsx
```diff
@@ -152,6 +170,22 @@ export const VolunteerGroup = ({
           isSuccessRequest={isSuccessRequest}
         />
 
+        {(isGroupOwner || !isGroupMember) && (
```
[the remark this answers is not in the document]
ardasnturk: Yeah, actually, you're right. I'm going to fix this. 👍

src/components/volunteer/VolunteerGroup.tsx
donni106: because with `owner.id` we already have some admin id to provide for group admins instead of nothing. so the first render cycle already has that information. if there is a group with only one member, which is the owner in that case, we can avoid filtering and mapping to check for further group admin to set the state with the one id of that particular owner/admin.

src/components/volunteer/VolunteerGroup.tsx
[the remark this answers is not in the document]
donni106: i suggest to not have a control. a group always has an owner which is an admin. so there could never be an empty list of admins. with having the owners id set as initial state, there can never be an empty `groupAdmins` anymore.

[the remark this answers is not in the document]
ardasnturk: If we can always fill this array, of course there is no need for control.

src/components/volunteer/VolunteerGroup.tsx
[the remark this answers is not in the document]
donni106: check for `isGroupOwner` is not necessary, as a group owner is always a group member too
updated with eade8fd

src/components/volunteer/VolunteerGroup.tsx
donni106: did it eade8fd

src/components/volunteer/VolunteerGroup.tsx
donni106: eade8fd

donni106 (approved):

PR #674, opened 2022-10-28
feat(ar): add light props for viro ambient light

ardasnturk:
- added light props from server to
  `ViroAmbientLight` component
- added uri check in `deleteObject`
  because the light prop has no
  downloadable link, otherwise
  `deleteAsync` will give an error
  because there is no uri
- added uri check to `downloadObject` 
  because `fileSystemDownload` will get 
  an error because the light prop does not 
  have a uri
- added new control to `maxDistance`, 
  `minDistance` and `physicalWidth` so 
  they don't take up unnecessary space in `JSON`
- added to the parser function in `ARShowSceen`
  to parse new light props

SVA-632

### DIFF
--- a/src/helpers/augmentedReality/deleteObject.js
+++ b/src/helpers/augmentedReality/deleteObject.js
@@ -24,7 +24,9 @@ export const deleteObject = async ({ index, data, setData }) => {
       const { storageName } = storageNameCreator({ dataItem, objectItem, sceneIndex });
 
       try {
-        await FileSystem.deleteAsync(objectItem?.uri);
+        if (objectItem?.uri) {
+          await FileSystem.deleteAsync(objectItem.uri);
+        }
         await AsyncStorage.removeItem(storageName);
 
         deletedData[index].payload.downloadType = DOWNLOAD_TYPE.DOWNLOADABLE;
--- a/src/helpers/augmentedReality/downloadObject.js
+++ b/src/helpers/augmentedReality/downloadObject.js
@@ -14,7 +14,9 @@ export const downloadObject = async ({ index, data, setData }) => {
     for (const objectItem of sceneItem?.downloadableUris) {
       const {
         chromaKeyFilteredVideo,
+        color,
         id,
+        intensity,
         isSpatialSound,
         maxDistance,
         minDistance,
@@ -23,6 +25,7 @@ export const downloadObject = async ({ index, data, setData }) => {
         rolloffModel,
         rotation,
         scale,
+        temperature,
         title,
         type,
         uri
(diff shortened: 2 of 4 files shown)


src/helpers/augmentedReality/downloadObject.js
```diff
@@ -42,6 +45,21 @@ export const downloadObject = async ({ index, data, setData }) => {
       );
 
       try {
+        /*
+        since light doesn't have a downloadable url, a control 
+        has been added here to save it to JSON
+        */
+        if (type === 'light') {
+          downloadedData[index].payload.scenes[sceneIndex].localUris?.push({
```
donni106, line 53: why are thee values pushed apart from the other values at

```
downloadedData[index].payload.scenes[sceneIndex].localUris?.push({
  chromaKeyFilteredVideo, // HEX Color Code
  id,
  isSpatialSound, // Boolean
  maxDistance: parseFloat(maxDistance),
  minDistance: parseFloat(minDistance),
  physicalWidth: parseFloat(physicalWidth),
  position, // Array [x,y,z]
  rolloffModel, // String (none, linear, or logarithmic)
  rotation, // Array [x,y,z]
  scale, // Array [x,y,z]
  size,
  title,
  type,
  uri: fileSystemDownload.uri
});
```
?

what happens if `color`, `intensity` and `temperature` would be added there as well?

donni106 (changes requested):

src/helpers/augmentedReality/downloadObject.js
ardasnturk: The reason for keeping the props separate is that the prop with light type has no uri. Since there is no uri, `downloadResumable` will give an error. So it will go to catch blog before adding to `localUris` and these props will not be added.

src/helpers/augmentedReality/deleteObject.js
```diff
@@ -24,7 +24,9 @@ export const deleteObject = async ({ index, data, setData }) => {
       const { storageName } = storageNameCreator({ dataItem, objectItem, sceneIndex });
 
       try {
-        await FileSystem.deleteAsync(objectItem?.uri);
+        if (objectItem.type !== 'light') {
+          await FileSystem.deleteAsync(objectItem?.uri);
+        }
```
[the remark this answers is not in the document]
ardasnturk: Yes, you're right! a general check will be better.

src/helpers/augmentedReality/downloadObject.js
ardasnturk: https://github.com/ikuseiGmbH/smart-village-app-app/pull/674/commits/94f2c4db4032096146eea25f4c89bf566ce7a02a

src/helpers/augmentedReality/downloadObject.js
```diff
@@ -74,11 +92,10 @@ export const downloadObject = async ({ index, data, setData }) => {
           type,
           uri: fileSystemDownload.uri
         });
-
-        addToStore(storageName, downloadedData[index]);
       } catch (e) {
         console.error(e);
       }
+      addToStore(storageName, downloadedData[index]);
```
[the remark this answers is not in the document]
ardasnturk: https://github.com/ikuseiGmbH/smart-village-app-app/pull/674/commits/94f2c4db4032096146eea25f4c89bf566ce7a02a

donni106 (approved):

PR #678, opened 2022-11-01
feat(volunteer): new personal screen

donni106:
The new personal screen is working with different static contents for texts and service tiles. This way we can configure multiple things on the server more dynamically and customer can access and change these data and information as well.

HDVT-87

|personal screen|logout screen|
|---|---|
|Simulator Screen Shot - iPhone 11 - 2022-11-01 at 16 10 59|Simulator Screen Shot - iPhone 11 - 2022-11-01 at 16 11 10|

### DIFF
--- a/src/config/navigation/defaultStackConfig.tsx
+++ b/src/config/navigation/defaultStackConfig.tsx
@@ -1,3 +1,4 @@
+import { CardStyleInterpolators } from '@react-navigation/stack';
 import React from 'react';
 
 import { HeaderLeft } from '../../components';
@@ -49,6 +50,7 @@ import {
   VolunteerHomeScreen,
   VolunteerIndexScreen,
   VolunteerLoginScreen,
+  VolunteerLogoutScreen,
   VolunteerMeScreen,
   VolunteerPersonalScreen,
   VolunteerRegisteredScreen,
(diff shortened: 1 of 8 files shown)


src/config/navigation/defaultStackConfig.tsx
```diff
@@ -344,10 +346,17 @@ export const defaultStackConfig = ({
       screenComponent: VolunteerLoginScreen,
       screenOptions: { title: texts.screenTitles.volunteer.home }
     },
+    {
+      routeName: ScreenName.VolunteerLogout,
+      screenComponent: VolunteerLogoutScreen,
+      screenOptions: getScreenOptions({
+        noHeaderLeft: true,
+        cardStyleInterpolator: CardStyleInterpolators.forScaleFromCenterAndroid
+      })
```
ardasnturk, line 355: I understand that we want to use a different animation for navigation. This was built to use this animation, right?

src/navigation/screenOptions.tsx
```diff
@@ -49,11 +53,13 @@ export const getScreenOptions =
...
-        ? () => <FavoritesHeader navigation={navigation} style={styles.icon} />
-        : HeaderLeft,
+      headerLeft:
+        !noHeaderLeft &&
+        (withFavorites
+          ? () => <FavoritesHeader navigation={navigation} style={styles.icon} />
+          : HeaderLeft),
       title: route.params?.title ?? '',
-      cardStyleInterpolator: CardStyleInterpolators.forHorizontalIOS
+      cardStyleInterpolator: cardStyleInterpolator ?? CardStyleInterpolators.forHorizontalIOS
```
ardasnturk, line 62: What are the differences between this and `forScaleFromCenterAndroid` used in `defaultStackConfig`?

donni106: https://reactnavigation.org/docs/5.x/stack-navigator#cardstyleinterpolators, sadly `forScaleFromCenterAndroid` is not listed and explained there. but a new screen is not coming from the right but is scaling from the center as the name suggests and you might already checked out.

src/config/navigation/defaultStackConfig.tsx
donni106: yes, this sets a different animation just for this route

src/navigation/screenOptions.tsx
ardasnturk: Thanks 👍

ardasnturk (approved):

PR #681, opened 2022-11-01
feat(volunteer): navigate from personal to login if not logged in

donni106:
HDVT-87

### DIFF
--- a/src/screens/volunteer/VolunteerPersonalScreen.tsx
+++ b/src/screens/volunteer/VolunteerPersonalScreen.tsx
@@ -1,15 +1,17 @@
 import { useFocusEffect } from '@react-navigation/native';
 import { StackScreenProps } from '@react-navigation/stack';
 import { DeviceEventEmitter } from 'expo-modules-core';
-import React, { useCallback } from 'react';
+import React, { useCallback, useEffect } from 'react';
 import { RefreshControl, ScrollView } from 'react-native';
 
 import { LoadingSpinner, SafeAreaViewFlex, ServiceTiles } from '../../components';
 import { colors } from '../../config';
-import { useStaticContent, VOLUNTEER_PERSONAL_REFRESH_EVENT } from '../../hooks';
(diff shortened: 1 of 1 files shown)


ardasnturk (approved):

src/screens/volunteer/VolunteerPersonalScreen.tsx
```diff
@@ -30,6 +32,13 @@ export const VolunteerPersonalScreen = ({ navigation }: StackScreenProps<any>) =
 
   useFocusEffect(refreshPersonal);
 
+  useEffect(() => {
+    if (!isLoading && !isLoggedIn) {
+      navigation.goBack();
+      navigation.navigate(ScreenName.VolunteerLogin);
```
donni106, line 38: @ardasnturk have you tested that on device? how did you feel about the behavior? any other ideas realizing the correct navigation?

ardasnturk: Yes, of course I tested it on the device. It's like navigating twice and that's not nice. I think I can recommend the [HDVT-89](https://github.com/ikuseiGmbH/smart-village-app-app/pull/676) for that.

donni106: How is the back navigation handled there in the PR? i can't see it. The problem here is not to navigate when not logged in but that the navigation needs to replace the current screen while navigating to the login screen, so that navigating back will not end up on the personal screen. please think about a few minutes and maybe do some research, if this can be achieved in a better way.

ardasnturk: I tried `navigation.replace(ScreenName.VolunteerLogin)` and it creates the same effect as `goBack()`. But I still think it's better than back and forth navigation.

donni106: i think the same and changed it

PR #685, opened 2022-11-02
feat(volunteer): update the icon of the scroll down button

ardasnturk:
- updated the icon of the scroll
  down button with `ArrowDown`

HDVT-85

## Screenshots:

|before|after|
|--|--|
Simulator Screen Shot - iPhone 14 Pro Max - 2022-11-02 at 13 47 53 | Simulator Screen Shot - iPhone 14 Pro Max - 2022-11-02 at 13 47 41

### DIFF
--- a/src/components/Chat.js
+++ b/src/components/Chat.js
@@ -100,6 +100,7 @@ export const Chat = ({
       minInputToolbarHeight={normalize(96)}
       placeholder={placeholder}
       scrollToBottom
+      scrollToBottomComponent={() => <Icon.ArrowDown />}
       user={{ _id: parseInt(userId) }}
       renderActions={(props) => {
         const mediaActionSheet = {


src/components/Chat.js
```diff
@@ -100,6 +100,7 @@ export const Chat = ({
       minInputToolbarHeight={normalize(96)}
       placeholder={placeholder}
       scrollToBottom
+      scrollToBottomComponent={() => <Icon.ArrowDown />}
```
donni106, line 103: does it need to be a function or would `Icon.ArrowDown` also work?

ardasnturk: This will unfortunately cause the application to crash. A function that will generate this prop return is needed.

image.png

donni106 (approved):

PR #691, opened 2022-11-04
fix(volunteer): event creation

donni106:
- not only owners can create events but also members, so used "my groups" to fill the select list with possible groups to create an event in

HDVT-87

### DIFF
--- a/src/components/volunteer/VolunteerFormCalendar.tsx
+++ b/src/components/volunteer/VolunteerFormCalendar.tsx
@@ -8,9 +8,9 @@ import { CheckBox } from 'react-native-elements';
 import { useMutation, useQuery } from 'react-query';
 
 import { colors, consts, texts } from '../../config';
-import { isOwner, jsonParser, momentFormat, volunteerUserData } from '../../helpers';
+import { jsonParser, momentFormat } from '../../helpers';
 import { QUERY_TYPES } from '../../queries';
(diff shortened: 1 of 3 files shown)


src/components/volunteer/VolunteerFormCalendar.tsx
```diff
@@ -218,26 +217,28 @@ export const VolunteerFormCalendar = ({
...
-                  name,
-                  label: texts.volunteer.group,
-                  placeholder: texts.volunteer.group,
-                  control
-                }}
-              />
-            )}
-            control={control}
-          />
+          groupDropdownData?.length && (
```
ardasnturk, line 220: Line 188 has an if check. If it passes that if, won't we already have `groupDropdownData?.length`? Do you think this check is necessary? 

Do you need the question mark in the `groupDropdownData?.length` control? Can't we just check `groupDropdownData.length`?

donni106: yes i think we need it because its initial state is `undefined`. line 188 checks in combination with `isSuccessGroupsMy`. so if there is success of the query we do not know that `groupDropdownData` is an array or undefined, or am i wrong?

ardasnturk: That's right. Let's keep it that way.

ardasnturk (approved):

PR #692, opened 2022-11-08
feat(push notifications): exclude data providers

donni106:
- added request to call a new endpoint on switching data provider settings with passing excluded data provider ids that will be respected for push notifications
- added `experienceId` needed to properly create expo push tokens
  - therefore added `owner` in app.json, which is the account name registered at expo.dev

SVA-710

|switches|passed object|
|---|---|
SCR-20221108-q2j|image|

### DIFF
--- a/app.json
+++ b/app.json
@@ -2,6 +2,7 @@
   "expo": {
     "name": "Smart Village App",
     "slug": "smart-village-app",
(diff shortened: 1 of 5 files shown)


ardasnturk (approved):
🎉

PR #697, opened 2022-11-18
feat(noticeboard): add `subQuery` to `nestedInfoScreen`

ardasnturk:
- added `subQuery` to show redirect button under `HtmlView` in `nestedInfoScreen`
  - added `subQuery` can be used to redirect from `nestedInfoScreen` to a desired page of the application

SVA-739

## How to test:
* [x] Android portrait mode
* [x] iOS portrait mode
* [x] Android landscape mode
* [x] iOS landscape mode

## Screenshots:

|`nestedInfoScreen` without `subQuery`|`nestedInfoScreen` with `subQuery`|
|--|--|
Simulator Screen Shot - iPhone 8 Plus - 2022-11-18 at 16 00 45 | Simulator Screen Shot - iPhone 8 Plus - 2022-11-18 at 16 00 39

### DIFF
--- a/src/components/MultiButtonWithSubQuery.tsx
+++ b/src/components/MultiButtonWithSubQuery.tsx
@@ -1,0 +1,99 @@
+import { StackNavigationProp } from '@react-navigation/stack';
+import React from 'react';
+
+import { SubQuery } from '../types';
+
+import { Button } from './Button';
+
+// action to open source urls or navigate to sub screens
+export const navigateWithSubQuery = ({
+  navigation,
+  title,
+  params,
+  rootRouteName,
+  subQuery
+}: {
+  navigation: StackNavigationProp<any, string>;
+  title?: string;
+  params?: { routeName: string; webUrl: string } | string;
+  rootRouteName?: string;
+  subQuery: SubQuery;
+}) => {
+  // if the `params` is a string, it is directly the web url to call
+  if (!!params && typeof params === 'string') {
+    return navigation.navigate({
+      name: 'Web',
+      params: {
+        rootRouteName,
+        title,
+        webUrl: params
+      }
+    });
+  }
+
+  // if the `params` is an object, it contains a `routeName` and a `webUrl`
+  if (!!params && typeof params === 'object') {
+    return navigation.navigate({
+      name: params.routeName,
+      params: {
+        rootRouteName,
+        title,
+        webUrl: params.webUrl
+      }
+    });
+  }
+
+  const subParams = { ...(subQuery.params ?? {}) };
+
+  // if there is no `params`, use the main `subQuery` values for `routeName` and a `webUrl` or
+  // `params` if the params contain a webUrl as well, the webUrl property of the subQuery
+  // will be ignored
+  return navigation.navigate({
+    name: subQuery.routeName,
+    params: {
+      ...subParams,
+      rootRouteName,
+      title,
+      webUrl: subQuery.webUrl
+    }
+  });
+};
+
+export const MultiButtonWithSubQuery = ({
+  navigation,
+  rootRouteName,
+  subQuery,
+  title
+}: {
+  navigation: StackNavigationProp<any>;
+  rootRouteName?: string;
+  subQuery: SubQuery;
+  title: string;
+}) => {
+  return (
+    <>
+      {!!subQuery && !!subQuery.routeName && (!!subQuery.webUrl || subQuery.params) && (
+        <Button
+          title={subQuery.buttonTitle || `${title} öffnen`}
+          onPress={() => navigateWithSubQuery({ navigation, subQuery, rootRouteName, title })}
+        />
+      )}
+
+      {subQuery?.buttons?.map((button, index) => (
+        <Button
+          key={`${index}-${button.webUrl}`}
+          title={button.buttonTitle || `${title} öffnen`}
+          onPress={() =>
+            navigateWithSubQuery({
+              navigation,
+              params: button,
+              rootRouteName,
+              subQuery,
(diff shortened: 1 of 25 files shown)


src/screens/NestedInfoScreen.tsx
```diff
@@ -42,6 +57,20 @@ const ListHeaderComponent = ({ loading, html }: { loading: boolean; html?: strin
...
+              navigation.navigate({
+                name: subQuery.routeName,
+                params: {
+                  ...subQuery.params,
+                  title: subQuery.params?.title
+                }
+              })
+            }
+          />
+        )}
```
donni106, line 73: how come that you have added this one button here? was this possible to be copied from elsewhere in the code base to be used here?
i have compared with `HtmlScreen` where a similar situation is already implemented.
what about reusing or copying that? this would also enable multiple buttons if necessary:

```
{!!subQuery && !!subQuery.routeName && (!!subQuery.webUrl || subQuery.params) && (
  <Button
    title={subQuery.buttonTitle || `${title} öffnen`}
    onPress={() => navigate()}
  />
)}

{!!subQuery &&
  !!subQuery.buttons &&
  subQuery.buttons.map((button, index) => (
    <Button
      key={`${index}-${button.webUrl}`}
      title={button.buttonTitle || `${title} öffnen`}
      onPress={() =>
        navigate({
          routeName: button.routeName,
          webUrl: button.webUrl
        })
      }
    />
  ))}
```

src/screens/NestedInfoScreen.tsx
```diff
@@ -42,6 +57,20 @@ const ListHeaderComponent = ({ loading, html }: { loading: boolean; html?: strin
...
         <HtmlView html={html} />
+        {!!subQuery && !!subQuery.routeName && subQuery.params && (
+          <Button
+            title={subQuery.buttonTitle}
+            onPress={() =>
+              navigation.navigate({
+                name: subQuery.routeName,
+                params: {
+                  ...subQuery.params,
+                  title: subQuery.params?.title
```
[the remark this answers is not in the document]
ardasnturk: i missed here. `params: subQuery.params` is more likely.

src/screens/NestedInfoScreen.tsx
ardasnturk: I worked with only one button logic because we don't need more buttons, but I can change that.

Yes, it is the same as in `HtmlScreen`, but navigation is not returned from a function.

src/screens/NestedInfoScreen.tsx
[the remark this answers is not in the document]
donni106: why have you decided to destruct the params of subquery here now instead of setting `params: subQuery.params` as is suggested? what do you think is the benefit?

src/screens/NestedInfoScreen.tsx
donni106: with more buttons, we can realize one button. with only implementing one button we are not prepared for requests to have more buttons. as we already have the code for more buttons, let's use it here as well please.

src/screens/NestedInfoScreen.tsx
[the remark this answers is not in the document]
ardasnturk: The code has been edited.
https://github.com/ikuseiGmbH/smart-village-app-app/commit/13f58939e0a9c13d144000a6c7285c32e8d4a385

src/screens/NestedInfoScreen.tsx
ardasnturk: i will create a component to avoid code repetition.

ardasnturk: a new component has been created to prevent code repetition. New `MultiButtonWithSubQuery` component has been added to `HtmlScreen` and `NestedInfoScreen`

https://github.com/ikuseiGmbH/smart-village-app-app/pull/697/commits/28605c4223fcf6a7db415de28d7ac01fbc86ee38

src/screens/NestedInfoScreen.tsx
```diff
@@ -53,9 +68,14 @@ export const NestedInfoScreen = ({ navigation, route }: StackScreenProps<any>) =
   const [refreshing, setRefreshing] = useState(false);
   const name = route.params?.name;
   const rootRouteName = route.params?.rootRouteName;
+  const subQuery = route.params?.subQuery ?? undefined;
```
donni106, line 71: we do not need to fallback to `undefined`. if there is no `subQuery` it's undefined. i will update that.

src/components/MultiButtonWithSubQuery.tsx
```diff
@@ -0,0 +1,106 @@
...
+
+import { Button } from './Button';
+
+export type TSubQueryType = {
+  buttons?: [{ buttonTitle: string; routeName: string; webUrl: string }];
+  buttonTitle: string;
+  params: { rootRouteName: string; routeName: string; title: string; webUrl: string };
+  routeName: string;
+  webUrl?: string;
+};
```
donni106, line 12: this should go to types as separate fie.
importing a type from components seems to be misplaced.

image

i will refactor that.

src/components/MultiButtonWithSubQuery.tsx
```diff
@@ -0,0 +1,106 @@
+import { StackNavigationProp } from '@react-navigation/stack';
+import React from 'react';
+
+import { Button } from './Button';
+
+export type TSubQueryType = {
```
donni106, line 6: "T" already means Type...so T...Type is a duplication. i will update while moving to types folder.

donni106 (approved):

PR #698, opened 2022-11-18
feat(noticeboard): add `noticeboardIndexScreen`

ardasnturk:
- added index screen for `noticeboard` `genericItem`
  - `noticeboardIndexScreen` differs from `nestedInfoScreen` in that it has a `genericItem` list with query
- added the necessary information for the navigation structure to `defaultStackConfig` and `Navigation`
- added `EmptyMessage` component to show on screen if there is no data in `noticeboardIndexScreen`

SVA-739

## How to test:
* [x] Android portrait mode
* [x] iOS portrait mode
* [x] Android landscape mode
* [x] iOS landscape mode

## Screenshots:

|Noticeboard Offers Screen|Noticeboard Searches Screen|Noticeboard NeighbourlyHelp Screen|
|--|--|--|
Simulator Screen Shot - iPhone 8 Plus - 2022-11-18 at 16 05 39 | Simulator Screen Shot - iPhone 8 Plus - 2022-11-18 at 16 05 51 | Simulator Screen Shot - iPhone 8 Plus - 2022-11-18 at 16 05 56

### DIFF
--- a/src/screens/NoticeboardIndexScreen.tsx
+++ b/src/screens/NoticeboardIndexScreen.tsx
@@ -77,10 +77,11 @@ export const NoticeboardIndexScreen = ({ navigation, route }: StackScreenProps<a
         }
         ListHeaderComponent={
           <ListHeaderComponent
-            loading={loadingHtml}
             html={dataHtml}
-            subQuery={subQuery}
+            loading={loadingHtml}
             navigation={navigation}
+            navigationTitle=""
+            subQuery={subQuery}
           />
         }
       />
--- a/src/screens/NoticeboardIndexScreen.tsx
+++ b/src/screens/NoticeboardIndexScreen.tsx
@@ -1,0 +1,99 @@
+import { StackScreenProps } from '@react-navigation/stack';
+import React, { useCallback, useContext, useState } from 'react';
+import { useQuery } from 'react-apollo';
+import { ActivityIndicator, RefreshControl, SectionList } from 'react-native';
+
+import { EmptyMessage, LoadingContainer, SafeAreaViewFlex, SectionHeader } from '../components';
+import { colors, texts } from '../config';
+import { graphqlFetchPolicy, parseListItemsFromQuery } from '../helpers';
+import { useStaticContent } from '../hooks';
+import { useRenderItem } from '../hooks/listHooks';
+import { NetworkContext } from '../NetworkProvider';
+import { getQuery, QUERY_TYPES } from '../queries';
+
+import { ListHeaderComponent } from './NestedInfoScreen';
+
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+/* eslint-disable complexity */
+export const NoticeboardIndexScreen = ({ navigation, route }: StackScreenProps<any>) => {
+  const { isConnected, isMainserverUp } = useContext(NetworkContext);
+  const fetchPolicy = graphqlFetchPolicy({ isConnected, isMainserverUp });
+  const [refreshing, setRefreshing] = useState(false);
+
+  const consentForDataProcessingText = route.params?.consentForDataProcessingText ?? '';
+  const content = route.params?.content ?? '';
+  const query = route.params?.query ?? '';
+  const queryVariables = route.params?.queryVariables ?? {};
+  const subQuery = route.params?.subQuery ?? '';
+  const title = route.params?.title ?? '';
+
+  const { data, loading, refetch } = useQuery(getQuery(query), {
+    fetchPolicy,
+    variables: queryVariables
+  });
+  const listItems = parseListItemsFromQuery(query, data, consentForDataProcessingText, {
+    queryVariables
+  });
+
+  const {
+    data: dataHtml,
+    loading: loadingHtml,
+    refetch: refetchHtml
+  } = useStaticContent<string>({
+    name: content,
+    type: 'html',
+    skip: !content
+  });
+
+  const onRefresh = useCallback(async () => {
+    setRefreshing(true);
+    if (isConnected) {
+      await refetch?.();
+      await refetchHtml?.();
+    }
+    setRefreshing(false);
+  }, [isConnected, refetch]);
+
+  const renderItem = useRenderItem(QUERY_TYPES.PUBLIC_JSON_FILE, navigation);
+
+  if (loading && !data)
+    return (
+      <LoadingContainer>
+        <ActivityIndicator color={colors.accent} />
+      </LoadingContainer>
+    );
+
+  if (!data) {
+    return <EmptyMessage title={texts.noticeboard.emptyTitle} />;
+  }
+
+  const sectionData = [{ title: title, data: listItems }];
+
+  return (
+    <SafeAreaViewFlex>
+      <SectionList
+        refreshControl={
+          <RefreshControl
+            refreshing={refreshing}
+            onRefresh={onRefresh}
+            colors={[colors.accent]}
+            tintColor={colors.accent}
+          />
+        }
+        ListHeaderComponent={
+          <ListHeaderComponent
+            loading={loadingHtml}
+            html={dataHtml}
+            subQuery={subQuery}
+            navigation={navigation}
+          />
+        }
+        sections={sectionData}
+        renderSectionHeader={({ section: { title } }) => <SectionHeader title={title} />}
+        renderItem={renderItem}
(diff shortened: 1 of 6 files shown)


src/screens/NoticeboardIndexScreen.tsx
```diff
@@ -0,0 +1,99 @@
...
+
+  if (!data) {
+    return <EmptyMessage title={texts.noticeboard.emptyTitle} />;
+  }
+
+  const sectionData = [{ title: title, data: listItems }];
+
+  return (
+    <SafeAreaViewFlex>
+      <SectionList
```
donni106, line 74: why is this a section list? what are the sections? there are no section or am i wrong? we have a list with three entries but no sections?

src/screens/NoticeboardIndexScreen.tsx
```diff
@@ -0,0 +1,99 @@
...
+        }
+        sections={sectionData}
+        renderSectionHeader={({ section: { title } }) => <SectionHeader title={title} />}
+        renderItem={renderItem}
+        keyExtractor={(item) => item.title + item.id}
+      />
+    </SafeAreaViewFlex>
+  );
+};
+/* eslint-enable complexity */
```
donni106, line 99: image

the three headlines should not contain "anzeigen" as this makes user think to be pressable like a button.
i think we can entirely get rid of the headline for the list. the main headline is already present on the screens below the image, like on the initial sketches, right?!

image

src/screens/NoticeboardIndexScreen.tsx
```diff
@@ -0,0 +1,99 @@
...
+            loading={loadingHtml}
+            html={dataHtml}
+            subQuery={subQuery}
+            navigation={navigation}
+          />
+        }
+        sections={sectionData}
+        renderSectionHeader={({ section: { title } }) => <SectionHeader title={title} />}
+        renderItem={renderItem}
+        keyExtractor={(item) => item.title + item.id}
```
donni106, line 94: how come that you have decided to add the id here to the title?

src/screens/NoticeboardIndexScreen.tsx
```diff
@@ -0,0 +1,99 @@
...
+import { ListHeaderComponent } from './NestedInfoScreen';
+
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+/* eslint-disable complexity */
+export const NoticeboardIndexScreen = ({ navigation, route }: StackScreenProps<any>) => {
+  const { isConnected, isMainserverUp } = useContext(NetworkContext);
+  const fetchPolicy = graphqlFetchPolicy({ isConnected, isMainserverUp });
+  const [refreshing, setRefreshing] = useState(false);
+
+  const consentForDataProcessingText = route.params?.consentForDataProcessingText ?? '';
```
donni106, line 23: what does this `consentForDataProcessingText` stand for and where will it be rendered?

src/screens/NoticeboardIndexScreen.tsx
```diff
@@ -0,0 +1,99 @@
...
+        ListHeaderComponent={
+          <ListHeaderComponent
+            loading={loadingHtml}
+            html={dataHtml}
+            subQuery={subQuery}
+            navigation={navigation}
+          />
+        }
+        sections={sectionData}
+        renderSectionHeader={({ section: { title } }) => <SectionHeader title={title} />}
```
donni106, line 92: there is a title already defined in the upper scope with `const title = route.params?.title ?? '';`. this was confusing me when reading. in `NestedInfoScreen` it was handled differently with `const navigationTitle = route.params?.title;`. why have you decided to change that?

src/screens/NoticeboardIndexScreen.tsx
ardasnturk: Yes, it looks like this in the drafts, but I used section list because I didn't like the look after I removed `SectionList`. I'll change it.

src/screens/NoticeboardIndexScreen.tsx
ardasnturk: I felt the need to add id to avoid key error in case of the same title.

> Warning: Encountered two children with the same key, `%s`. Keys should be unique so that components maintain their identity across updates. Non-unique keys may cause children to be duplicated and/or omitted — the behavior is unsupported and could change in a future version.%s, .$0=2Searches,

src/screens/NoticeboardIndexScreen.tsx
ardasnturk: It is the text that the customer wants on the `Checkbox` we use for data processing permission. This will be displayed on the form.

`consentForDataProcessingText` is sent as the `titleDetail` prop in the `parseListItemsFromQuery` function and displayed on the Form page.

src/screens/NoticeboardIndexScreen.tsx
ardasnturk: There is no extra reason for me to change this. There is just a difference between how it is used in `NestedInfoScreen` and how it is used in `NoticeboardIndexScreen`.

In `NestedInfoScreen` the `NavigationTitle` is used in the `children.map()` function inside the `sectionData` and is added as a param for the `Header` on the next screen.

In `NoticeboardScreen` it is only used as the `title` of the `SectionList`. The title used for the `Header` comes from the `parseListItemFromQuery` function.

src/screens/NoticeboardIndexScreen.tsx
ardasnturk: I can remove `SectionList`. Yes, there is no section, but I don't like the look when I don't use `SectionList`.

src/screens/NoticeboardIndexScreen.tsx
donni106: but that is a problem of the individual data and not the code. having the same title on a service screen makes not much sense, right? and we would need to change other code places as well, where we have that, right?

src/screens/NoticeboardIndexScreen.tsx
donni106: is it possible to request the text directly on the form screen? why should it be handled on screens, where we do not need it?

src/screens/NoticeboardIndexScreen.tsx
donni106: this is what i saw from the code, yes. but you still have the problem of two `title` definitions.

src/screens/NoticeboardIndexScreen.tsx
donni106: yes please, we do not have sections here

ardasnturk: I think everything's all right now.

https://github.com/ikuseiGmbH/smart-village-app-app/pull/698/commits/00a19f8f9ba995dcfbf34843f7404fcef1a7816f

Simulator Screen Shot - iPhone 8 Plus - 2022-11-21 at 12 44 13

src/screens/NoticeboardIndexScreen.tsx
ardasnturk: https://github.com/ikuseiGmbH/smart-village-app-app/pull/698/commits/00a19f8f9ba995dcfbf34843f7404fcef1a7816f

Simulator Screen Shot - iPhone 8 Plus - 2022-11-21 at 12 44 13

src/screens/NoticeboardIndexScreen.tsx
ardasnturk: I don't think we have that problem now.

https://github.com/ikuseiGmbH/smart-village-app-app/pull/698/commits/00a19f8f9ba995dcfbf34843f7404fcef1a7816f

src/screens/NoticeboardIndexScreen.tsx
ardasnturk: Yes, that's right. This problem disappeared when I deleted `SectionList`.

https://github.com/ikuseiGmbH/smart-village-app-app/pull/698/commits/00a19f8f9ba995dcfbf34843f7404fcef1a7816f

src/screens/NoticeboardIndexScreen.tsx
ardasnturk: I don't think this is possible. The data here comes as param from the `noticeboardMainContent` `staticContent`. So we only have this data on this page. But if you want I can create a new `HtmlContent` just for this `consentForDataProcessingText`.

PR #701, opened 2022-11-21
feat(price card): show 'kostenlos' instead of nothing

donni106:
- there can be free entries, which we were not showing because it resulted in `false`

SVA-822

|before|after|
|---|---|
|Simulator Screen Shot - iPhone 12 mini - 2022-11-21 at 11 59 39|simulator_screenshot_FE2E8DF8-3BBC-44A8-872C-007904303AB0|


ardasnturk (approved):
It works well. Thank you 🎉

PR #702, opened 2022-11-29
feat(ar): multiple scene logic

ardasnturk:
- added `multipleSceneIndexGenerator` to calculate texture and model indexes for a model containing more than one scene
- added moment-range package to measure day difference
- added stable key to `downloadObject` helper to save stable textures coming from server
- added selecting the required model from the scenes array with the index information from the `multipleSceneIndexGenerator` function
- added the required texture data to the newly created `variableTextures` array after the `modalIndex` and `textureIndex` data have the number type and added to the `parsedObject.textures` array
- renamed the `texture` array to `textures` for clarity
- changed the `multipleSceneIndexGenerator` function to `void` format as there is no need for `async/await` function
- filtering the texture files in a single loop because two different loops are too many
- added to `parsedObject.textures` in case there is an item in the `variableTextures` array

SVA-565

⚠️ use the following procedure to test the function ⚠️

- download the [dummyData.js](https://github.com/ikuseiGmbH/smart-village-app-app/files/10115149/dummyData.js.zip) file to test.
- add `tourStops: dummyData` as a prop inside the `AugmentedReality` component in `Tour.js`
-  add the following code blog inside the `parsedObject` inside `ARShowScreen`
```
console.warn({
    modelIndex,
    textureIndex,
    variableTextures: variableTextures[textureIndex || 0] || [],
    parsedObject: JSON.parse(JSON.stringify(parsedObject))
  });
```
- this code blog will only give you an idea whether the function is working correctly or not. Since there is no `vrx` file inside the `dummyData`, the `AugmentedReality` object will not be visible.

### DIFF
--- a/src/screens/augmentedReality/ARShowScreen.js
+++ b/src/screens/augmentedReality/ARShowScreen.js
@@ -149,19 +149,36 @@ const multipleSceneIndexGenerator = async ({ startDate, timePeriodInDays, scenes
   });
 
 const objectParser = async ({ item, setObject, setIsLoading, onPress }) => {
-  const parsedObject = { texture: [] };
+  const parsedObject = { textures: [] };
+  const variableTextures = [];
 
-  // TODO: in the future the index 0 used here will be changed according to time logic!
-  const localUris = item?.payload?.scenes?.[0]?.localUris;
+  const { modalIndex, textureIndex } = await multipleSceneIndexGenerator({
+    startDate: item?.payload?.startDate,
+    timePeriodInDays: item?.payload?.timePeriodInDays,
+    scenes: item?.payload?.scenes
+  });
+
+  const localUris =
+    item?.payload?.scenes?.[modalIndex]?.localUris || item?.payload?.scenes?.[0]?.localUris;
+
+  localUris.filter(
+    ({ stable, uri, type }) => stable && type === 'texture' && parsedObject.textures.push({ uri })
+  );
+
+  // combine stable textures with variable textures
+  if (typeof modalIndex === 'number' && typeof textureIndex === 'number') {
+    localUris.filter(
+      ({ stable, uri, type }) => !stable && type === 'texture' && variableTextures.push({ uri })
+    );
+    parsedObject.textures.push(variableTextures[textureIndex || 0] || []);
+  }
 
   if (localUris?.animationName) {
     parsedObject.animationName = localUris?.animationName;
   }
 
   localUris?.forEach((item) => {
-    if (item.type === 'texture') {
-      parsedObject[item.type]?.push({ uri: item?.uri });
-    } else {
+    if (item.type !== 'texture') {
       parsedObject[item.type] = {
         chromaKeyFilteredVideo: item?.chromaKeyFilteredVideo,
         color: item?.color,
@@ -177,10 +194,12 @@ const objectParser = async ({ item, setObject, setIsLoading, onPress }) => {
         temperature: item?.temperature,
         uri: item?.uri
       };
+    } else {
+      return;
     }
   });
 
-  if (!parsedObject?.texture?.length || !parsedObject?.vrx) {
+  if (!parsedObject?.textures?.length || !parsedObject?.vrx) {
     return Alert.alert(
       texts.augmentedReality.modalHiddenAlertTitle,
       texts.augmentedReality.invalidModelError,
--- a/src/screens/augmentedReality/ARShowScreen.js
+++ b/src/screens/augmentedReality/ARShowScreen.js
@@ -1,10 +1,14 @@
 import { ViroARSceneNavigator } from '@viro-community/react-viro';
+import moment from 'moment';
+import { extendMoment } from 'moment-range';
 import PropTypes from 'prop-types';
 import React, { useCallback, useEffect, useRef, useState } from 'react';
 import { Alert, Animated, StyleSheet, TouchableOpacity, View } from 'react-native';
 
 import { AugmentedRealityView, LoadingSpinner } from '../../components';
 import { colors, Icon, normalize, texts } from '../../config';
+
+const extendedMoment = extendMoment(moment);
 
 export const ARShowScreen = ({ navigation, route }) => {
   const [isLoading, setIsLoading] = useState(true);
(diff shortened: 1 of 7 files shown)


src/screens/augmentedReality/ARShowScreen.js
```diff
@@ -141,10 +194,12 @@ const objectParser = async ({ item, setObject, setIsLoading, onPress }) => {
         temperature: item?.temperature,
         uri: item?.uri
       };
+    } else {
+      return;
```
donni106, line 198: what is this return useful for?

src/screens/augmentedReality/ARShowScreen.js
```diff
@@ -112,20 +116,69 @@ export const ARShowScreen = ({ navigation, route }) => {
...
+/**
+ * index creation function for model and texture
+ *
+ * @param {string} startDate           start date of the model
+ * @param {number} timePeriodInDays    time period in days of the model
+ * @param {array}  scenes              array of models
+ *
+ * @return {object} both parsed values as an object, like { modalIndex: 1, textureIndex: 1 }
+ */
+const multipleSceneIndexGenerator = async ({ startDate, timePeriodInDays, scenes }) =>
```
donni106, line 128: please move everything but the component and styles into other files. the screen is too bloated.
and as i already mentioned once, having logics under the component in the same file is bad practice.

src/screens/augmentedReality/ARShowScreen.js
```diff
@@ -112,20 +116,69 @@ export const ARShowScreen = ({ navigation, route }) => {
...
+    item?.payload?.scenes?.[modalIndex]?.localUris || item?.payload?.scenes?.[0]?.localUris;
+
+  localUris.filter(
+    ({ stable, uri, type }) => stable && type === 'texture' && parsedObject.textures.push({ uri })
+  );
 
-  // TODO: in the future the index 0 used here will be changed according to time logic!
-  const localUris = item?.payload?.scenes?.[0]?.localUris;
+  // combine stable textures with variable textures
+  if (typeof modalIndex === 'number' && typeof textureIndex === 'number') {
```
donni106, line 169: why are you checking `typeof`s here? what is the case if the types are different? can they be string boolean or anything else?

src/screens/augmentedReality/ARShowScreen.js
```diff
@@ -112,20 +116,69 @@ export const ARShowScreen = ({ navigation, route }) => {
...
+    localUris.filter(
+      ({ stable, uri, type }) => !stable && type === 'texture' && variableTextures.push({ uri })
+    );
+    parsedObject.textures.push(variableTextures[textureIndex || 0] || []);
```
donni106, line 173: what does `variableTextures[textureIndex || 0] || []` stand for. can you give it a name and make it a const for example?

src/screens/augmentedReality/ARShowScreen.js
```diff
@@ -112,20 +116,69 @@ export const ARShowScreen = ({ navigation, route }) => {
...
+  localUris.filter(
+    ({ stable, uri, type }) => stable && type === 'texture' && parsedObject.textures.push({ uri })
+  );
 
-  // TODO: in the future the index 0 used here will be changed according to time logic!
-  const localUris = item?.payload?.scenes?.[0]?.localUris;
+  // combine stable textures with variable textures
+  if (typeof modalIndex === 'number' && typeof textureIndex === 'number') {
+    localUris.filter(
+      ({ stable, uri, type }) => !stable && type === 'texture' && variableTextures.push({ uri })
```
donni106, line 171: you are first filtering `localUris` for stable entries pushing in parsed objects, right?
and then looping over `localUris` to filter for not stable entries and push in variables textures. why?
are there two loops necessary? and what is the difference between `parsedObjects` and `variableTextures`?

src/screens/augmentedReality/ARShowScreen.js
```diff
@@ -112,20 +116,69 @@ export const ARShowScreen = ({ navigation, route }) => {
...
+ *
+ * @param {string} startDate           start date of the model
+ * @param {number} timePeriodInDays    time period in days of the model
+ * @param {array}  scenes              array of models
+ *
+ * @return {object} both parsed values as an object, like { modalIndex: 1, textureIndex: 1 }
+ */
+const multipleSceneIndexGenerator = async ({ startDate, timePeriodInDays, scenes }) =>
+  new Promise((resolve) => {
+    let modalIndex, textureIndex;
```
donni106, line 130: do you mean "model" with "modal"? a modal is like a window/popup in tech speech: https://en.wikipedia.org/wiki/Modal_window

src/screens/augmentedReality/ARShowScreen.js
```diff
@@ -112,20 +116,69 @@ export const ARShowScreen = ({ navigation, route }) => {
...
+ * index creation function for model and texture
+ *
+ * @param {string} startDate           start date of the model
+ * @param {number} timePeriodInDays    time period in days of the model
+ * @param {array}  scenes              array of models
+ *
+ * @return {object} both parsed values as an object, like { modalIndex: 1, textureIndex: 1 }
+ */
+const multipleSceneIndexGenerator = async ({ startDate, timePeriodInDays, scenes }) =>
+  new Promise((resolve) => {
```
donni106, line 129: why are you using a promise here and async function? i cannot find any async requirements the code mathematics?

src/screens/augmentedReality/ARShowScreen.js
```diff
@@ -112,20 +116,69 @@ export const ARShowScreen = ({ navigation, route }) => {
...
+    let modalIndex, textureIndex;
+
+    /* all models must have the same number of variable textures. Therefore, in order to obtain the 
+       number of textures, the textures of the first model were calculated. */
+    if (scenes.length > 1) {
+      const today = new Date(),
+        differenceDate = extendedMoment.range(startDate, today).diff('days'),
+        textureCount = scenes[0]?.localUris?.filter(
+          ({ type, stable }) => !stable && type === 'texture'
```
donni106, line 138: why are you checking the type here? can there be different types we do not want to consider?

donni106 (changes requested):

src/screens/augmentedReality/ARShowScreen.js
ardasnturk: Yeah, you're right, I will. 👍

src/screens/augmentedReality/ARShowScreen.js
ardasnturk: Data of type undefined or number can come here. If the data is of type number, I can create an array for variable textures and add it to the main texture array.

src/screens/augmentedReality/ARShowScreen.js
ardasnturk: Yes I can. 👍

`variableTextures[textureIndex || 0] || []`
- If there is a `textureIndex` number from the `multipleSceneIndexGenerator` function, the directory where this number is located is selected.
- Otherwise, the data at index 0 is added.
- If there is none, empty array is added.

In this way, a control is provided for variable textures.

src/screens/augmentedReality/ARShowScreen.js
ardasnturk: I think I can do it in one loop.

There is not only texture data in `parsedObjects`. All the data required for `AugmentedReality` is in this object. Therefore, fixed texture files are added in `parsebObjects.textures`.

If there are variable texture files, they are separated later and added to `parsedObjects.textures`.

src/screens/augmentedReality/ARShowScreen.js
ardasnturk: Yeah, I'll change that. 👍

src/screens/augmentedReality/ARShowScreen.js
ardasnturk: After running the function, I realised that the application continued without the information I needed from the function and the data was not arriving properly.
So I felt the need to use `async/await`.

src/screens/augmentedReality/ARShowScreen.js
ardasnturk: To get the index number I need the number of variable textures a model has.
I don't think I need the other types.

src/screens/augmentedReality/ARShowScreen.js
ardasnturk: I guess it's not doing anything. I'm already checking above that the type is a texture.

donni106 (approved):
let's get the 🎉  started

PR #703, opened 2022-11-29
feat: use data resource settings for toggling url links/buttons

donni106:
SVA-764

|without setting|with setting|
|---|---|
|Simulator Screen Shot - iPhone 12 mini - 2022-11-29 at 17 41 28|Simulator Screen Shot - iPhone 12 mini - 2022-11-29 at 17 41 40|

### DIFF
--- a/src/components/screens/EventRecord.js
+++ b/src/components/screens/EventRecord.js
@@ -11,12 +11,13 @@ import { Logo } from '../Logo';
 import { Title, TitleContainer, TitleShadow } from '../Title';
 import { Touchable } from '../Touchable';
 import { Wrapper, WrapperHorizontal, WrapperWithOrientation } from '../Wrapper';
-import { matomoTrackingString, trimNewLines } from '../../helpers';
+import { matomoTrackingString, openLink, trimNewLines } from '../../helpers';
 import { useMatomoTrackScreenView, useOpenWebScreen } from '../../hooks';
 import { DataProviderNotice } from '../DataProviderNotice';
 import { ImageSection } from '../ImageSection';
 import { InfoCard } from '../infoCard';
 import { DataProviderButton } from '../DataProviderButton';
+import { Button } from '../Button';
 
 import { OperatingCompany } from './OperatingCompany';
 import { OpeningTimesCard } from './OpeningTimesCard';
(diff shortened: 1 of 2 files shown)


src/components/screens/EventRecord.js
```diff
@@ -182,6 +184,22 @@ export const EventRecord = ({ data, route }) => {
           title={texts.eventRecord.operatingCompany}
         />
 
+        {settings?.displayOnlySummary === 'true' && !!settings?.onlySummaryLinkText && (
```
ardasnturk, line 187: Is it possible to replace "true" with another variable? Why is the string value coming?

src/components/screens/EventRecord.js
```diff
@@ -182,6 +184,22 @@ export const EventRecord = ({ data, route }) => {
...
+              const { url } = item;
+
+              return (
+                <Button
+                  key={index}
+                  title={settings.onlySummaryLinkText}
+                  onPress={() => openLink(url, openWebScreen)}
+                />
+              );
+            })}
```
ardasnturk, line 199: Here we only use the `url` information from the item and there is no control inside the `Button` component. How about using it as `webUrls.map(({ url }, index) => {...}` ?
for example:

```
{webUrls.map(({ url }, index) => (
  <Button
    key={index}
    title={settings.onlySummaryLinkText}
    onPress={() => openLink(url, openWebScreen)}
  />
))}
```

ardasnturk (commented):
Please give me more information so I can try.

src/components/screens/EventRecord.js
donni106: i don't understand what you mean with replacing true?
`displayOnlySummary` comes as `'true'` or `'false'` per GraphQL and i don't know why it is not boolean.

src/components/screens/EventRecord.js
donni106: good idea! i like the simplicity about this.

src/components/screens/EventRecord.js
ardasnturk: Okay, I got my answer. Only `'true'` or `'false'` string values are coming. 👍

src/components/screens/EventRecord.js
donni106: 8987358

ardasnturk (approved):

PR #704, opened 2022-12-02
feat(ar): make the sound infinite loop

ardasnturk:
- added `loop` prop to `ViroSound` and `ViroSpatialSound` component for unlimited looping
- deleted the `onFinish` function in the `ViroSound` and `ViroSpatialSound` component to prevent the audio file from being stopped after finishing

SVA-565


donni106 (approved):
🔉 👂

PR #706, opened 2022-12-06
feature(ar): show multiple models in one scene

ardasnturk:
- added multiple models and textures from the `scene` array to the `models` and `textures` arrays in order to show the models using the map function on the `AugmentedRealityView` screen
- deleted objects according to their types with `_remove` method to prevent unnecessary texture and model information
- creating `variableTextures` based on index 0 of the scene and we were constantly getting variable textures of the model at index 0. To avoid this, `variableTextures` has been updated with `modelIndex`
- deleted the direct return of `localUris` in case `variableTexturesCount` is `false` because we need `models` and `textures` arrays inside `localUris`
- added `models` array in `parsedObject` to parse one or more models
- changed `parsedObject?.vrx` control to `parsedObject?.models?.length` as it can no longer be done
- added to arrays in `parsedObject` since all `models` and `textures` are contained in an array
- added `Viro3DObject` to the map function in `AugmentedRealityView` to show all models inside the models array
- removed extra controls on `AugmentedRealityView` page to improve performance and because they have already been checked

SVA-565

📝 note: with this code, one or more models can be displayed in a scene at the same time. 

‼️ attention! displaying more than one model at the same time may cause RAM problems. ‼️

## How to test:
* [x] Android portrait mode
* [x] iOS portrait mode

⚠️ use the following procedure to test the function ⚠️

- download the [dummyData.js](https://github.com/ikuseiGmbH/smart-village-app-app/files/10166934/dummyData.js.zip) file to test.
- extract `dummyData` from zip
- add dummyData into a folder of your choice
- add `tourStops: dummyData` as a prop inside the `AugmentedReality` component in `Tour.js`
- go to any `InternationalerKunstwanderweg` screen from the App
- you can download the model you want from a list of all possible `scenes` in different situations and try it out

### DIFF
--- a/src/components/augmentedReality/AugmentedRealityView.js
+++ b/src/components/augmentedReality/AugmentedRealityView.js
@@ -93,7 +93,7 @@ const ViroSoundAnd3DObject = (item) => {
     // if the `chromaKeyFilteredVideo` prop is undefined, the default color of a green screen is set
     ViroMaterials.createMaterials({
       chromaKeyFilteredVideo: {
-        chromaKeyFilteringColor: object?.mp4?.chromaKeyFilteredVideo || '#00FF00'
+        chromaKeyFilteringColor: object.mp4?.chromaKeyFilteredVideo || '#00FF00'
       }
     });
   }
@@ -103,19 +103,19 @@ const ViroSoundAnd3DObject = (item) => {
       {!isObjectLoading && (
         <>
           {!!object?.mp3 &&
-            (object?.mp3?.isSpatialSound ? (
+            (object.mp3?.isSpatialSound ? (
               <ViroSpatialSound
-                source={{ uri: object?.mp3?.uri }}
+                source={{ uri: object.mp3?.uri }}
                 paused={!isStartAnimationAndSound}
                 loop
-                maxDistance={object?.mp3?.maxDistance}
-                minDistance={object?.mp3?.minDistance}
-                position={object?.mp3?.position}
-                rolloffModel={object?.mp3?.rolloffModel}
+                maxDistance={object.mp3?.maxDistance}
+                minDistance={object.mp3?.minDistance}
+                position={object.mp3?.position}
+                rolloffModel={object.mp3?.rolloffModel}
               />
             ) : (
               <ViroSound
-                source={{ uri: object?.mp3?.uri }}
+                source={{ uri: object.mp3?.uri }}
                 paused={!isStartAnimationAndSound}
                 loop
               />
@@ -125,26 +125,26 @@ const ViroSoundAnd3DObject = (item) => {
             <ViroVideo
               loop
               materials={['chromaKeyFilteredVideo']}
-              position={object?.mp4?.position}
-              rotation={object?.mp4?.rotation}
-              scale={object?.mp4?.scale}
-              source={{ uri: object?.mp4?.uri }}
+              position={object.mp4?.position}
+              rotation={object.mp4?.rotation}
+              scale={object.mp4?.scale}
+              source={{ uri: object.mp4?.uri }}
             />
           )}
 
           {!!object?.image && (
             <ViroImage
-              position={object?.image?.position}
-              rotation={object?.image?.rotation}
-              scale={object?.image?.scale}
-              source={{ uri: object?.image?.uri }}
+              position={object.image?.position}
+              rotation={object.image?.rotation}
+              scale={object.image?.scale}
+              source={{ uri: object.image?.uri }}
             />
           )}
         </>
       )}
 
-      {!!object?.models &&
-        !!object?.textures &&
+      {!!object?.models?.length &&
+        !!object?.textures?.length &&
         object.models.map((model, index) => (
           <Viro3DObject
             key={index}
--- a/src/components/augmentedReality/AugmentedRealityView.js
+++ b/src/components/augmentedReality/AugmentedRealityView.js
@@ -143,22 +143,27 @@ const ViroSoundAnd3DObject = (item) => {
         </>
       )}
 
-      <Viro3DObject
-        source={{ uri: object?.vrx?.uri }}
-        resources={object?.textures}
-        type="VRX"
-        position={object?.vrx?.position}
-        rotation={object?.vrx?.rotation}
-        scale={object?.vrx?.scale}
-        onLoadStart={() => setIsObjectLoading(true)}
-        onLoadEnd={() => setIsObjectLoading(false)}
-        onError={() => alert(texts.augmentedReality.arShowScreen.objectLoadErrorAlert)}
-        animation={{
-          loop: true,
-          name: object?.animationName,
-          run: isStartAnimationAndSound
-        }}
-      />
+      {!!object?.models &&
+        !!object?.textures &&
+        object.models.map((model, index) => (
+          <Viro3DObject
+            key={index}
+            source={{ uri: model.uri }}
+            resources={object.textures}
+            type="VRX"
+            position={model.position}
+            rotation={model.rotation}
+            scale={model.scale}
+            onLoadStart={() => setIsObjectLoading(true)}
+            onLoadEnd={() => setIsObjectLoading(false)}
+            onError={() => alert(texts.augmentedReality.arShowScreen.objectLoadErrorAlert)}
+            animation={{
+              loop: true,
+              name: object?.animationName,
+              run: isStartAnimationAndSound
+            }}
+          />
+        ))}
     </>
   );
 };
--- a/src/helpers/augmentedReality/multipleSceneIndexGenerator.ts
+++ b/src/helpers/augmentedReality/multipleSceneIndexGenerator.ts
@@ -55,6 +55,18 @@ export const multipleSceneIndexGenerator = ({
     );
     const variableTexturesCount = variableTextures?.length;
 
+    if (!variableTexturesCount) {
+      // nothing to do if there are no unstable textures
+      // keep original localUris with modelIndex = 0
+      localUris.forEach((subItem) => {
+        modelsAndTexturesArrayGenerator(subItem, models, textures);
+      });
+
+      _remove(localUris, ({ type }) => type === 'texture' || type === 'vrx');
+
+      return { localUris, models, textures };
+    }
+
     const today = new Date();
     // the current day number is the number of running days since the given start date
     const currentDayNumber = extendedMoment.range(startDate, today).diff('days');
@@ -62,8 +74,8 @@ export const multipleSceneIndexGenerator = ({
     const model = Math.floor(texture / variableTexturesCount);
     modelIndex = model % scenesCount;
 
-    // When VariableTextures are created, they are created according to index 0 of the scene array.
-    // Once the modelIndex is found, it must be updated to select the correct texture file.
+    // When `variableTextures` are created, they are created according to index 0 of the scene array.
+    // Once the `modelIndex` is found, it must be updated to select the correct texture file.
     if (modelIndex > 0) {
       variableTextures = scenes[modelIndex]?.localUris?.filter(
         ({ type, stable }) => type === 'texture' && !stable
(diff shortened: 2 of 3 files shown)


src/components/augmentedReality/AugmentedRealityView.js
```diff
@@ -105,17 +105,17 @@ const ViroSoundAnd3DObject = (item) => {
           {!!object?.mp3 &&
             (object?.mp3?.isSpatialSound ? (
```
donni106, line 106: remove the `?` here as well for `object.mp3?.isSpatialSound`?

src/helpers/augmentedReality/multipleSceneIndexGenerator.ts
```diff
@@ -24,36 +42,73 @@ export const multipleSceneIndexGenerator = ({
...
       ({ type, stable }) => type === 'texture' && !stable
     );
     const variableTexturesCount = variableTextures?.length;
 
-    if (!variableTexturesCount) {
-      // nothing to do if there are no unstable textures
-      // keep original localUris with modelIndex = 0
-      return { localUris };
-    }
-
```
donni106, line 43: what if `variableTexturesCount` is zero now?

src/helpers/augmentedReality/multipleSceneIndexGenerator.ts
```diff
@@ -24,36 +42,73 @@ export const multipleSceneIndexGenerator = ({
...
-    const variableTexture = variableTextures?.[textureIndex];
+
+    // When VariableTextures are created, they are created according to index 0 of the scene array.
+    // Once the modelIndex is found, it must be updated to select the correct texture file.
+    variableTextures = scenes[modelIndex]?.localUris?.filter(
+      ({ type, stable }) => type === 'texture' && !stable
+    );
     localUris = scenes?.[modelIndex]?.localUris;
+    const textureIndex = texture % variableTexturesCount;
+    const variableTexture = variableTextures[textureIndex];
```
donni106, line 72: what if there is something wrong with `modelIndex` and no `variableTextures`?
`const variableTexture = variableTextures[textureIndex];` could crash.

src/helpers/augmentedReality/multipleSceneIndexGenerator.ts
```diff
@@ -24,36 +42,73 @@ export const multipleSceneIndexGenerator = ({
...
     const model = Math.floor(texture / variableTexturesCount);
     modelIndex = model % scenesCount;
-    const textureIndex = texture % variableTexturesCount;
-    const variableTexture = variableTextures?.[textureIndex];
+
+    // When VariableTextures are created, they are created according to index 0 of the scene array.
+    // Once the modelIndex is found, it must be updated to select the correct texture file.
+    variableTextures = scenes[modelIndex]?.localUris?.filter(
+      ({ type, stable }) => type === 'texture' && !stable
+    );
```
donni106, line 69: if `modelIndex` is 0 `variableTextures` is the same as already created in line 53, so we would not need to process with filter again.

src/helpers/augmentedReality/multipleSceneIndexGenerator.ts
```diff
@@ -24,36 +42,73 @@ export const multipleSceneIndexGenerator = ({
...
+            textures.push(item as TTextures[0]);
+            break;
+          default:
+            break;
+        }
+      });
+    });
 
-    _remove(localUris, ({ type, stable }) => type === 'texture' && !stable);
+    _remove(localUris, ({ type }) => type === 'texture' || type === 'vrx');
```
donni106, line 109: why is vrx removed and than pushed again?

src/helpers/augmentedReality/multipleSceneIndexGenerator.ts
```diff
@@ -24,36 +42,73 @@ export const multipleSceneIndexGenerator = ({
...
+          break;
+        default:
+          break;
+      }
+    });
+
+    _remove(localUris, ({ type }) => type === 'texture' || type === 'vrx');
+
+    localUris.push({ models, textures });
+  } else {
```
donni106, line 93: there is a lot repetition in the if and in the else body. can this be cleanup to be mnore readable?

src/components/augmentedReality/AugmentedRealityView.js
```diff
@@ -125,40 +125,45 @@ const ViroSoundAnd3DObject = (item) => {
...
-        onLoadEnd={() => setIsObjectLoading(false)}
-        onError={() => alert(texts.augmentedReality.arShowScreen.objectLoadErrorAlert)}
-        animation={{
-          loop: true,
-          name: object?.animationName,
-          run: isStartAnimationAndSound
-        }}
-      />
+      {!!object?.models &&
+        !!object?.textures &&
```
donni106, line 147: `?.length` for both to ensure entries in these arrays?

src/components/augmentedReality/AugmentedRealityView.js
```diff
@@ -125,40 +125,45 @@ const ViroSoundAnd3DObject = (item) => {
...
+            type="VRX"
+            position={model.position}
+            rotation={model.rotation}
+            scale={model.scale}
+            onLoadStart={() => setIsObjectLoading(true)}
+            onLoadEnd={() => setIsObjectLoading(false)}
+            onError={() => alert(texts.augmentedReality.arShowScreen.objectLoadErrorAlert)}
+            animation={{
+              loop: true,
+              name: object?.animationName,
```
donni106, line 162: animation name stays directly at `object`?

donni106 (changes requested):

src/helpers/augmentedReality/multipleSceneIndexGenerator.ts
ardasnturk: I think we're gonna run into error right now. Now we need models and textures arrays inside localUris. I'll think of a new logic for this.

src/helpers/augmentedReality/multipleSceneIndexGenerator.ts
ardasnturk: would direct `variableTextures = scenes[modelIndex]?.localUris` work?

src/helpers/augmentedReality/multipleSceneIndexGenerator.ts
ardasnturk: I need to extract the model and textures from the `localUris` array. Otherwise, objects with `vrx` and `texture` keys will be created unnecessarily in `objectParser`.

src/helpers/augmentedReality/multipleSceneIndexGenerator.ts
donni106: i think you want all local uris for a specific model index filtered by `!stable`. but not if model index is 0. 
is that right? try to translate that into code to run the filter only for model indexes greater than 0.

src/components/augmentedReality/AugmentedRealityView.js
ardasnturk: Yes, you are right 👍

src/helpers/augmentedReality/multipleSceneIndexGenerator.ts
ardasnturk: I'll update it in a readable format. 👍

src/components/augmentedReality/AugmentedRealityView.js
ardasnturk: Yes, it is directly inside the object. But in the last update, when updating `localUris`, we forgot to update here. So I'm updating it immediately.

Actually `animationName` information is in the `payload`. Right now we want `localUris.animationName`. But `localUris` is an array and there is no `animationName` information in it. Now I wrote `payload` instead of `localUris` and the problem is fixed.

https://github.com/ikuseiGmbH/smart-village-app-app/pull/706/commits/958199c5725ff2f7d22e0199b00113d37c3c744e

src/helpers/augmentedReality/multipleSceneIndexGenerator.ts
donni106: > But how can I add models or textures array inside the localUris array?

try to explain what you want to have exactly. and then translate it in code. i can help you with that, but i do not know yet what the startpoint is, wat the transformation should be.
something like `localUris.models = localUris.filter .... type == vrx...`?

ardasnturk: As we discussed in Slack, the `models` and `textures` arrays returned

https://github.com/ikuseiGmbH/smart-village-app-app/pull/706/commits/baf73b38379a872911ca84651baab6e508211dcf

src/helpers/augmentedReality/multipleSceneIndexGenerator.ts
```diff
@@ -24,21 +42,29 @@ export const multipleSceneIndexGenerator = ({
...
+  const models: TModels = [];
+  const textures: TTextures = [];
 
   // if we have multiple scenes, we want to calculate the model and texture based on the current day
   // compared to the `startDate`
   if (scenesCount > 1 && startDate && timePeriodInDays) {
     // all models must have the same number of variable textures. therefore, in order to obtain the
     // number of textures, the textures of the first model were calculated.
-    const variableTextures = scenes[0]?.localUris?.filter(
+    let variableTextures = scenes[modelIndex]?.localUris?.filter(
```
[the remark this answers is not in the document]
donni106: ah ok, `modelIndex` is 0 here because of the initial value `let modelIndex = 0;`

src/helpers/augmentedReality/multipleSceneIndexGenerator.ts
```diff
@@ -47,14 +73,58 @@ export const multipleSceneIndexGenerator = ({
...
+        modelsAndTexturesArrayGenerator(subItem, models, textures, true);
+      });
+      break;
+    } else {
+      localUris = scenes?.[i]?.localUris;
+      localUris.forEach((subItem) => {
+        modelsAndTexturesArrayGenerator(subItem, models, textures);
+      });
+    }
+  }
```
donni106, line 104: i think this can be optimized to reduce repetition and conditions:

```
for (let i = 0; i < scenes.length; i++) {
  if (scenesCount > 1 && !startDate && !timePeriodInDays) {
    localUris = scenes?.[i]?.localUris;
  }

  localUris.forEach((subItem) => {
    modelsAndTexturesArrayGenerator(subItem, models, textures, subItem.stable);
  });
}
```

src/helpers/augmentedReality/multipleSceneIndexGenerator.ts
```diff
@@ -47,14 +73,58 @@ export const multipleSceneIndexGenerator = ({
...
+  stableTextureLogic?: boolean
+) => {
+  switch (subItem.type) {
+    case 'vrx':
+      models.push(subItem as TModels[0]);
+      break;
+    case 'texture':
+      stableTextureLogic
+        ? subItem.stable && textures.push(subItem as TTextures[0])
+        : textures.push(subItem as TTextures[0]);
```
donni106, line 125: and then we can get rid of the duplicated `subItem.stable` condition here maybe?

```
stableTextureLogic ? textures.push(subItem as TTextures[0]) : textures.push(subItem as TTextures[0]);
```

but äääh if else are the same.
sowhat should be the logic behind this condition exactly?

src/helpers/augmentedReality/multipleSceneIndexGenerator.ts
```diff
@@ -47,14 +73,58 @@ export const multipleSceneIndexGenerator = ({
...
-    !!variableTexture && localUris.push(variableTexture);
+    localUris = scenes?.[modelIndex]?.localUris;
+    const textureIndex = texture % variableTexturesCount;
+    textures.push(variableTextures?.[textureIndex] as TTextures[0]);
   }
 
-  return { localUris };
+  for (let i = 0; i < scenes.length; i++) {
+    if (scenesCount > 1 && startDate && timePeriodInDays) {
+      // INFO: localUris = scenes?.[modelIndex]?.localUris;
```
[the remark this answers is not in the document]
ardasnturk: I think I can move the `for` loop here into the if above. I can return directly from there.

src/helpers/augmentedReality/multipleSceneIndexGenerator.ts
ardasnturk: It's absolutely the right perspective.

src/helpers/augmentedReality/multipleSceneIndexGenerator.ts
ardasnturk: only the `stable` variable is inserted into the `textures` array.

The condition you wrote does not do this, but if you look at the code in detail, a new texture is added to the `textures` array only if `subItem.stable` is true.

donni106: so `stableTextureLogic && textures.push(subItem as TTextures[0])`? but then there is a missing push in which case?

ardasnturk: If there is only 1 stable texture then I only need to send 1 texture into the `textures` array.
But if there are no stable textures then I have to send all textures into the `textures` array.
Therefore I need to do 2 checks. Is there a stable texture and is the texture I send to the `textures` array stable or not?

src/helpers/augmentedReality/multipleSceneIndexGenerator.ts
ardasnturk: I split the if controls inside the for loop here and I think the code is more understandable now.

https://github.com/ikuseiGmbH/smart-village-app-app/pull/706/commits/ab256e92e32672bea38b55043e6cd5d8f428f5ca

src/helpers/augmentedReality/multipleSceneIndexGenerator.ts
ardasnturk: I'm sorry, but I guess it doesn't get any better than this. Of course, I'm always open to new ideas.😁

https://github.com/ikuseiGmbH/smart-village-app-app/pull/706/commits/ab256e92e32672bea38b55043e6cd5d8f428f5ca

donni106: the question here is to push a texture or not, right? so what about: `shouldPushTexture && textures.push(subItem as TTextures[0]);` with `shouldPushTexture?: boolean` instead of `stableTextureLogic?: boolean`.

if there are no stable textures, `shouldPushTexture` should be true, right? so this could be the default `shouldPushTexture = true`.

this would result in the following staying the same:

```
...
for (let i = 0; i < scenes.length; i++) {
  localUris = scenes?.[i]?.localUris;
  localUris.forEach((subItem) => {
    modelsAndTexturesArrayGenerator(subItem, models, textures);
  });
}
...
```

and in the other cases, first setting to false:

```
...
if (!variableTexturesCount) {
  // nothing to do if there are no unstable textures
  // keep original localUris with modelIndex = 0
  localUris.forEach((subItem) => {
    modelsAndTexturesArrayGenerator(subItem, models, textures, false);
  });

  _remove(localUris, ({ type }) => type === 'texture' || type === 'vrx');

  return { localUris, models, textures };
}
...
```

and passing theboolean of `subItem.stable`:

```
...
const textureIndex = texture % variableTexturesCount;
textures.push(variableTextures?.[textureIndex] as TTextures[0]);

localUris.forEach((subItem) => {
  modelsAndTexturesArrayGenerator(subItem, models, textures, !!subItem.stable);
});
...
```

does this handle all cases as expected?

ardasnturk: I think I know what you mean. Yes, it makes perfect sense. I tested it and it works fine. I'm looking forward to your tests.

https://github.com/ikuseiGmbH/smart-village-app-app/commit/4f204b4ec2dafbf5b67de593447fbb0fdf920f60

src/helpers/augmentedReality/multipleSceneIndexGenerator.ts
```diff
@@ -47,14 +73,56 @@ export const multipleSceneIndexGenerator = ({
...
+  _remove(localUris, ({ type }) => type === 'texture' || type === 'vrx');
+
+  return { localUris, models, textures };
+};
+
+const modelsAndTexturesArrayGenerator = (
+  subItem: TScenes['localUris'][0],
+  models: TModels,
+  textures: TTextures,
+  shouldPushTexture: boolean = true
```
donni106, line 116: image

`shouldPushTexture = true` should  be enough

donni106 (approved):
🥇

src/helpers/augmentedReality/multipleSceneIndexGenerator.ts
ardasnturk: 👍
https://github.com/ikuseiGmbH/smart-village-app-app/commit/d7995b7f7ee9213263362f5fdb7ab6b04cf89d0d

PR #709, opened 2022-12-15
feat(defect report): new form with image upload

donni106:
- created new feature form to report defects
- refactored `ImageSelector` with a `isConsul` in addition to `isVolunteer` and added a default for our main server
- refactored location settings abort button to be the same as in the new screen being less significant

you need to add a new service icon in homeService for example
```
{
  "title": "Mängelmelder",
  "routeName": "Html",
  "iconName": "warning-outline",
  "params": {
    "title": "Mängelmelder",
    "rootRouteName": "Mängelmelder",
    "query": "publicHtmlFile",
    "queryVariables": {
      "name": "defectreport"
    },
    "subQuery": {
      "routeName": "DefectReportForm",
      "buttonTitle": "Mängel melden",
      "params": {
        "rootRouteName": "Mängelmelder",
        "consentForDataProcessingText": "Maecenas sed diam eget risus varius blandit sit amet non magna. Aenean lacinia bibendum nulla sed consectetur. Nullam quis risus eget urna mollis ornare vel eu leo"
      }
    }
  }
}
```

you also need to create a new category with some sub categories.

image

the main category id needs to be put in global settings:

image

SVA-738

Simulator Screen Shot - iPhone 12 mini - 2022-12-15 at 15 15 29-tile

### DIFF
--- a/src/components/consul/selectors/ImageSelector.js
+++ b/src/components/consul/selectors/ImageSelector.js
@@ -244,8 +244,8 @@ ImageSelector.propTypes = {
 
 const styles = StyleSheet.create({
   image: {
-    height: imageHeight(imageWidth() * 0.8),
-    width: imageWidth() * 0.8
+    height: imageHeight(imageWidth() * 0.6),
+    width: imageWidth() * 0.6
   },
   volunteerContainer: {
     marginBottom: normalize(8)
--- a/src/components/consul/selectors/ImageSelector.js
+++ b/src/components/consul/selectors/ImageSelector.js
@@ -42,7 +42,7 @@ const deleteImageAlert = (onPress) =>
     ]
   );
 
-export const ImageSelector = ({ control, field, imageId, isVolunteer, item }) => {
+export const ImageSelector = ({ control, field, imageId, isConsul, isVolunteer, item }) => {
   const { buttonTitle, infoText } = item;
   const { name, onChange, value } = field;
 
@@ -65,6 +65,19 @@ export const ImageSelector = ({ control, field, imageId, isVolunteer, item }) =>
   }, [imagesAttributes]);
 
   const onDeleteImage = async (index) => {
+    if (isConsul) {
+      if (imageId) {
+        try {
+          await deleteImage({ variables: { id: imageId } });
+        } catch (err) {
+          console.error(err);
+        }
+      }
+
+      onChange('');
+      setInfoAndErrorText({});
+    }
+
     if (isVolunteer) {
       const isURL = URL_REGEX.test(imagesAttributes[index].uri);
 
@@ -81,16 +94,8 @@ export const ImageSelector = ({ control, field, imageId, isVolunteer, item }) =>
       return;
     }
 
-    if (imageId) {
-      try {
-        await deleteImage({ variables: { id: imageId } });
-      } catch (err) {
-        console.error(err);
-      }
-    }
-
-    onChange('');
-    setInfoAndErrorText({});
+    setImagesAttributes(deleteArrayItem(imagesAttributes, index));
+    setInfoAndErrorText(deleteArrayItem(infoAndErrorText, index));
   };
 
   const imageSelect = async () => {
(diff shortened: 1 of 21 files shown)


donni106:
Category selection added

Simulator Screen Shot - iPhone 12 mini - 2022-12-15 at 15 55 53

src/components/defectReport/DefectReportLocationForm.tsx
```diff
@@ -0,0 +1,115 @@
...
+    if (
+      shouldGetPosition &&
+      !loadingPosition &&
+      !loadingLastKnownPosition &&
+      (position || lastKnownPosition)
+    ) {
+      setIsLocationSelect(false);
+      setSelectedPosition((position || lastKnownPosition)?.coords);
+    }
+  }, [lastKnownPosition, loadingPosition, loadingLastKnownPosition, position, shouldGetPosition]);
```
ardasnturk, line 53: If the location permissions in the system are turned off, the location value is `undefined`. 

Is it supposed to be like this? Or should we ask the user to grant location permission?

src/components/defectReport/DefectReportCreateForm.tsx
```diff
@@ -0,0 +1,309 @@
...
+                  name: 'images',
+                  label: texts.volunteer.images,
+                  buttonTitle: texts.volunteer.addImage
+                }
+              }}
+            />
+          )}
+          control={control}
+        />
+      </Wrapper>
```
ardasnturk, line 208: After selecting an image, I get a screen like this. The delete image button is not visible on the screen and is not centred exactly.

IMG_D5142462639D-1

src/components/defectReport/DefectReportCreateForm.tsx
```diff
@@ -0,0 +1,309 @@
...
+              phone: defectReportNewData.phone
+            }
+          ],
+          contentBlocks: [{ body: defectReportNewData.body, title: defectReportNewData.title }],
+          addresses: [
+            {
+              geoLocation: {
+                latitude: selectedPosition?.latitude,
+                longitude: selectedPosition?.longitude
+              }
```
ardasnturk, line 113: If I press the `Auf der Karte auswählen` button and then press the `weiter` button without selecting a new position on the map, the `selectedPosition` is `undefined`.

image

ardasnturk (commented):
I must say it works well. I think it will be much better after a little refactor.

src/components/defectReport/DefectReportCreateForm.tsx
donni106: ok nice catch, we need a condition for wider screens

src/components/defectReport/DefectReportCreateForm.tsx
donni106: very good case, i will have a look and try to set a fallback with the pin we see on the map

src/components/consul/selectors/ImageSelector.js
```diff
@@ -205,14 +237,15 @@ ImageSelector.propTypes = {
...
   isVolunteer: PropTypes.bool,
   item: PropTypes.object
 };
 
 const styles = StyleSheet.create({
   image: {
-    height: imageHeight(imageWidth() * 0.6),
-    width: imageWidth() * 0.6
+    height: imageHeight(imageWidth() * 0.8),
+    width: imageWidth() * 0.8
```
donni106, line 248: i will revert this to result in 

image

src/components/defectReport/DefectReportLocationForm.tsx
ardasnturk: I think it's better now.

src/components/defectReport/DefectReportCreateForm.tsx
ardasnturk: Very good 👍

src/components/consul/selectors/ImageSelector.js
ardasnturk: At least the delete button can be pressed. 😂👍

ardasnturk (approved):

PR #712, opened 2022-12-20
fix(ar): add height to modal

ardasnturk:
- added height to modal because modal cannot be switched off on small screen phones if there are too many models

SVA-565

## Screenshots:

|before|after|
|--|--|
IMG_0098 | IMG_0100

### DIFF
--- a/src/components/augmentedReality/ARModal.js
+++ b/src/components/augmentedReality/ARModal.js
@@ -47,6 +47,7 @@ export const ARModal = ({
   return (
     <Modal
       isListView={isListView}
+      height={isListView && '85%'}
       onModalVisible={() => {
         if (onModalVisible) {
           onModalVisible();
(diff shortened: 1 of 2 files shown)


src/components/augmentedReality/ARModal.js
```diff
@@ -47,6 +47,7 @@ export const ARModal = ({
   return (
     <Modal
       isListView={isListView}
+      height={isListView && '70%'}
```
donni106, line 50: we talked about 80% and i think it would be good to have a bit more space, as we  will have many items in the list. maybe 85% also works out?

ardasnturk: Yeah, you're right. I tried 85% and it looks good. I'm updating to 85%.
B1F85860-71D7-45D0-9130-05F520E2D4A0

donni106 (approved):

PR #713, opened 2022-12-21
fix(ar): click on map marker issue fixed

ardasnturk:
- fixed the problem that the marker could not be pressed again after returning from the `ArtworkDetailScreen` page after clicking on the marker on the map
- added `navigationToArtworksDetailScreen` function to `onMarkerPress`

SVA-565

### DIFF
--- a/src/screens/augmentedReality/ArtworkDetailScreen.js
+++ b/src/screens/augmentedReality/ArtworkDetailScreen.js
@@ -25,6 +25,13 @@ export const ArtworkDetailScreen = ({ route, navigation }) => {
   const index = route?.params?.index;
   const description = data[index]?.description;
   const { downloadType } = data[index]?.payload;
+  const setModelId = route?.params?.setModelId;
+
+  useEffect(() => {
+    if (setModelId) {
+      setModelId(null);
+    }
+  }, []);
 
   useEffect(() => {
     if (data?.length) {
--- a/src/screens/augmentedReality/ArtworkDetailScreen.js
+++ b/src/screens/augmentedReality/ArtworkDetailScreen.js
@@ -25,13 +25,6 @@ export const ArtworkDetailScreen = ({ route, navigation }) => {
   const index = route?.params?.index;
   const description = data[index]?.description;
   const { downloadType } = data[index]?.payload;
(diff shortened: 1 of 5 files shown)


src/screens/augmentedReality/ArtworkDetailScreen.js
```diff
@@ -25,6 +25,13 @@ export const ArtworkDetailScreen = ({ route, navigation }) => {
   const index = route?.params?.index;
   const description = data[index]?.description;
   const { downloadType } = data[index]?.payload;
+  const setModelId = route?.params?.setModelId;
+
+  useEffect(() => {
+    if (setModelId) {
+      setModelId(null);
```
donni106, line 32: why do we need to reset the model id? what do we need the model id for in which cases?

ardasnturk: `modelId` is a data that updates itself only when you click on the marker on the map.

For example, if we clicked on the model with id 1 before, we cannot click on the model with id 1 again. Because modelId does not update itself. Therefore, we need to reset the modelId each time or update it with another value.

donni106: but if it is the 1 already and we click again, then the 1 is still ok to navigate to, or am i wrong?
where is the code for the marker click? i would like to have a look and think about it for a few minutes.

ardasnturk: In `AugmentedReality.js`, routing is done with `useEffect` in line 70. The `useEffect` here works according to the `modelId`. In `navigationToArtworksDetailScreen`, navigation is directed to the `ArtworkDetailScreen`.

ardasnturk: I think it's much better now.

https://github.com/ikuseiGmbH/smart-village-app-app/pull/713/commits/d5ee2b6511e16e3ef6200fa6b6a4f1a96928a69a

donni106: nice, it looks so much clearer now

donni106 (approved):

PR #721, opened 2023-01-03
feat(ar): add `ViroSpotLight` and `ViroQuad` to scene

ardasnturk:
- added `ViroSpotLight` component which is required to show a shadow effect under the model
- added `ViroQuad` to create a virtual floor
- added the `arShadowReceiver` prop to `ViroQuad`, which is required for the shadow to mask the model
- added `shadowCastingBitMask` prop to `Viro3DObject` component to show the shadow of the model on the ground
- added new properties to `downloadObject.js` to be able to add the new information needed for `ViroSpotLight` and ViroQuad` to `localUris` to show the shadow
- edited `objectParser.js` to move newly added features to `AugmentedRealityView` screen
- removed the extra checks in the `AugmentedRealityView` screen, as we have already checked that the `object` contains the required information

SVA-565

## How to test:
* [x] Android portrait mode
* [x] iOS portrait mode

## Screenshots:

|before|after|
|--|--|
AugmentedReality_1672767027545 | AugmentedReality_1672766989042

### DIFF
--- a/src/components/augmentedReality/AugmentedRealityView.js
+++ b/src/components/augmentedReality/AugmentedRealityView.js
@@ -6,8 +6,10 @@ import {
   ViroARTrackingTargets,
   ViroImage,
   ViroMaterials,
+  ViroQuad,
   ViroSound,
   ViroSpatialSound,
+  ViroSpotLight,
   ViroVideo
 } from '@viro-community/react-viro';
 import PropTypes from 'prop-types';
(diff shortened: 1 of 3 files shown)


src/components/augmentedReality/AugmentedRealityView.js
```diff
@@ -157,13 +159,36 @@ const ViroSoundAnd3DObject = (item) => {
...
+        shadowOpacity={0.6}
+      />
+
+      <ViroQuad
+        arShadowReceiver
+        height={1000}
+        width={1000}
+        lightReceivingBitMask={2}
+        position={[0, 0, 0]}
+        rotation={[-90, 0, 0]}
```
donni106, line 190: which props do we want editable from the CMS? `height`, `width`, `position`?

ardasnturk: Only the `position` feature should be adjustable. 👍

ardasnturk: Do we plan to add a setting for the degree of `SpotLight` as we discussed yesterday?

donni106: we wanted to ask the artists first on Monday, right?

donni106: new fields are done in the cms

src/helpers/augmentedReality/downloadObject.js
```diff
@@ -71,22 +79,29 @@ export const downloadObject = async ({ index, data, setData }) => {
...
+          outerAngle: outerAngle && parseInt(outerAngle), // Number
           physicalWidth: physicalWidth && parseFloat(physicalWidth),
           position, // Array [x,y,z]
           rolloffModel, // String (none, linear, or logarithmic)
           rotation, // Array [x,y,z]
           scale, // Array [x,y,z]
-          size,
-          stable,
+          shadowMapSize: shadowMapSize && parseInt(shadowMapSize), // Number
+          shadowOpacity: shadowOpacity && parseFloat(shadowOpacity), // Number
```
donni106, line 97: at distance floats we do not have a `// Number` comment afterwards. do we need it here?

donni106 (approved):

src/helpers/augmentedReality/downloadObject.js
ardasnturk: Yes, you are right. But it's just a little information for later. ☺️

donni106: isn't `parseInt` or `parseFloat` enough information already?

PR #722, opened 2023-01-10
feat(accessibility): add the german language option

ardasnturk:
- added `locales` section to `app.json` to ensure that the iOS operating system understands that the app is in German

SVA-879


src/config/locales/de.json
```diff
@@ -0,0 +1 @@
+{}
```
donni106, line 1: you wrote:

> added locales section to app.json to ensure that the iOS operating system understands that the app is in German

but here is a new file as well. why is it empty? what can be a content here?
can you point to some expo documentation about this topic?

ardasnturk: This method is actually to make the application multilingual with the i18n package. After adding a language JSON to Locales, the language information required for Xcode is written in the new build.

But now I tried a new build and found out that it doesn't work correctly. I think I need to try a different method for this. 

I will let you know when I try the method.

This problem may be more intractable than I thought.

[Expo Doc](https://docs.expo.dev/versions/v45.0.0/config/app/#locales)
According to `Expo`'s documentation, if `locales` are added to `app.json`, the information we added for `infoPlist` will be overwritten.

[Stack Overflow](https://stackoverflow.com/questions/45962000/set-up-different-language-for-ios-app-using-react-native-and-expo)

ardasnturk:
This PR has not added the German language option that was supposed to be added to the app. It is unnecessary for now.

PR #723, opened 2023-01-10
refactor(accessibility): change bookmark alternative text

ardasnturk:
- updated the alternative text read by the phone when pressing the bookmark button with VoiceOver to be more understandable
- removed settings not only from the description but also from the key section since settings is no longer among the tasks of the button

SVA-867


src/config/consts.js
```diff
@@ -50,8 +50,8 @@ export const consts = {
     settingsIcon: 'Einstellungen (Taste)',
     settingsHint: 'Anpassen des Layouts von',
     settingsIconHint: 'Zu den Einstellungen wechseln',
-    settingsBookmarksIcon: 'Einstellungen und Lesezeichen (Taste)',
-    settingsBookmarksHint: 'Zu den Einstellungen und Lesezeichen wechseln',
+    settingsBookmarksIcon: 'Lesezeichen (Taste)',
+    settingsBookmarksHint: 'Zu den Lesezeichen wechseln',
```
donni106, line 54: we do not have settings here anymore. please also rename the key to make it fit.

donni106 (changes requested):

src/config/consts.js
ardasnturk: Updated 👍

https://github.com/smart-village-solutions/smart-village-app-app/commit/76c9de881368178dcf3a183b4ff6c2f9aad18d0a

donni106 (approved):

PR #725, opened 2023-01-11
refactor(accessibility): convert weather forecast to one text

ardasnturk:
- changed daily weather forecasts from two text components to one text component to enable `VoiceOver` to read them continuously
- changed `showsHorizontalScrollIndicator` to false to avoid preventing VoiceOver from reading the degree of weather in hourly weather

SVA-885

### DIFF
--- a/src/components/weather/DailyWeather.tsx
+++ b/src/components/weather/DailyWeather.tsx
@@ -51,22 +51,22 @@ export const DailyWeather = ({ date, description, icon, temperatures }: Props) =
           </WrapperRow>
           <RegularText />
           <WrapperRow spaceBetween>
-            <View style={styles.dayTimeEntry}>
-              <RegularText>Morgens</RegularText>
-              <RegularText>{morn.toFixed(1)}°C</RegularText>
-            </View>
-            <View style={styles.dayTimeEntry}>
-              <RegularText>Mittags</RegularText>
-              <RegularText>{day.toFixed(1)}°C</RegularText>
-            </View>
-            <View style={styles.dayTimeEntry}>
-              <RegularText>Abends</RegularText>
-              <RegularText>{eve.toFixed(1)}°C</RegularText>
-            </View>
-            <View style={styles.dayTimeEntry}>
-              <RegularText>Nachts</RegularText>
-              <RegularText>{night.toFixed(1)}°C</RegularText>
-            </View>
+            <RegularText center>
+              Morgens{'\n'}
+              {morn.toFixed(1)}°C
+            </RegularText>
+            <RegularText center>
+              Mittags{'\n'}
+              {day.toFixed(1)}°C
+            </RegularText>
+            <RegularText center>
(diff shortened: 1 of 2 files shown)


src/components/weather/DailyWeather.tsx
```diff
@@ -52,20 +52,24 @@ export const DailyWeather = ({ date, description, icon, temperatures }: Props) =
           <RegularText />
           <WrapperRow spaceBetween>
             <View style={styles.dayTimeEntry}>
-              <RegularText>Morgens</RegularText>
-              <RegularText>{morn.toFixed(1)}°C</RegularText>
+              <RegularText center>
```
donni106, line 55: why was `center` added? was ist aligned left with the new content?
and then the `styles.dayTimeEntry` of the parent view can be removed? or maybe the entire wrapping view, which was used to center?

src/components/weather/DailyWeather.tsx
```diff
@@ -52,20 +52,24 @@ export const DailyWeather = ({ date, description, icon, temperatures }: Props) =
...
+                Morgens {'\n'} {morn.toFixed(1)}°C
```
donni106, line 56: do we need additional spaces for voice over here?

src/screens/WeatherScreen.js
```diff
@@ -131,6 +131,7 @@ export const WeatherScreen = () => {
               horizontal
               keyExtractor={hourlyKeyExtractor}
               renderItem={renderHourlyWeather}
+              showsHorizontalScrollIndicator={false}
```
donni106, line 134: can you explain more in detail, what the difference is between having it `false` here compared to before? how does the look be different and how does it effect voice over?

donni106 (changes requested):

src/components/weather/DailyWeather.tsx
ardasnturk: If I don't give RegularText the centre prop, the text is written left-aligned. I need to give the centre prop so that the design is the same as before.

Yes, you are right. We can remove the Views here and just use RegularText.

src/components/weather/DailyWeather.tsx
ardasnturk: We don't need extra spaces for VoiceOver here. I just added it to make the code look nicer. Unnecessary spaces can be deleted.

src/screens/WeatherScreen.js
ardasnturk: If we set this value to true, an indicator will appear under the FlatList. When a person using VoiceOver clicks on this area, he clicks on this indicator instead of on the degree and cannot listen to the data he wants. That's why I set the value of the indicator to false.

As you can see in the screenshot below, there is no big difference in appearance.


|true|false|
|--|--|
IMG_0110 | IMG_0111

donni106 (approved):

PR #726, opened 2023-01-16
chore: update snap carousel package

ardasnturk:
- updated the `react-native-snap-carousel` package to a beta version to remove the `ViewPropTypes` warning and avoid a bug during future updates

SVA-890

## How to test:
* [x] Android portrait mode
* [x] iOS portrait mode
* [x] Android landscape mode

### DIFF
--- a/package.json
+++ b/package.json
@@ -108,7 +108,7 @@
     "react-native-render-html": "4.2.5",
     "react-native-safe-area-context": "4.2.4",
     "react-native-screens": "~3.11.1",
-    "react-native-snap-carousel": "~3.9.0",
+    "react-native-snap-carousel": "4.0.0-beta.6",
     "react-native-svg": "12.3.0",
     "react-native-web": "0.17.7",
     "react-native-webview": "11.18.1",


package.json
```diff
@@ -108,7 +108,7 @@
     "react-native-render-html": "4.2.5",
     "react-native-safe-area-context": "4.2.4",
     "react-native-screens": "~3.11.1",
-    "react-native-snap-carousel": "~3.9.0",
+    "react-native-snap-carousel": "4.0.0-beta.6",
```
donni106, line 111: 4.0.0-beta.6 is from August 2020, wow :/
Last commit on master is from February 2021.

I cannot find some changelogs for the 4.0.0 beta versions nor issues about that. There are mostly about 3.9.1 and patch package. Did you find some changelogs to know about differences between 3.9.x and 4.0.0-beta.x versions?

ardasnturk: I'm sorry, but unfortunately I couldn't find these changes either. Unfortunately, I had to update to the beta version because we had this ViewPropTypes problem even in the latest stable version. But the beta version was updated earlier than the latest stable version. I can't understand it either. 🤷‍♂️

PR #727, opened 2023-01-17
chore: update react native elements

ardasnturk:
- added a style change for the height of the `Input` as I read in the update changes so that the inputs have the same design as before the update
- edited because as I read in the update notes of the `rn-elements` package, the `height`, `width` and `borderRadius` props for the `Overlay` component have been added to the `overlayStyle`
- fixed text error in the title of the `Button` in `NoticeboardCreateForm`
- added new prop instead of deleted `ListItem` props in `rn-elements` package
- updated `style` prop in `Image` component to `childrenContainerStyle`
- added `accessibilityRole` prop to prevent application crash due to `Checkbox` component of `rn-elements` package
- added package `react-native-status-bar-height` deleted from package `rn-elements` to `package.json`

SVA-888

## How to test: 
please navigate through the app and report any UI bugs you see. The biggest changes have been made to pages with lists.

* [x] Android portrait mode
* [x] iOS portrait mode
* [x] Android landscape mode

## Screenshots:

|before (textliste)|after (textliste)|
|--|--|
IMG_1982 | IMG_1988
|before (list with small pictures)|after (list with small pictures)|
IMG_1983 | IMG_1989
|before (list with large images)|after (list with large images)|
IMG_1984 | IMG_1990


src/components/Modal.js
```diff
@@ -44,7 +41,11 @@ export const Modal = ({
 
 const styles = StyleSheet.create({
   overlay: {
+    borderRadius: normalize(8),
```
donni106, line 44: why is this changed to be normalized? do we need this? can you find some changes when trying different values or even no border radius? i cannot see differences.

src/components/form/Input.tsx
```diff
@@ -114,7 +115,7 @@ export const Input = ({
         multiline && device.platform === 'ios' && styles.multiline,
         !isValid && !!errorMessage && styles.inputError
       ]}
-      errorStyle={styles.inputError}
+      errorStyle={[styles.inputError, !errorMessage && styles.inputErrorHeight]}
```
donni106, line 118: why were this changes with heights added?

i can see a difference in the paddings between some form inputs/label now:

|before|after|
|---|---|
|IMG_87AE59423535-1|Simulator Screen Shot - iPhone 13 Pro Max - 2023-03-01 at 18 28 39|

you can see the differences between the first one and at the last ones.

donni106 (changes requested):

src/screens/EncounterRegistrationScreen.tsx
```diff
@@ -204,6 +204,7 @@ export const EncounterRegistrationScreen = ({ navigation }: StackScreenProps<any
             <Wrapper style={styles.noPaddingTop}>
               <WrapperRow style={styles.privacyContainer}>
                 <CheckBox
+                  accessibilityRole="button"
```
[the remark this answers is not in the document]
ardasnturk: this difference is due to the difference between the devices. this branch comes from behind the branch master. this problem is solved in https://github.com/smart-village-solutions/smart-village-app-app/pull/740. please test on devices of the same size.

|before|after|
|--|--|
Simulator Screen Shot - iPhone 8 Plus - 2023-03-02 at 09 08 20 | image

package.json
```diff
@@ -109,6 +109,7 @@
     "react-native-safe-area-context": "4.2.4",
     "react-native-screens": "~3.11.1",
     "react-native-snap-carousel": "~3.9.0",
+    "react-native-status-bar-height": "^2.6.0",
```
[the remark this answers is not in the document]
ardasnturk: you are right 👍

src/components/form/Input.tsx
ardasnturk: I have found the cause of the problem and I will fix it. I will edit the size of the input again.

src/components/Modal.js
```diff
@@ -44,7 +41,11 @@ export const Modal = ({
...
     padding: normalize(30)
+  },
+  overlayWidth: {
+    width: '80%'
```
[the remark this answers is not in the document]
ardasnturk: because the `overlay` style is only applied if `isListView` is `false`. We always need the `width` style.

src/components/Modal.js
ardasnturk: `borderRadius` style added to match the previous design

|without `borderRadius`|with `borderRadius`|
|--|--|
IMG_2632 | IMG_2631

src/components/form/Input.tsx
ardasnturk: increased the size of the input container from `normalize(65)` to `normalize(75)`

https://github.com/smart-village-solutions/smart-village-app-app/commit/5abeb5b34d774762c238b6dda8492f996627c635

package.json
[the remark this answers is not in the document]
ardasnturk: yarn.lock added
https://github.com/smart-village-solutions/smart-village-app-app/commit/5abeb5b34d774762c238b6dda8492f996627c635

package.json
```diff
@@ -94,7 +94,7 @@
     "react-native-autocomplete-input": "^5.1.0",
     "react-native-calendars": "1.1276.0",
     "react-native-collapsible": "^1.6.0",
-    "react-native-elements": "~1.2.0",
+    "react-native-elements": "^3.4.3",
```
donni106, line 97: another difference i found on the big tiles:

|before|after|
|---|---|
|IMG_ABB68C178E5D-1|Simulator Screen Shot - iPhone 13 Pro Max - 2023-03-02 at 10 53 50|

they overlap in width a bit.

ardasnturk: this is a really interesting problem. actually, `ServiceTiles` has nothing to do with this update. i've searched for the problem, but I can't solve it.

donni106: Maybe screens or wrapper or paddings changed somehow? We could inspect the screen elements to see if there are differences.

ardasnturk: okay, I found the cause of the problem. it is caused by prop change.
in the latest version of `rn-elements`, the `childrenContainerStyle` prop was added for the `Image` component. 
the prop we used to use as style is now called `childrenContainerStyle`. but we can still use the `style` prop. 
the `style` prop we have used so far in our code will now be the `childrenContainerStyle` prop.

donni106: well done

donni106 (approved):

PR #729, opened 2023-01-20
feat(events): calendar toggle and calendar view

donni106:
- adapted calendar functionality from volunteer for main event records lists
- added section header click to navigate to a specific day from list like clicking on a day in calendar
- refactored some concerning code 

SVA-665

---

ℹ️ When navigating in a specific day or selecting to show just the current day, the calendar option disappears, because there is no need for a calendar having data for just one day.

Simulator Screen Shot - iPhone 13 Pro Max - 2023-01-20 at 19 54 42-tile


src/screens/IndexScreen.js
```diff
@@ -365,67 +375,88 @@ export const IndexScreen = ({ navigation, route }) => {
...
-                        <DropdownHeader
-                          {...{
-                            query,
-                            queryVariables,
-                            data: initialNewsItemsFetch && loading ? {} : data,
-                            updateListData: updateListDataByDropdown
-                          }}
+              <>
+                {isCalendar && !hasDailyFilterSelection && (
+                  <CalendarListToggle
```
ardasnturk, line 380: Why is there a hasDailyFilterSelection control? Is it desirable that the toggle disappears when this filter is active?
|list toggle - filter: false|list toggle - filter: true|calendar toggle - filter: false|calendar toggle - filter: true|
|--|--|--|--|
IMG_2040 | IMG_2041 | IMG_2042 | IMG_2043

ardasnturk, line 380: Why is there a hasDailyFilterSelection control? Is it desirable that the toggle disappears when this filter is active?
|list toggle - filter: false|list toggle - filter: true|calendar toggle - filter: false|calendar toggle - filter: true|
|--|--|--|--|
IMG_2040 | IMG_2041 | IMG_2042 | IMG_2043

src/components/EventList.js
```diff
@@ -71,7 +76,20 @@ export const EventList = ({
...
+            navigation.push(ScreenName.Index, {
+              title: texts.homeTitles.events,
+              query,
+              queryVariables: {
+                ...queryVariables,
+                dateRange: [momentFormat(title, 'YYYY-MM-DD'), momentFormat(title, 'YYYY-MM-DD')]
+              },
+              rootRouteName: ROOT_ROUTE_NAMES.EVENT_RECORDS
+            })
+          }
```
ardasnturk, line 91: I'm getting an error here for Volunteer. I think a redirect needs to be added for the `ScreenName.VolunteerIndex` screen.

IMG_2044

ardasnturk (changes requested):
Generally, it looks good. 👍

src/screens/IndexScreen.js
donni106: good point. with daily filter we should show the list always, right? even if we have selected the calendar before.
or we keep the toggle visible as well and let the users handle and know everything on their own?

ardasnturk: Yes, but neither is possible at the moment. I have activated the daily filter after opening the calendar view. 

Calendar appears. toggle does not appear.

According to the ticket, if the daily event filter is activated, the list view should be switched to. So I think it would be better to remove the toggle and replace the calendar with the list.

Or am I thinking wrong?

donni106: i think we agree on this:

https://user-images.githubusercontent.com/1942953/214077838-a7c3ec89-acf1-4336-b4fd-ff92a63ba684.mp4

ardasnturk: Yeah, it's better now.  👍

src/components/EventList.js
donni106: there is a weird error on volunteer calendar sections, thanks!
but it is not located in the `EventList`, as tis component is not used for volunteer. i need to find the right place.

donni106: one problem was the pressable header for volunteer lists, which i have reverted and restricted to default events: 1c1e9e1

donni106: found the bad spot: 96a259c
changed condition on `query` instead of `contentContainerId`, which can be empty for volunteer as well

ardasnturk (approved):
It looks good. 👍

PR #742, opened 2023-02-02
refactor(defect report): add star for required field

ardasnturk:
- added `*` to the label of the category because it is required

SVA-738

* [x] Android portrait mode
* [x] iOS portrait mode
* [x] Android landscape mode
* [x] iOS landscape mode

## Screenshots:

|before|after|
|--|--|
Simulator Screen Shot - iPhone 14 - 2023-02-02 at 15 44 44 | Simulator Screen Shot - iPhone 14 - 2023-02-02 at 15 44 47

### DIFF
--- a/src/components/defectReport/DefectReportCreateForm.tsx
+++ b/src/components/defectReport/DefectReportCreateForm.tsx
@@ -149,7 +149,7 @@ export const DefectReportCreateForm = ({
                 onChange,
                 name,
                 required: true,
-                label: texts.defectReport.categoryName,
+                label: `${texts.defectReport.categoryName} *`,
                 placeholder: texts.defectReport.categoryName,
                 control
               }}


src/components/defectReport/DefectReportCreateForm.tsx
```diff
@@ -149,7 +149,7 @@ export const DefectReportCreateForm = ({
                 onChange,
                 name,
                 required: true,
-                label: texts.defectReport.categoryName,
+                label: texts.defectReport.categoryName + ' *',
```
donni106, line 152: please see other lines in this file being formatted like this 
```
label={`${texts.defectReport.inputTitle} *`}
```

do you think it's fine to have the two different syntaxes? dann fell free to merge, otherwise you could change to an uniform format.

ardasnturk: I wrote the `placeholder` in the same format.

https://github.com/smart-village-solutions/smart-village-app-app/commit/537a83a3ec675e2563d4c93994bfdba2dce14873

donni106: and why? what was you goal? the placeholder has no string concatenation or anything.

ardasnturk: So that all inputs are in the same format. But you're right. It doesn't need any extra logic.
I've reversed the change.

https://github.com/smart-village-solutions/smart-village-app-app/commit/d9a9f57fc10b44ca8700325a106c249dcd0ff009

donni106: the same format should count for strings and values with concatenation, not for single variables

donni106 (approved):

PR #743, opened 2023-02-02
feat(ar): add missing information

ardasnturk:
- added to `localUris` by finding the information in the first index for more stable operation of the AR feature

SVA-571

* [x] Android portrait mode
* [x] iOS portrait mode
* [x] Android landscape mode
* [x] iOS landscape mode

### DIFF
--- a/src/helpers/augmentedReality/multipleSceneIndexGenerator.ts
+++ b/src/helpers/augmentedReality/multipleSceneIndexGenerator.ts
@@ -79,7 +79,40 @@ export const multipleSceneIndexGenerator = ({
       );
     }
 
+    // localUris = [{ type: 'vrx' }, { type: 'texture' }, { type: 'texture' }];
     localUris = scenes?.[modelIndex]?.localUris;
+
+    if (modelIndex > 0) {
+      /*
+      the first index of scenes is selected because all data except models and textures 
+      are in the first index
+      firstSceneLocalUris = [
+        { type: 'vrx' },
+        { type: 'texture' },
+        { type: 'texture' },
+        // benötigen wir Folgendes.
+        { type: 'quad' },
+        { type: 'target' },
+        { type: 'mp3' }
+      ];
+      */
+      const firstSceneLocalUris = scenes?.[0]?.localUris;
+
+      /*
+      according to the date logic, all types in the selected index are equal to uniqueTypes
+      uniqueTypes = {'vrx', 'texture'};
+       */
+      const uniqueTypes = new Set(localUris.map((item) => item.type));
+
+      /* all the data we need except model and texture are merged with localUris
+      localUris = [...localUris, { type: 'quad' }, { type: 'target' }, { type: 'mp3' }]; 
+      */
+      localUris = [
+        ...localUris,
+        ...firstSceneLocalUris.filter((item) => !uniqueTypes.has(item.type))
+      ];
+    }
+
     const textureIndex = texture % variableTexturesCount;
     textures.push(variableTextures?.[textureIndex] as TTextures[0]);
 


src/helpers/augmentedReality/multipleSceneIndexGenerator.ts
```diff
@@ -80,6 +80,14 @@ export const multipleSceneIndexGenerator = ({
     }
 
     localUris = scenes?.[modelIndex]?.localUris;
+
+    if (modelIndex !== 0) {
```
donni106, line 84: we want this for > 0, and not for negative values, or am i wrong? comparison with > 0 would read better here i guess?

src/helpers/augmentedReality/multipleSceneIndexGenerator.ts
```diff
@@ -80,6 +80,14 @@ export const multipleSceneIndexGenerator = ({
...
+      const firstIndex = scenes?.[0]?.localUris;
```
donni106, line 85: the const name is not correct. getting `localUris` from the first index does not result in `firstIndex`. first index makes me think of a `0`. can you find a more precise name here?

src/helpers/augmentedReality/multipleSceneIndexGenerator.ts
```diff
@@ -80,6 +80,14 @@ export const multipleSceneIndexGenerator = ({
...
+      const uniqueTypes = new Set(localUris.map((item) => item.type));
+
+      localUris = [...localUris, ...firstIndex.filter((item) => !uniqueTypes.has(item.type))];
```
donni106, line 88: the commit message 

> added to localUris by finding the information in the first index for more stable operation of the AR feature

does not help me to understand this line.

can you please tell me what happens here? thanks.

src/helpers/augmentedReality/multipleSceneIndexGenerator.ts
ardasnturk: That's right. Index will never be a negative number. 👍

src/helpers/augmentedReality/multipleSceneIndexGenerator.ts
ardasnturk: How about `firstLocalUris`?

donni106: first of what? maybe more precisely `firstSceneLocalUris`?

src/helpers/augmentedReality/multipleSceneIndexGenerator.ts
ardasnturk: All the files we need are in the first index.

We change the `modelIndex` according to the date logic. In this case, the modelIndex we choose does not have the files we want. For example `targetImage`, `mp3` etc. We add the types in `localUris` that we use to `uniqueTypes`.

Then we filter the types that are not in `localUris[x]` and add them to `localUris[x]`. This way we have all the files we want.

donni106: what is included in unique types and what not?

src/helpers/augmentedReality/multipleSceneIndexGenerator.ts
ardasnturk: Yes, it's more understandable.

https://github.com/smart-village-solutions/smart-village-app-app/commit/48355bfc73c3eb1184fd467ccd846dc334c027d7

src/helpers/augmentedReality/multipleSceneIndexGenerator.ts
ardasnturk: unique types: `target`, `mp4`, `mp3`, `image`, `spotlight`, `light`, `quad`

donni106: why aren't textures included in `const uniqueTypes = new Set(localUris.map((item) => item.type));`?

ardasnturk: There are `textures` and `vrx` types in `uniqueTypes`. But the types we want to add are the types I mentioned above. Therefore, we go through a filtering process.

For our date logic to work properly, we need the `textures` and `vrx` in the `x` index. We compare these types with the first index and add the types we don't have to `localUris`.

donni106: > There are textures and vrx types in uniqueTypes. But the types we want to add are the types I mentioned above. Therefore, we go through a filtering process.

i understand that there is a filtering process and what the result should be. but i cannot understand, how this happens. can you please explain, what the set and filter does with all the localUris based on their item types!?

ardasnturk: this is what I did, step by step.

`scenes` is an array and we select an index from this array according to the date logic. we equate `localUris` to the `localUris` of this index. (line 82)

But there are models and textures in the index we selected. The other types we need remain in the first index.

If the `modelIndex` is greater than 0, we set the `localUris` of the first index equal to the `firstSceneLocalUris`.

In `uniqueTypes`, we match the types in the x index we select (line 86). `uniqueTypes` contains the types of models and textures. The types we need are in `firstSceneLocalUris`. But `firstSceneLocalUris` also contains models and textures. 

In line 88 we create `localUris` again. We filter and add the types that are in `uniqueTypes` on the `localUris` and not in `localUris`.

After this process, the `localUris` array consists of model, textures and other necessary file types.

ardasnturk: I added more descriptive text and examples to the code.

https://github.com/smart-village-solutions/smart-village-app-app/commit/90d9edfbe35db5999b228500a95de950838f9a70

donni106 (approved):

PR #746, opened 2023-02-08
fix: add control in case of no data

ardasnturk:
- added `undefined` control for data so that the app does not crash in case of no data

SBB-7


donni106:
thanks, done with 058b2a4d9cb476ef0031951a687fb3a3c362cbc7

PR #747, opened 2023-02-08
feat(accessibility): add new accessibility to dropdown

ardasnturk:
- added accessibility features to the `View` that wraps the entire dropdown so that it can be read by `VoiceOver` along with the dropdown's title
- added accessibility features to the` Wrapper` so that dropdown menu elements can be read properly by `VoiceOver`
- disabled the `accessible` feature in the `Dropdown` component because it prevents the accessibility features we have made

SVA-882

|before|before|before|
|--|--|--|
IMG_2449 | IMG_2450 | IMG_2451

|after|after|
|--|--|
IMG_2444 | IMG_2445


src/components/DropdownSelect.js
```diff
@@ -68,7 +74,11 @@ export const DropdownSelect = ({
   }, [selectedData]);
 
   return (
-    <View>
+    <View
+      accessible
+      accessibilityLabel={`(${label}) (${multipleSelect ? selectedMultipleValues : selectedValue})
```
donni106, line 79: please order the props `accessible` and `accessibilityLabel` always in the same way

src/config/consts.js
```diff
@@ -41,6 +41,7 @@ export const consts = {
     mail: '(E-Mail-Adresse)',
     mailHint: '(Wechselt zur E-Mail-App)',
     mapHint: '(Wechselt zur Karten-App)',
+    menuItem: 'Menüelement (Taste)',
```
donni106, line 44: i would not create a new "context" here with `Menü`. i relates to the dropdown, right? so what about `dropDownMenuItem: Dropdown-Menü-Eintrag (Taste)`?

ardasnturk: yes you are right. It's better now.
https://github.com/smart-village-solutions/smart-village-app-app/commit/0a6c8ce7fa5ed299446fb3ab7a910a08b673a271

src/components/DropdownSelect.js
ardasnturk: I can delete the brackets in the label because brackets are used in the rest of the sentence. 👍

https://github.com/smart-village-solutions/smart-village-app-app/commit/0a6c8ce7fa5ed299446fb3ab7a910a08b673a271

donni106: > please order the props accessible and accessibilityLabel always in the same way

you missed that, right?

ardasnturk: I think it's done now. 🤷‍♂️

https://github.com/smart-village-solutions/smart-village-app-app/commit/2b70e4c58f4da0633c16f4fb6de915210e644970

donni106 (approved):

PR #749, opened 2023-02-10
feat(category): make category intro text changeable from server

ardasnturk:
- added `categoryListIntroText` to the `sections` section on the server to be able to change the text on the `CategoryList` screen via the server
- added intro text used in `CategoryList` component to `IndexScreen`

SBB-19

## How to test:
To try it out, add the following line to the `sections` section in `globalSettings` on Main-Server.

```
"categoryListIntroText": "Intro Text",
```

* [x] Android portrait mode
* [x] iOS portrait mode
* [x] Android landscape mode
* [x] iOS landscape mode

### DIFF
--- a/src/components/CategoryList.js
+++ b/src/components/CategoryList.js
@@ -7,9 +7,7 @@ import { colors, consts, device, texts } from '../config';
 
 import { CategoryListItem } from './CategoryListItem';
 import { LoadingContainer } from './LoadingContainer';
-import { RegularText } from './Text';
 import { Title, TitleContainer, TitleShadow } from './Title';
-import { Wrapper } from './Wrapper';
 
 export class CategoryList extends React.PureComponent {
   keyExtractor = (item, index) => `index${index}-id${item.id}`;
@@ -30,7 +28,7 @@ export class CategoryList extends React.PureComponent {
   };
 
   render() {
-    const { data, navigation, noSubtitle, refreshControl, hasSectionHeader } = this.props;
+    const { data, navigation, noSubtitle, refreshControl, ListHeaderComponent } = this.props;
 
     if (!data?.length) {
       return (
@@ -72,14 +70,7 @@ export class CategoryList extends React.PureComponent {
           />
         )}
         renderSectionHeader={this.renderSectionHeader}
-        ListHeaderComponent={
-          hasSectionHeader &&
-          !!texts.categoryList.intro && (
-            <Wrapper>
-              <RegularText>{texts.categoryList.intro}</RegularText>
-            </Wrapper>
-          )
-        }
+        ListHeaderComponent={ListHeaderComponent}
         stickySectionHeadersEnabled
         refreshControl={refreshControl}
       />
@@ -92,7 +83,8 @@ CategoryList.propTypes = {
   data: PropTypes.array,
   noSubtitle: PropTypes.bool,
   refreshControl: PropTypes.object,
-  hasSectionHeader: PropTypes.bool
+  hasSectionHeader: PropTypes.bool,
+  ListHeaderComponent: PropTypes.object
 };
 
 CategoryList.defaultProps = {
--- a/src/screens/IndexScreen.js
+++ b/src/screens/IndexScreen.js
@@ -18,7 +18,9 @@ import {
   LoadingContainer,
   LocationOverview,
   OptionToggle,
-  SafeAreaViewFlex
+  RegularText,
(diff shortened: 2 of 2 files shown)


src/screens/IndexScreen.js
```diff
@@ -3,7 +3,7 @@ import moment from 'moment';
 import PropTypes from 'prop-types';
 import React, { useCallback, useContext, useEffect, useState } from 'react';
 import { Query } from 'react-apollo';
-import { ActivityIndicator, RefreshControl, View } from 'react-native';
+import { ActivityIndicator, RefreshControl, Text, View } from 'react-native';
```
donni106, line 6: this `Text` is not used, right?

src/screens/IndexScreen.js
```diff
@@ -83,10 +85,11 @@ export const IndexScreen = ({ navigation, route }) => {
   const [topFilter, setTopFilter] = useState(INITIAL_TOP_FILTER);
   const { isConnected, isMainserverUp } = useContext(NetworkContext);
   const { globalSettings } = useContext(SettingsContext);
-  const { filter = {}, hdvt = {}, settings = {} } = globalSettings;
+  const { filter = {}, hdvt = {}, settings = {}, sections = {} } = globalSettings;
   const { news: showNewsFilter = false, events: showEventsFilter = true } = filter;
   const { events: showVolunteerEvents = false } = hdvt;
   const { calendarToggle = false } = settings;
+  const { categoryListIntroText } = sections;
```
donni106, line 92: please add the fallback for `texts.categoryList.intro` here to not have it twice in the render part

src/components/CategoryList.js
```diff
@@ -92,7 +83,8 @@ CategoryList.propTypes = {
   data: PropTypes.array,
   noSubtitle: PropTypes.bool,
   refreshControl: PropTypes.object,
-  hasSectionHeader: PropTypes.bool
+  hasSectionHeader: PropTypes.bool,
```
donni106, line 86: if you remove the usage of `hasSectionHeader`, it can be removed here as well, right?

src/screens/IndexScreen.js
```diff
@@ -416,6 +419,14 @@ export const IndexScreen = ({ navigation, route }) => {
...
                         />
                       )}
+                      {query === QUERY_TYPES.CATEGORIES &&
+                        (!!categoryListIntroText || !!texts.categoryList.intro) && (
+                          <Wrapper>
+                            <RegularText>
+                              {categoryListIntroText || texts.categoryList.intro}
+                            </RegularText>
+                          </Wrapper>
+                        )}
```
donni106, line 429: this should be ListHeaderComponent of the `CategoryList` or how does that work?
you have addd the usage of `ListHeaderComponent={ListHeaderComponent}` in `CategoryList` but are not passing it here in line 419

donni106 (changes requested):

src/components/CategoryList.js
ardasnturk: I haven't deprecated `hasSectionHeader`, it just gets deleted and re-added when I add a new line.

`hasSectionHeader` is still in use above.

```
renderSectionHeader = ({ section: { title, data } }) => {
    const { hasSectionHeader } = this.props;

    if (!data?.length || !hasSectionHeader) return null;
...
```

src/screens/IndexScreen.js
ardasnturk: Good idea. 👍

src/screens/IndexScreen.js
ardasnturk: Yes 👍

src/screens/IndexScreen.js
ardasnturk: In line 419, the `query` is named `pointOfInterest`. The header here is only used if there is a subcategory. Therefore, there is no situation that will affect the `intro` text.

I am sharing screenshots.

|categoryList|subCategoryList|
|--|--|
Simulator Screen Shot - iPhone 14 Plus - 2023-02-10 at 16 24 32 | Simulator Screen Shot - iPhone 14 Plus - 2023-02-10 at 16 24 42

src/components/CategoryList.js
donni106: ah i see

src/screens/IndexScreen.js
donni106: ok 🤞

donni106 (approved):

PR #759, opened 2023-03-02
feat(waste calendar): add possibility for two step ui

donni106:
if there should be two steps, users need to select a city prior to the street

- added new boolean flag `wasteAddresses.twoStep` in global settings, which indicates the functionality or not
  - also added a `wasteAddresses.streetCount` to be able to adjust the amount of streets suggested
- refactored code to consider the new flag in behavior and renderings
- moved methods in the component scope to be able to rely on state variables
- updated behavior of the screen to show the calendar only if a street is selected and hide it while user change their city or street until a new street is chosen
- added alphabetical sorting for streets
- removed absolute styling of the input as we need two inputs
  - with changing the visibility of the calendar depending on a selected street there is no need to render result lists of autocomplete fields above it

SVA-833

|1|2|3|4|5|
|---|---|---|---|---|
|Simulator Screen Shot - iPhone 13 Pro Max - 2023-03-02 at 13 45 45|Simulator Screen Shot - iPhone 13 Pro Max - 2023-03-02 at 13 46 02|Simulator Screen Shot - iPhone 13 Pro Max - 2023-03-02 at 13 46 10|Simulator Screen Shot - iPhone 13 Pro Max - 2023-03-02 at 13 46 17|Simulator Screen Shot - iPhone 13 Pro Max - 2023-03-02 at 13 46 22|


src/screens/WasteCollectionScreen.js
```diff
@@ -143,6 +132,79 @@ export const WasteCollectionScreen = ({ navigation }) => {
...
+  // return empty list on an exact match (except for capitalization)
+  const filterCities = useCallback(
+    (currentInputValue, addressesData) => {
+      if (currentInputValue === '') return [];
+
+      const cities = addressesData?.filter(
+        (address) =>
+          address.city !== currentInputValue &&
+          address.city.toLowerCase().includes(currentInputValue.toLowerCase())
+      );
```
ardasnturk, line 159: i think there should be a check for the `null` value of `city`.
like this:
`address.city?.toLowerCase()`

IMG_8C3B5F24396B-1

src/screens/WasteCollectionScreen.js
```diff
@@ -143,6 +132,79 @@ export const WasteCollectionScreen = ({ navigation }) => {
...
+      if (currentInputValue === '') return [];
+
+      const streets = addressesData
+        ?.filter((address) => {
+          const street = getStreetString(address);
+
+          return (
+            street !== currentInputValue &&
+            street.toLowerCase().includes(currentInputValue.toLowerCase())
+          );
```
ardasnturk, line 181: can `street` get a `null` value?

src/screens/WasteCollectionScreen.js
```diff
@@ -306,8 +397,8 @@ const styles = StyleSheet.create({
   autoCompleteList: {
     margin: 0
   },
-  topMarginContainer: {
-    marginTop: normalize(44)
+  noBottomBorder: {
+    borderBottomWidth: 0
   }
```
ardasnturk, line 402: the name has been changed but not used. can it be deleted?

src/screens/WasteCollectionScreen.js
```diff
@@ -232,26 +302,48 @@ export const WasteCollectionScreen = ({ navigation }) => {
...
+              disableFullscreenUI
+              onChangeText={(text) => setInputValue(text)}
+              placeholder="Straße"
+              flatListProps={{
+                renderItem: renderSuggestion
+              }}
+              inputContainerStyle={styles.autoCompleteInput}
+              listStyle={styles.autoCompleteList}
+            />
+          )}
```
ardasnturk, line 345: can it be alphabetised to make it easier to read? 🤷‍♂️☺️

src/screens/WasteCollectionScreen.js
```diff
@@ -143,6 +132,79 @@ export const WasteCollectionScreen = ({ navigation }) => {
...
+  );
+
+  // show streets that contain the string in it
+  // consider the city if it is set because of the two step process
+  // return empty list on an exact match (except for capitalization)
+  const filterStreets = useCallback(
+    (currentInputValue, addressesData) => {
+      if (wasteAddressesTwoStep && inputValueCity === '') return [];
+      if (currentInputValue === '') return [];
+
```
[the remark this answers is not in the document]
donni106: no, brackets would be missing in your example around `wasteAddressesTwoStep && inputValueCity === ''`.
and i think it is not a better solution. i see different scenarios here. having two step, than it is necessary to check the `inputValueCity`. in every other case, like not having two step or two step with `inputValueCity !== ''`, `currentInputValue === ''` needs to be checked. because of this two point of view, i decided to have two conditions with return here.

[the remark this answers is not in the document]
ardasnturk: then we do it this way. 🎉

ardasnturk (approved):

PR #778, opened 2023-03-14
feat(accessibility): add bold option to all text styles

ardasnturk:
- edited according to arrow function format to be able to use `useContext` in html style
- updated `RegularText` to arrow function format to be able to replace text using `RegularText` `styled-component` with bold font
- updated `RegularText` to `RegularTextStyledComponent` to use `RegularText` `styled-component`
- updated `styles.html` in `HtmlView` because html style was updated to arrow function format
- added `fontFamily` style for the `p` tag to `StorySection` since the `p` tag is inside the `div` tag and uses the style of the `p` tag as a priority
- added `bold` or `regular` `fontFamily` option to all text using html style according to `isBoldTextEnabled`

SVA-886

## Screenshots:

|before (home screen)|after (home screen)|before (news screen)|after (news screen)|
|--|--|--|--|
IMG_2724 | IMG_2728 | IMG_2725 | IMG_2729

|before (POI screen)|after (POI screen)|before (impressum screen)|after (impressum screen)|
|--|--|--|--|
IMG_2726 | IMG_2730 | IMG_2727 | IMG_2731

### DIFF
--- a/src/components/Text.js
+++ b/src/components/Text.js
@@ -33,23 +33,32 @@ function parseText(text) {
   return result;
 }
 
-export const Text = ({ children, ...props }) => {
-  return (
-    <RNText {...props}>{typeof children === 'string' ? parseText(children) : children}</RNText>
-  );
-};
-
-export const RegularText = ({ children, ...props }) => {
+export const Text = ({ children, style, italic, ...props }) => {
   const { isBoldTextEnabled } = useContext(AccessibilityContext);
 
-  if (isBoldTextEnabled) {
-    return <BoldText {...props}>{children}</BoldText>;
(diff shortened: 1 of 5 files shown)


src/components/Text.js
donni106, line 1: this is a bit too confusing for me in different points. 1. naming of the component, 2. new component just for checking a context variable

i thought about it and used the already existing `Text` component to handle the context variable. this way everything else stays untouched: 1f312e1

ardasnturk: It's a very nice change. Thank you. 👍

donni106 (approved):

PR #779, opened 2023-03-15
feat(accessibility): increase contrast for AZ filter

ardasnturk:
- added `AccessibilityContext` to `AZFilter` to know whether `isReduceTransparencyEnabled` is `true` or `false`
- added `blackBackgroundForContrast` prop to `AZFilterElement` component to increase contrast
- set `blackBackgroundForContrast` to `true` according to `isReduceTransparencyEnabled` and `item.selected` control in `AZFilterElement` component
- added `surface` color to `BoldText` to increase contrast with background if `isReduceTransparencyEnabled` and `item.selected` is `true`

SBB-25

## Screenshots:

|before|after|
|--|--|
Simulator Screen Shot - iPhone 14 Pro Max - 2023-03-15 at 09 37 45 | Simulator Screen Shot - iPhone 14 Pro Max - 2023-03-15 at 09 37 38
image | image


src/components/BB-BUS/AZFilterElement.js
```diff
@@ -32,4 +32,10 @@ export const AZFilterElement = styled.View`
     css`
       margin-right: ${normalize(14)}px;
     `};
+
+  ${(props) =>
+    props.blackBackgroundForContrast &&
+    css`
+      background-color: ${colors.darkText};
```
donni106, line 39: why have you decided to make it black here instead of `colors.overlayRgba` like in https://github.com/smart-village-solutions/smart-village-app-app/pull/777/files?

src/components/BB-BUS/AZFilterElement.js
```diff
@@ -32,4 +32,10 @@ export const AZFilterElement = styled.View`
     css`
       margin-right: ${normalize(14)}px;
     `};
+
+  ${(props) =>
+    props.blackBackgroundForContrast &&
```
donni106, line 37: please don't call it `black`, as this can be different in some cases. `backgroundForContrast` should be enough to tell what is going on.

src/components/BB-BUS/AZFilterElement.js
ardasnturk: I chose black to increase the contrast ratio. But in `colors.overlayRgba` we can get a contrast ratio of 6,9:1 and that should be enough. 👍

|darkText|overlayRgba|
|--|--|
image | image

https://github.com/smart-village-solutions/smart-village-app-app/commit/c8b47067d02ff0a1dc6b6832058238818b26ea70

src/components/BB-BUS/AZFilterElement.js
ardasnturk: https://github.com/smart-village-solutions/smart-village-app-app/commit/c8b47067d02ff0a1dc6b6832058238818b26ea70

donni106 (approved):

PR #780, opened 2023-03-15
feat(accessibility): add style to inputs

ardasnturk:
- added `borderColor` style so that the borders of the inputs in the `Feedback` form have a better contrast ratio with the background after turning on the `Reduce Transparency` feature in the accessibility settings

SBB-29

## Screenshots:

|before|after|
|--|--|
Simulator Screen Shot - iPhone 14 Pro Max - 2023-03-15 at 10 26 09 | Simulator Screen Shot - iPhone 14 Pro Max - 2023-03-15 at 10 26 19
image | image


src/screens/FeedbackScreen.js
```diff
@@ -149,6 +164,9 @@ export const FeedbackScreen = () => {
 };
 
 const styles = StyleSheet.create({
+  inputBorderContrastColor: {
```
donni106, line 167: please do not name it `...Color`, as it is a style and can change its content. if that happens, you would not like to change the name. please be accurate with `backgroundContrast` [here](https://github.com/smart-village-solutions/smart-village-app-app/pull/779) and name it `inputContrast` for example

donni106 (changes requested):

src/screens/FeedbackScreen.js
ardasnturk: https://github.com/smart-village-solutions/smart-village-app-app/commit/449a788e159ff48ac0d38841c88ee5ca5f4baef5

donni106 (approved):

PR #782, opened 2023-03-15
feat(accessibility): increase contrast for hourly weather

ardasnturk:
- updated the background color to `primary` and `RegularText` to `lightest` if `isReduceTransparencyEnabled` is `true` to increase contrast in the hourly weather section

SBB-24

## Screenshots:

|before|after|
|--|--|
Simulator Screen Shot - iPhone 14 Pro Max - 2023-03-15 at 11 32 17 | Simulator Screen Shot - iPhone 14 Pro Max - 2023-03-15 at 11 32 08
|background contrast|background contrast|
image | image
|text contrast|text contrast|
image | image

### DIFF
--- a/src/components/weather/HourlyWeather.tsx
+++ b/src/components/weather/HourlyWeather.tsx
@@ -1,6 +1,7 @@
-import React from 'react';
+import React, { useContext } from 'react';
 import { StyleSheet, View } from 'react-native';
 
+import { AccessibilityContext } from '../../AccessibilityProvider';
 import { colors, normalize } from '../../config';
 import { momentFormat } from '../../helpers';
 import { Image } from '../Image';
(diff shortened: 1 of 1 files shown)


src/components/weather/HourlyWeather.tsx
```diff
@@ -34,7 +43,9 @@ export const HourlyWeather = ({
         style={styles.icon}
         resizeMode="contain"
       />
-      <RegularText>{temperature.toFixed(1)}°C</RegularText>
+      <RegularText lightest={isNow && isReduceTransparencyEnabled}>
+        {temperature.toFixed(1)}°C
+      </RegularText>
```
donni106, line 48: the text color needs to be checked when merging with all the customers, because it can be that `lightest` and `primary` are not suitable in some cases.

donni106 (approved):

src/components/weather/HourlyWeather.tsx
donni106: yes, we need to be aware of this in the future when merging things

PR #783, opened 2023-03-16
chore(expo sdk): upgrade to 46

ardasnturk:
This PR upgrades Expo SDK to 46. It was hard to find out which caused the problem with babel stuff. After deleting yarn.lock and node_modules to reinstall everything from scratch the problem was solved. So maybe you also need to do these steps.

SVA-851


src/index.js
```diff
@@ -38,6 +37,7 @@ import { SettingsProvider } from './SettingsProvider';
 const { LIST_TYPES } = consts;
 
 const MainAppWithApolloProvider = () => {
+  const { isConnected, isMainserverUp } = useContext(NetworkContext);
```
donni106, line 40: @ardasnturk please have a few checks on your side, if everything works as before?!

ardasnturk: i have not encountered any problems. i can say that it works well. 👍

PR #784, opened 2023-03-17
chore(expo sdk): upgrade to 47

ardasnturk:
This PR upgrades Expo SDK to 47. After deleting yarn.lock and node_modules to reinstall everything from scratch the problem was solved. So maybe you also need to do these steps.

please make a new build to try the new version:
`yarn prebuild`
`yarn ios -d` or `yarn android -d`
⚠️ added command to delete yarn.lock in prebuild script ⚠️

[CHANGELOG](https://github.com/expo/expo/blob/main/CHANGELOG.md#4700--2022-10-28)
⚠️ With the Expo 47 update, the minimum version requirement for iOS is now 13. ⚠️

SVA-852


package.json
```diff
@@ -86,52 +86,52 @@
...
     "react-native-calendars": "1.1276.0",
     "react-native-collapsible": "^1.6.0",
     "react-native-elements": "^3.4.3",
     "react-native-expo-image-cache": "4.1.0",
-    "react-native-gesture-handler": "~2.5.0",
+    "react-native-gesture-handler": "~2.8.0",
     "react-native-get-random-values": "~1.8.0",
     "react-native-gifted-chat": "0.16.3",
-    "react-native-maps": "0.31.1",
+    "react-native-maps": "1.3.2",
```
donni106, line 103: whuu big step. did you check that in detail?

ardasnturk: yes it's a big update but when I read the changes I don't think we have to make any changes for us.
[releases](https://github.com/react-native-maps/react-native-maps/releases)

donni106 (approved):

package.json
donni106: all right

PR #785, opened 2023-03-17
chore(expo sdk): upgrade to 48

ardasnturk:
This PR upgrades Expo SDK to 48. After deleting yarn.lock and node_modules to reinstall everything from scratch the problem was solved. So maybe you also need to do these steps.

please make a new build to try the new version:
`yarn prebuild`
`yarn ios -d` or `yarn android -d`

Expo [CHANGELOG](https://github.com/expo/expo/blob/main/CHANGELOG.md#4800--2023-02-09)

⚠️ orientation bug fixed in this version ⚠️

SVA-937


app.json
```diff
@@ -9,7 +9,6 @@
     "platforms": ["ios", "android"],
     "version": "3.1.2",
     "runtimeVersion": "nativeVersion",
-    "orientation": "portrait",
```
donni106, line 12: why was this removed?

https://docs.expo.dev/versions/latest/config/app/#orientation

App.js
```diff
@@ -12,6 +12,7 @@ const sentryApi = secrets[namespace].sentryApi;
 if (sentryApi?.dsn) {
   Sentry.init({
     dsn: sentryApi.dsn,
+    enableInExpoDevelopment: true,
```
donni106, line 15: what is the reason for the decision to report dev errors to sentry?

app.json
```diff
@@ -67,6 +67,12 @@
           "microphonePermission": false
         }
       ],
+      [
+        "expo-screen-orientation",
+        {
+          "initialOrientation": "DEFAULT"
+        }
```
donni106, line 74: does this make our orientation provider superfluous?

app.json
ardasnturk: since the `initialOrientation` value is set to `DEFAULT` in the plugin section, we do not need the `orientation` value here. if we set the `orientation` value to `portrait`, we cannot switch the application to landscape mode in our builds for iOS.

App.js
donni106: please check the line two lines after this one, where a comment describes the situation

package.json
donni106, line 1: did you read about `expo-image`? do we want to give it a try?

https://blog.expo.dev/expo-sdk-48-ccb8302e231#d1dc
https://docs.expo.dev/versions/v48.0.0/sdk/image

ardasnturk: of course we can try. But I'm not sure it works perfectly. 🤷‍♂️ there is a video example in the tweet I sent below. but I don't know which version of `expo-image` is used in the video.
https://twitter.com/vadimnotjustdev/status/1636631047868653571?s=46&t=TWDbqoteDDeGe0cb3GWRlQ

app.json
donni106: good info.
1. buttons should not be full width on large screens. that is too large.
2. according full width of contents we decided, that we do not want to smallen anymore, right?!
so we still need the provider but maybe need some adjustments?
but this should go in a separate ticket and process.

App.js
ardasnturk: I did rebase and deleted `enableInExpoDevelopment`.

https://github.com/smart-village-solutions/smart-village-app-app/commit/2c91de40ca84dc76fb38888ce7606499e6ea5f5e

app.json
donni106: i don't have in mind if it's just about the button or maybe anything else. the carousel image is also something i can think of.

app.json
ardasnturk: yes you are right the lock has been removed in `OnboardingManager`. but why don't we want the application to be rotated at startup?

donni106: i think it was related to splash screen images being bad at landscape mode. maybe you find some git history entry about it from the past?

ardasnturk: I have tried different builds and now I think that we don't need to set `initialOrientation` to `DEFAULT`.
It is enough to add `expo-screen-orientation` as a plugin.
Initially the application is locked in portrait mode, but afterwards it can be switched to landscape mode when locked to `DEFAULT` mode.

app.json
donni106: can you please check if there is anything else orientation aware like the button? the carousel or html image on top of screens?

package.json
donni106: can you take some time to try it out and spot some differences?

app.json
donni106: i built new development apps locally for android and ios.
on android, the orientation works fine.
sadly i have a lot of the following warnings, which i did not knew from earlier versions:
image

on ios, i cannot change orientation in the dev app. when i do an upside down, i see that the status bar goes away, but the screen stays.

any idea, why?

ardasnturk: I don't know the cause of the image source null error. This is the first time I've seen it.

---
I'm sorry to hear that. It worked fine for me. I'll try it and let you know more.

app.json
donni106: and what about androids input change on landscape? is this normal?

|portrait|landscape|
|---|---|
|Screenshot_1681733850|Screenshot_1681733863|

do we and the users want/expect it the same on landscape as on portrait?

app.json
ardasnturk: After my recent tests, I think this is the code that makes Orientation work the most correctly.

image

The application initially opens in Portrait mode and can then be switched to Landscape mode.

|landscape|portrait|
|--|--|
Simulator Screen Shot - iPhone 14 Plus - 2023-04-19 at 13 04 57 | Simulator Screen Shot - iPhone 14 Plus - 2023-04-19 at 13 04 59

app.json
ardasnturk: In this version we use `WrapperWithOrientation` component. In version `3.2.1` this component will be removed. We use `WrapperWithOrientation` component in many places but `noFlex` property is used only by `TextBlock`. We use `needLandscapeStyle` for Landscape. This is used by `Button`, `Checkbox` and `OrientationAwareIcon`.

---
I think it is caused by the `react-native-elements` update. If you try it in version `3.1.3`, it looks the same.
I have encountered Input in this way in other Android applications.

package.json
ardasnturk: Yes, I will try and write about my experiences. 👍

ardasnturk: I must say I'm very impressed with the `expo-image` package. I think the props are very useful and future-proof.
The image component of the `rn-elements` we currently use usually contains style properties. The `expo-image` package has functional features in addition to style features. It's hard to predict which props we'll use and which we won't at the moment, but I can tell you that it will make things a lot easier for a feature we want to do in the future.

For example, I think the `enableLiveTextInteraction` prop is very useful. The `expo-image` package also has its own cache system. I think we can use this as well.

I have not encountered a performance problem like the twitter link I sent before. It may be because I have not done a very big experiment. I am not sure.

I think we can use it.

app.json
donni106: Is there a prop to avoid that?

package.json
donni106: > For example, I think the enableLiveTextInteraction prop is very useful. The expo-image package also has its own cache system. I think we can use this as well.

Why do you think these are useful?

app.json
ardasnturk: I think this little code will solve our problem. 👍
image

package.json
ardasnturk: I think we will be able to use this feature in the future for any requests from customers. (`enableLiveTextInteraction`)

donni106: ok i don't fully understand, but let's keep that in mind

app.json
donni106: and how does it look like now with the change? what is the effect?

ardasnturk: Looks the same as before.

donni106 (approved):

PR #786, opened 2023-03-17
fix(VirtualisedLists): replace `ScrollView` with `FlatList`

ardasnturk:
- replaced `ScrollView` with `FlatList` to remove `VirtualisedLists` error on `HomeScreen`
- added `flatListData` to show the content of `FlatList`
- added `ConnectedImagesCarousel` and `Widgets` shown in the top section of `HomeScreen` as `ListHeaderComponent` to `FlatList`
- added `HomeService` and `About`, shown at the bottom of `HomeScreen`, to `FlatList` as `ListFooterComponent`
- added the `HomeSection` component shown in the centre section of `HomeScreen` to the `renderItem` section of `FlatList`

## How to test:
* [x] Android portrait mode
* [x] iOS portrait mode
* [x] Android landscape mode
* [x] iOS landscape mode

## Screenshots:

|home screen header|home screen (home sections)|home screen (home sections)|home screen (footer)|
|--|--|--|--|
IMG_0135 | IMG_0136 | IMG_0137 | IMG_0138

### DIFF
--- a/src/screens/HomeScreen.js
+++ b/src/screens/HomeScreen.js
@@ -57,7 +57,13 @@ const renderItem = ({ item }) => {
         rootRouteName: ROOT_ROUTE_NAMES.EVENT_RECORDS
       }
     },
-    NEWS_ITEMS_INDEX: ({ categoryId, title, titleDetail, indexCategoryIds, rootRouteName }) => {
+    NEWS_ITEMS_INDEX: ({
+      categoryId,
+      title,
+      titleDetail,
+      indexCategoryIds,
+      rootRouteName = ROOT_ROUTE_NAMES.NEWS_ITEMS
+    }) => {
       const queryVariables = { limit: 15 };
 
       if (indexCategoryIds?.length) {
@@ -73,7 +79,7 @@ const renderItem = ({ item }) => {
           titleDetail,
           query: QUERY_TYPES.NEWS_ITEMS,
           queryVariables,
-          rootRouteName: rootRouteName || ROOT_ROUTE_NAMES.NEWS_ITEMS
+          rootRouteName
         }
       };
     }
--- a/src/screens/HomeScreen.js
+++ b/src/screens/HomeScreen.js
@@ -29,11 +29,11 @@ const renderItem = ({ item }) => {
     buttonTitle,
     categoriesNews,
     fetchPolicy,
-    showData,
     navigate,
     navigation,
     query,
     queryVariables,
+    showData,
     showVolunteerEvents,
     title
   } = item;
@@ -57,13 +57,7 @@ const renderItem = ({ item }) => {
         rootRouteName: ROOT_ROUTE_NAMES.EVENT_RECORDS
       }
     },
-    NEWS_ITEMS_INDEX: ({
-      categoryId,
-      categoryTitle,
-      categoryTitleDetail,
-      indexCategoryIds,
-      rootRouteName
-    }) => {
+    NEWS_ITEMS_INDEX: ({ categoryId, title, titleDetail, indexCategoryIds, rootRouteName }) => {
       const queryVariables = { limit: 15 };
 
       if (indexCategoryIds?.length) {
@@ -75,8 +69,8 @@ const renderItem = ({ item }) => {
       return {
         name: ScreenName.Index,
         params: {
-          title: categoryTitle,
-          titleDetail: categoryTitleDetail,
+          title,
+          titleDetail,
           query: QUERY_TYPES.NEWS_ITEMS,
           queryVariables,
           rootRouteName: rootRouteName || ROOT_ROUTE_NAMES.NEWS_ITEMS
@@ -93,10 +87,10 @@ const renderItem = ({ item }) => {
     return categoriesNews.map(
       (
         {
-          categoryButton,
+          categoryButton: buttonTitle,
           categoryId,
-          categoryTitle,
-          categoryTitleDetail,
+          categoryTitle: title,
+          categoryTitleDetail: titleDetail,
           indexCategoryIds,
           rootRouteName
         },
@@ -105,7 +99,7 @@ const renderItem = ({ item }) => {
         // eslint-disable-next-line react/jsx-key
         <HomeSection
           {...{
-            buttonTitle: categoryButton,
+            buttonTitle,
             categoryId,
             fetchPolicy,
             key: index,
@@ -114,18 +108,18 @@ const renderItem = ({ item }) => {
               navigation.navigate(
                 NAVIGATION.NEWS_ITEMS_INDEX({
                   categoryId,
-                  categoryTitle,
-                  categoryTitleDetail,
+                  title,
+                  titleDetail,
                   indexCategoryIds,
                   rootRouteName
                 })
               ),
-            placeholder: <NewsSectionPlaceholder navigation={navigation} title={categoryTitle} />,
+            placeholder: <NewsSectionPlaceholder navigation={navigation} title={title} />,
             query,
             queryVariables: { ...queryVariables, categoryId },
             showVolunteerEvents,
-            title: categoryTitle,
-            titleDetail: categoryTitleDetail
+            title,
+            titleDetail
           }}
         />
       )
@@ -227,29 +221,29 @@ export const HomeScreen = ({ navigation, route }) => {
     {
       categoriesNews,
       fetchPolicy,
-      showData: showNews,
       navigation,
       query: QUERY_TYPES.NEWS_ITEMS,
-      queryVariables: { limit: 3, excludeDataProviderIds }
+      queryVariables: { limit: 3, excludeDataProviderIds },
+      showData: showNews
     },
     {
       buttonTitle: buttonPointsOfInterestAndTours,
       fetchPolicy,
-      showData: showPointsOfInterestAndTours,
       navigate: 'CATEGORIES_INDEX',
       navigation,
       query: QUERY_TYPES.POINTS_OF_INTEREST_AND_TOURS,
       queryVariables: { limit: 10, orderPoi: 'RAND', orderTour: 'RAND' },
+      showData: showPointsOfInterestAndTours,
       title: headlinePointsOfInterestAndTours
     },
     {
       buttonTitle: buttonEvents,
       fetchPolicy,
-      showData: showEvents,
       navigate: 'EVENT_RECORDS_INDEX',
       navigation,
       query: QUERY_TYPES.EVENT_RECORDS,
       queryVariables: { limit: 3, order: 'listDate_ASC' },
+      showData: showEvents,
       showVolunteerEvents,
       title: headlineEvents
     }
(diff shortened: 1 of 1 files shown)


src/screens/HomeScreen.js
```diff
@@ -112,7 +197,7 @@ export const HomeScreen = ({ navigation, route }) => {
     EVENT_RECORDS_INDEX: {
       name: ScreenName.Index,
       params: {
-        title: headlineEvents,
+        title: title,
```
donni106, line 200: just `title`

src/screens/HomeScreen.js
```diff
@@ -99,11 +99,96 @@ export const HomeScreen = ({ navigation, route }) => {
     }, 500);
   };
 
+  const flatListData = [
+    {
+      categoriesNews,
+      fetchPolicy,
+      isShow: showNews,
```
donni106, line 106: i think "is show" is weird to read and understand. what about `showData`?

src/screens/HomeScreen.js
```diff
@@ -99,11 +99,96 @@ export const HomeScreen = ({ navigation, route }) => {
...
+      navigation,
+      query: QUERY_TYPES.NEWS_ITEMS,
+      queryVariables: { limit: 3, excludeDataProviderIds }
+    },
+    {
+      buttonTitle: buttonPointsOfInterestAndTours,
+      fetchPolicy: fetchPolicy,
```
donni106, line 113: just `fetchPolicy`

src/screens/HomeScreen.js
```diff
@@ -99,11 +99,96 @@ export const HomeScreen = ({ navigation, route }) => {
...
+      isShow: showPointsOfInterestAndTours,
+      navigate: 'CATEGORIES_INDEX',
+      navigation,
+      query: QUERY_TYPES.POINTS_OF_INTEREST_AND_TOURS,
+      queryVariables: { limit: 10, orderPoi: 'RAND', orderTour: 'RAND' },
+      title: headlinePointsOfInterestAndTours
+    },
+    {
+      buttonTitle: buttonEvents,
+      fetchPolicy: fetchPolicy,
```
donni106, line 123: just `fetchPolicy`

src/screens/HomeScreen.js
```diff
@@ -146,104 +231,65 @@ export const HomeScreen = ({ navigation, route }) => {
...
+        index
+      ) => (
+        <HomeSection
+          key={index}
+          {...{
+            buttonTitle: categoryButton,
+            categoryId: categoryId,
+            fetchPolicy,
+            navigation,
+            navigate: () =>
```
donni106, line 254: why have you changed the way of passing the props? and how come, that you mix the style of `key` and the others?

src/screens/HomeScreen.js
```diff
@@ -99,11 +99,96 @@ export const HomeScreen = ({ navigation, route }) => {
     }, 500);
   };
 
+  const flatListData = [
```
donni106, line 102: why not simply `data` name here?

src/screens/HomeScreen.js
```diff
@@ -146,104 +231,65 @@ export const HomeScreen = ({ navigation, route }) => {
...
-            tintColor={colors.accent}
-          />
-        }
-      >
-        <ConnectedImagesCarousel
-          alternateAspectRatio
-          navigation={navigation}
-          publicJsonFile="homeCarousel"
-          refreshTimeKey="publicJsonFile-homeCarousel"
+  if (isShow && !!categoriesNews) {
```
donni106, line 234: `categoriesNews` is an array with objects. even ifit has no objects and is an empty array, this check would give `true`. as we `map` over the data in the condition, i suggest to check for `.length` here.

src/screens/HomeScreen.js
```diff
@@ -146,104 +231,65 @@ export const HomeScreen = ({ navigation, route }) => {
...
-        {route.params?.isDrawer && (
-          <>
-            <HomeService />
-            <About navigation={navigation} withHomeRefresh />
-          </>
-        )}
-      </ScrollView>
-    </SafeAreaViewFlex>
+  return (
+    isShow && (
```
donni106, line 277: this is the second `isShow` check. what about having an early return of `null` when it is not true to avoid checking it multiple times?

src/screens/HomeScreen.js
```diff
@@ -99,11 +99,96 @@ export const HomeScreen = ({ navigation, route }) => {
...
+          />
+        }
+        renderItem={(item) => renderItem(item)}
+      />
+    </SafeAreaViewFlex>
+  );
+};
+/* eslint-enable complexity */
+
+const renderItem = ({ item }) => {
```
donni106, line 173: please put `renderItem` method before the list, like in every other file i can think of in the code base. and pleae use `useCallback` for it.

src/screens/HomeScreen.js
```diff
@@ -99,11 +99,96 @@ export const HomeScreen = ({ navigation, route }) => {
...
+        }
+        refreshControl={
+          <RefreshControl
+            refreshing={refreshing}
+            onRefresh={refresh}
+            colors={[colors.accent]}
+            tintColor={colors.accent}
+          />
+        }
+        renderItem={(item) => renderItem(item)}
```
donni106, line 166: this can be simplified as you only take the param and passes the same param => `renderItem={renderItem}`

src/screens/HomeScreen.js
ardasnturk: Yes, it'll be more understandable.

src/screens/HomeScreen.js
ardasnturk: Because some of the props here come from the map function. So I had to change it.

The reason I wrote key outside the object was to remove the error `Missing "key" prop for element in iterator`.

src/screens/HomeScreen.js
ardasnturk: I named it `flatListData` to show that data is the data of `FlatList`, but I'll change it.

src/screens/HomeScreen.js
ardasnturk: I never thought of that, but yes, I should add that check. 👍

src/screens/HomeScreen.js
```diff
@@ -24,6 +24,130 @@ import { ScreenName } from '../types';
...
+        }
+      };
+    }
+  };
+
+  if (!showData) {
+    return null;
+  }
+
+  if (showData && !!categoriesNews?.length) {
```
donni106, line 92: this `showData` is not needed anymore, as it is always `true` when the code reaches this line

src/screens/HomeScreen.js
donni106: and if we change the flat list to another list or a scrollview somewhen, you would rename the variable here? the name should not contain something about where it is used but what it is.

ardasnturk: No, I wouldn't. You're right. 👍☺️

src/screens/HomeScreen.js
ardasnturk: https://github.com/smart-village-solutions/smart-village-app-app/commit/81aa9283fae19a01bde21a5b5012d7b7bdab7531

src/screens/HomeScreen.js
ardasnturk: https://github.com/smart-village-solutions/smart-village-app-app/commit/18f60d09e11b4f24a6b9f44081fb22ee4ae412f1

src/screens/HomeScreen.js
ardasnturk: https://github.com/smart-village-solutions/smart-village-app-app/commit/18f60d09e11b4f24a6b9f44081fb22ee4ae412f1

src/screens/HomeScreen.js
ardasnturk: https://github.com/smart-village-solutions/smart-village-app-app/commit/18f60d09e11b4f24a6b9f44081fb22ee4ae412f1

src/screens/HomeScreen.js
ardasnturk: https://github.com/smart-village-solutions/smart-village-app-app/commit/18f60d09e11b4f24a6b9f44081fb22ee4ae412f1

src/screens/HomeScreen.js
```diff
@@ -24,6 +24,130 @@ import { ScreenName } from '../types';
...
+          indexCategoryIds,
+          rootRouteName
+        },
+        index
+      ) => (
+        // eslint-disable-next-line react/jsx-key
+        <HomeSection
+          {...{
+            buttonTitle: categoryButton,
+            categoryId: categoryId,
```
donni106, line 109: `categoryId: categoryId,` -> `categoryId` ;)

src/screens/HomeScreen.js
```diff
@@ -24,6 +24,130 @@ import { ScreenName } from '../types';
...
+        queryVariables: { limit: 15, order: 'listDate_ASC' },
+        rootRouteName: ROOT_ROUTE_NAMES.EVENT_RECORDS
+      }
+    },
+    NEWS_ITEMS_INDEX: ({
+      categoryId,
+      categoryTitle,
+      categoryTitleDetail,
+      indexCategoryIds,
+      rootRouteName
```
donni106, line 65: what about setting the fallback here `rootRouteName = ROOT_ROUTE_NAMES.NEWS_ITEMS` and have just `rootRouteName` in line 82?

src/screens/HomeScreen.js
```diff
@@ -24,6 +24,130 @@ import { ScreenName } from '../types';
...
+        title,
+        query: QUERY_TYPES.EVENT_RECORDS,
+        queryVariables: { limit: 15, order: 'listDate_ASC' },
+        rootRouteName: ROOT_ROUTE_NAMES.EVENT_RECORDS
+      }
+    },
+    NEWS_ITEMS_INDEX: ({
+      categoryId,
+      categoryTitle,
+      categoryTitleDetail,
```
donni106, line 63: we could rename them here directly

```
categoryTitle: title,
categoryTitleDetail: titleDetail,
```

in order to have a cleaner return with

```
params: {
  title,
  titleDetail,
  query: QUERY_TYPES.NEWS_ITEMS,
  queryVariables,
  rootRouteName
}
```

what do you think?

src/screens/HomeScreen.js
ardasnturk: How about adding `rootRouteName` into the `data` array? I suppose it would be more useful? What do you think?

src/screens/HomeScreen.js
ardasnturk: https://github.com/smart-village-solutions/smart-village-app-app/pull/786/commits/48dc731a6e9a172ff8def17024414d4f0145c493

src/screens/HomeScreen.js
ardasnturk: I thought it would be enough to update only the data array in case of a future change. `ROOT_ROUTE_NAMES` comes from `consts`.

src/screens/HomeScreen.js
ardasnturk: https://github.com/smart-village-solutions/smart-village-app-app/commit/d6e7c412b88c3fce0989ac0a8f870b98973e96c8

src/screens/HomeScreen.js
ardasnturk: yes, that's right 🤦‍♂️

donni106 (approved):

PR #807, opened 2023-03-28
feat: add location filter for event list

ardasnturk:
- added `location` variable to `eventRecords` queries to add `location` filter via `queryVariables`
- added `isLocationFilter` prop to find out if `DropdownHeader` has a location filter
- added `isLocationFilter` prop to `dropdownEntries` function to create items in the dropdown menu and filter the list with the selected item
- added `DropdownHeader` depending on the `showEventLocationFilter` control in the `ListHeaderComponent` prop of the `ListComponent` component in `IndexScreen` to show the location filter
- alphabetised `DropdownHeader`'s props
- added `isLocationFilter` to the `selectedKey` object because the location filter is filtered according to the `value` value
- added `dropdownData` empty array to prevent the data in the location filter from changing after changing the data
- added `isLocationFilter` to make `queryVariables` `true` or `false` according to the data in it after applying the location filter of selected
- added to the `dropdownEntries` function to create the items of the location dropdown menu, if `isLocationFilter` is `true`, the `addresses` information in `data[query]` is pushed into the `dropdownData` array.
- used the `_uniqBy` function according to the `value` variable when creating entries
- returned the `keyForSelectedValueByQuery` object to arrow function format because I need two different filters in the `eventRecords` query and added the `isLocationFilter` prop to see if the filter is a location filter
- added `isLocationFilter` prop to `getAdditionalQueryVariables` function for use in `keyForSelectedValueByQuery`
- added `eventsLocationFilter` variable to filter section to switch on and off the location filter on main server
- returned to the `useCallback` function to resolve the error that the list was not refreshed after selecting the `- Alle -` filter because the `hasCategoryFilterSelection` variable did not recognise the location variable during filtering
- added `isLocationFilter` prop so that the `updateListDataByDropdown` function can detect whether it is location or category when updating the `queryVariables` state
- changed the name of `categoryFilter` to `dropdownFilter` in the texts file since there is no longer only category filter and added `location` to appear in the header of the location filter

SVA-952

## How to test:
* [x] Android portrait mode
* [x] iOS portrait mode
* [x] Android landscape mode
* [x] iOS landscape mode

## Screenshots:

|before|after|after (with location filter)|after (with category & location filter)|
|--|--|--|--|
Simulator Screen Shot - iPhone 14 - 2023-03-28 at 16 22 33 | Simulator Screen Shot - iPhone 14 - 2023-03-28 at 16 20 18 |Simulator Screen Shot - iPhone 14 - 2023-03-28 at 16 20 23 | Simulator Screen Shot - iPhone 14 - 2023-03-28 at 16 20 27

|after (with category filter)|after (with category & day filter)|after (with location & day filter)|news filter|
|--|--|--|--|
Simulator Screen Shot - iPhone 14 - 2023-03-28 at 16 20 30 |Simulator Screen Shot - iPhone 14 - 2023-03-28 at 16 20 33 |Simulator Screen Shot - iPhone 14 - 2023-03-28 at 16 20 41 | Simulator Screen Shot - iPhone 14 - 2023-03-28 at 16 21 56

|calendar view|calendar view (with location filter)|calendar view (with category filter)|calendar view (with location & category filter)|
|--|--|--|--|
Simulator Screen Shot - iPhone 14 - 2023-03-28 at 16 21 25 |Simulator Screen Shot - iPhone 14 - 2023-03-28 at 16 21 19 |Simulator Screen Shot - iPhone 14 - 2023-03-28 at 16 21 37 | Simulator Screen Shot - iPhone 14 - 2023-03-28 at 16 21 43

### DIFF
--- a/src/components/DropdownHeader.js
+++ b/src/components/DropdownHeader.js
@@ -79,8 +79,10 @@ export const DropdownHeader = ({
   updateListData
 }) => {
   const dropdownLabel = {
-    [QUERY_TYPES.EVENT_RECORDS]: texts.categoryFilter.category,
-    [QUERY_TYPES.NEWS_ITEMS]: texts.categoryFilter.dataProvider
+    [QUERY_TYPES.EVENT_RECORDS]: isLocationFilter
+      ? texts.dropdownFilter.location
+      : texts.dropdownFilter.category,
+    [QUERY_TYPES.NEWS_ITEMS]: texts.dropdownFilter.dataProvider
   }[query];
 
   const selectedKey = {
--- a/src/components/DropdownHeader.js
+++ b/src/components/DropdownHeader.js
@@ -10,11 +10,15 @@ import { QUERY_TYPES } from '../queries';
 import { DropdownSelect } from './DropdownSelect';
 import { Wrapper } from './Wrapper';
 
-const dropdownEntries = (query, queryVariables, data, excludedDataProviders) => {
+const dropdownData = [];
+
+const dropdownEntries = (query, queryVariables, data, excludedDataProviders, isLocationFilter) => {
   // check if there is something set in the certain `queryVariables`
   // if not, - Alle - will be selected in the `dropdownData`
   const selected = {
-    [QUERY_TYPES.EVENT_RECORDS]: !queryVariables.categoryId,
+    [QUERY_TYPES.EVENT_RECORDS]: isLocationFilter
+      ? !queryVariables.location
+      : !queryVariables.categoryId,
     [QUERY_TYPES.NEWS_ITEMS]: !queryVariables.dataProvider
   }[query];
 
@@ -28,14 +32,28 @@ const dropdownEntries = (query, queryVariables, data, excludedDataProviders) => 
   let entries = [];
 
   if (query === QUERY_TYPES.EVENT_RECORDS && data?.categories?.length) {
-    entries = _uniqBy(data.categories, 'name')
-      .filter((category) => !!category.upcomingEventRecordsCount)
-      .map((category, index) => ({
+    if (isLocationFilter) {
+      data[query].forEach(({ addresses }) => {
+        if (addresses.length) {
+          dropdownData.push({ value: addresses?.[0]?.city });
+        }
+      });
+
+      entries = _uniqBy(dropdownData, 'value').map((location, index) => ({
         index: index + 1,
-        id: category.id,
-        value: category.name,
-        selected: category.id === queryVariables.categoryId
+        value: location.value,
+        selected: location.value === queryVariables.location
       }));
+    } else {
+      entries = _uniqBy(data.categories, 'name')
+        .filter((category) => !!category.upcomingEventRecordsCount)
+        .map((category, index) => ({
+          index: index + 1,
+          id: category.id,
+          value: category.name,
+          selected: category.id === queryVariables.categoryId
+        }));
+    }
   } else if (query === QUERY_TYPES.NEWS_ITEMS) {
     const filteredDataProviders = data?.dataProviders?.filter(
       (dataProvider) => !excludedDataProviders.includes(dataProvider.id)
@@ -66,7 +84,7 @@ export const DropdownHeader = ({
   }[query];
 
   const selectedKey = {
-    [QUERY_TYPES.EVENT_RECORDS]: 'id',
+    [QUERY_TYPES.EVENT_RECORDS]: isLocationFilter ? 'value' : 'id',
     [QUERY_TYPES.NEWS_ITEMS]: 'value'
   }[query];
 
--- a/src/components/DropdownHeader.js
+++ b/src/components/DropdownHeader.js
@@ -53,7 +53,13 @@ const dropdownEntries = (query, queryVariables, data, excludedDataProviders) => 
   return [blankEntry, ...entries];
 };
 
-export const DropdownHeader = ({ query, queryVariables, data, updateListData }) => {
+export const DropdownHeader = ({
+  data,
+  isLocationFilter,
+  query,
+  queryVariables,
+  updateListData
+}) => {
   const dropdownLabel = {
     [QUERY_TYPES.EVENT_RECORDS]: texts.categoryFilter.category,
     [QUERY_TYPES.NEWS_ITEMS]: texts.categoryFilter.dataProvider
@@ -105,8 +111,9 @@ export const DropdownHeader = ({ query, queryVariables, data, updateListData }) 
 };
 
 DropdownHeader.propTypes = {
+  data: PropTypes.object.isRequired,
+  isLocationFilter: PropTypes.bool,
   query: PropTypes.string.isRequired,
   queryVariables: PropTypes.object.isRequired,
-  data: PropTypes.object.isRequired,
   updateListData: PropTypes.func.isRequired
 };
--- a/src/components/DropdownHeader.js
+++ b/src/components/DropdownHeader.js
@@ -91,7 +91,7 @@ export const DropdownHeader = ({
   const { state: excludedDataProviders } = usePermanentFilter();
 
   const [dropdownData, setDropdownData] = useState(
-    dropdownEntries(query, queryVariables, data, excludedDataProviders)
+    dropdownEntries(query, queryVariables, data, excludedDataProviders, isLocationFilter)
   );
 
   const selectedDropdownData = dropdownData?.find((entry) => entry.selected) || {};
@@ -100,7 +100,9 @@ export const DropdownHeader = ({
   const initialRender = useRef(true);
 
   useEffect(() => {
-    setDropdownData(dropdownEntries(query, queryVariables, data, excludedDataProviders));
+    setDropdownData(
+      dropdownEntries(query, queryVariables, data, excludedDataProviders, isLocationFilter)
+    );
   }, [data]);
 
   // influence list data when changing selected dropdown value,
@@ -113,7 +115,10 @@ export const DropdownHeader = ({
       // not pass the value, if the index is 0, because we do not want to use "- Alle -" or "0"
       // inside of `updateListData`
       !_isEmpty(selectedDropdownData) &&
-        updateListData(!!selectedDropdownData.index && selectedDropdownData[selectedKey]);
+        updateListData(
+          !!selectedDropdownData.index && selectedDropdownData[selectedKey],
+          isLocationFilter
+        );
     }
   }, [selectedDropdownData]);
 
--- a/src/screens/IndexScreen.js
+++ b/src/screens/IndexScreen.js
@@ -56,13 +56,22 @@ const isMapSelected = (query, topFilter) =>
   query === QUERY_TYPES.POINTS_OF_INTEREST &&
   topFilter.find((entry) => entry.selected).id === TOP_FILTER.MAP;
 
-const keyForSelectedValueByQuery = {
-  [QUERY_TYPES.EVENT_RECORDS]: 'categoryId',
-  [QUERY_TYPES.NEWS_ITEMS]: 'dataProvider'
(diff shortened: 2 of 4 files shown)


src/screens/IndexScreen.js
```diff
@@ -120,9 +133,14 @@ export const IndexScreen = ({ navigation, route }) => {
       [QUERY_TYPES.NEWS_ITEMS]: showNewsFilter
     }[query];
 
-  const hasCategoryFilterSelection = !!Object.prototype.hasOwnProperty.call(queryVariables, [
-    keyForSelectedValueByQuery?.[query]
-  ]);
+  const hasCategoryFilterSelection = useCallback(
+    (query, isLocationFilter) => {
```
donni106, line 137: how is the location filter related to the category filter?
i guess it needs an own selection callback keeping the `hasCategoryFilterSelection` untouched.

src/screens/IndexScreen.js
```diff
@@ -367,15 +390,15 @@ export const IndexScreen = ({ navigation, route }) => {
...
+                keyForSelectedValueByQuery(query)
               ]) &&
               !Object.prototype.hasOwnProperty.call(queryVariables, 'refetch');
 
             // apply additional data if volunteer events should be presented and
             // no category selection is made, because the category has nothing to do with
             // volunteer data
             const additionalData =
-              isCalendarWithVolunteerEvents && !hasCategoryFilterSelection
+              isCalendarWithVolunteerEvents && !hasCategoryFilterSelection(query)
```
donni106, line 401: i guess `dataVolunteerEvents` should also be ignored in case of location filter selection.

src/screens/IndexScreen.js
```diff
@@ -394,12 +417,25 @@ export const IndexScreen = ({ navigation, route }) => {
...
                               queryVariables,
-                              data: initialNewsItemsFetch && loading ? {} : data,
                               updateListData: updateListDataByDropdown
                             }}
                           />
+
+                          {query === QUERY_TYPES.EVENT_RECORDS && showEventLocationFilter && (
+                            <DropdownHeader
+                              {...{
+                                data: initialNewsItemsFetch && loading ? {} : data,
```
donni106, line 430: how are news items related here?

src/screens/IndexScreen.js
```diff
@@ -394,12 +417,25 @@ export const IndexScreen = ({ navigation, route }) => {
...
+                                isLocationFilter: true,
```
donni106, line 431: again, `queryVariables` should be enough inside the `DropdownHeader`

donni106 (changes requested):

src/components/DropdownHeader.js
```diff
@@ -28,14 +32,28 @@ const dropdownEntries = (query, queryVariables, data, excludedDataProviders) =>
   let entries = [];
 
   if (query === QUERY_TYPES.EVENT_RECORDS && data?.categories?.length) {
-    entries = _uniqBy(data.categories, 'name')
-      .filter((category) => !!category.upcomingEventRecordsCount)
-      .map((category, index) => ({
+    if (isLocationFilter) {
+      data[query].forEach(({ addresses }) => {
```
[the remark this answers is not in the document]
donni106: also with this, the navigation from POIs index to events index per drawer crashes. this should be checked again please.

src/screens/IndexScreen.js
ardasnturk: The problem here is that the `query` does not change. The `query` remains the same but the `queryVariables` change. Since the query remains the same, it is difficult to understand whether it is a `category` filter or a `location` filter.

src/screens/IndexScreen.js
ardasnturk: Yes, you're right. I'll add it. 👍

src/screens/IndexScreen.js
ardasnturk: actually the filter here has nothing to do with newsItem. I just forgot it that way. I'll fix it. 👍

src/components/DropdownHeader.js
```diff
@@ -10,11 +10,15 @@ import { QUERY_TYPES } from '../queries';
...
+const dropdownData = [];
+
+const dropdownEntries = (query, queryVariables, data, excludedDataProviders, isLocationFilter) => {
   // check if there is something set in the certain `queryVariables`
   // if not, - Alle - will be selected in the `dropdownData`
   const selected = {
-    [QUERY_TYPES.EVENT_RECORDS]: !queryVariables.categoryId,
+    [QUERY_TYPES.EVENT_RECORDS]: isLocationFilter
+      ? !queryVariables.location
+      : !queryVariables.categoryId,
```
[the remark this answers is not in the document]
ardasnturk: Unfortunately I don't think it will work, because we have two different filters and 4 different options.

The user can apply no filter at all,
the user can only apply a location filter,
the user can only apply a category filter,
the user can apply both filters.

We cannot control all of these options in an interdependent manner. We need to maintain the condition that one is selected while the other is deselected. Otherwise, after one filter is selected, the other filter is automatically selected, but without any value in the dropdown.

src/components/DropdownHeader.js
```diff
@@ -28,14 +32,28 @@ const dropdownEntries = (query, queryVariables, data, excludedDataProviders) =>
   let entries = [];
 
   if (query === QUERY_TYPES.EVENT_RECORDS && data?.categories?.length) {
-    entries = _uniqBy(data.categories, 'name')
-      .filter((category) => !!category.upcomingEventRecordsCount)
-      .map((category, index) => ({
+    if (isLocationFilter) {
+      data[query].forEach(({ addresses }) => {
+        if (addresses.length) {
```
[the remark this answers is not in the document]
ardasnturk: Yes, the address section can be empty. I'll add a check for that. 👍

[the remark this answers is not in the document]
donni106: the city is the only thing which we are interested in, right?
so we should check for this presence right away instead only for addresses being there:

`addresses?.[0]?.city && dropdownData.push({ value: addresses?.[0]?.city });` 

we do not want to push an empty value if there is an address 0 without a city, right?

src/components/DropdownHeader.js
```diff
@@ -53,21 +71,29 @@ const dropdownEntries = (query, queryVariables, data, excludedDataProviders) =>
...
+  query,
+  queryVariables,
+  updateListData
+}) => {
   const dropdownLabel = {
-    [QUERY_TYPES.EVENT_RECORDS]: texts.categoryFilter.category,
-    [QUERY_TYPES.NEWS_ITEMS]: texts.categoryFilter.dataProvider
+    [QUERY_TYPES.EVENT_RECORDS]: isLocationFilter
+      ? texts.dropdownFilter.location
+      : texts.dropdownFilter.category,
```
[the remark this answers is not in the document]
donni106: true, so let's think about a stable working solution for our cases.

src/screens/IndexScreen.js
```diff
@@ -120,9 +133,14 @@ export const IndexScreen = ({ navigation, route }) => {
       [QUERY_TYPES.NEWS_ITEMS]: showNewsFilter
     }[query];
```
donni106, line 134: isn't the new filter missing here?

src/components/DropdownHeader.js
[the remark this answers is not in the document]
donni106: true, so let's think about a stable working solution for our cases.

src/components/DropdownHeader.js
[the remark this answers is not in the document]
ardasnturk: now we have an extra query for addresses and a separate branch has been created for the refactor as we discussed.
https://github.com/smart-village-solutions/smart-village-app-app/commit/88d315d19937921c6729c08bef7e3d3ba2e03260

src/screens/IndexScreen.js
ardasnturk: https://github.com/smart-village-solutions/smart-village-app-app/pull/807/commits/167609db8f6e3f2c14dcb590081f5a57abfa2985

src/screens/IndexScreen.js
ardasnturk: https://github.com/smart-village-solutions/smart-village-app-app/pull/807/commits/167609db8f6e3f2c14dcb590081f5a57abfa2985

src/screens/IndexScreen.js
ardasnturk: now we have an extra query for addresses and a separate branch has been created for the refactor as we discussed.
https://github.com/smart-village-solutions/smart-village-app-app/commit/88d315d19937921c6729c08bef7e3d3ba2e03260

src/components/DropdownHeader.js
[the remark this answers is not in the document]
ardasnturk: https://github.com/smart-village-solutions/smart-village-app-app/pull/807/commits/167609db8f6e3f2c14dcb590081f5a57abfa2985

src/components/DropdownHeader.js
```diff
@@ -10,11 +10,15 @@ import { QUERY_TYPES } from '../queries';
 import { DropdownSelect } from './DropdownSelect';
 import { Wrapper } from './Wrapper';
 
-const dropdownEntries = (query, queryVariables, data, excludedDataProviders) => {
+const dropdownData = [];
+
+const dropdownEntries = (query, queryVariables, data, excludedDataProviders, isLocationFilter) => {
```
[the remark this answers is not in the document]
ardasnturk: https://github.com/smart-village-solutions/smart-village-app-app/pull/807/commits/167609db8f6e3f2c14dcb590081f5a57abfa2985

src/screens/IndexScreen.js
ardasnturk: now we have an extra query for addresses and a separate branch has been created for the refactor as we discussed.
https://github.com/smart-village-solutions/smart-village-app-app/commit/88d315d19937921c6729c08bef7e3d3ba2e03260

src/screens/IndexScreen.js
```diff
@@ -86,7 +95,11 @@ export const IndexScreen = ({ navigation, route }) => {
   const { isConnected, isMainserverUp } = useContext(NetworkContext);
   const { globalSettings } = useContext(SettingsContext);
   const { filter = {}, hdvt = {}, settings = {}, sections = {} } = globalSettings;
-  const { news: showNewsFilter = false, events: showEventsFilter = true } = filter;
+  const {
+    news: showNewsFilter = false,
+    events: showEventsFilter = true,
+    eventsLocationFilter: showEventLocationFilter = true
```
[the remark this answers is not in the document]
donni106: this should still be the change! and it needs to be `false` by default to not have this feature activated automatically for all customers. i will change that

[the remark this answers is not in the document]
donni106: 95ba92

donni106 (approved):

src/screens/IndexScreen.js
```diff
@@ -394,12 +431,32 @@ export const IndexScreen = ({ navigation, route }) => {
...
                           />
+
+                          {query === QUERY_TYPES.EVENT_RECORDS && !!eventRecordsAddressesData && (
+                            <DropdownHeader
+                              {...{
+                                data:
+                                  (initialEventRecordsItemsFetch && loading) ||
+                                  eventRecordsAddressesLoading
+                                    ? []
+                                    : eventRecordsAddressesData,
```
ardasnturk, line 451: `eventRecordsAddressesData` is actually an object. `eventRecordsAddressesData.eventRecordsAddresses` is an array.

Therefore, if we give array as default here, we encounter key problem. I will fix this as object.
image

ardasnturk: https://github.com/smart-village-solutions/smart-village-app-app/commit/3a54535c361440fd3d017100f8e37b5a30b6037e

PR #810, opened 2023-05-09
fix: solve android app hanging on splash screen error after update

ardasnturk:
- added `try/catch` inside the `isTokenValid` function because
  we cannot extract `ACCESS_TOKEN_EXPIRE_TIME` data from
  `SecureStore` after the update
  - getting the token fails, but the application gets stuck here because of await.
     `try/catch` prevents it from getting stuck. since the token is deleted, it is recreated
     and added to the `SecureStore`
- added `try/catch ` to `SecureStore` items used for volunteer so
  that the data in the application can be loaded properly
- added `ACCESS_TOKEN_EXPIRE_TIME` constant because we use more than one in the code

SVA-704

## How to test:
* [x] Android portrait mode

### DIFF
--- a/src/auth.js
+++ b/src/auth.js
@@ -2,13 +2,25 @@ import * as SecureStore from 'expo-secure-store';
 
 import { namespace, secrets } from './config';
 
+const ACCESS_TOKEN_EXPIRE_TIME = 'ACCESS_TOKEN_EXPIRE_TIME';
+
 /**
  * Check if a auth token is expired or still valid.
  * The expire time is saved as seconds from 01.01.1970 00:00:00 UTC, so
  * we need to divide Date.now() by 1000, which otherwise would return miliseconds.
  */
 const isTokenValid = async () => {
-  const accessTokenExpireTime = await SecureStore.getItemAsync('ACCESS_TOKEN_EXPIRE_TIME');
+  let accessTokenExpireTime = null;
+
+  // The reason for the problem of staying in SplashScreen that occurs after the application is
+  // updated on the Android side is the inability to obtain the token here.
+  // For this reason, try/catch is used here and the problem of getting stuck in SplashScreen is solved.
+  try {
+    accessTokenExpireTime = await SecureStore.getItemAsync(ACCESS_TOKEN_EXPIRE_TIME);
+  } catch {
+    // Token deleted here so that it can be recreated
+    await SecureStore.deleteItemAsync(ACCESS_TOKEN_EXPIRE_TIME);
+  }
 
   if (!accessTokenExpireTime) return false;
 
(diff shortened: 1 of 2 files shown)


src/helpers/volunteerHelper.ts
```diff
@@ -28,7 +28,16 @@ export const storeVolunteerAuthToken = (authToken?: string) => {
   }
 };
 
-export const volunteerAuthToken = () => SecureStore.getItemAsync(VOLUNTEER_AUTH_TOKEN);
+export const volunteerAuthToken = async () => {
```
donni106, line 31: is the volunteer section a problem? why have you added code changes here? this has nothing todo with app start and splash screen behavior

src/auth.js
```diff
@@ -8,7 +8,13 @@ import { namespace, secrets } from './config';
  * we need to divide Date.now() by 1000, which otherwise would return miliseconds.
  */
 const isTokenValid = async () => {
-  const accessTokenExpireTime = await SecureStore.getItemAsync('ACCESS_TOKEN_EXPIRE_TIME');
+  let accessTokenExpireTime = null;
+  try {
+    accessTokenExpireTime = await SecureStore.getItemAsync('ACCESS_TOKEN_EXPIRE_TIME');
+  } catch (e) {
+    // this only throws if they reinstall on android
```
donni106, line 15: who is "they"? and how is our code effected on reinstalls on android? should the code have a condition for android devices if this only effects android? what happens on ios?

src/auth.js
```diff
@@ -8,7 +8,13 @@ import { namespace, secrets } from './config';
...
+    await SecureStore.deleteItemAsync('ACCESS_TOKEN_EXPIRE_TIME');
```
donni106, line 16: why are you deleting a token if getting that token fails?

donni106 (changes requested):

src/helpers/volunteerHelper.ts
ardasnturk: That's what I thought. But as far as I can see, it prevents the data in the application from loading.
This line of code causes a warning when the application is first launched. I updated the code to remove this warning.

src/auth.js
ardasnturk: On Android, after the application is reinstalled, the value we want from the `SecureStore` must be deleted to recreate it.
I haven't tried it for iOS. That's my fault. I will try and write about my experiences. Maybe an extra logic can be added for Android. But I don't think we will have a problem on the iOS side. it will not enter the catch blog.

src/auth.js
ardasnturk: Getting the token fails, but the application gets stuck here because of await. `trycatch` prevents it from getting stuck. Since the token is deleted, it is recreated and added to the `SecureStore`.

donni106: ok i understand more now. please write something like this in the commit message and a comment, that explains the whole try catch token stuff and why we are doing it

src/helpers/volunteerHelper.ts
donni106: when we first start the app there is some warning about volunteer auth token?

src/auth.js
ardasnturk: https://github.com/smart-village-solutions/smart-village-app-app/commit/7b1495e9105b7a17df02662514498c9d7ab6468f

src/auth.js
ardasnturk: https://github.com/smart-village-solutions/smart-village-app-app/commit/fc0c2bbe7a779cb79d601672e1f4baa4b5eef5a5

donni106 (approved):

PR #831, opened 2023-06-01
feat(waste calendar): show results when street input is pressed

ardasnturk:
- added `isStreetInputFocused` to recognise if input is pressed
- updated the `isStreetInputFocused` value to `true` in the `onFocus` prop to
  recognise that it is focused on the street input
- added `isStreetInputFocused` control in `filterStreets` function for automatic
  showing of streets
- updated `isStreetInputFocused` to `false` in the `onFocus` prop in the city input
  to ensure that the streets are not visible
- added `isInputAutoFocus` value to `globalSettings` to activate the feature on the server
- removed `streetCount` limitation for listing all streets after auto-focus feature is activated
- added `height` property to `flatListProps` to display all elements in the list and make
  the list scrollable

SVA-833

## Screenshots:

|before|after|
|--|--|
IMG_3945 | IMG_3947

### DIFF
--- a/src/screens/WasteCollectionScreen.js
+++ b/src/screens/WasteCollectionScreen.js
@@ -34,6 +34,7 @@ import { NetworkContext } from '../NetworkProvider';
 import { getInAppPermission, showPermissionRequiredAlert } from '../pushNotifications';
 import { getQuery, QUERY_TYPES } from '../queries';
 import { SettingsContext } from '../SettingsProvider';
+import { OrientationContext } from '../OrientationProvider';
 
 const dotSize = 6;
 
@@ -83,10 +84,12 @@ const getLocationData = (streetData) => {
 // eslint-disable-next-line complexity
 export const WasteCollectionScreen = ({ navigation }) => {
   const { isConnected, isMainserverUp } = useContext(NetworkContext);
+  const { dimensions } = useContext(OrientationContext);
   const { globalSettings } = useContext(SettingsContext);
   const [inputValueCity, setInputValueCity] = useState('');
   const [inputValueCitySelected, setInputValueCitySelected] = useState(false);
   const [inputValue, setInputValue] = useState('');
+  const [isStreetInputFocused, setIsStreetInputFocused] = useState(false);
   const [selectedStreetId, setSelectedStreetId] = useState();
 
   const addressesRefreshTime = useRefreshTime('waste-addresses');
@@ -132,6 +135,7 @@ export const WasteCollectionScreen = ({ navigation }) => {
     }
   );
 
+  const isInputAutoFocus = globalSettings?.settings?.wasteAddresses?.isInputAutoFocus;
   const wasteAddressesStreetCount = globalSettings?.settings?.wasteAddresses?.streetCount || 5;
   const wasteAddressesTwoStep = !!globalSettings?.settings?.wasteAddresses?.twoStep;
 
@@ -169,7 +173,7 @@ export const WasteCollectionScreen = ({ navigation }) => {
   const filterStreets = useCallback(
     (currentInputValue, addressesData) => {
       if (wasteAddressesTwoStep && inputValueCity === '') return [];
-      if (currentInputValue === '') return [];
+      if (isInputAutoFocus && currentInputValue === '' && !isStreetInputFocused) return [];
 
       const streets = addressesData
         ?.filter((address) => {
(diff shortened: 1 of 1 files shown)


src/screens/WasteCollectionScreen.js
donni106, line 1: I am missing a global setting to activate this feature. Please consider a flag for activating/deactivating that.

In my test with city "Rathenow" when the street overlay appears, I can only see a specific amount of streets: `wasteAddressesStreetCount`. 

Simulator Screen Shot - iPhone 12 - 2023-06-07 at 10 38 35

When typing something this makes sense. But with activated feature of showing the overlay without typing we need to be able to see all the streets and scroll them entirely. Otherwise it makes no sense from my point of view, trying to think as a user.

donni106 (changes requested):

src/screens/WasteCollectionScreen.js
ardasnturk: - `isInputAutoFocus` value has been added to `globalSettings` to use this feature. 
- If `isInputAutoFocus` is true, the `wasteAddressesStreetCount` limitation in the list is removed.
- Also added `height` property to `flatListProps` to make the list scrollable.

https://github.com/smart-village-solutions/smart-village-app-app/commit/b7cb7effac8f2e9129ffdda828b87e9172f66366

src/screens/WasteCollectionScreen.js
```diff
@@ -169,7 +173,7 @@ export const WasteCollectionScreen = ({ navigation }) => {
   const filterStreets = useCallback(
     (currentInputValue, addressesData) => {
       if (wasteAddressesTwoStep && inputValueCity === '') return [];
-      if (currentInputValue === '') return [];
+      if (currentInputValue === '' && !isStreetInputFocused && isInputAutoFocus) return [];
```
donni106, line 176: please check the gobal setting first, before checking the values for the logic. if the global setting is false, we do not have to check the other values, right? this is more naturally readable in my eyes.

src/screens/WasteCollectionScreen.js
```diff
@@ -182,9 +186,15 @@ export const WasteCollectionScreen = ({ navigation }) => {
         })
         ?.filter((address) => (wasteAddressesTwoStep ? address.city === inputValueCity : true));
 
-      return _sortBy(streets, 'street').slice(0, wasteAddressesStreetCount);
+      const sortedList = _sortBy(streets, 'street');
```
donni106, line 189: i would suggest `sortedStreets` here instead of `list` which does not say anything about the object.

src/screens/WasteCollectionScreen.js
```diff
@@ -349,10 +365,16 @@ export const WasteCollectionScreen = ({ navigation }) => {
             data={filteredStreets}
             disableFullscreenUI
             flatListProps={{
-              renderItem: renderSuggestion
+              renderItem: renderSuggestion,
+              height: dimensions.height - normalize(220)
```
donni106, line 369: you wrote

> - added `height` property to `flatListProps` to display all elements in the list and make
  the list scrollable

but why `dimensions.height - normalize(220)`? this is not described in the commit.

ardasnturk: I set it to `normalise(220)` so that the list does not overflow the bottom of the screen.

src/screens/WasteCollectionScreen.js
ardasnturk: You're talking about this control?
```
if (isInputAutoFocus && currentInputValue === '' && !isStreetInputFocused) return [];
```
 or 
```
if (isInputAutoFocus) return [];
if (currentInputValue === '' && !isStreetInputFocused) return [];
```

What do you mean by globalSettings?

src/screens/WasteCollectionScreen.js
ardasnturk: https://github.com/smart-village-solutions/smart-village-app-app/commit/9e1dd698a77e4699502272048afcdf6b63799d17

src/screens/WasteCollectionScreen.js
donni106: > What do you mean by globalSettings?

`isInputAutoFocus `

you understood correctly with `if (isInputAutoFocus && currentInputValue === '' && !isStreetInputFocused) return [];`

donni106 (approved):

src/screens/WasteCollectionScreen.js
ardasnturk: I can't say it's perfect, but we can see the full list on the screen.

|iPad|iPhone 14|iPhone SE|iPhone 14 Pro Max|
|--|--|--|--|
Simulator Screenshot - iPad mini (6th generation) - 2023-06-14 at 14 18 07 | Simulator Screenshot - iPhone 14 - 2023-06-14 at 14 18 08 | Simulator Screenshot - iPhone SE (3rd generation) - 2023-06-14 at 14 18 09 | Simulator Screenshot - iPhone 14 Pro Max - 2023-06-14 at 14 18 11

donni106: ok

src/screens/WasteCollectionScreen.js
ardasnturk: https://github.com/smart-village-solutions/smart-village-app-app/commit/4e36378f9f2ac4c5bc8c431ed435eddc4c91ecf0

PR #833, opened 2023-06-05
fix(waste calendar): the type problem in the waste calendar reminder

ardasnturk:
- updated the types of `city` and `zip` before validation to prevent them
  from coming from the server with different types
- added to `WasteCalendar` types to use `ReminderSettingJson` from anywhere

SVA-1028

It was not possible to create a new reminder because of a previously saved reminder. The reason for this problem is that the types here are different from what we expected.

**Difference between data**
|with `city` & `zip`|without `city` & `zip`|
|--|--|
image | image

### DIFF
--- a/src/jsonValidation/reminderSettingsValidation.ts
+++ b/src/jsonValidation/reminderSettingsValidation.ts
@@ -1,18 +1,8 @@
 import _isObjectLike from 'lodash/isObjectLike';
 
-import { ReminderSettings } from '../types';
+import { ReminderSettingJson, ReminderSettings } from '../types';
 
 import { isArrayOfType } from './basicTypeValidation';
-
-type ReminderSettingJson = {
-  city: string;
-  id: number;
-  notify_at: string;
-  notify_days_before: number;
-  notify_for_waste_type: string;
-  street: string;
-  zip: string;
-};
 
 const isValidReminderSetting = (data: unknown): data is ReminderSettingJson => {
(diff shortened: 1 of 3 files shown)


donni106:
> * updated `ReminderSettingJson` types because `city` and `zip` data can have `null` values

can we force `""` instead of `null` to not have the need of changing the types?

src/jsonValidation/reminderSettingsValidation.ts
```diff
@@ -29,14 +29,17 @@ const isValidReminderSetting = (data: unknown): data is ReminderSettingJson => {
     zip
   } = data as ReminderSettingJson;
 
+  const cityString = city ?? '';
+  const zipString = zip ?? '';
```
donni106, line 33: naming a variable with a type is generally bad practice. why is this needed for validation?
can't we ensure the correct value without touching the validation?

ardasnturk: I added a new method to update the data after pulling it from the server. In this method, if the city and zip values are null, they are returned to empty string and sent to validation in that form.

https://github.com/smart-village-solutions/smart-village-app-app/commit/13ebc43362b441c180cd3ca313b2a44efe3c9bb6

donni106: yeah! i like that ver much more

donni106 (approved):

PR #840, opened 2023-06-13
fix: remove spaces in the link

ardasnturk:
- deleted spaces in links because the links could not be opened by the application due to spaces in some links

SVA-1036

### DIFF
--- a/src/helpers/linkHelper.js
+++ b/src/helpers/linkHelper.js
@@ -14,7 +14,7 @@ import * as Linking from 'expo-linking';
  */
 function ensureProtocol(link) {
   const protocolRegExpMatch = link.match(/^https?:|tel:|mailto:|maps:|geo:/);
-  let linkWithProtocol = link;
+  let linkWithProtocol = link.replaceAll(' ', '');
 
   if (!protocolRegExpMatch) {
     linkWithProtocol = `http://${link}`;


src/helpers/linkHelper.js
```diff
@@ -20,7 +20,12 @@ function ensureProtocol(link) {
     linkWithProtocol = `http://${link}`;
   }
 
-  return linkWithProtocol;
+  return spaceDeletion(linkWithProtocol);
```
donni106, line 23: i suggest `link.replaceAll(' ', '');` when setting `let linkWithProtocol = link` instead of creating a method that is used only once? and i see no reason in removing spaces from `linkWithProtocol` as it is the `link` prefixed with http://

donni106, line 23: i suggest `link.replaceAll(' ', '');` when setting `let linkWithProtocol = link` instead of creating a method that is used only once? and i see no reason in removing spaces from `linkWithProtocol` as it is the `link` prefixed with http://

donni106 (changes requested):

src/helpers/linkHelper.js
ardasnturk: Yes. It is unnecessary to create a new method for a feature that is used once. I just created it to fit the structure in the code. I have updated it here: https://github.com/smart-village-solutions/smart-village-app-app/commit/b44a3ca6bb3c3e1d8354bf7f505c88719a9f0cba

donni106 (approved):

PR #855, opened 2023-06-21
fix: solve serial events display error in the list

ardasnturk:
- added `refetch` to refresh the data in the list after selecting any day from the calendar
- added `fetchPolicy` for `refetch` to work
- added `loading` control to the `data` props of `ListComponent` to show the user that the data has been refreshed

SVA-665

### DIFF
--- a/src/components/Calendar.tsx
+++ b/src/components/Calendar.tsx
@@ -13,7 +13,7 @@ import { SettingsContext } from '../SettingsProvider';
 import { colors, consts, normalize, texts } from '../config';
 import { graphqlFetchPolicy, parseListItemsFromQuery } from '../helpers';
 import { setupLocales } from '../helpers/calendarHelper';
-import { QUERY_TYPES, getQuery } from '../queries';
+import { QUERY_TYPES, getFetchMoreQuery, getQuery } from '../queries';
 import { ScreenName, Calendar as TCalendar } from '../types';
 
 import { EmptyMessage } from './EmptyMessage';
@@ -92,11 +92,28 @@ export const Calendar = ({ query, queryVariables, calendarData, isLoading, navig
     getMarkedDates(calendarData, dotCount, today)
   );
 
-  const { data, loading, refetch } = useQuery(getQuery(QUERY_TYPES.EVENT_RECORDS), {
+  const { data, loading, refetch, fetchMore } = useQuery(getQuery(QUERY_TYPES.EVENT_RECORDS), {
     fetchPolicy,
     variables: queryVariableWithDateRange,
     skip: !subList
   });
+
+  const fetchMoreData = () =>
+    fetchMore({
+      query: getFetchMoreQuery(query),
+      variables: {
+        ...queryVariableWithDateRange,
+        offset: data?.[query]?.length
+      },
+      updateQuery: (prevResult, { fetchMoreResult }) => {
+        if (!fetchMoreResult || !fetchMoreResult[query].length) return prevResult;
+
+        return {
+          ...prevResult,
+          [query]: [...prevResult[query], ...fetchMoreResult[query]]
+        };
+      }
+    });
 
   const onDayPress = useCallback(
     async (day: DateData) => {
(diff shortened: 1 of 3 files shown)


src/components/EventList.js
```diff
@@ -59,7 +59,7 @@ export const EventList = ({
   return (
     <SectionList
       initialNumToRender={
-        data?.length < MAX_INITIAL_NUM_TO_RENDER ? data.length : MAX_INITIAL_NUM_TO_RENDER
+        data?.length < MAX_INITIAL_NUM_TO_RENDER ? data.length + 1 : MAX_INITIAL_NUM_TO_RENDER
```
donni106, line 62: do we need to add +1 always, also for `MAX_INITIAL_NUM_TO_RENDER` in order to render 20 instead of 19 maybe?

ardasnturk: Good point. I'm not sure that's necessary. I think pagination comes into play when you get to the end of the list anyway. 👍

donni106: yes, at the end, when we load more available entries to lists

ardasnturk: So I don't think we need `MAX_INITIAL_NUM_TO_RENDER` + 1.

donni106: but if we want to render always 20 and this does not work we maybe miss some entries?

ardasnturk: I have tried it for more than 20 event days (28.06) without any problems.

donni106: what if we use always `initialNumToRender={MAX_INITIAL_NUM_TO_RENDER}`? if we have just 5 events, it is also ok to have initial number to 20, right?

src/components/Calendar.tsx
donni106, line 1: in calendar view we do not need the date in the overline of list items.
this should be the same as in the list view under each section:

image

src/components/Calendar.tsx
```diff
@@ -153,7 +160,7 @@ export const Calendar = ({ query, queryVariables, calendarData, isLoading, navig
 
       {subList && (
         <ListComponent
-          data={buildListItems(data)}
+          data={loading ? [] : buildListItems(data)}
           horizontal={false}
```
donni106, line 164: I found out, that `horizontal` is not needed here, or am I wrong?

donni106: same for `showBackToTop`, which does not exist for `EventList`

ardasnturk: Yes, you are right! https://github.com/smart-village-solutions/smart-village-app-app/pull/855/commits/2a84d5989617f2fd9b25e76f484c2352ccab602c

src/components/EventList.js
ardasnturk: I don't think it'll be a problem, but I'd still prefer to keep it like this. I think it could cause errors that we can't see at the moment.

src/components/Calendar.tsx
ardasnturk: https://github.com/smart-village-solutions/smart-village-app-app/pull/855/commits/e7ce03c00dc8cffcaf8597a27b89f433da6c2143

src/components/EventList.js
donni106: but what is the benefit of having a more complex logic with some random +1?
if we have 5 data, it should not matter if we have initial num to render 5,6 or 20.
if we have 25 data, it should be 20.
so having 20 as initalnumtorender should be fine all the time without confusing conditions or logics.
you know what i mean?

ardasnturk: Yeah, I understand. I think you're absolutely right. It shouldn't matter 5 or 6 or 30. Anyway, if the size of the data is more than 20, 20 cells are rendered. If there are 5 data, there will be 5.

ardasnturk: https://github.com/smart-village-solutions/smart-village-app-app/pull/855/commits/1588ac1d17b0e4914c02cd9ffe7acbe0bae91239

src/components/Calendar.tsx
ardasnturk: |before|after|
|--|--|
Simulator Screenshot - iPhone 14 Pro Max - 2023-06-28 at 14 18 31 | Simulator Screenshot - iPhone 14 Pro Max - 2023-06-28 at 14 18 04

donni106: looking awesome now

donni106 (approved):

PR #856, opened 2023-06-26
feat: add whatsapp link redirection to ios app

ardasnturk:
- added `LSApplicationQueriesSchemes` to `info.plist` to redirect whatsapp links in the webview in the application to whatsapp application

SVA-1047

## How to test:
* [x] iOS portrait mode
* [x] iOS landscape mode

### DIFF
--- a/app.json
+++ b/app.json
@@ -26,7 +26,8 @@
         "NSPhotoLibraryUsageDescription": "Diese App nutzt die Medienbibliothek für die Auswahl eines Profilbildes.",
         "NSLocationWhenInUseUsageDescription": "Diese App nutzt die Standortbestimmung um Ihre aktuelle Position auf der Karte darzustellen.",
         "CFBundleLocalizations": ["de"],
-        "CFBundleDevelopmentRegion": "de_DE"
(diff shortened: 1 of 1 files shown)


app.json
```diff
@@ -26,7 +26,8 @@
         "NSPhotoLibraryUsageDescription": "Diese App nutzt die Medienbibliothek für die Auswahl eines Profilbildes.",
         "NSLocationWhenInUseUsageDescription": "Diese App nutzt die Standortbestimmung um Ihre aktuelle Position auf der Karte darzustellen.",
         "CFBundleLocalizations": ["de"],
-        "CFBundleDevelopmentRegion": "de_DE"
+        "CFBundleDevelopmentRegion": "de_DE",
+        "LSApplicationQueriesSchemes": ["whatsapp"]
```
donni106, line 30: what is the source of this? where do you got it from? what is the default value? do we need to add more here to make other things still work?

ardasnturk: It's mentioned in the expo documents. If we want to add more redirects to the application then we need to add them here.

https://docs.expo.dev/guides/linking/#custom-url-schemes

donni106 (approved):

PR #857, opened 2023-06-27
feat: integrate custom links for android

ardasnturk:
- added `withAndroidMailQueriesAndWhatsappPackage` config for `mailto:` and `whatsapp:` custom link in webview to work on Android devices

SVA-1047

In order to run any custom link inside the webview, it is necessary to update the native files of Android. With Expo config-plugins, I updated the AndroidManifest.xml file in the Android folder (android/app/src/main/AndroidManifest.xml) so that the application can be redirected with `mailto:` and` whatsapp:` links.

I got help from the following issue.
[React-Native Issue](https://github.com/facebook/react-native/issues/30909)

To try it out:
1. select branch and run `yarn prebuild` command
2. connect the android device to the computer with the cable, run the `yarn android -d` command and select your device
3. while waiting for the build to finish, perform the following actions
4. change the `settings` prop of the `StorySection` component in `NewsItem.js` as follows.
```
settings={{ displayOnlySummary: 'true', onlySummaryLinkText: 'Weiterlesen' }}
```
5. after the build process is successfully completed, follow these steps
6. then enter the news "Test News mit Link"
7. click on the "Weiterlesen" button
8. you will see links for mail and whatsapp on the page that opens. you can try whether the feature works with these links.

### DIFF
--- a/app.json
+++ b/app.json
@@ -74,7 +74,8 @@
           "initialOrientation": "PORTRAIT"
         }
       ],
-      "sentry-expo"
+      "sentry-expo",
+      "./config-plugins/withAndroidMailQueriesAndWhatsappPackage"
     ]
   }
 }
--- a/config-plugins/withAndroidMailQueriesAndWhatsappPackage.js
+++ b/config-plugins/withAndroidMailQueriesAndWhatsappPackage.js
@@ -1,0 +1,43 @@
+// eslint-disable-next-line @typescript-eslint/no-var-requires
+const { withAndroidManifest } = require('@expo/config-plugins');
+
+function addIntentToQuery({ mod, intent }) {
+  const copyMod = { ...mod };
+  if (!copyMod.modResults.manifest.queries) {
+    copyMod.modResults.manifest.queries = [{ intent: [] }];
+  } else if (!copyMod.modResults.manifest.queries[0]) {
+    copyMod.modResults.manifest.queries = [];
+  }
+
+  copyMod.modResults.manifest.queries.push({
+    ...(!intent.dataScheme && { package: [{ $: { 'android:name': intent.actionName } }] }),
+    ...(intent.dataScheme && {
+      intent: {
+        action: [{ $: { 'android:name': intent.actionName } }],
+        data: [{ $: { 'android:scheme': intent.dataScheme } }]
+      }
+    })
+  });
+
+  return copyMod;
+}
+
+const withAndroidMailQueriesAndWhatsappPackage = (config) => {
+  const mod = withAndroidManifest(config, (mod) => {
+    let modified = addIntentToQuery({
+      mod,
+      intent: { actionName: 'android.intent.action.SENDTO', dataScheme: 'mailto' }
+    });
+
+    modified = addIntentToQuery({
+      mod,
+      intent: { actionName: 'com.whatsapp' }
+    });
+
+    return modified;
+  });
+
+  return mod;
+};
+
+module.exports = withAndroidMailQueriesAndWhatsappPackage;


app.json
```diff
@@ -74,7 +74,9 @@
           "initialOrientation": "PORTRAIT"
         }
       ],
-      "sentry-expo"
+      "sentry-expo",
+      "./config-plugins/withAndroidMailQueries",
```
donni106, line 78: this does not match the file name! so how could it work? please rename the file accordingly or remove it, if not necessary.

config-plugins/withAndroidWhatsappPackage.js
```diff
@@ -0,0 +1,30 @@
+// eslint-disable-next-line @typescript-eslint/no-var-requires
+const { withAndroidManifest } = require('@expo/config-plugins');
+
+function addIntentToQuery({ mod, intent }) {
```
donni106, line 4: why is this `addIntentToQuery` a bit different than in the other file? couldn't this be the same function exported in a file and used in both config-plugin files?

config-plugins/withAndroidWhatsappPackage.js
```diff
@@ -0,0 +1,30 @@
...
+  });
+  return copyMod;
+}
+
+const withAndroidWhatsappPackage = (config) => {
+  const mod = withAndroidManifest(config, (mod) => {
+    return addIntentToQuery({
+      mod,
+      intent: { actionName: 'com.whatsapp' }
+    });
```
donni106, line 24: this is the only difference, right? couldn't mail and whatsapp happen in one method, that calls `addIntentToQuery` twice?

donni106 (changes requested):

config-plugins/withAndroidWhatsappPackage.js
donni106: if we add both intents in one file, you can also keep `addIntentToQuery` in that one remaining file and do not need to export import

app.json
ardasnturk: Yes, you're right. I will update.

config-plugins/withAndroidWhatsappPackage.js
ardasnturk: I updated the code in this way. 

```
export const addIntentToQuery = ({ mod, intent }) => {
  const copyMod = { ...mod };
  if (!copyMod.modResults.manifest.queries) {
    copyMod.modResults.manifest.queries = [{ intent: [] }];
  } else if (!copyMod.modResults.manifest.queries[0]) {
    copyMod.modResults.manifest.queries = [];
  } else if (!copyMod.modResults.manifest.queries[0].intent?.[0]) {
    copyMod.modResults.manifest.queries[0].intent = [];
  }

  copyMod.modResults.manifest.queries.push({
    intent: { action: [{ $: { 'android:name': intent.actionName } }] },
    package: [{ $: { 'android:name': intent.actionName } }],
    ...(intent.dataScheme && { data: [{ $: { 'android:scheme': intent.dataScheme } }] })
  });

  return copyMod;
};
```

There is also a difference between Whatsapp and Mail plugins. I need to add Whatsapp to `queries` as a `package`. I need to add the Mail feature as an `action` in the `intent`. I didn't want to use it in one place because I thought extra controls might make the code more complex.

I edited `addIntentToQuery` and exported it. When I ran the `prebuild` command, I got the following error.

image

---

After running the plugin, AndroidManifest needs to be updated as follows.
|before|after|
|--|--|
image | SCR-20230703-ksgb

config-plugins/withAndroidWhatsappPackage.js
ardasnturk: I think I can find a way round that.

image

I will update the code in this way. Yes, it will be a bit complex, but this way we can solve everything in one place.

ardasnturk: https://github.com/smart-village-solutions/smart-village-app-app/commit/001f367f34ae03dade0fa83279b273fd37231cee

config-plugins/withAndroidWhatsappPackage.js
ardasnturk: https://github.com/smart-village-solutions/smart-village-app-app/commit/001f367f34ae03dade0fa83279b273fd37231cee

app.json
ardasnturk: https://github.com/smart-village-solutions/smart-village-app-app/commit/001f367f34ae03dade0fa83279b273fd37231cee

config-plugins/withAndroidWhatsappPackage.js
ardasnturk: AndroidManifest after updating the code.
|before|after|
|--|--|
SCR-20230703-ksgb | image

donni106 (approved):

PR #858, opened 2023-06-27
fix: resolve the error preventing the bbnavi website from opening

ardasnturk:
- changed `applicationNameForUserAgent` to `undefined` for bbnavi website because bbnavi prevents the website from opening

BBNAV-321

### DIFF
--- a/src/screens/WebScreen.js
+++ b/src/screens/WebScreen.js
@@ -42,7 +42,9 @@ export const WebScreen = ({ route }) => {
         // https://github.com/react-native-webview/react-native-webview/blob/19980d888d66554875f3ac64b3e8a35bd7ad998b/src/WebViewTypes.ts#L378-L389
         decelerationRate="normal"
         // https://github.com/react-native-webview/react-native-webview/blob/master/docs/Reference.md#applicationnameforuseragent
-        applicationNameForUserAgent={appJson.expo.scheme}
+        applicationNameForUserAgent={
+          !webUrl.includes('bbnavi.de') ? appJson.expo.scheme : undefined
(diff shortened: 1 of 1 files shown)


src/screens/WebScreen.js
```diff
@@ -42,7 +42,7 @@ export const WebScreen = ({ route }) => {
         // https://github.com/react-native-webview/react-native-webview/blob/19980d888d66554875f3ac64b3e8a35bd7ad998b/src/WebViewTypes.ts#L378-L389
         decelerationRate="normal"
         // https://github.com/react-native-webview/react-native-webview/blob/master/docs/Reference.md#applicationnameforuseragent
-        applicationNameForUserAgent={appJson.expo.scheme}
+        applicationNameForUserAgent={!webUrl.includes('bbnavi') ? appJson.expo.scheme : ''}
```
donni106, line 45: this prop is not required, so I would suggest to set `undefined` when not used instead of an empty string

donni106, line 45: please check for `bbnavi.de` as this is more explicit...just bbnavi could be in some other url as well maybe

ardasnturk: https://github.com/smart-village-solutions/smart-village-app-app/commit/051c46c54eabbbea189e34caecce23c33ad76bd3

ardasnturk: https://github.com/smart-village-solutions/smart-village-app-app/commit/051c46c54eabbbea189e34caecce23c33ad76bd3

donni106 (approved):

PR #860, opened 2023-07-03
feat(opening times): dates with parsed TMB year

donni106:
In the future, TMB pois will have adjusted dates for specific cases, being returned from the main server with year 1970. This way we can identify them to adjust checking opening time and rendering.

|before|after|
|---|---|
|Simulator Screen Shot - iPhone 12 - 2023-07-03 at 17 20 55|Simulator Screen Shot - iPhone 12 - 2023-07-03 at 17 18 41|

In order to test with the POI Eiscafé "Zur Postmeile" you need to set the `TMB_YEAR_TO_PARSE` to 2020.

SVA-985


ardasnturk (approved):
It looks good.

PR #861, opened 2023-07-04
fix: a space error between two inputs

ardasnturk:
- added not to show the results if `inputValueCitySelected` is `true` to the city `Autocomplete` component to solve the problem of the space between the two inputs that occurs if there are more suggestions related to the selected city name after a place is selected from the city input

SVA-833

## How to test:
* [x] Android
* [x] iOS

|Android before|Android after|Android before|Android after|
|--|--|--|--|
Screenshot_20230704-112858 | Screenshot_20230704-112802 | Screenshot_20230704-112903 | Screenshot_20230704-112808

|iOS before|iOS after|iOS before|iOS after|
|--|--|--|--|
IMG_4106 | IMG_4108 | IMG_4107 | IMG_4109


donni106 (approved):
I have marked it as part of 3.2.4, to not need a new release with versioning for this bugfix.

PR #876, opened 2023-07-25
fix: add control to prevent image duplication for android

ardasnturk:
- added platform control to the `Image` component to prevent the `TalkBack` feature used in Android from saying `bild` twice at the end of any image

SBB-50


src/components/Image.js
```diff
@@ -104,7 +104,7 @@ export const Image = ({
         placeholderStyle={styles.placeholderStyle}
         accessible={!!sourceProp?.captionText}
         accessibilityLabel={`${sourceProp.captionText ? sourceProp.captionText : ''} ${
-          consts.a11yLabel.image
+          device.platform === 'ios' && consts.a11yLabel.image
```
donni106, line 107: for `captionText` we have `...? ... : ''` and now you have used `&&`, why? could it be possible for some android, that TalkBack reads the `undefined`? i suggest having the same way of handling conditions for the same prop here and go for `...? ... : ''`, as we can be sure to have an empty string, that will not be read.

ardasnturk: Yeah, you're right. I changed it.

https://github.com/smart-village-solutions/smart-village-app-app/commit/f77c9eb27445cfab59a71cdb207e97cdd743d320

donni106 (approved):

PR #884, opened 2023-08-03
feat: adjust zoom value of maps

ardasnturk:
- added `zoomLevelForMaps` in `globalSettings` to `Map.tsx` so that zoom values of maps can be set from main server
- added the `isMultipleMarkersMap` prop to `Map.tsx` because we want the maps to have a different zoom value for POI and added the zoom value of the POI map according to this prop
- moved `Map.tsx` up in the code as we now need to use `globalSettings` for the zoom value
- deleted `zoom` prop as the `zoom` value is now set from the `Map` component
- added the `isMultipleMarkersMap` prop as `true` to the `Maps` component in `LocationOverview` because it is only used in POI maps

SVA-1095

## How to test:
* [x] Android portrait mode
* [x] iOS portrait mode
* [x] Android landscape mode
* [x] iOS landscape mode

## Screenshots:

`globalSettings: { settings: { "zoomLevelForMaps": { "multipleMarkers": 4, "singleMarker": 1 } } }`

|defect report|settings screen|POI map|
|--|--|--|
Simulator Screenshot - iPhone 14 Pro Max - 2023-08-03 at 10 46 04 | Simulator Screenshot - iPhone 14 Pro Max - 2023-08-03 at 10 46 14 | Simulator Screenshot - iPhone 14 Pro Max - 2023-08-03 at 10 46 30

`globalSettings: { settings: { "zoomLevelForMaps": { "singleMarker": 1 } } }`

|defect report|settings screen|POI map|
|--|--|--|
Simulator Screenshot - iPhone 14 Pro Max - 2023-08-03 at 10 50 38 | Simulator Screenshot - iPhone 14 Pro Max - 2023-08-03 at 10 50 48 | Simulator Screenshot - iPhone 14 Pro Max - 2023-08-03 at 10 51 05

`globalSettings: { settings: { "zoomLevelForMaps": { "multipleMarkers": 4 } } }`

|defect report|settings screen|POI map|
|--|--|--|
Simulator Screenshot - iPhone 14 Pro Max - 2023-08-03 at 10 52 42 | Simulator Screenshot - iPhone 14 Pro Max - 2023-08-03 at 10 52 50 | Simulator Screenshot - iPhone 14 Pro Max - 2023-08-03 at 10 53 03

`globalSettings: { settings: {}}`

|defect report|settings screen|POI map|
|--|--|--|
Simulator Screenshot - iPhone 14 Pro Max - 2023-08-03 at 10 54 32 | Simulator Screenshot - iPhone 14 Pro Max - 2023-08-03 at 10 54 39 | Simulator Screenshot - iPhone 14 Pro Max - 2023-08-03 at 10 54 55

### DIFF
--- a/src/components/map/Map.tsx
+++ b/src/components/map/Map.tsx
@@ -11,6 +11,7 @@ import { MapMarker } from '../../types';
 type Props = {
   geometryTourData?: LatLng[];
   isMaximizeButtonVisible?: boolean;
+  isMultipleMarkersMap?: boolean;
   locations?: MapMarker[];
   mapCenterPosition?: { latitude: number; longitude: number };
   mapStyle?: StyleProp<ViewStyle>;
@@ -19,7 +20,6 @@ type Props = {
   onMaximizeButtonPress?: () => void;
   showsUserLocation?: boolean;
   style?: StyleProp<ViewStyle>;
-  zoom?: number;
 };
 
 const MARKER_ICON_SIZE = normalize(40);
@@ -27,6 +27,7 @@ const MARKER_ICON_SIZE = normalize(40);
 export const Map = ({
   geometryTourData,
   isMaximizeButtonVisible,
+  isMultipleMarkersMap = false,
   locations,
   mapCenterPosition,
   mapStyle,
@@ -34,9 +35,17 @@ export const Map = ({
   onMarkerPress,
   onMaximizeButtonPress,
   style,
-  zoom = 0,
   ...otherProps
 }: Props) => {
+  const { globalSettings } = useContext(SettingsContext);
+  const { settings = {} } = globalSettings;
+  const { zoomLevelForMaps = {}, locationService = {} } = settings;
+
+  const showsUserLocation = otherProps.showsUserLocation ?? !!locationService;
+  const zoom = isMultipleMarkersMap
+    ? zoomLevelForMaps.multipleMarkers
+    : zoomLevelForMaps.singleMarker;
+
   const refForMapView = useRef<MapView>(null);
   // LATITUDE_DELTA handles the zoom, see: https://github.com/react-native-maps/react-native-maps/issues/2129#issuecomment-457056572
   const LATITUDE_DELTA = zoom || 0.0922;
@@ -64,10 +73,6 @@ export const Map = ({
       longitude: locations[0].position.longitude
     };
   }
-
-  const { globalSettings } = useContext(SettingsContext);
-  const showsUserLocation =
-    otherProps.showsUserLocation ?? !!globalSettings?.settings?.locationService;
 
   return (
     <View style={[styles.container, style]}>
--- a/src/components/defectReport/DefectReportLocationForm.tsx
+++ b/src/components/defectReport/DefectReportLocationForm.tsx
@@ -84,7 +84,6 @@ export const DefectReportLocationForm = ({
               ...nativeEvent.coordinate
             });
           }}
-          zoom={4}
         />
         <Wrapper>
           <Button
--- a/src/components/map/LocationOverview.tsx
+++ b/src/components/map/LocationOverview.tsx
@@ -113,7 +113,11 @@ export const LocationOverview = ({
   return (
     <SafeAreaViewFlex>
       <ScrollView>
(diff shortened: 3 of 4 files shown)


src/components/map/Map.tsx
```diff
@@ -11,6 +11,7 @@ import { MapMarker } from '../../types';
 type Props = {
   geometryTourData?: LatLng[];
   isMaximizeButtonVisible?: boolean;
+  isPOI?: boolean;
```
donni106, line 14: i think this should be renamed to something that makes it clear to be applied for a map overview with multiple markers, right? a POI is one point of interest but the condition we implement is for "a map with a lot markers" or "a map with just one marker". `isPOI` does not fit that in my eyes.

src/components/map/Map.tsx
```diff
@@ -19,24 +20,30 @@ type Props = {
...
   style,
-  zoom = 0,
   ...otherProps
 }: Props) => {
+  const { globalSettings } = useContext(SettingsContext);
+  const { settings = {} } = globalSettings;
+  const { zoomLevelForMaps = {}, locationService = {} } = settings;
+
+  const showsUserLocation = otherProps.showsUserLocation ?? !!locationService;
+  const zoom = isPOI ? zoomLevelForMaps?.zoomForPOI : zoomLevelForMaps?.zoom;
```
donni106, line 45: this way we have no fallback value anymore, right? how does the map render now for screens, that had zoom=0 or zoom=4 before, when there are no global settings set? i would suggest a destruction for these two variables here with certain default values.

donni106 (changes requested):

src/components/map/Map.tsx
donni106: `zoomForPOI` and `zoom` could also be renamed to something that makes "a map with a lot markers" or "a map with just one marker" more clear, as i wrote in the other comment already.

ardasnturk: But I think we need to specify the zoom here, so why don't we change it like this?

`zoomValueForMultipleMarkerMap` und `zoomValueForOneMarkerMap`

src/components/map/Map.tsx
ardasnturk: How about `isMultipleMarkersMap`? I think it's right for this place.

donni106: i like it

src/components/map/Map.tsx
ardasnturk: Yes, you are right. We're already talking about the zoom value for maps. No need to repeat it again.
How about this one?

`"zoomLevelForMaps": { "multipleMarkers": 4, "singleMarker": 1 }`

donni106: great, I can't think of anything better

ardasnturk: I changed it.

https://github.com/smart-village-solutions/smart-village-app-app/commit/34ff12eaf22df69c02cde6db188b3c78f214a07c

src/components/map/Map.tsx
ardasnturk: I changed it.

https://github.com/smart-village-solutions/smart-village-app-app/commit/34ff12eaf22df69c02cde6db188b3c78f214a07c

src/components/map/Map.tsx
```diff
@@ -19,24 +20,32 @@ type Props = {
...
   ...otherProps
 }: Props) => {
+  const { globalSettings } = useContext(SettingsContext);
+  const { settings = {} } = globalSettings;
+  const { zoomLevelForMaps = {}, locationService = {} } = settings;
+
+  const showsUserLocation = otherProps.showsUserLocation ?? !!locationService;
+  const zoom = isMultipleMarkersMap
+    ? zoomLevelForMaps?.multipleMarkers
+    : zoomLevelForMaps?.singleMarker;
```
donni106, line 47: if you fallback to `{}` i guess you do not have to use the conditional chaining here, because `zoomLevelForMaps` is always present an accessible as object.

ardasnturk: Yeah, that's right. I fixed it.

https://github.com/smart-village-solutions/smart-village-app-app/commit/9710bf33bdc67175807eab8b3c663c27fcca6fc5

donni106 (approved):

PR #891, opened 2023-08-09
feat(opening times): respect `useYear` instead of 1970-hack

donni106:
the main server removed the 1970-hack and introduced a `useYear` flag, that we will be able to use in order to render without the year if false

SVA-985

| `useYear = true`| `useYear = false`|
|---|---|
| image | image |


ardasnturk (approved):
I don't see any problem. I tested it and I can say that it works well. 🎉

PR #893, opened 2023-08-11
fix: event list showing only 15 events

ardasnturk:
- removed the limit value in `queryVariables` to fix the issue of not showing more than 15 events in the event list

SVA-1114

---
the `useIsFocused()` hook in `IndexScreen` was causing the query on the event page to constantly re-run, and since the limit value was set to 15 in `queryVariables`, no more than 15 events could be listed in the list. To fix this issue, `{ limit: 15, ... }` in `HomeScreen` has been removed.

this is a temporary code update for now. A more detailed code update may need to be investigated.


donni106:
We will not use this, as we planned to make the `limit`s editable per global settings.

PR #894, opened 2023-08-11
feat: add query variables that can be set from the server for the event

ardasnturk:
- added editing the `queryVariables` `limit` in `IndexScreen` for all sections shown in `HomeScreen` via `globalSettings`
- added default limit value of all sections as 15
- added `queryVariables` limit when going to the event screen via widget
- added default limit value as 15 on the event screen navigated using widget
- added `limit` to `WidgetProps` to prevent type error

SVA-1119

---

make the following adjustments to the `globalSettings` to try the feature.

the `queryVariables` required to go to the Event screen with the "Alle Veranstaltungen anzeigen" button on the `HomeScreen`
```
"sections": {
    ...
    "limitEvents": 20,
    "limitNews": 20,
    "limitPOI": 20
    ...
  },
```

to go to the Event screen with the widget `queryVariables`
```
{
  "widgetName": "event",
  "additionalProps": {
   ...
    "limit": 20
    ...
  }
},
```

### DIFF
--- a/src/screens/HomeScreen.js
+++ b/src/screens/HomeScreen.js
@@ -171,7 +171,7 @@ export const HomeScreen = ({ navigation, route }) => {
     buttonEvents = texts.homeButtons.events,
     limitEvents = 15,
     limitNews = 15,
-    limitPOI = 15
+    limitPointsOfInterestAndTours = 15
   } = sections;
   const { events: showVolunteerEvents = false } = hdvt;
   const [refreshing, setRefreshing] = useState(false);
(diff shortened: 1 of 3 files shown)


src/screens/HomeScreen.js
```diff
@@ -235,6 +240,7 @@ export const HomeScreen = ({ navigation, route }) => {
     {
       buttonTitle: buttonPointsOfInterestAndTours,
       fetchPolicy,
+      limit: limitPOI,
```
donni106, line 243: should be `limitPointsOfInterestAndTours` to match the structure from news and events. i will adjust that.

donni106 (approved):

PR #896, opened 2023-08-14
refactor: use `SectionHeader` everywhere

donni106:
- added early return if check for title presence fails
- added style props, that were used for some title headers

### DIFF
--- a/src/screens/volunteer/VolunteerRegistrationScreen.tsx
+++ b/src/screens/volunteer/VolunteerRegistrationScreen.tsx
@@ -14,17 +14,15 @@ import {
   LoadingModal,
   RegularText,
   SafeAreaViewFlex,
-  Title,
-  TitleContainer,
-  TitleShadow,
+  SectionHeader,
   Wrapper,
   WrapperHorizontal
 } from '../../components';
-import { consts, device, secrets, texts } from '../../config';
+import { consts, secrets, texts } from '../../config';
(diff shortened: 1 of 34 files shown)


src/screens/volunteer/VolunteerRegistrationScreen.tsx
```diff
@@ -93,16 +91,7 @@ export const VolunteerRegistrationScreen = ({ navigation }: StackScreenProps<any
...
-            <Title
-              big
-              center
-              accessibilityLabel={`${texts.volunteer.registrationTitle} ${a11yLabel.heading}`}
-            >
-              {texts.volunteer.registrationTitle}
-            </Title>
-          </TitleContainer>
-          {device.platform === 'ios' && <TitleShadow />}
+          <SectionHeader title={texts.consul.registrationTitle} big center />
```
d4v3d4vE, line 94: Possibly not correct import, shouldn't it be `texts.volunteer.registrationTitle`?

d4v3d4vE (changes requested):

src/screens/volunteer/VolunteerRegistrationScreen.tsx
donni106: uh yes

d4v3d4vE (approved):

ardasnturk (approved):
I have not seen any problems in the code.

PR #897, opened 2023-08-17
feat: add setting to show or hide filter by opening times

donni106:
There is a new global setting that allows hiding the filter for opening times, which results in an index screen without that:

Simulator Screen Shot - iPhone 12 - 2023-08-17 at 10 40 58

SVA-1127


d4v3d4vE (commented):
in my case the filter option isn't hidden altough its set to false on the mainserver 
console.log from mainserver settings:
image

ardasnturk (approved):
Let's send this code to the master. 🎉

PR #898, opened 2023-08-17
feat: new map/list switcher floating button

donni106:
- removed "TOP" naming in variables, as the filter should not be named with some rendering position, as it can be used in different place (like now with this change)
- added setting to switch between rendering of top filters or bottom floating button on index screens for points of interest
- created new component `IndexMapSwitch` for bottom floating button using the same action as the top filter to switch the screen content

|list|map|
|---|---|
|Simulator Screen Shot - iPhone 12 - 2023-08-17 at 14 59 52|Simulator Screen Shot - iPhone 12 - 2023-08-17 at 14 59 56|

MQGB-34

### DIFF
--- a/src/components/IndexMapSwitch.js
+++ b/src/components/IndexMapSwitch.js
@@ -1,0 +1,47 @@
+import PropTypes from 'prop-types';
+import React from 'react';
+import { StyleSheet, View } from 'react-native';
+
+import { texts } from '../config';
+
+import { Button } from './Button';
+
+export const IndexMapSwitch = ({ filter, setFilter }) => (
+  <View style={styles.floatingButtonContainer}>
+    <Button
+      onPress={() => {
+        const selectedFilter = filter.find((entry) => entry.selected);
+        const updatedFilter = filter.map((entry) => {
+          if (entry.id !== selectedFilter.id) {
+            entry.selected = true;
+          } else {
+            entry.selected = false;
+          }
+
+          return entry;
+        });
+
+        setFilter(updatedFilter);
+      }}
+      title={
+        filter.find((entry) => entry.selected).title === texts.locationOverview.list
+          ? texts.locationOverview.map
+          : texts.locationOverview.list
+      }
+      notFullWidth
+    />
+  </View>
+);
+
+const styles = StyleSheet.create({
+  floatingButtonContainer: {
+    alignSelf: 'center',
+    bottom: '5%',
+    position: 'absolute'
+  }
+});
+
+IndexMapSwitch.propTypes = {
+  filter: PropTypes.array.isRequired,
+  setFilter: PropTypes.func.isRequired
+};
--- a/src/components/IndexMapSwitch.js
+++ b/src/components/IndexMapSwitch.js
@@ -1,45 +1,61 @@
 import PropTypes from 'prop-types';
-import React from 'react';
+import React, { useContext } from 'react';
 import { StyleSheet, View } from 'react-native';
 
 import { texts } from '../config';
+import { SettingsContext } from '../SettingsProvider';
 
 import { Button } from './Button';
 
-export const IndexMapSwitch = ({ filter, setFilter }) => (
-  <View style={styles.floatingButtonContainer}>
-    <Button
-      onPress={() => {
-        const selectedFilter = filter.find((entry) => entry.selected);
-        const updatedFilter = filter.map((entry) => {
-          if (entry.id !== selectedFilter.id) {
-            entry.selected = true;
-          } else {
-            entry.selected = false;
-          }
+export const IndexMapSwitch = ({ filter, setFilter }) => {
+  const { globalSettings } = useContext(SettingsContext);
+  const { navigation = 'drawer' } = globalSettings;
 
-          return entry;
-        });
+  return (
+    <View style={[styles.floatingButtonContainer, stylesWithProps({ navigation }).position]}>
+      <Button
+        onPress={() => {
+          const selectedFilter = filter.find((entry) => entry.selected);
+          const updatedFilter = filter.map((entry) => {
+            if (entry.id !== selectedFilter.id) {
+              entry.selected = true;
+            } else {
+              entry.selected = false;
+            }
 
-        setFilter(updatedFilter);
-      }}
-      title={
-        filter.find((entry) => entry.selected).title === texts.locationOverview.list
-          ? texts.locationOverview.map
-          : texts.locationOverview.list
-      }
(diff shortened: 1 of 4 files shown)


d4v3d4vE (approved):
Works great!

src/components/IndexMapSwitch.js
```diff
@@ -0,0 +1,63 @@
...
+
+/* eslint-disable react-native/no-unused-styles */
+/* this works properly, we do not want that warning */
+const stylesWithProps = ({ navigation }) => {
+  return StyleSheet.create({
+    position: {
+      bottom: navigation === 'tab' ? 0 : '5%'
+    }
+  });
+};
```
ardasnturk, line 57: Wouldn't it be more correct to use `drawer` instead of `tab` here? Because there is a possibility that the navigation value is `undefined` in `globalSettings`. But we use `drawer` as standard. If the navigation value is not used in `globalSettings`, the navigation structure of the application will be tabbar. What kind of view will we get in this case?

|globalSettings.navigation = drawer|globalSettings.navigation = tab|globalSettings.navigation = undefined|
|--|--|--|
Simulator Screenshot - iPhone 14 - 2023-08-21 at 16 24 20 | Simulator Screenshot - iPhone 14 - 2023-08-21 at 16 24 50 | Simulator Screenshot - iPhone 14 - 2023-08-21 at 16 25 10

src/screens/IndexScreen.js
```diff
@@ -602,6 +616,9 @@ export const IndexScreen = ({ navigation, route }) => {
           }}
         </Query>
       )}
+      {query === QUERY_TYPES.POINTS_OF_INTEREST && (
+        <IndexMapSwitch filter={filterType} setFilter={setFilterType} />
+      )}
```
ardasnturk, line 621: Can this be wrapped with view or can the style in IndexMapSwitch be updated? The Zurück Nach Oben button is not accessible because it is behind the filter. 

Simulator Screenshot - iPhone 14 - 2023-08-21 at 16 09 09

ardasnturk (changes requested):

src/screens/IndexScreen.js
donni106: thanks, i will change to not showing a "back to top" in case of POI list + floating button

src/components/IndexMapSwitch.js
donni106: i guess there is a development ongoing to change the default navigation to tabs. when we are not defining a navigation, the fallback value needs to be aligned with the resulting navigation. so in your third case the value should result in `tab`, which would result in a correct positioning with `0`. so we do not need to change the code here, or am i wrong?

ardasnturk: I'm a little confused about this. Looks like you're right. If we don't specify a navigation, we will automatically have tab navigation. I can't say anything as the development is not finished yet. I will merge the two branches in my local and share the results with you. 👍

ardasnturk: The result looks the same as the screenshots above. The reason is that we set `navigation` as `drawer` in line 12 of `IndexMapSwitch`.

|globalSettings.navigation = drawer|globalSettings.navigation = tab|globalSettings.navigation = undefined|
|--|--|--|
Simulator Screenshot - iPhone 14 Pro Max - 2023-09-14 at 12 00 12 | Simulator Screenshot - iPhone 14 Pro Max - 2023-09-14 at 12 02 26 | Simulator Screenshot - iPhone 14 Pro Max - 2023-09-14 at 12 02 48

how about adding a control like this?
```
return StyleSheet.create({
    position: {
      bottom: !navigation || navigation === 'tab' ? 0 : '5%'
    }
  });
```

src/screens/IndexScreen.js
ardasnturk: if we remove "back to top" then how will the list look like? The button appears above the list. Do we want it to be like this?

|list with back to top|list without back to top|
|--|--|
Simulator Screenshot - iPhone 14 Pro Max - 2023-09-14 at 12 11 31 | Simulator Screenshot - iPhone 14 Pro Max - 2023-09-14 at 12 11 38

donni106: yes, we cannot position the button more to the middle of the screen. it needs to stay at the most possible button, which sure overlapps the last entry. but the content is still visible and pressable.

one thing could be to add some whitespace below the last entry if we do not show a "back to top", so that the last entry is scrollable the way it is fully visible?

src/components/IndexMapSwitch.js
donni106: `!navigation` is never the case, look at your logs. it is always drawer or tab. that's why i mean to not change anything. if we correctly fallback to tab in case nothing is defined in global settings, your third case would result in `{"navigation":"tab"}` and the condition works.

ardasnturk: If we don't define anything for navigation in `globalSettings`, `navigation = 'drawer'`. Because in line 12 of the code, that is here:
```
const { navigation = 'drawer' } = globalSettings; 
```
we say that navigation is `drawer`.
Therefore the warning text in the screenshots says `{ 'navigation': 'drawer' }`. But nothing is set in `globalSettings`.
I guess we can't agree on that. 😂

src/screens/IndexScreen.js
ardasnturk: Just as an idea, I added View instead of BackToTop component in a very simple way and the whole list can be shown smoothly as in the screenshot.
image

src/components/IndexMapSwitch.js
donni106: correct, which needs to be changed to fallback to tabs with our change of tab navigation as default.

src/screens/IndexScreen.js
ardasnturk: That's a good idea. But we'll need an extra design for that. 🤷‍♂️

donni106: Thanks for your suggestions. @d4v3d4vE this would be too much changes visually, which would mean an extra task for design + development.

I decided to keep the "back to top" and add a spacer for POIs:

Simulator Screen Shot - iPhone 14 - 2023-09-14 at 14 32 28

ardasnturk: This way we can solve the problem in less time. I think it's a good solution. 👍

d4v3d4vE: True, this way it also looks nice!

donni106:
superseded by #921

PR #899, opened 2023-08-17
refactor: unify structures

donni106:
- corrected order of code in files
- moved styles before prop types in files
- moved consts to the top instead of the bottom
- organized imports


ardasnturk (approved):
It looks clean. 👍

d4v3d4vE (approved):

PR #900, opened 2023-08-18
refactor: add own color for refresh control

donni106:
We are more flexible having an own color for refresh controls and activity indicators.


ardasnturk (approved):
It looks good. But after the 3.3.0 update, we should remember to check the color of the `colors.refreshControl` for each build.

d4v3d4vE (approved):

PR #903, opened 2023-08-21
feat: add ability for a factor to size tiles

donni106:
- if tiles are not 1:1 aspect ratio, we can now have a factor to properly size them
- `tileSizeFactor` can be set through `settings` in global settings

|without factor|with factor|
|---|---|
|image|image|

MQGB-27


ardasnturk (approved):
I don't have any questions about the code. It looks good. 🎉

PR #905, opened 2023-08-22
feat: add a multi-button feature for navigating screens in the app

ardasnturk:
- added navigation according to `params.paramsForMultiButtons` object to be able to navigate to different screens inside the application
- added new types for `paramsForMultiButtons` object to eliminate type error
- edited comment lines according to the new feature

SVA-1120

## Screenshots:

|`staticContent` where the property can be tested|`HtmlScreen` with multi button prepared with different navigation types|`WebScreen`|`IndexScreen` with `categories` query|`HtmlScreen` with button|
|--|--|--|--|--|
SCR-20230822-krdv | SCR-20230822-krgz | SCR-20230822-krja | SCR-20230822-krky | SCR-20230822-krmy

### DIFF
--- a/src/components/MultiButtonWithSubQuery.tsx
+++ b/src/components/MultiButtonWithSubQuery.tsx
@@ -35,13 +35,10 @@ export const navigateWithSubQuery = ({
 }) => {
   // if the `params` is a string, it is directly the web url to call
   if (!!params && typeof params === 'string') {
-    return navigation.navigate({
-      name: 'Web',
-      params: {
-        rootRouteName,
-        title,
-        webUrl: params
-      }
+    return navigation.push('Web', {
+      rootRouteName,
+      title,
+      webUrl: params
     });
   }
 
@@ -49,24 +46,18 @@ export const navigateWithSubQuery = ({
     // if `params` is an object and contains the `paramsForButton` object,
     // the values here are added to the navigation
     if (params.paramsForButton) {
-      return navigation.navigate({
-        name: params.routeName,
-        params: {
-          ...params.paramsForButton,
-          rootRouteName
-        }
+      return navigation.push(params.routeName, {
+        ...params.paramsForButton,
+        rootRouteName
       });
     }
 
     // if `params` is an object and does not contain the `paramsForButton` object,
     // it contains a `routeName` and a `webUrl`
-    return navigation.navigate({
-      name: params.routeName,
-      params: {
-        rootRouteName,
-        title,
-        webUrl: params.webUrl
-      }
+    return navigation.push(params.routeName, {
+      rootRouteName,
+      title,
+      webUrl: params.webUrl
     });
   }
 
@@ -75,14 +66,11 @@ export const navigateWithSubQuery = ({
   // if there is no `params`, use the main `subQuery` values for `routeName` and a `webUrl` or
   // `params` if the params contain a webUrl as well, the webUrl property of the subQuery
   // will be ignored
-  return navigation.navigate({
-    name: subQuery.routeName,
-    params: {
-      ...subParams,
-      rootRouteName,
-      title,
-      webUrl: subQuery.webUrl
-    }
+  return navigation.push(subQuery.routeName, {
+    ...subParams,
+    rootRouteName,
+    title,
+    webUrl: subQuery.webUrl
   });
 };
 
--- a/src/components/MultiButtonWithSubQuery.tsx
+++ b/src/components/MultiButtonWithSubQuery.tsx
@@ -15,7 +15,21 @@ export const navigateWithSubQuery = ({
 }: {
   navigation: StackNavigationProp<any, string>;
   title?: string;
-  params?: { routeName: string; webUrl: string } | string;
+  params?:
+    | {
+        routeName: string;
+        webUrl: string;
+        paramsForButton?: {
+          query: string;
+          queryVariables: { name: string };
+          rootRouteName: string;
+          routeName: string;
+          subQuery: SubQuery;
+          title: string;
+          webUrl?: string;
+        };
+      }
+    | string;
   rootRouteName?: string;
   subQuery: SubQuery;
 }) => {
@@ -31,8 +45,21 @@ export const navigateWithSubQuery = ({
     });
   }
 
-  // if the `params` is an object, it contains a `routeName` and a `webUrl`
   if (!!params && typeof params === 'object') {
+    // if `params` is an object and contains the `paramsForButton` object,
+    // the values here are added to the navigation
+    if (params.paramsForButton) {
+      return navigation.navigate({
+        name: params.routeName,
+        params: {
+          ...params.paramsForButton,
+          rootRouteName
+        }
+      });
+    }
+
+    // if `params` is an object and does not contain the `paramsForButton` object,
+    // it contains a `routeName` and a `webUrl`
     return navigation.navigate({
       name: params.routeName,
       params: {


src/components/MultiButtonWithSubQuery.tsx
```diff
@@ -15,7 +15,21 @@ export const navigateWithSubQuery = ({
 }: {
   navigation: StackNavigationProp<any, string>;
   title?: string;
-  params?: { routeName: string; webUrl: string } | string;
+  params?:
+    | {
```
donni106, line 19: What does `| {` mean here?

src/components/MultiButtonWithSubQuery.tsx
```diff
@@ -15,7 +15,21 @@ export const navigateWithSubQuery = ({
...
+        routeName: string;
+        webUrl: string;
+        paramsForMultiButtons?: {
```
donni106, line 22: Why not short `params` here as well?

src/components/MultiButtonWithSubQuery.tsx
```diff
@@ -15,7 +15,21 @@ export const navigateWithSubQuery = ({
 }: {
   navigation: StackNavigationProp<any, string>;
   title?: string;
-  params?: { routeName: string; webUrl: string } | string;
+  params?:
+    | {
+        routeName: string;
+        webUrl: string;
```
donni106, line 21: `webUrl` should be changed to an optional param here, right?

src/components/MultiButtonWithSubQuery.tsx
donni106: Isn't `{ routeName: string; webUrl: string }` missing in first place for cases, where not "multi button"?

src/components/MultiButtonWithSubQuery.tsx
ardasnturk: I left this part of the code as it was before because I didn't want to have incompatibility with older versions later. If there is no webUrl, we don't show the first button anyway. When you say optional, do you mean it can be `undefined`? Then I think I need to change the following code as well.

```
 {!!subQuery && !!subQuery.routeName && (!!subQuery.webUrl || subQuery.params) && (
        <Button
          title={subQuery.buttonTitle || `${title} öffnen`}
          onPress={() => navigateWithSubQuery({ navigation, subQuery, rootRouteName, title })}
        />
      )}
```

src/components/MultiButtonWithSubQuery.tsx
ardasnturk: I did not want to use params.params in the code. 

```
 if (!!params && typeof params === 'object') {
    if (params.paramsForMultiButtons) {
        ...
        params: {
          ...params.paramsForMultiButtons,
        }
      });
    }
```
if I use params this code will look like the following.
```
if (!!params && typeof params === 'object') {
    if (params.params) {
      ...
        params: {
          ...params.params,
        }
      });
    }
```

src/components/MultiButtonWithSubQuery.tsx
ardasnturk: I don't know why you added the `| {` code. I think it was added during automatic formatting

---

If it is not multi button, the params object is not sent to this function as a prop.

```
return (
    <>
      {!!subQuery && !!subQuery.routeName && (!!subQuery.webUrl || subQuery.params) && (
        <Button
          title={subQuery.buttonTitle || `${title} öffnen`}
          onPress={() => navigateWithSubQuery({ navigation, subQuery, rootRouteName, title })}
        />
      )}

     ...
    </>
  );
```

donni106: i cannot understand how "If it is not multi button, the params object is not sent to this function as a prop." relates to the code you have sent? can you elaborate a bit more please. the `params` are part of `subQuery` which is passed to `navigateWithSubQuery`.

src/components/MultiButtonWithSubQuery.tsx
donni106: ok, keep it as is

src/components/MultiButtonWithSubQuery.tsx
```diff
@@ -31,8 +45,21 @@ export const navigateWithSubQuery = ({
     });
   }
 
-  // if the `params` is an object, it contains a `routeName` and a `webUrl`
   if (!!params && typeof params === 'object') {
```
donni106, line 48: with having a new condition checking something in params this gets more complex. it is very hard to understand when which params do what.
could you think of refactorings to make this whole part easier to read and less complex?
regarding all cases we can have, how can they be unified or defined with an object, that is more flexible, to reduce the conditions and different possibilities we need to think of and remember all the time.
maybe in a session together with @d4v3d4vE?

src/components/MultiButtonWithSubQuery.tsx
ardasnturk: The `params` prop specified in the question is only valid for multi buttons. We do not send the `params` prop to the `navigateWithSubQuery` function with the first button on the screen. instead we create a variable called `subParams` in the `navigateWithSubQuery` function and integrate `subQuery.params` into the navigation structure. That's why I thought the params prop is not necessary for the first button, so I sent the code required for the first button.

Or I have misunderstood the question you wanted to ask. 🤷‍♂️
```
export const navigateWithSubQuery = ({
...
const subParams = { ...(subQuery.params ?? {}) };

  // if there is no `params`, use the main `subQuery` values for `routeName` and a `webUrl` or
  // `params` if the params contain a webUrl as well, the webUrl property of the subQuery
  // will be ignored
  return navigation.navigate({
    name: subQuery.routeName,
    params: {
      ...subParams,
      rootRouteName,
      title,
      webUrl: subQuery.webUrl
    }
  });

...
return (
    <>
      {!!subQuery && !!subQuery.routeName && (!!subQuery.webUrl || subQuery.params) && (
        <Button
          title={subQuery.buttonTitle || `${title} öffnen`}
          onPress={() => navigateWithSubQuery({ navigation, subQuery, rootRouteName, title })}
        />
      )}

      {subQuery?.buttons?.map((button, index) => (
        <Button
          key={`${index}-${button.webUrl}`}
          title={button.buttonTitle || `${title} öffnen`}
          onPress={() =>
            navigateWithSubQuery({
              navigation,
              params: button,
              rootRouteName,
              subQuery,
              title
            })
          }
        />
      ))}
    </>
  );
```

src/components/MultiButtonWithSubQuery.tsx
ardasnturk: Okay, I can rename this prop this way. It'll be shorter and clearer. 👍

src/components/MultiButtonWithSubQuery.tsx
ardasnturk: As a result of a small conversation with ChatGPT, such a code came out. Do you think it's better or more complex? Sure, we can have a talk with David about it.
```
// Action to navigate based on provided parameters
export const navigateWithSubQuery = ({
  navigation,
  title,
  params,
  rootRouteName,
  subQuery
}: {
  navigation: StackNavigationProp<any, string>;
  title?: string;
  params?: ParamsType;
  rootRouteName?: string;
  subQuery: SubQuery;
}) => {
  // Common navigation parameters
  const commonParams = { rootRouteName, title };

  // Navigate based on the type of `params`
  if (typeof params === 'string') {
    navigateTo('Web', { webUrl: params });
  } else if (params?.paramsForMultiButtons) {
    navigateTo(params.routeName, { ...params.paramsForMultiButtons });
  } else if (params) {
    navigateTo(params.routeName, { webUrl: params.webUrl });
  } else {
    const subParams = { ...(subQuery.params ?? {}) };
    navigateTo(subQuery.routeName, { ...subParams, webUrl: subQuery.webUrl });
  }

  // Helper function for navigation
  function navigateTo(routeName: string, additionalParams: object) {
    navigation.navigate({
      name: routeName,
      params: { ...commonParams, ...additionalParams }
    });
  }
};
```

src/components/MultiButtonWithSubQuery.tsx
ardasnturk: https://github.com/smart-village-solutions/smart-village-app-app/commit/1fd55b6e7ac650ac6929042ae0f3310262c94971

src/components/MultiButtonWithSubQuery.tsx
```diff
@@ -31,8 +45,21 @@ export const navigateWithSubQuery = ({
...
+    // if `params` is an object and contains the `paramsForButton` object,
+    // the values here are added to the navigation
+    if (params.paramsForButton) {
+      return navigation.navigate({
```
donni106, line 52: If we are on a Html-Screen and navigate to a Html-Screen there is no navigation but exchange of screen contents and we cannot navigate back. We will need `push`s here instead of `navigate`s, to ensure clear navigation through the app.

src/components/MultiButtonWithSubQuery.tsx
donni106: got it

src/components/MultiButtonWithSubQuery.tsx
donni106: 1. we would use `const navigateTo...` instead of `function`.
2. this would not need an extra function i guess, but setting the full params in the cases and doing a `navigation.navigate` to the created full params?
3. i think it has no benefit to the current version, as it does completely the same with a bit less code, right? and we are missing the comments.

so let's keep our current version here.

src/components/MultiButtonWithSubQuery.tsx
donni106: 6169dd9

https://github.com/smart-village-solutions/smart-village-app-app/assets/1942953/50082c0d-e518-49c1-b884-276c4aa975fc

donni106 (approved):

src/components/MultiButtonWithSubQuery.tsx
ardasnturk: Thank you very much for the change. 👍

d4v3d4vE (changes requested):
the buttons: “WebScreen öffnen - 2”, “IndexScreen öffnen mit Categories query“, “HtmlScreen öffnen mit button” und “HtmlScreen öffnen mit multiButtons” are not working anymore & there is a render bug: 
Simulator Screenshot - iPhone 14 Pro - 2023-09-04 at 09 52 47

ardasnturk:
> the buttons: “WebScreen öffnen - 2”, “IndexScreen öffnen mit Categories query“, “HtmlScreen öffnen mit button” und “HtmlScreen öffnen mit multiButtons” are not working anymore & there is a render bug:
> Simulator Screenshot - iPhone 14 Pro - 2023-09-04 at 09 52 47

I have no such problem. I looked at the html content and only the text "Zur" is wrapped with the strong tag. So I don't think this is a problem. What size device did you try it on? Maybe it looks different on different sized devices. Also, did you pull the last changes in the branch?

Simulator Screenshot - iPhone 14 - 2023-09-04 at 10 17 42

Also a note for the future. The next time you share a screenshot, it would be better if you reduce the size of the screenshot a bit. Otherwise it looks too big.
To do that: Open the screenshot, then select `Werkzeuge` in the menu above and in `Größenkorrektur` select 320x320 or 640x640 and save.

d4v3d4vE (approved):

ardasnturk:
We talked over Slack and solved the problem. So it can be merged.

PR #906, opened 2023-08-23
refactor: change navigation structure

ardasnturk:
- deleted `companyTabConfig` tab navigation as we don't need it
- added `pointOfInterest` screen as the 3rd tab because we want to see it in tab navigation
- added `EventScreen` as the 4rd tab because we want to see it in tab navigation
- updated `texts.js` to show the title and tab name of `EventScreen` and `pointsOfInterest` screens
- sorted `tabBarLabel` alphabetically to improve readability

SVA-1116

to try this feature, please update the codes in `config.ts` according to the tab navigation structure
## Screenshots:

|before|after|POI Screen|EventScreen|
|--|--|--|--|
Simulator Screenshot - iPhone 14 - 2023-08-23 at 12 46 59 | Simulator Screenshot - iPhone 14 - 2023-08-23 at 12 41 07 | Simulator Screenshot - iPhone 14 - 2023-08-23 at 12 41 11 | Simulator Screenshot - iPhone 14 - 2023-08-23 at 12 41 14

### DIFF
--- a/src/config/navigation/defaultStackConfig.tsx
+++ b/src/config/navigation/defaultStackConfig.tsx
@@ -255,7 +255,12 @@ export const defaultStackConfig = ({
     },
     {
       routeName: ScreenName.Index,
-      screenComponent: IndexScreen
+      screenComponent: IndexScreen,
+      // NOTE: is used as initial screen for the points of interest tab
+      inititalParams: {
+        title: texts.screenTitles.pointsOfInterest,
+        query: QUERY_TYPES.CATEGORIES
+      }
     },
     {
       routeName: ScreenName.Lunch,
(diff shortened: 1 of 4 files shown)


d4v3d4vE (approved):
Works great!

src/config/navigation/defaultStackConfig.tsx
```diff
@@ -255,7 +265,12 @@ export const defaultStackConfig = ({
     },
     {
       routeName: ScreenName.Index,
-      screenComponent: IndexScreen
+      screenComponent: IndexScreen,
+      // NOTE: is used as initial screen for the points of interest tab
```
donni106, line 269: i wonder if we really need this note? what was your intention?

donni106 (approved):

src/config/navigation/defaultStackConfig.tsx
ardasnturk: This note is not necessary. But I'd still prefer to keep it so that in the future we can understand why it happened the way it did.

donni106: but what does it help? it says that it is the initial screen for points of interest. yes true thats why it is configured.
or other question, why do we not have NOTEs at the other initial screens? why only for POI? i do not understand.

PR #907, opened 2023-08-23
refactor: change query for index screen

ardasnturk:
- replaced the `Query` component structure in `IndexScreen` with `useQuery` to use the `useFocusEffect` hook
- added the props used in `Query` to `useQuery`
- moved the functions, variables and controls used in the `Query` component above return to fix the problem of the `useIsFocused` hook being run continuously inside the `Query` component every time we open `IndexScreen`
- moved `fetchPolicy` variable above useQuery

SVA-1114


d4v3d4vE (approved):
Looks good to me!

src/screens/IndexScreen.js
donni106, line 1: sadly I have duplicated events in calendar view on a date.

Simulator Screen Shot - iPhone 12 - 2023-09-08 at 15 10 40-tile

I will look deeper into it.

donni106 (changes requested):

src/screens/IndexScreen.js
donni106: and after pull to refresh I have some entries of earlier days visible

Simulator Screen Shot - iPhone 12 - 2023-09-08 at 15 14 12

donni106: and sometimes the last entry is duplicated

Simulator Screen Shot - iPhone 12 - 2023-09-08 at 15 22 14-tile

src/screens/IndexScreen.js
```diff
@@ -374,223 +444,150 @@ export const IndexScreen = ({ navigation, route }) => {
...
+            }
+            horizontal={false}
+            sectionByDate={isCalendar ? !showCalendar : true}
+            query={query}
+            queryVariables={queryVariables}
+            fetchMoreData={isConnected ? fetchMoreData : null}
+            refreshControl={
+              <RefreshControl
+                refreshing={refreshing}
+                onRefresh={() => refresh(refetch)}
```
donni106, line 583: as `refetch` comes from hook `const { data, loading, fetchMore, refetch } = useQuery(` now, we do not need to pass it to the `refresh` method, at is already defined.

src/screens/IndexScreen.js
donni106: refactoring done

PR #908, opened 2023-08-24
feat: add html section above the survey list

ardasnturk:
- added `HtmlView` inside `ListHeaderComponent` to `SectionList` to show html on the survey list
- added `Props` type to remove the type error of the `route` prop in `SurveyOverviewScreen`
- added as param to use the `additionalProps.htmlName` added to the survey widget in `globalSettings` on the `SurveyOverview` screen
- added `useStaticContent` to get the html content from the main server added to `additionalProps.htmlName` to show html in `HtmlView`
- added `htmlName` type as string to `WidgetProps` to avoid type error

SBB-109

## Screenshots:

|before|after|
|--|--|
Simulator Screenshot - iPhone 14 - 2023-08-24 at 12 50 03 | Simulator Screenshot - iPhone 14 - 2023-08-24 at 12 49 53

### DIFF
--- a/src/screens/SurveyOverviewScreen.tsx
+++ b/src/screens/SurveyOverviewScreen.tsx
@@ -4,16 +4,30 @@ import React, { useCallback } from 'react';
 import { useQuery } from 'react-apollo';
 import { SectionList } from 'react-native';
 
-import { SafeAreaViewFlex, SectionHeader, TextListItem } from '../components';
+import { HtmlView, SafeAreaViewFlex, SectionHeader, TextListItem, Wrapper } from '../components';
 import { LoadingSpinner } from '../components/LoadingSpinner';
 import { texts } from '../config';
 import { combineLanguages } from '../helpers';
-import { usePullToRefetch, useSurveyLanguages } from '../hooks';
+import { usePullToRefetch, useStaticContent, useSurveyLanguages } from '../hooks';
 import { SURVEYS } from '../queries/survey';
 import { Survey } from '../types';
 
+type Props = {
+  route: {
+    params?: {
+      additionalProps?: {
+        htmlName?: string;
+      };
+    };
+  };
+};
+
 const useSurveySections = () => {
-  const { data: surveys, loading, refetch } = useQuery<{
+  const {
+    data: surveys,
+    loading,
+    refetch
+  } = useQuery<{
     ongoing: Survey[];
     archived: Survey[];
   }>(SURVEYS, { fetchPolicy: 'cache-and-network' });
@@ -58,11 +72,19 @@ const renderSectionHeader = ({
   return <SectionHeader title={title} />;
 };
 
-export const SurveyOverviewScreen = () => {
+export const SurveyOverviewScreen = ({ route }: Props) => {
   const { loading, refetch, surveySections } = useSurveySections();
   const RefreshControl = usePullToRefetch(refetch);
   const languages = useSurveyLanguages();
   const navigation = useNavigation<StackNavigationProp<any>>();
+  const { additionalProps } = route.params ?? {};
+
+  const { data } = useStaticContent({
+    name: additionalProps?.htmlName,
+    type: 'html',
+    refreshTimeKey: `publicHtmlFile-${additionalProps?.htmlName}`,
+    skip: !additionalProps?.htmlName
+  });
 
   const renderSurvey = useCallback(
     ({ item: survey }: { item: Survey }) => {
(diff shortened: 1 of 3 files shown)


d4v3d4vE (approved):
Works great!

src/screens/SurveyOverviewScreen.tsx
```diff
@@ -58,11 +72,18 @@ const renderSectionHeader = ({
...
-export const SurveyOverviewScreen = () => {
+export const SurveyOverviewScreen = ({ route }: Props) => {
   const { loading, refetch, surveySections } = useSurveySections();
   const RefreshControl = usePullToRefetch(refetch);
   const languages = useSurveyLanguages();
   const navigation = useNavigation<StackNavigationProp<any>>();
+  const { additionalProps } = route.params ?? {};
+
+  const { data } = useStaticContent({
+    name: additionalProps?.htmlName ?? '',
```
donni106, line 83: Why `??` instead of `||`? this would result in `false` for example if `additionalProps?.htmlName` would be `false`.

image

I guess we want a fallback in any falsey case here to an empty string, right?

src/screens/SurveyOverviewScreen.tsx
```diff
@@ -58,11 +72,18 @@ const renderSectionHeader = ({
...
+    type: 'html',
+    refreshTimeKey: 'publicHtmlFile-survey'
```
donni106, line 85: with this hardcoded refresh time key, different `name`s use the same key. having a variable `name` in params to gather different html contents, we should also consider variable refresh time keys. we do this in other places already, have a look around next time please, for example in `HtmlScreen`:

```
const { data, loading, refetch } = useStaticContent({
  name: queryVariables.name,
  type: 'html',
  refreshTimeKey: `${query}-${queryVariables.name}`
});
```

src/screens/SurveyOverviewScreen.tsx
```diff
@@ -80,6 +101,15 @@ export const SurveyOverviewScreen = () => {
...
       <SectionList
+        ListHeaderComponent={
+          <>
+            {!!data && (
+              <Wrapper>
+                <HtmlView html={data} />
+              </Wrapper>
+            )}
+          </>
+        }
```
donni106, line 112: I suggest passing `null` here instead of having an empty fragment always for rendering, even if there is no `data`. We do not want any `ListHeaderComponent` for that case, right:

```
ListHeaderComponent={
  !!data ? (
    <Wrapper>
      <HtmlView html={data} />
    </Wrapper>
  ) : null
}
```

src/screens/SurveyOverviewScreen.tsx
ardasnturk: Yes, you are right. I didn't think of Skip for this place. I guess if we use skip, we don't need extra checks for name, right?

src/screens/SurveyOverviewScreen.tsx
ardasnturk: Yes, that's right. But since we don't have `query` or `queryVariables` here, I prefer to write hard code. What do you think about that?

```
refreshTimeKey: `publicHtmlFile-${additionalProps?.htmlName}`,
```

src/screens/SurveyOverviewScreen.tsx
ardasnturk: Yeah, you're right. I think I missed that.

ardasnturk: https://github.com/smart-village-solutions/smart-village-app-app/commit/d3e2bb5ccf4ca1da43e67f69877b7b68207df161

src/screens/SurveyOverviewScreen.tsx
ardasnturk: https://github.com/smart-village-solutions/smart-village-app-app/commit/d3e2bb5ccf4ca1da43e67f69877b7b68207df161

src/screens/SurveyOverviewScreen.tsx
ardasnturk: https://github.com/smart-village-solutions/smart-village-app-app/commit/d3e2bb5ccf4ca1da43e67f69877b7b68207df161

src/screens/SurveyOverviewScreen.tsx
donni106: yes exactly what i thought of, the variable part is represented

src/screens/SurveyOverviewScreen.tsx
ardasnturk: Ow. I apologise. It was my carelessness. 🫣
https://github.com/smart-village-solutions/smart-village-app-app/commit/57301acb3d4274e87de430fd447912e8242b4342

donni106 (approved):

d4v3d4vE (approved):
works great!

PR #909, opened 2023-08-25
feat: new module disturber

ardasnturk:
- added `Distruber` component to show distruber module on `HomeScreen`
- added `isActive` and `findClosesItem` helpers to show the active and latest distruber
- added `useStaticContent` hook to pull data from json content added to main server
- added a close button to ensure that the disturber is not shown and when the button is pressed, the disturber's ID is saved in the device's memory and a useEffect is added to ensure that it is not shown when the app is reopened until a new disturber is added
- added null and type checks to handle invalid or missing data gracefully
- used JavaScript's `reduce` and `map` methods for efficient computation
- utilized `Moment.js` for date manipulations
- added `containerStyle` prop to change the style of the button
- updated `PropTypes` to include the new `containerStyle` prop

MQGB-32

To test this feature:
Create a JSON `staticContent` with the name `homeDisurber`.
Here is an example:

```
[
  {
    "picture": {
      "uri": "", // url
      "aspectRatio": {
        "HEIGHT": 0.7,
        "WIDTH": 1
      }
    },
    "backgroundColor": "", // hex
    "id": 1,
    "title": "", // string
    "headline": "", // string
    "description": "", // string
    "dates": [
      {
        "dateStart": "2023-08-22",
        "dateEnd": "2023-08-24"
      }
    ],
    "button": {
      "navigationTo": "Web",
      "title": "", // string
      "params": {
        "webUrl": "" // url
      }
    }
  },
]
```

## Screenshots:

||||
|--|--|--|
Simulator Screenshot - iPhone 14 Pro Max - 2023-09-11 at 12 37 22 | Simulator Screenshot - iPhone 14 Pro Max - 2023-09-11 at 12 38 21 | Simulator Screenshot - iPhone 14 Pro Max - 2023-09-11 at 12 39 11

### DIFF
--- a/src/components/ConnectedImagesCarousel.js
+++ b/src/components/ConnectedImagesCarousel.js
@@ -11,14 +11,9 @@ import { SettingsContext } from '../SettingsProvider';
 import { ImagesCarousel } from './ImagesCarousel';
 import { LoadingContainer } from './LoadingContainer';
 
-export const ConnectedImagesCarousel = ({
-  alternateAspectRatio,
-  navigation,
-  publicJsonFile,
-  refreshTimeKey
-}) => {
+export const ConnectedImagesCarousel = ({ alternateAspectRatio, navigation, publicJsonFile }) => {
   const { data, loading, refetch } = useStaticContent({
-    refreshTimeKey,
+    refreshTimeKey: `publicJsonFile-${publicJsonFile}`,
     name: publicJsonFile,
     type: 'json'
   });
@@ -38,7 +33,7 @@ export const ConnectedImagesCarousel = ({
     <ImagesCarousel
       navigation={navigation}
       data={_shuffle(data)}
-      refreshTimeKey={refreshTimeKey}
+      refreshTimeKey={`publicJsonFile-${publicJsonFile}`}
       aspectRatio={
         alternateAspectRatio
           ? parsedImageAspectRatio(globalSettings?.homeCarousel?.imageAspectRatio)
@@ -51,8 +46,7 @@ export const ConnectedImagesCarousel = ({
 ConnectedImagesCarousel.propTypes = {
   alternateAspectRatio: PropTypes.bool,
   navigation: PropTypes.object.isRequired,
-  publicJsonFile: PropTypes.string.isRequired,
-  refreshTimeKey: PropTypes.string.isRequired
+  publicJsonFile: PropTypes.string.isRequired
 };
 
 ConnectedImagesCarousel.defaultProps = {
--- a/src/components/Disturber.tsx
+++ b/src/components/Disturber.tsx
@@ -1,6 +1,6 @@
 import { DrawerNavigationProp } from '@react-navigation/drawer';
 import React, { useEffect, useState } from 'react';
-import { StyleSheet, Text, TouchableOpacity, View } from 'react-native';
+import { StyleSheet, TouchableOpacity, View } from 'react-native';
 
 import { Icon, colors, normalize } from '../config';
 import { addToStore, findClosestItem, isActive, readFromStore } from '../helpers';
@@ -13,7 +13,7 @@ import { Wrapper, WrapperHorizontal } from './Wrapper';
 
 type Props = {
   navigation: DrawerNavigationProp<any>;
-  name: string;
+  staticContentName: string;
 };
 
 interface DateRange {
@@ -28,7 +28,7 @@ interface DataItem {
     params: { title?: string; webUrl: string };
     title?: string;
   };
-  category?: string;
+  headline?: string;
   dates: DateRange[];
   description: string;
   id: number;
@@ -37,43 +37,54 @@ interface DataItem {
 }
 
 // eslint-disable-next-line complexity
-export const Disturber = ({ navigation, name }: Props) => {
-  const [isVisible, setIsVisiblle] = useState(false);
-  const refreshTimeKey = `publicJsonFile-${name}`;
+export const Disturber = ({ navigation, staticContentName }: Props) => {
+  const [isVisible, setIsVisible] = useState(false);
+  const refreshTimeKey = `publicJsonFile-${staticContentName}`;
 
   const { data, refetch } = useStaticContent<DataItem[]>({
     refreshTimeKey,
-    name,
+    name: staticContentName,
     type: 'json'
   });
 
   useHomeRefresh(refetch);
 
-  // filter data for present items and items with active date/time periods
-  const isActiveData = data?.filter((item) => item && isActive(item));
-
   // find the closest item to the current date/time
-  const closestItem: DataItem | null = findClosestItem(isActiveData || []);
+  const closestItem: DataItem | null = findClosestItem(
+    data?.filter((item) => item && isActive(item)) || []
+  );
 
-  const setDistruberComplete = () => {
-    setIsVisiblle((prev) => !prev);
-    addToStore(name, closestItem?.id.toString());
+  const setDisturberComplete = () => {
+    setIsVisible((prev) => !prev);
+    addToStore(staticContentName, closestItem?.id.toString());
   };
 
   useEffect(() => {
-    readFromStore(name).then((value) => {
-      if (closestItem?.id && value !== closestItem?.id.toString()) setIsVisiblle(true);
-    });
+    const disturberStatus = async () => {
+      try {
+        const disturberComplete = await readFromStore(staticContentName);
+
+        if (closestItem?.id && disturberComplete !== closestItem?.id.toString()) {
+          setIsVisible(true);
+        }
+      } catch (e) {
+        setIsVisible(false);
+
+        console.error(e);
+      }
+    };
+
+    disturberStatus();
   }, [closestItem]);
 
   if (!isVisible || !closestItem) return null;
 
   return (
-    <Wrapper style={{ padding: normalize(24) }}>
+    <Wrapper style={styles.wrapperPadding}>
       <View style={[styles.containerRadius, { backgroundColor: closestItem.backgroundColor }]}>
         {!!closestItem && (
           <>
-            <TouchableOpacity onPress={setDistruberComplete} style={styles.closeButton}>
+            <TouchableOpacity onPress={setDisturberComplete} style={styles.closeButton}>
               <Icon.Close color={colors.darkText} size={normalize(16)} />
             </TouchableOpacity>
 
@@ -82,15 +93,15 @@ export const Disturber = ({ navigation, name }: Props) => {
                 source={closestItem.picture}
                 borderRadius={normalize(8)}
                 aspectRatio={closestItem.picture.aspectRatio || { HEIGHT: 0.7, WIDTH: 1 }}
-                resizeMode={'cover'}
+                resizeMode="cover"
               />
             )}
 
             <Wrapper style={styles.noPaddingBottom}>
               <WrapperHorizontal>
-                {!!closestItem.category && (
-                  <BoldText center uppercase style={styles.categoryText}>
-                    {closestItem.category}
+                {!!closestItem.headline && (
+                  <BoldText center uppercase style={styles.headlineText}>
+                    {closestItem.headline}
                   </BoldText>
                 )}
               </WrapperHorizontal>
@@ -116,17 +127,19 @@ export const Disturber = ({ navigation, name }: Props) => {
 
             <Wrapper>
               <WrapperHorizontal>
-                {!!closestItem.button && (
+                {!!closestItem.button && closestItem.button.title && (
                   <Button
                     extraLarge
                     containerStyle={styles.containerRadius}
-                    title={closestItem.button.title || 'Mehr'}
-                    onPress={() =>
-                      navigation.navigate(
-                        closestItem.button.navigationTo,
-                        closestItem.button.params
-                      )
-                    }
+                    title={closestItem.button.title}
+                    onPress={() => {
+                      if (closestItem.button.navigationTo && closestItem.button.params) {
+                        navigation.navigate(
+                          closestItem.button.navigationTo,
+                          closestItem.button.params
+                        );
+                      }
+                    }}
                   />
                 )}
               </WrapperHorizontal>
@@ -139,12 +152,6 @@ export const Disturber = ({ navigation, name }: Props) => {
 };
 
 const styles = StyleSheet.create({
-  categoryText: {
-    fontSize: normalize(14),
-    fontWeight: '700',
-    lineHeight: normalize(16),
-    textTransform: 'uppercase'
-  },
   closeButton: {
     alignItems: 'center',
     backgroundColor: colors.gray20,
@@ -161,10 +168,19 @@ const styles = StyleSheet.create({
   containerRadius: {
     borderRadius: normalize(8)
   },
+  headlineText: {
+    fontSize: normalize(14),
+    fontWeight: '700',
+    lineHeight: normalize(16),
+    textTransform: 'uppercase'
+  },
   noPaddingBottom: {
     paddingBottom: 0
   },
   noPaddingTop: {
     paddingTop: 0
+  },
+  wrapperPadding: {
+    padding: normalize(24)
   }
 });
--- a/src/components/Disturber.tsx
+++ b/src/components/Disturber.tsx
@@ -1,0 +1,170 @@
+import { DrawerNavigationProp } from '@react-navigation/drawer';
+import React, { useEffect, useState } from 'react';
+import { StyleSheet, Text, TouchableOpacity, View } from 'react-native';
+
+import { Icon, colors, normalize } from '../config';
+import { addToStore, findClosestItem, isActive, readFromStore } from '../helpers';
+import { useHomeRefresh, useStaticContent } from '../hooks';
+
+import { Button } from './Button';
+import { Image } from './Image';
+import { BoldText, RegularText } from './Text';
+import { Wrapper, WrapperHorizontal } from './Wrapper';
+
+type Props = {
+  navigation: DrawerNavigationProp<any>;
+  name: string;
+};
+
+interface DateRange {
+  dateEnd: string;
+  dateStart: string;
+}
+
+interface DataItem {
+  backgroundColor?: string;
+  button: {
+    navigationTo: string;
+    params: { title?: string; webUrl: string };
+    title?: string;
+  };
+  category?: string;
+  dates: DateRange[];
+  description: string;
+  id: number;
+  picture?: { aspectRatio?: { HEIGHT: number; WIDTH: number }; uri: string };
+  title: string;
+}
+
+// eslint-disable-next-line complexity
+export const Disturber = ({ navigation, name }: Props) => {
+  const [isVisible, setIsVisiblle] = useState(false);
+  const refreshTimeKey = `publicJsonFile-${name}`;
+
+  const { data, refetch } = useStaticContent<DataItem[]>({
+    refreshTimeKey,
+    name,
+    type: 'json'
+  });
+
+  useHomeRefresh(refetch);
+
+  // filter data for present items and items with active date/time periods
+  const isActiveData = data?.filter((item) => item && isActive(item));
+
+  // find the closest item to the current date/time
+  const closestItem: DataItem | null = findClosestItem(isActiveData || []);
+
+  const setDistruberComplete = () => {
+    setIsVisiblle((prev) => !prev);
+    addToStore(name, closestItem?.id.toString());
+  };
+
+  useEffect(() => {
+    readFromStore(name).then((value) => {
+      if (closestItem?.id && value !== closestItem?.id.toString()) setIsVisiblle(true);
+    });
+  }, [closestItem]);
+
+  if (!isVisible || !closestItem) return null;
+
+  return (
+    <Wrapper style={{ padding: normalize(24) }}>
+      <View style={[styles.containerRadius, { backgroundColor: closestItem.backgroundColor }]}>
+        {!!closestItem && (
+          <>
+            <TouchableOpacity onPress={setDistruberComplete} style={styles.closeButton}>
+              <Icon.Close color={colors.darkText} size={normalize(16)} />
+            </TouchableOpacity>
+
+            {!!closestItem.picture && (
+              <Image
+                source={closestItem.picture}
+                borderRadius={normalize(8)}
+                aspectRatio={closestItem.picture.aspectRatio || { HEIGHT: 0.7, WIDTH: 1 }}
+                resizeMode={'cover'}
+              />
+            )}
+
+            <Wrapper style={styles.noPaddingBottom}>
+              <WrapperHorizontal>
+                {!!closestItem.category && (
+                  <BoldText center uppercase style={styles.categoryText}>
+                    {closestItem.category}
+                  </BoldText>
+                )}
+              </WrapperHorizontal>
+            </Wrapper>
+
+            <Wrapper style={styles.noPaddingTop}>
+              <WrapperHorizontal>
+                {!!closestItem.title && (
+                  <BoldText center large>
+                    {closestItem.title}
+                  </BoldText>
+                )}
+              </WrapperHorizontal>
+            </Wrapper>
+
+            <Wrapper style={styles.noPaddingTop}>
+              <WrapperHorizontal>
+                {!!closestItem.description && (
+                  <RegularText center>{closestItem.description}</RegularText>
+                )}
+              </WrapperHorizontal>
+            </Wrapper>
+
+            <Wrapper>
+              <WrapperHorizontal>
+                {!!closestItem.button && (
+                  <Button
+                    extraLarge
+                    containerStyle={styles.containerRadius}
+                    title={closestItem.button.title || 'Mehr'}
+                    onPress={() =>
+                      navigation.navigate(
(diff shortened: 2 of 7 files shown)


d4v3d4vE (approved):

src/components/Disturber.tsx
```diff
@@ -0,0 +1,170 @@
...
+  useEffect(() => {
+    readFromStore(name).then((value) => {
+      if (closestItem?.id && value !== closestItem?.id.toString()) setIsVisiblle(true);
+    });
+  }, [closestItem]);
+
+  if (!isVisible || !closestItem) return null;
+
+  return (
+    <Wrapper style={{ padding: normalize(24) }}>
```
donni106, line 72: can the wrapper style be excluded in styles at the bottom, like in `<Wrapper style={styles.noPaddingTop}>` or at some additional prop in `Wrapper`?

src/components/Disturber.tsx
```diff
@@ -0,0 +1,170 @@
...
+            <TouchableOpacity onPress={setDistruberComplete} style={styles.closeButton}>
+              <Icon.Close color={colors.darkText} size={normalize(16)} />
+            </TouchableOpacity>
+
+            {!!closestItem.picture && (
+              <Image
+                source={closestItem.picture}
+                borderRadius={normalize(8)}
+                aspectRatio={closestItem.picture.aspectRatio || { HEIGHT: 0.7, WIDTH: 1 }}
+                resizeMode={'cover'}
```
donni106, line 85: we have a string here, so i suggest `resizeMode="cover"` for readability here

src/components/Disturber.tsx
```diff
@@ -0,0 +1,170 @@
...
+              </WrapperHorizontal>
+            </Wrapper>
+
+            <Wrapper>
+              <WrapperHorizontal>
+                {!!closestItem.button && (
+                  <Button
+                    extraLarge
+                    containerStyle={styles.containerRadius}
+                    title={closestItem.button.title || 'Mehr'}
```
donni106, line 123: can you move `'Mehr'` in our texts file please.
or we avoid a fallback and say the button title is required to not show a button if no title is given. i do not think that "Mehr" is something the clients want to have in any case.

src/components/Disturber.tsx
```diff
@@ -0,0 +1,170 @@
...
+            </Wrapper>
+          </>
+        )}
+      </View>
+    </Wrapper>
+  );
+};
+
+const styles = StyleSheet.create({
+  categoryText: {
```
donni106, line 142: please adjust naming here as well, considering that "category" is the wrong naming

donni106 (changes requested):

src/screens/HomeScreen.js
```diff
@@ -173,6 +175,7 @@ export const HomeScreen = ({ navigation, route }) => {
     limitNews = 15,
     limitPointsOfInterestAndTours = 15
   } = sections;
+  const { show: showDisturber = false, content: disturberContent = 'Disturber' } = disturber;
```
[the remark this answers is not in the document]
ardasnturk: Yeah, you're absolutely right. I'll change that.

src/screens/HomeScreen.js
```diff
@@ -276,6 +279,8 @@ export const HomeScreen = ({ navigation, route }) => {
             />
 
             <Widgets widgetConfigs={widgetConfigs} />
+
+            {showDisturber && <Disturber navigation={navigation} name={disturberContent} />}
```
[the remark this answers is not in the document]
ardasnturk: Since there is a `name` prop in `useStaticContent`, I added the `name` prop directly. But I will change it to make it more understandable. 👍

src/components/Disturber.tsx
```diff
@@ -0,0 +1,170 @@
...
+  const setDistruberComplete = () => {
+    setIsVisiblle((prev) => !prev);
+    addToStore(name, closestItem?.id.toString());
+  };
+
+  useEffect(() => {
+    readFromStore(name).then((value) => {
+      if (closestItem?.id && value !== closestItem?.id.toString()) setIsVisiblle(true);
+    });
+  }, [closestItem]);
```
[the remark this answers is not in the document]
ardasnturk: I think there's a typo. I want to hide the storer module after pressing the x button until a new storer is created.

src/components/Disturber.tsx
```diff
@@ -0,0 +1,170 @@
...
+    type: 'json'
+  });
+
+  useHomeRefresh(refetch);
+
+  // filter data for present items and items with active date/time periods
+  const isActiveData = data?.filter((item) => item && isActive(item));
+
+  // find the closest item to the current date/time
+  const closestItem: DataItem | null = findClosestItem(isActiveData || []);
```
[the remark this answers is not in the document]
ardasnturk: I developed this logic for the storer module, to find the most recent storer module to show.

src/components/Button.js
```diff
@@ -46,26 +59,47 @@ export const Button = ({ disabled, invert, notFullWidth, onPress, title }) => {
       titleStyle={[
         styles.titleStyle,
         invert && styles.titleStyleInvert,
-        needLandscapeStyle && styles.titleStyleLandscape
+        needLandscapeStyle && styles.titleStyleLandscape,
+        (extraLarge || large || medium) && styles.spatialButtonTitle,
```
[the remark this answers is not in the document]
ardasnturk: All 3 different button sizes have the same text style. So I added a single text style for 3 different button sizes. I can update the name to make it clearer. Do you have any suggestions for that?

src/components/Disturber.tsx
```diff
@@ -0,0 +1,170 @@
...
+  dates: DateRange[];
+  description: string;
+  id: number;
+  picture?: { aspectRatio?: { HEIGHT: number; WIDTH: number }; uri: string };
+  title: string;
+}
+
+// eslint-disable-next-line complexity
+export const Disturber = ({ navigation, name }: Props) => {
+  const [isVisible, setIsVisiblle] = useState(false);
```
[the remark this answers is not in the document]
ardasnturk: Ow. Updating.

src/components/Disturber.tsx
```diff
@@ -0,0 +1,170 @@
...
+
+  useHomeRefresh(refetch);
+
+  // filter data for present items and items with active date/time periods
+  const isActiveData = data?.filter((item) => item && isActive(item));
+
+  // find the closest item to the current date/time
+  const closestItem: DataItem | null = findClosestItem(isActiveData || []);
+
+  const setDistruberComplete = () => {
```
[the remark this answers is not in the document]
ardasnturk: Ow. Updating.

src/components/Disturber.tsx
ardasnturk: Yeah, you're absolutely right. I missed it. I'll fix it.

src/components/Disturber.tsx
ardasnturk: Ahahah, that's right. I don't know why I did that.

src/components/Disturber.tsx
ardasnturk: Yes, you're right. I guess we should add some controls to the onPress prop, right? Otherwise the application could crash? Like this:
```
onPress={() => {
  if (closestItem.button.navigationTo && closestItem.button.params) {
    navigation.navigate(
      closestItem.button.navigationTo,
      closestItem.button.params
    );
  }
}}
```

src/components/Disturber.tsx
ardasnturk: Why don't you like `category`? Should it be `headline`?

src/screens/HomeScreen.js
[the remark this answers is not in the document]
ardasnturk: I can fix this. But won't this change cause confusion in the future? Figma says that this is a `category`, but in our code it will be a `headline`. 🤷‍♂️ Yes, we don't have to apply everything in Figma directly to the code, but then I suggest to change what is in Figma to avoid confusion later on.

src/components/Disturber.tsx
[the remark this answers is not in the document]
donni106: can we get rid of `isActiveData` which is just a buffer variable that is only used once in my eyes? couldn't we throw `data?.filter((item) => item && isActive(item))` directly in `findClosestItem`?

src/components/Disturber.tsx
donni106: yes, because it is not a category but context free text

src/screens/HomeScreen.js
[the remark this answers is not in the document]
donni106: we do not have a context to any category in this case. it is just a whatever text. it is not setup correctly in Figma in my eyes.

src/components/Disturber.tsx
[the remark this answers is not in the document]
ardasnturk: Yeah, we can do that.

src/components/Button.js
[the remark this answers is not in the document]
donni106: again: will there be conflicts? if we merge this, do we have incompleted MQGB-47 stuff? or do we need to finish MQGB-47 first?

src/components/Button.js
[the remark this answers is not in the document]
donni106: yes, like `small && styles.smallButtonTitle` all its own, to be able to change all of those separately depending on requirements.
i guess spatial isn't the word you wanted to use? did you mean special?

src/components/Disturber.tsx
```diff
@@ -0,0 +1,186 @@
...
+              <WrapperHorizontal>
+                {!!closestItem.description && (
+                  <RegularText center>{closestItem.description}</RegularText>
+                )}
+              </WrapperHorizontal>
+            </Wrapper>
+
+            <Wrapper>
+              <WrapperHorizontal>
+                {!!closestItem.button && closestItem.button.title && (
```
donni106, line 130: as the title is a string, please add `!!` to boolean it

src/components/Disturber.tsx
donni106: let's check for `navigateTo` and `params` already with the title instead of in the onPress. if we do not have values set, we do not need to render the button at all, right?!

ardasnturk: I've added the controls you wanted for the button. Don't you think this control is too much?

https://github.com/smart-village-solutions/smart-village-app-app/pull/909/commits/790533a892a73135a641652cecc7dd521444636d

src/components/Disturber.tsx
ardasnturk: Yes, you are right. I fixed it.

https://github.com/smart-village-solutions/smart-village-app-app/pull/909/commits/790533a892a73135a641652cecc7dd521444636d

src/components/Button.js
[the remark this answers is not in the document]
ardasnturk: I added new styles for different text styles and left the styles empty. Since all styles are the same, I added the necessary styles to `titleStyle`. 

Yes special would be more correct.

https://github.com/smart-village-solutions/smart-village-app-app/pull/909/commits/790533a892a73135a641652cecc7dd521444636d

src/components/Button.js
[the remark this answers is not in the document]
ardasnturk: It would be better to finish MQGB-47 first, but I don't think there will be a big conflict. There will be a conflict with my last commit.

src/components/Disturber.tsx
donni106: what should be too much? you do not need to check more than we will use but at least the values, we need, to show a proper button with working navigation.

src/components/Button.js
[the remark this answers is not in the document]
donni106: i cannot follow this decision. if there are no styles to apply, we do not need any code for that.
with changing `titleStyle` alle buttons will look different after merge? please compare the implicit values with the one you added with the commit. do they differ?

src/components/Disturber.tsx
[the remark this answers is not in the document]
donni106: can you correct the wording in that commit please?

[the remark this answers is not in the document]
ardasnturk: I have edited and resent the commit message. 👍

https://github.com/smart-village-solutions/smart-village-app-app/pull/909/commits/7c2d05bf4e11b569ab029b69e66966b13792e2b9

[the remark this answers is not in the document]
donni106: and adjust in the PR description as well please

src/components/Button.js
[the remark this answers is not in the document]
ardasnturk: But this time we need to consider that now all buttons will appear in this style as standard. I don't know if this is the right way to do it.

|before|after|
|--|--|
Simulator Screenshot - iPhone 14 Pro - 2023-09-18 at 12 33 18 | Simulator Screenshot - iPhone 14 Pro - 2023-09-18 at 12 33 23


As you can see in the screenshots, the size of the buttons shrinks as the size of the text inside the buttons shrinks. I think it would be more correct to add `specialButtonStyle`. Or extra styles adjusted for all button sizes. But all styles will be the same. For example:

```
  extraLargeButtonTitle: {
    fontSize: normalize(14),
    fontWeight: '600',
    lineHeight: normalize(18),
  },
  largeButtonTitle: {
    fontSize: normalize(14),
    fontWeight: '600',
    lineHeight: normalize(18),
  },
  mediumButtonTitle: {
    fontSize: normalize(14),
    fontWeight: '600',
    lineHeight: normalize(18),
  },
```

or: 
```
  specialButtonTitle: {
    fontSize: normalize(14),
    fontWeight: '600',
    lineHeight: normalize(18),
  }
```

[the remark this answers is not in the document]
ardasnturk: I think the code is as it should be right now, isn't it?

new styles created for title are the same, but I have added new styles according to the props to make it easier to update in the future

https://github.com/smart-village-solutions/smart-village-app-app/commit/1a047b5e20ef37fbd3576cd8d20ff7a8ac819a46

src/screens/HomeScreen.js
```diff
@@ -276,6 +277,8 @@ export const HomeScreen = ({ navigation, route }) => {
             />
 
             <Widgets widgetConfigs={widgetConfigs} />
+
+            <Disturber navigation={navigation} staticContentName="homeDisturber" />
```
donni106, line 281: compared to the lines before for connected images carousel, what about using the same prop instead of `staticContentName`? 

```
<ConnectedImagesCarousel
  alternateAspectRatio
  navigation={navigation}
  publicJsonFile="homeCarousel"
  refreshTimeKey="publicJsonFile-homeCarousel"
/>

<Widgets widgetConfigs={widgetConfigs} />

<Disturber navigation={navigation} staticContentName="homeDisturber" />
```

do we use `staticContentName` already somewhere else?
how can we achieve the most aligned result for the code base?

src/components/Button.js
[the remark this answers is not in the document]
ardasnturk: I sent the update for button to MQGB-47 branch and rebase this branch with MQGB-47. Then I updated the target.

src/screens/HomeScreen.js
ardasnturk: Yes, that's exactly right! I made this change to improve the understandability of the code. Now the components are more similar to each other.

https://github.com/smart-village-solutions/smart-village-app-app/pull/909/commits/2852592d533e68dc4016d32fb7c84d40c0faee51

src/components/Button.js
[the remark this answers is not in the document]
donni106: and the additional `containerStyle` is correctly placed here in this PR and not as part of #911?

src/components/ConnectedImagesCarousel.js
```diff
@@ -11,12 +11,9 @@ import { SettingsContext } from '../SettingsProvider';
...
-  alternateAspectRatio,
-  navigation,
-  publicJsonFile,
-  refreshTimeKey
-}) => {
+export const ConnectedImagesCarousel = ({ alternateAspectRatio, navigation, publicJsonFile }) => {
+  const refreshTimeKey = `publicJsonFile-${publicJsonFile}`;
+
   const { data, loading, refetch } = useStaticContent({
     refreshTimeKey,
```
donni106, line 18: please do it the same way everywhere: 

```
refreshTimeKey: `publicJsonFile-${publicJsonFile}`,
```

we do not need an extra variable `refreshTimeKey` which is only used once and builds a string.

src/helpers/momentHelper.js
```diff
@@ -74,3 +74,56 @@ export const isActive = (item) => {
...
+ *
+ * @param {DataItem[]} data an array of objects, each with a `dates` array
+ *
+ * @return {DataItem | null} the item with the closest date range to today, or null if no such item exists
+ */
+export const findClosestItem = (data) => {
+  // Check if data is null or undefined or not an array
+  if (!Array.isArray(data)) {
+    return null;
+  }
```
donni106, line 90: the comment does not fit the code, or am i wrong? this checks for "not an array" only?
i suggest not using a comment that translates the code in english. better would be a comment explaining what is it useful for?

src/helpers/momentHelper.js
```diff
@@ -74,3 +74,56 @@ export const isActive = (item) => {
...
   return true;
 };
+
+/**
+ *
+ * Find the item with the closest date range to today.
+ *
+ * @param {DataItem[]} data an array of objects, each with a `dates` array
+ *
+ * @return {DataItem | null} the item with the closest date range to today, or null if no such item exists
```
donni106, line 84: do your prefer `null` over `undefined` in general? what is the benefit of having `null`s here for `findClosestItem`?

src/helpers/momentHelper.js
```diff
@@ -74,3 +74,56 @@ export const isActive = (item) => {
...
+          }
+
+          // Convert the start and end dates to time in milliseconds.
+          const dateStart = moment(dateRange.dateStart).valueOf();
+          const dateEnd = moment(dateRange.dateEnd).valueOf();
+
+          // Calculate the absolute difference between today and both the start and end dates,
+          // and return the smaller of the two.
+          return Math.min(
+            Math.abs(today.getTime() - dateStart),
```
donni106, line 116: `today` is only needed here, right? so you could move `const today = new Date();` right before this return, to avoid having the variable created in all the other cases, where it is not needed.

src/helpers/momentHelper.js
```diff
@@ -74,3 +74,56 @@ export const isActive = (item) => {
...
+            Math.abs(today.getTime() - dateEnd)
+          );
```
donni106, line 118: `today` is only needed here, right? so you could move `const today = new Date();` right before this return, to avoid having the variable created in all the other cases, where it is not needed.

src/helpers/momentHelper.js
```diff
@@ -74,3 +74,56 @@ export const isActive = (item) => {
...
+        })
+      );
+
+      // If the calculated minimum difference is smaller than the current smallest difference,
+      // update the closest object.
+      return minDifference < closest.minDifference ? { minDifference, item } : closest;
+    },
+    // Start with an initial closest object where minDifference is Infinity and item is null.
```
donni106, line 126: this again translates code into english, right? i think we do not need it.

src/helpers/momentHelper.js
ardasnturk: Yes, you're right, I'll update the comment line.

src/helpers/momentHelper.js
ardasnturk: I don't think there is any advantage or disadvantage of using null over undefined, but I believe that undefined would be more accurate. You are right.

src/components/ConnectedImagesCarousel.js
ardasnturk: https://github.com/smart-village-solutions/smart-village-app-app/commit/80ef948ed83acb0dcc7d7280796271fc8b02d6da

src/helpers/momentHelper.js
ardasnturk: https://github.com/smart-village-solutions/smart-village-app-app/pull/909/commits/d664c3f0a762f98b5502f2da02380b3c1cbd2e5b

src/helpers/momentHelper.js
ardasnturk: https://github.com/smart-village-solutions/smart-village-app-app/pull/909/commits/d664c3f0a762f98b5502f2da02380b3c1cbd2e5b

src/helpers/momentHelper.js
ardasnturk: https://github.com/smart-village-solutions/smart-village-app-app/pull/909/commits/d664c3f0a762f98b5502f2da02380b3c1cbd2e5b

src/helpers/momentHelper.js
ardasnturk: https://github.com/smart-village-solutions/smart-village-app-app/pull/909/commits/d664c3f0a762f98b5502f2da02380b3c1cbd2e5b

src/helpers/momentHelper.js
```diff
@@ -74,3 +74,56 @@ export const isActive = (item) => {
...
+    { minDifference: Infinity, item: null }
```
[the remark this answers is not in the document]
ardasnturk: Initial value in reduce: The `reduce` function starts with an initial value for the accumulator (`closest`). In this initial value, `minDifference` is set to `Infinity`. The idea is that as you iterate over the data array, you can compare this Infinity value with the actual time differences between the date ranges and today's date. Any real number will be smaller than Infinity, so the first time a valid `minDifference` is calculated, it will replace Infinity.

src/components/Button.js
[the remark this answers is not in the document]
ardasnturk: I have reorganised the commits. 👍

src/components/Disturber.tsx
```diff
@@ -0,0 +1,186 @@
...
+
+  // find the closest item to the current date/time
+  const closestItem: DataItem | null = findClosestItem(
+    data?.filter((item) => item && isActive(item)) || []
+  );
+
+  const setDisturberComplete = () => {
+    setIsVisible((prev) => !prev);
+    addToStore(publicJsonFile, closestItem?.id.toString());
+  };
```
donni106, line 59: line 53: fallback should be `{}` instead of `[]`, or am i wrong? DataItem is not an array.
on the other hand the type is `DataItem | null`, so there is no fallback needed, as it is ok to be null, right?

line 58: if `closestItem` would be an `[]` you wouldn't be able to call `?.id` on it. anyways, if there is no `closestItem` we don't want to add anything to store, right? so we should call `addToStore` just in case we have an item to store?

src/components/Disturber.tsx
```diff
@@ -0,0 +1,186 @@
...
+
+  useEffect(() => {
+    const disturberStatus = async () => {
+      try {
+        const disturberComplete = await readFromStore(publicJsonFile);
+
+        if (closestItem?.id && disturberComplete !== closestItem?.id.toString()) {
```
donni106, line 66: with optional chaining `disturberComplete !== closestItem?.id?.toString()` would be enough here. it will check the ability to call `id` and `toString` implicitly

src/components/Disturber.tsx
```diff
@@ -0,0 +1,186 @@
...
+
+    disturberStatus();
+  }, [closestItem]);
+
+  if (!isVisible || !closestItem) return null;
+
+  return (
+    <Wrapper style={styles.wrapperPadding}>
+      <View style={[styles.containerRadius, { backgroundColor: closestItem.backgroundColor }]}>
+        {!!closestItem && (
```
donni106, line 84: there is already an return if there is no item in line 79, so we do not need to check for presence here.

donni106 (changes requested):

src/components/Disturber.tsx
ardasnturk: I think you are wrong. Because here I need to give the format that `findClosesItem` expects. `findClosestItem` is expecting an array. I think there should be an empty array. `closestItem` does not have a fallback, we give a value for the `findClosestItem` helper. This value should be an array.

The `setDisturberComplete` function only works when we press the button to close the disturber. If there is no `closestItem`, the disturber does not appear. For this reason, if `closestItem.id` does not exist, there is no disturber. So I think there is no need to add an extra control here.

donni106: 1. oh yes i see about the array. thanks.

2. ok, let's remove the ?

ardasnturk: Yes, we can remove the ? 👍

ardasnturk: But in this case, we get an error because closestItem may return null value.
image

donni106: could this be fixed with `closestItem && addToStore`?

ardasnturk: Yeah!! ✌️
https://github.com/smart-village-solutions/smart-village-app-app/pull/909/commits/4dd1f20a139e88749b4a6e145ef93f008948fcab

src/components/Disturber.tsx
ardasnturk: https://github.com/smart-village-solutions/smart-village-app-app/pull/909/commits/4dd1f20a139e88749b4a6e145ef93f008948fcab

src/components/Disturber.tsx
ardasnturk: https://github.com/smart-village-solutions/smart-village-app-app/pull/909/commits/4dd1f20a139e88749b4a6e145ef93f008948fcab

donni106 (approved):

PR #910, opened 2023-08-28
feature: add poi preview in full screen map

ardasnturk:
- added `zoomLevelForMaps` and `locationService` from settings to the `Map` component for configurable zoom and location features
- introduced `activeMarker` state to keep track of the currently selected map marker
- added `onViewableItemsChanged` ref function to update `activeMarker` based on viewable items in a `FlatList`
- updated `useEffect` to animate map focus based on the `activeMarker`
- extended marker `onPress` to scroll the `FlatList` to the clicked marker
- included a `FlatList` to display points of interest when `showPOIsFullScreenMap` is true
- created a new `POIMapView` component to display points of interest in a `FlatList`
- included image and text display within a touchable area
- added navigation to detail screen on touch
- extended the `mapToMapMarkers` function to include images, title, and details in the returned object for better map marker information
- added `showPOIsFullScreenMap` prop to the `Map` component to conditionally adjust map height
- introduced `stylesWithProps` function to dynamically set map height based on `navigationType`
- added `selectedPointOfInterest` state to manage selected points of interest
- added `showPOIsFullScreenMap` to `settings` object with default value as false
- added `useLayoutEffect` to conditionally set `headerLeft` based on query and `showMap`
- added close icon as `backImage` when query is points of interest and `showMap` is true
- passed `showPOIsFullScreenMap` prop to `LocationOverview` component
- modified condition for rendering `IndexMapSwitch` based on selected `filterType`
- added the `showPOIsFullScreenMap` prop to the `LocationOverview` component to allow control over the display of POIs on a full screen map
- added `backImage` prop to `HeaderLeft` component
- modified the `TouchableOpacity` to conditionally render `backImage` or default `Icon.ArrowLeft`

MQGB-34

To make this feature available, add the following code to the `settings` section in `globalSettings` on the main server.

```
"showPOIsFullScreenMap": true
```
### Drawer navigation:
|list|map|map|
|--|--|--|
Simulator Screenshot - iPhone 14 Pro - 2023-08-28 at 17 47 53 | Simulator Screenshot - iPhone 14 Pro - 2023-08-28 at 17 47 57 | Simulator Screenshot - iPhone 14 Pro - 2023-08-28 at 17 48 06

### Tab navigation:
|list|map|map|
|--|--|--|
Simulator Screenshot - iPhone 14 Pro - 2023-08-28 at 17 48 54 | Simulator Screenshot - iPhone 14 Pro - 2023-08-28 at 17 48 58 | Simulator Screenshot - iPhone 14 Pro - 2023-08-28 at 17 49 05

### DIFF
--- a/src/components/map/Map.tsx
+++ b/src/components/map/Map.tsx
@@ -1,5 +1,5 @@
-import React, { useContext, useRef } from 'react';
-import { StyleProp, StyleSheet, TouchableOpacity, View, ViewStyle } from 'react-native';
+import React, { useContext, useEffect, useRef, useState } from 'react';
+import { FlatList, StyleProp, StyleSheet, TouchableOpacity, View, ViewStyle } from 'react-native';
 import MapView, { LatLng, MAP_TYPES, Marker, Polyline, Region, UrlTile } from 'react-native-maps';
 import { SvgXml } from 'react-native-svg';
 
@@ -7,6 +7,7 @@ import { colors, device, Icon, normalize } from '../../config';
 import { imageHeight, imageWidth } from '../../helpers';
 import { SettingsContext } from '../../SettingsProvider';
 import { MapMarker } from '../../types';
+import { POIMapView } from '../POIMapView';
 
 type Props = {
   geometryTourData?: LatLng[];
@@ -18,6 +19,7 @@ type Props = {
   onMapPress?: () => void;
   onMarkerPress?: (arg0?: string) => void;
   onMaximizeButtonPress?: () => void;
+  showPOIsFullScreenMap?: boolean;
   showsUserLocation?: boolean;
   style?: StyleProp<ViewStyle>;
 };
@@ -26,7 +28,7 @@ const MARKER_ICON_SIZE = normalize(40);
 
 export const Map = ({
   geometryTourData,
-  isMaximizeButtonVisible,
+  isMaximizeButtonVisible = false,
   isMultipleMarkersMap = false,
   locations,
   mapCenterPosition,
@@ -34,6 +36,7 @@ export const Map = ({
   onMapPress,
   onMarkerPress,
   onMaximizeButtonPress,
+  showPOIsFullScreenMap,
   style,
   ...otherProps
 }: Props) => {
@@ -41,12 +44,16 @@ export const Map = ({
   const { settings = {} } = globalSettings;
   const { zoomLevelForMaps = {}, locationService = {} } = settings;
 
+  // state to keep track of the currently active marker
+  const [activeMarker, setActiveMarker] = useState(null);
+
   const showsUserLocation = otherProps.showsUserLocation ?? !!locationService;
   const zoom = isMultipleMarkersMap
     ? zoomLevelForMaps.multipleMarkers
     : zoomLevelForMaps.singleMarker;
 
   const refForMapView = useRef<MapView>(null);
+  const flatListRef = useRef<FlatList>(null);
   // LATITUDE_DELTA handles the zoom, see: https://github.com/react-native-maps/react-native-maps/issues/2129#issuecomment-457056572
   const LATITUDE_DELTA = zoom || 0.0922;
   // example for longitude delta: https://github.com/react-native-maps/react-native-maps/blob/0.30.x/example/examples/DisplayLatLng.js#L18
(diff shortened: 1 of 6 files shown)


ardasnturk:
> Works and looks great! In addition, would it be possible to change the marker color to for example `colors.secondary` of the marker that is selected for better recognition of the selected location?

Yes, you're absolutely right. I tried that but I couldn't change the colors of the marker icons. 👍

src/components/map/Map.tsx
```diff
@@ -74,6 +81,37 @@ export const Map = ({
...
+    if (firstViewableItem) {
+      setActiveMarker(firstViewableItem.id);
+    }
+  }).current;
+
+  // useEffect to handle changes in activeMarker
+  // it animates the map to focus on the coordinates of the active marker
+  useEffect(() => {
+    if (activeMarker && refForMapView.current) {
+      const marker = locations?.find((loc) => loc.id === activeMarker);
```
donni106, line 97: coming back to the marker hint of @d4v3d4vE, commenting it here as I cannot find a better place, but to be able to have a connected conversation:
we need that active marker with a different color to finish this PR.

src/screens/IndexScreen.js
```diff
@@ -367,6 +369,40 @@ export const IndexScreen = ({ navigation, route }) => {
...
+            onPress={() => {
+              const selectedFilter = filterType.find((entry) => entry.selected);
+              const updatedFilter = filterType.map((entry) => {
+                if (entry.id !== selectedFilter.id) {
+                  entry.selected = true;
+                } else {
+                  entry.selected = false;
+                }
+                return entry;
+              });
```
donni106, line 386: this will not work if the filters are more than two, because then more than 1 would be set to true here, as only 1 can be selected at the same time.
so thinking again, what do we want here. we want to "jump" in the previous screen view which is the same as before. we want to kind of "rest" the header left situation. in this case, we do not need to have any logics but to set the initial state: `onPress={() => setFilterType(INITIAL_FILTER)}`

src/screens/IndexScreen.js
```diff
@@ -616,9 +653,10 @@ export const IndexScreen = ({ navigation, route }) => {
           }}
         </Query>
       )}
-      {query === QUERY_TYPES.POINTS_OF_INTEREST && (
-        <IndexMapSwitch filter={filterType} setFilter={setFilterType} />
-      )}
+      {query === QUERY_TYPES.POINTS_OF_INTEREST &&
+        filterType.find((i) => i.selected && i.title === 'Listenansicht') && (
```
donni106, line 657: please do not use `i` as it suggests some `index` counter. 
earlier in this file you have used `filterType.find((entry) => entry.selected);`. this would be the better alternative here as well.

donni106, line 657: no hardcoded text please, we already have `texts.locationOverview.list` for this

donni106, line 657: the logic you implemented with `find` returns an object or not. luckily that works and resolves to true. i suggest a refactoring to really rely on a boolean here: `filterType.find((entry) => entry.title == texts.locationOverview.list)?.selected`

src/screens/IndexScreen.js
```diff
@@ -367,6 +369,40 @@ export const IndexScreen = ({ navigation, route }) => {
...
+                color={colors.lightestText}
+                style={{ paddingHorizontal: normalize(14) }}
+              />
+            }
+          />
+        )
+      });
+    } else {
+      navigation.setOptions({
+        headerLeft: () => <HeaderLeft onPress={() => navigation.goBack()} />
```
donni106, line 401: sadly this effect is running all the time on index screens now. but ok, with having our index screen refactored soon, we will have this logic only remaining for the index of points of interest.

src/components/map/Map.tsx
```diff
@@ -26,27 +28,32 @@ const MARKER_ICON_SIZE = normalize(40);
...
   const { settings = {} } = globalSettings;
   const { zoomLevelForMaps = {}, locationService = {} } = settings;
 
+  // state to keep track of the currently active marker
+  const [activeMarker, setActiveMarker] = useState(null);
+
   const showsUserLocation = otherProps.showsUserLocation ?? !!locationService;
   const zoom = isMultipleMarkersMap
     ? zoomLevelForMaps.multipleMarkers
     : zoomLevelForMaps.singleMarker;
```
donni106, line 53: you wrote

> - added `zoomLevelForMaps` and `locationService` from settings to the `Map` component for
  configurable zoom and location features

but this was already implemented, right?!

src/components/map/Map.tsx
```diff
@@ -120,26 +158,61 @@ export const Map = ({
...
       </MapView>
+
       {isMaximizeButtonVisible && (
         <TouchableOpacity style={styles.maximizeMapButton} onPress={onMaximizeButtonPress}>
           <Icon.ExpandMap size={normalize(18)} />
         </TouchableOpacity>
       )}
+
+      {showPOIsFullScreenMap && (
+        <FlatList
```
donni106, line 184: list logics were not requested at this time. but it is a cool way, we can get back to in the future. i saved it in a new branch https://github.com/smart-village-solutions/smart-village-app-app/tree/feature/MQGB-34-flat-list-markers-overlay

src/components/map/Map.tsx
donni106: should be done in #921 now

src/screens/IndexScreen.js
donni106: done with #921

donni106: done with #921

donni106:
superseded by #921

PR #914, opened 2023-08-30
feat: add secret icons screen to servicetiles

d4v3d4vE:
- added `SecretIconsScreen.tsx` to `src/screens/` & `index.js`
- added `SecretIconsScreen.tsx` to `defaultStackConfig.tsx`, set routeName to `SecretIcons`
- added `SecretIcons` to `Navigation.ts`
- added `SecretIcons` screen to servicetiles
- adjusted icon names in `Icon.tsx`

SVA-1135

|  |  |  |
| ------------- | ------------- | ------------- |
| Simulator Screenshot - iPhone SE (3rd generation) - 2023-08-30 at 15 57 26 | Simulator Screenshot - iPhone SE (3rd generation) - 2023-08-30 at 15 57 32 | Simulator Screenshot - iPhone SE (3rd generation) - 2023-08-30 at 15 57 44 |


src/screens/SecretIconsScreen.tsx
```diff
@@ -0,0 +1,42 @@
...
+
+const iconNames = Object.keys(Icon);
+
+const IconItem = ({ iconName }: TIconItem) => {
+  const SelectedIcon = Icon[iconName];
+  return (
+    <View style={styles.iconField}>
+      <Text>
+        <SelectedIcon size={normalize(35)} color={colors.secondary} />
+      </Text>
```
ardasnturk, line 19: Why did you wrap this component with `Text`?

src/screens/SecretIconsScreen.tsx
```diff
@@ -0,0 +1,42 @@
...
+      numColumns={3}
+    />
+  );
+};
+
+const styles = StyleSheet.create({
+  iconField: {
+    margin: 10,
+    flex: 1,
+    alignItems: 'center'
```
ardasnturk, line 40: I recommend alphabetical sorting. 👍

src/screens/SecretIconsScreen.tsx
```diff
@@ -0,0 +1,42 @@
...
+
+type TIconItem = {
+  iconName: string;
+};
+
+const iconNames = Object.keys(Icon);
+
+const IconItem = ({ iconName }: TIconItem) => {
+  const SelectedIcon = Icon[iconName];
+  return (
```
ardasnturk, line 15: I recommend leaving a line space between SelectedIcon and return section.

src/screens/SecretIconsScreen.tsx
ardasnturk: In my tests, when I removed the Text component, the icons were displayed without any problems.

ardasnturk (changes requested):
Overall, it looks good. I think it will be better after we make small improvements.

src/screens/SecretIconsScreen.tsx
d4v3d4vE: You're right! I got a error before that seems to be resolved now.

src/screens/SecretIconsScreen.tsx
d4v3d4vE: Adjusted it!

src/screens/SecretIconsScreen.tsx
d4v3d4vE: Adjusted it!

ardasnturk (approved):
I think it's better now, but there's still one thing I'd like to add. Please include a link to the commit with the changes you made after answering the open questions, it will be easier to test. 👍

d4v3d4vE:
> I think it's better now, but there's still one thing I'd like to add. Please include a link to the commit with the changes you made after answering the open questions, it will be easier to test. 👍 I also changed the milestone as I think we can use this feature in version 3.2.7.

Just to make sure, do you mean to add the commit link to the code comments like this [commit](https://github.com/smart-village-solutions/smart-village-app-app/pull/914/commits/30d2172ce7e35cd02bc00a0fc2c5a2695724b1de)?

src/screens/SecretIconsScreen.tsx
donni106, line 1: i suggest to move this screen in a new folder, as it is a special screen for testing or documentation purposes. it has not the same priority as the other screens.
also "secret" in the name is a bit "hard" i would say. "secret" is used differntly in our project, as we have secrets for credentials.
what about `src/screens/doc/IconsScreen`? or `src/screens/about/IconsScreen`? or `src/screens/help/IconsScreen`. 
i can think of growing screens in that folder in the future.

src/types/Navigation.ts
```diff
@@ -48,6 +48,7 @@ export enum ScreenName {
   OParlOverview = 'OParlOverview',
   OParlPersons = 'OParlPersons',
   OParlSearch = 'OParlSearch',
+  SecretIcons = 'SecretIcons',
```
donni106, line 51: let's use a prefix for Icons depending on the path structure we chose, `DocIcons`/`DocumentationIcons`, `AboutIcons` or `HelpIcons`?

src/screens/SecretIconsScreen.tsx
```diff
@@ -0,0 +1,41 @@
...
+      numColumns={3}
+    />
+  );
+};
+
+const styles = StyleSheet.create({
+  iconField: {
+    alignItems: 'center',
+    flex: 1,
+    margin: 10
```
donni106, line 39: can we `normalize` the number as usual?

src/screens/SecretIconsScreen.tsx
```diff
@@ -0,0 +1,41 @@
...
+      data={iconNames}
+      renderItem={({ item }) => <IconItem iconName={item} />}
+      keyExtractor={(item) => item}
+      numColumns={3}
+    />
+  );
+};
+
+const styles = StyleSheet.create({
+  iconField: {
```
donni106, line 36: i am wondering about `field` here. i think we often use `wrapper` or `container`. could you think of using `iconWrapper` or `iconContainer` here for consistency?

src/screens/SecretIconsScreen.tsx
```diff
@@ -0,0 +1,41 @@
...
+import { colors, Icon, normalize } from '../config';
+
+type TIconItem = {
+  iconName: string;
+};
+
+const iconNames = Object.keys(Icon);
+
+const IconItem = ({ iconName }: TIconItem) => {
+  const SelectedIcon = Icon[iconName];
```
donni106, line 14: TypeScript complains here:

image

Please take some time to describe the situation for yourself and try to find the solution.

src/screens/SecretIconsScreen.tsx
```diff
@@ -0,0 +1,41 @@
...
+      <SelectedIcon size={normalize(35)} color={colors.secondary} />
+      <RegularText smallest>{iconName}</RegularText>
+    </View>
+  );
+};
+
+export const SecretIconsScreen = () => {
+  return (
+    <FlatList
+      data={iconNames}
```
donni106, line 27: as `iconNames` is only used once and there is no computation or anything, you can throw `Object.keys(Icon)` in here directly.

donni106: and we need to filter out `NamedIcon` i guess:

image

src/screens/SecretIconsScreen.tsx
```diff
@@ -0,0 +1,41 @@
...
+      renderItem={({ item }) => <IconItem iconName={item} />}
+      keyExtractor={(item) => item}
```
donni106, line 29: what is the `item`, that is used as a key here? is this following best practices?

image

can we keep it or can you think of something else?

donni106 (changes requested):

src/screens/SecretIconsScreen.tsx
d4v3d4vE: sound good, adjusted it to your first proposal [29186de](https://github.com/smart-village-solutions/smart-village-app-app/pull/914/commits/29186de714d420a15bf9284a4e17cbc553609203)

src/types/Navigation.ts
d4v3d4vE: adjusted the prefix matching with the new path structure [29186de](https://github.com/smart-village-solutions/smart-village-app-app/pull/914/commits/29186de714d420a15bf9284a4e17cbc553609203)

src/screens/SecretIconsScreen.tsx
d4v3d4vE: yes, added normalize [29186de](https://github.com/smart-village-solutions/smart-village-app-app/pull/914/commits/29186de714d420a15bf9284a4e17cbc553609203)

src/screens/SecretIconsScreen.tsx
d4v3d4vE: you're right, changed it to iconContainer [29186de](https://github.com/smart-village-solutions/smart-village-app-app/pull/914/commits/29186de714d420a15bf9284a4e17cbc553609203)

src/screens/SecretIconsScreen.tsx
d4v3d4vE: should be solved now in the new commit [29186de](https://github.com/smart-village-solutions/smart-village-app-app/pull/914/commits/29186de714d420a15bf9284a4e17cbc553609203)

src/screens/SecretIconsScreen.tsx
d4v3d4vE: filtered it out in the new commit [29186de](https://github.com/smart-village-solutions/smart-village-app-app/pull/914/commits/29186de714d420a15bf9284a4e17cbc553609203)

src/screens/SecretIconsScreen.tsx
d4v3d4vE: adjusted it to keyExtractor like its common practice in our codebase [29186de](https://github.com/smart-village-solutions/smart-village-app-app/pull/914/commits/29186de714d420a15bf9284a4e17cbc553609203)

src/screens/doc/DocIconsScreen.tsx
```diff
@@ -0,0 +1,39 @@
...
+      <SelectedIcon size={normalize(35)} color={colors.secondary} />
+      <RegularText smallest>{iconName}</RegularText>
+    </View>
+  );
+};
+
+export const DocIconsScreen = () => {
+  return (
+    <FlatList
+      data={Object.keys(Icon).filter((name) => name !== "NamedIcon")}
```
donni106, line 25: image

src/types/Navigation.ts
donni106: DocIcons let me think of doctors somehow 👨‍⚕️ 👩‍⚕️ but all right, let's keep it!

src/screens/SecretIconsScreen.tsx
donni106: well done, how did you find out about the solution?

src/screens/doc/DocIconsScreen.tsx
```diff
@@ -0,0 +1,39 @@
...
+type TIconItem = {
+  iconName: keyof typeof Icon;
+};
+
+const IconItem = ({ iconName }: TIconItem) => {
+  const SelectedIcon = Icon[iconName];
+
+  return (
+    <View style={styles.iconContainer}>
+      <SelectedIcon size={normalize(35)} color={colors.secondary} />
```
donni106, line 16: what about `<SelectedIcon />` to have it as short as possible? there is no need for special adjustment, right? that way we have the icons 1:1 from the configs:

Simulator Screen Shot - iPhone 14 - 2023-09-15 at 16 20 03

donni106, line 16: there is an ts problem here. did you try to resolve it?

d4v3d4vE: I figured some icons have a better visibility with a darker color on the white background, like e.g. EmptySection in your screenshot above. 
Tell me what you think about this idea, here is a comparison screenshot to yours above:

src/screens/doc/DocIconsScreen.tsx
d4v3d4vE: Adjusted it!

src/screens/SecretIconsScreen.tsx
d4v3d4vE: I think I now got it right, before using the code you wrote above the the items in the flatlist hat no unique key.  `keyExtractor={(index) => index.toString()}` thats my new approach and this way all items should have a unique key since the index of each iconname in the array should be unique.

donni106: And how come, that you are converting the index to a string?
Can you read https://reactnative.dev/docs/flatlist#keyextractor again please, what's the same as in my screenshot earlier.

> ...The default extractor checks item.key, then item.id, and then falls back to using the index, like React does.

This way i see no benefit in passing an index manually.

Another point is the problem with using index, which should be avoided. The following article points it out as well: https://dev.to/shiv1998/why-not-to-use-index-as-key-in-react-lists-practical-example-3e66#:~:text=Always%20prefer%20using%20a%20unique,of%20sorting%2C%20filtering%20might%20happen.

src/screens/doc/DocIconsScreen.tsx
donni106: Where? I cannot see it.

donni106: ah i see, please do not comment on local development, as we cannot follow it.

src/screens/SecretIconsScreen.tsx
d4v3d4vE: using `keyExtractor={(index) => index.id}` or `keyExtractor={(index) => index.key}` i get the warning below and the vs code error on both, therefore I chose index.toString() even tho it might not be the best solution
image
image

donni106: do you know what the passed param is? please log the variable you want to use. is it an object? does it have `id` or `key` prop? what do you think `index.toString()` results to?

src/screens/doc/DocIconsScreen.tsx
d4v3d4vE: yes i tried to but have not found a solution yet.

donni106: ok, then please add a ts lint ignore comment for that

d4v3d4vE: managed to solve it, added  name={'link'} to SelectedIcon

d4v3d4vE: @donni106

donni106: it's not about best visibility in first place, but we want to see icons. to understand when looking on them, i suggest not to modify them. when modifying for documentation we cannot draw conclusions to the rest of the code.
color: mostly shown as primary in the app -> should be primary here
size: mostly shown in default size configured in the app -> should be the same here as much as possible
the empty sections icon is special in size, yes. we do not need a properly visible size in the documentation, as it is the priority to see the icon and to know which icons are available.

src/screens/SecretIconsScreen.tsx
d4v3d4vE: should be fixed now in https://github.com/smart-village-solutions/smart-village-app-app/pull/914/commits/74026a36793ec9e05d851202baaf819b672fe17c

src/screens/doc/DocIconsScreen.tsx
```diff
@@ -0,0 +1,40 @@
...
+      numColumns={3}
+    />
+  );
+};
+
+const styles = StyleSheet.create({
+  iconContainer: {
+    alignItems: 'center',
+    flex: 1,
+    margin: normalize(9)
```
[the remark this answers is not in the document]
donni106: ok, let's keep 9. but there will be devices, where 9 will also not fit all the possible name cases. a breaking line is not a problem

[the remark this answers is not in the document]
d4v3d4vE: okay

src/screens/SecretIconsScreen.tsx
donni106: did you answer my questions for yourself? please do it in text here next time, so i can follow your thoughts and way for finding solutions. did you check our code for other `keyExtractor`s first? the latest code seems very familiar to me.

src/screens/doc/DocIconsScreen.tsx
```diff
@@ -0,0 +1,40 @@
...
+  iconName: keyof typeof Icon;
+};
+
+const IconItem = ({ iconName }: TIconItem) => {
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  const SelectedIcon: any = Icon[iconName];
+
+  return (
+    <View style={styles.iconContainer}>
+      <SelectedIcon />
```
donni106, line 17: how come that you have added the ts lint ignore comment for line 13 and not for 17? i thought line 17 was making the issue?
why is `: any` added in line 13? i think it wasn't there before?

src/screens/SecretIconsScreen.tsx
d4v3d4vE: in the process I forgot about the question, sorry for not answering it. I've checked your code before and in a call with Arda we've gone through the code together again to do the right changes since I had a mental block after quite some time.

src/screens/doc/DocIconsScreen.tsx
d4v3d4vE: with the type `: any` the ts error for `SelectedIcon` was solved but a eslint ts error occured with for `any`, therefore the ts lint ignore comment for line 13, for the type `any`

donni106: so there is an `any` added that needs to be ignored to fix the initial ts issue? why not ignoring the initial complain directly?

d4v3d4vE: cause it was a ts issue not a es lint issue, I haven't found a way and neither did Arda to ignore a ts issue. Therefore we "fixed" it with giving it a type and ignored the type issue of `any` which was now a es lint issue that could be disabled via the eslint-disable comment

donni106: ok but i understand any type issue as TS as well?
we already have ts error comments in our code, please have a look for `// @ts-expect-error`s, which would be a better case here instead if working around.

d4v3d4vE: thank you for the tip, updated the code accordingly.
https://github.com/smart-village-solutions/smart-village-app-app/pull/914/commits/281ea8084181c28f9fec438f40df3491f44e3057

src/screens/doc/DocIconsScreen.tsx
```diff
@@ -0,0 +1,39 @@
...
+import { RegularText } from '../../components';
+import { Icon, normalize } from '../../config';
+
+type TIconItem = {
+  iconName: keyof typeof Icon;
+};
+
+const IconItem = ({ iconName }: TIconItem) => {
+  const SelectedIcon = Icon[iconName];
+  return (
```
donni106, line 13: empty line please between const and return for better readability flow

donni106 (approved):

ardasnturk (approved):

PR #915, opened 2023-08-31
feat: add dynamic intro content for points of interest list

ardasnturk:
- added `poiListIntro` to the destructured `sections` object to fetch intro text for points of interest
- introduced `htmlContentName` and `htmlContent` to fetch and display HTML content conditionally for points of interest
- modified `ListComponent` to include the new HTML content as a header when displaying points of interest

to be able to try this feature, please edit the `sections` section in `globalSettings` as follows.

```
"poiListIntro": {
  "Bildung": "noticeboardMainContent"
},
```

the key in the object of `poiListIntro` must belong to a category. In this way, if you go to the correct poi category screen, the html content here will be automatically displayed above the poi list. then you should create the htmlContent that we named here on the main server.

SBB-110

## Screenshots:

|bildung category with html|burgen category without html|
|--|--|
Simulator Screenshot - iPhone 14 - 2023-08-31 at 12 52 42 | image

### DIFF
--- a/src/screens/IndexScreen.js
+++ b/src/screens/IndexScreen.js
@@ -15,6 +15,7 @@ import {
   CategoryList,
   DropdownHeader,
   EmptyMessage,
+  HtmlView,
   IndexFilterWrapperAndList,
   ListComponent,
   LoadingContainer,
@@ -37,6 +38,7 @@ import {
   useOpenWebScreen,
   usePermanentFilter,
   usePosition,
+  useStaticContent,
   useTrackScreenViewAsync,
   useVolunteerData
 } from '../hooks';
@@ -105,15 +107,13 @@ export const IndexScreen = ({ navigation, route }) => {
     eventLocations: showEventLocationsFilter = false
   } = filter;
   const { events: showVolunteerEvents = false } = hdvt;
-  const {
-    calendarToggle = false,
-    showFilterByOpeningTimes = true
-  } = settings;
+  const { calendarToggle = false, showFilterByOpeningTimes = true } = settings;
   const {
     categoryListIntroText = texts.categoryList.intro,
     categoryListFooter,
     categoryTitles,
-    eventListIntro
+    eventListIntro,
+    poiListIntro
   } = sections;
   const [queryVariables, setQueryVariables] = useState(route.params?.queryVariables ?? {});
   const [showCalendar, setShowCalendar] = useState(false);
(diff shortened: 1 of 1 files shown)


src/screens/IndexScreen.js
```diff
@@ -145,6 +148,14 @@ export const IndexScreen = ({ navigation, route }) => {
...
       [QUERY_TYPES.NEWS_ITEMS]: showNewsFilter
     }[query];
+  const htmlContentName =
+    (query === QUERY_TYPES.POINTS_OF_INTEREST && poiListIntro[queryVariables.category]) || '';
+
+  const { data: htmlContent } = useStaticContent({
+    name: htmlContentName,
+    type: 'html',
+    refreshTimeKey: `${query}-${queryVariables.category}`
+  });
```
d4v3d4vE, line 158: could you explain what this part of the code does since I'm not very familiar with queries yet?

d4v3d4vE (approved):
Feature is working and looks good

donni106:
This has no need for 3.2.7, it is no critical feature which needs to be integrated as soon as possible, so we will add this in 3.3.0!

src/screens/IndexScreen.js
```diff
@@ -145,6 +148,14 @@ export const IndexScreen = ({ navigation, route }) => {
       [QUERY_TYPES.EVENT_RECORDS]: showEventsFilter,
       [QUERY_TYPES.NEWS_ITEMS]: showNewsFilter
     }[query];
+  const htmlContentName =
+    (query === QUERY_TYPES.POINTS_OF_INTEREST && poiListIntro?.[queryVariables.category]) || '';
```
donni106, line 152: we do not need a fallback here. if we do not have the POI query and a `poiListIntro`, `htmlContentName` should be undefined

src/screens/IndexScreen.js
```diff
@@ -145,6 +148,14 @@ export const IndexScreen = ({ navigation, route }) => {
...
+
+  const { data: htmlContent } = useStaticContent({
+    name: htmlContentName,
+    type: 'html',
+    refreshTimeKey: `${query}-${queryVariables.category}`
+  });
```
donni106, line 158: this is the index screen, which is very much too complex already. with this, everytime accessing the index screen the hook is called and performs something. we need to avoid such cases with high priority! please add a `skip` here, if `htmlContentName` is not defined to nut run the hook and perform anything.

src/screens/IndexScreen.js
```diff
@@ -528,6 +539,13 @@ export const IndexScreen = ({ navigation, route }) => {
                           <RegularText>{categoryListIntroText}</RegularText>
                         </Wrapper>
                       )}
+                      {query === QUERY_TYPES.POINTS_OF_INTEREST &&
+                        !!htmlContentName &&
+                        !!htmlContent && (
+                          <Wrapper>
+                            <HtmlView html={htmlContent} />
+                          </Wrapper>
+                        )}
```
donni106, line 548: these are too much checks. please only regard the part what is rendered. we want to render `htmlContent`, so `htmlContent` needs to be defined. the other checks do not matter here. in the render part we try to avoid as much logics as possible. at this point all logics have to be done, to result in a defined or undefined `htmlContent`

donni106 (changes requested):

src/screens/IndexScreen.js
ardasnturk: I fixed it.
https://github.com/smart-village-solutions/smart-village-app-app/commit/a30ae59e770597a7b078e8ba7e44919e29c990c6

src/screens/IndexScreen.js
ardasnturk: Yes, you are right. We used it for the previous feature. I'm adding it here.
https://github.com/smart-village-solutions/smart-village-app-app/commit/a30ae59e770597a7b078e8ba7e44919e29c990c6

donni106 (approved):
smart, fresh and easily readable now!

PR #916, opened 2023-08-31
feat: add link to share feature

ardasnturk:
- added `url` data returned from `mergeWebUrls` to be able to add links to share messages if there are any
- exported from `InfoCard` to use the `mergeWebUrls` helper

SBB-115

## Screenshots:

|share without link|share with one link|share with multiple link|
|--|--|--|
IMG_4486 | IMG_4485 | IMG_4484

### DIFF
--- a/src/components/infoCard/InfoCard.tsx
+++ b/src/components/infoCard/InfoCard.tsx
@@ -4,6 +4,7 @@ import { StyleSheet, View } from 'react-native';
 import { Icon as RNEIcon } from 'react-native-elements';
 
 import { colors, consts, normalize } from '../../config';
+import { mergeWebUrls } from '../../helpers';
 import { Address, Contact, OpeningHour, WebUrl } from '../../types';
 import { RegularText } from '../Text';
 import { InfoBox } from '../Wrapper';
@@ -26,31 +27,6 @@ type Props = WebUrlProps & {
   name?: string;
   openingHours?: OpeningHour[];
   openWebScreen: (link: string) => void;
-};
-
-export const mergeWebUrls = ({ webUrls, contact, contacts }: WebUrlProps) => {
-  const mergedWebUrls = webUrls ? [...webUrls] : [];
-
-  if (contact?.www) {
-    mergedWebUrls.unshift({ url: contact.www });
-  }
-
-  // merge a `contact`s `webUrls` to `webUrls`
-  if (contact?.webUrls?.length) {
-    mergedWebUrls.push(...contact.webUrls);
-  }
-
-  // iterate all `contacts` and merge every `contact`s `webUrls` to `webUrls`
-  contacts?.forEach((contact) => {
-    if (contact?.www) {
-      mergedWebUrls.push({ url: contact.www });
-    }
-    if (contact?.webUrls?.length) {
-      mergedWebUrls.push(...contact.webUrls);
-    }
-  });
-
-  return mergedWebUrls;
 };
 
 /* TODO: add a logic to display info category and url that fit the screen even if long text
--- a/src/components/infoCard/InfoCard.tsx
+++ b/src/components/infoCard/InfoCard.tsx
@@ -28,7 +28,7 @@ type Props = WebUrlProps & {
   openWebScreen: (link: string) => void;
 };
 
-const mergeWebUrls = ({ webUrls, contact, contacts }: WebUrlProps) => {
+export const mergeWebUrls = ({ webUrls, contact, contacts }: WebUrlProps) => {
   const mergedWebUrls = webUrls ? [...webUrls] : [];
 
   if (contact?.www) {
--- a/src/helpers/shareHelper.js
+++ b/src/helpers/shareHelper.js
@@ -1,6 +1,7 @@
 import { Share } from 'react-native';
 
 import appJson from '../../app.json';
+import { mergeWebUrls } from '../components';
 import { QUERY_TYPES } from '../queries';
 
 import { momentFormat } from './momentHelper';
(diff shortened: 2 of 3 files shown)


d4v3d4vE (approved):

src/components/infoCard/InfoCard.tsx
```diff
@@ -28,7 +28,7 @@ type Props = WebUrlProps & {
   openWebScreen: (link: string) => void;
 };
 
-const mergeWebUrls = ({ webUrls, contact, contacts }: WebUrlProps) => {
+export const mergeWebUrls = ({ webUrls, contact, contacts }: WebUrlProps) => {
```
donni106, line 31: please move this in our helpers area, as this method is not used just for info cards anymore. the component should stay clean for its purpose.

src/helpers/shareHelper.js
```diff
@@ -38,20 +39,26 @@ export const openShare = async ({ message, title, url }) => {
 };
 
 export const shareMessage = (data, query) => {
+  const urls = mergeWebUrls({
+    webUrls: data.webUrls,
+    contact: data.contact,
+    contacts: data.contacts
+  })?.map(({ url }) => `\n${url}`);
```
donni106, line 46: nice for finding and using `mergeWebUrls` here!
i am missing the "Link: " part in front of every link, that was requested in the ticket comments.

src/helpers/shareHelper.js
```diff
@@ -38,20 +39,26 @@ export const openShare = async ({ message, title, url }) => {
...
         return `${momentFormat(data.publishedAt)} | ${data.dataProvider?.name}: ${
           data.contentBlocks?.[0]?.title
-        }`;
+        }${`\n${urls}`}`;
       case QUERY_TYPES.POINT_OF_INTEREST:
-        return `${data.category?.name}: ${data.name}`;
+        return `${data.category?.name}: ${data.name}${`\n${urls}`}`;
       case QUERY_TYPES.TOUR:
-        return `${data.category?.name}: ${data.name}`;
+        return `${data.category?.name}: ${data.name}${`\n${urls}`}`;
```
donni106, line 61: please do not add line breaks if there are no urls. this causes an additional empty line in the share message.

donni106 (changes requested):

src/components/infoCard/InfoCard.tsx
ardasnturk: Yeah, you're right. You can find this fix in this commit.https://github.com/smart-village-solutions/smart-village-app-app/pull/916/commits/081c27876bf3cbd07a86a776fd5aff7f5aef0b28

src/helpers/shareHelper.js
ardasnturk: I thought there might be more than one Link. At the moment I have added the `Link` text at the beginning of each link. If you want, I can make the `Links` write in case there is more than one link with more control.
https://github.com/smart-village-solutions/smart-village-app-app/pull/916/commits/081c27876bf3cbd07a86a776fd5aff7f5aef0b28

src/helpers/shareHelper.js
ardasnturk: Yes, that's right. I overlooked that. I've edited it now.
https://github.com/smart-village-solutions/smart-village-app-app/pull/916/commits/081c27876bf3cbd07a86a776fd5aff7f5aef0b28

src/components/infoCard/InfoCard.tsx
```diff
@@ -4,7 +4,9 @@ import { StyleSheet, View } from 'react-native';
 import { Icon as RNEIcon } from 'react-native-elements';
 
 import { colors, consts, normalize } from '../../config';
-import { Address, Contact, OpeningHour, WebUrl } from '../../types';
+import { mergeWebUrls } from '../../helpers/mergeWebUrls';
+import { Address, OpeningHour } from '../../types';
+import { WebUrlProps } from '../../types/WebUrlProps';
```
donni106, line 9: what is the difference between them to be imported differently?

src/helpers/shareHelper.js
```diff
@@ -3,6 +3,7 @@ import { Share } from 'react-native';
 import appJson from '../../app.json';
 import { QUERY_TYPES } from '../queries';
 
+import { mergeWebUrls } from './mergeWebUrls';
```
donni106, line 6: the structure of the helper file names is mostly ...Helper.js, please apply here as well.
and what about placing it in `linkHelper` instead of creating a new file?

src/helpers/shareHelper.js
donni106: did you ensure, that rendering will not result in "...false" if there are no urls?

src/helpers/shareHelper.js
donni106: I do not think that something with "Links" is necessary as "Link: ..." was requested. Each link can have its "Link: ".

src/helpers/shareHelper.js
ardasnturk: we have 3 different situations. One is the case of more than one link, the second is the case of only one link and the third is the case of no link at all. All three cases look like in the screenshot.
image

donni106: In my eyes the last one hast two empty lines?

src/components/infoCard/InfoCard.tsx
ardasnturk: This is because I didn't add `WebUrlProps` to index.js. But now I've moved `mergeWebUrls` to `linkHelper.js` and added `WebUrlProps` back to `InfoCard.tsx` since I only use it in one place. 

https://github.com/smart-village-solutions/smart-village-app-app/commit/0de11ba779b6755711897c4ca4939cada2092108

src/helpers/shareHelper.js
ardasnturk: Yes, it makes sense. I've made that change.
https://github.com/smart-village-solutions/smart-village-app-app/commit/0de11ba779b6755711897c4ca4939cada2092108

src/helpers/shareHelper.js
ardasnturk: Yes, you're right. That's not supposed to happen. I'm trying to solve this problem. 👍

ardasnturk: I fixed the problem and now the 3 different options that can occur when the share button is pressed are as in the screenshot.
image

https://github.com/smart-village-solutions/smart-village-app-app/pull/916/commits/a65c1e0cfa0acb3da5fa37faa921a6e5f2cad86a

donni106: There is a } at the end of every title now

ardasnturk: Yes, that's right. I fixed that problem.

image

https://github.com/smart-village-solutions/smart-village-app-app/commit/ab5b6bec9d66d8aab4869a5b39603472b288d1b1

src/helpers/shareHelper.js
```diff
@@ -38,26 +39,43 @@ export const openShare = async ({ message, title, url }) => {
...
+    webUrls: data.webUrls,
+    contact: data.contact,
+    contacts: data.contacts
+  })?.map(({ url }) => `Link: ${url}`);
+
+  const urlSection = urls.length ? urls.join('\n') : '';
+  const spacer = urlSection ? '\n' : '';
+
   const buildMessage = (query) => {
+    let baseMessage = '';
```
donni106, line 52: what about just `message`? the method `buildMessage` builds a `message`? what was your thought with `base`?

ardasnturk: yes, it can be. there's nothing extra I want to say with `base`

---
yes it makes sense. this way we don't need the default section in the switch

here's the commit I've applied the two changes to
https://github.com/smart-village-solutions/smart-village-app-app/commit/8306f3e91b6144eb3771fd40c8b84cdde753ad68

donni106: but empty line still missing :(

image

ardasnturk: yes, you are right. i fixed it right away. 👍

https://github.com/smart-village-solutions/smart-village-app-app/commit/229db627ae2309d82340bb5c902e21cf30d1fa90

PR #901, opened 2023-08-18
chore(expo): update expo version to 49

d4v3d4vE:
- ran `yarn add expo@^49.0.0`
- ran `npx expo install --fix` to upgrade all dependencies to match SDK 49
- added `"expo-localization"` to `app.json` as part of the expo 49 update
- https://blog.expo.dev/expo-sdk-49-c6d398cdf740

- updated `react-navigation` from 5.X to 6.X manually as part of the expo 49 update -> changed the version numbers of `react-navigation` from 5.X to 6.X inside the `package.json`, deleted the `node_modules` folder, deleted the `yarn.lock` file and ran the `yarn` command to ensure a clean install
- adjusted the headerStyle in `screenOptions.tsx` to match the `react-navigation` 6.X to the before used design with `react-navigation` 5.X
- changed in `MainTabNavigator.tsx` the `tabBarOptions` to `screenOptions` and the options names to match with the `react-navigation` 6.X update
- added `headerShown: false` to `screenOptions` in `MainTabNavigator.tsx` to ensure the before used design with `react-navigation` 5.X and not show a tab screen header with the new `react-navigation` 6.X update
- added `screenOptions` to the `Drawer.Navigator` in `DrawerNavigator.tsx` with `drawerPosition: "right"` and `headerShown: false` to ensure the before used design with `react-navigation` 5.X and not show a drawer screen header with the new `react-navigation` 6.X update

- changed the navigation from drawer to tabnavigation

- to setup the new expo version execute the following commands in the project folder: 

1. run `yarn prebuild`

2. run `npx pod-install`

3. run `yarn ios -d`

SVA-1081

| TabNav before | TabNav now | Drawer before | Drawer now |
| ------------- | ------------- | ------------- | ------------- |
| Simulator Screenshot - iPhone 14 Pro - 2023-08-18 at 18 12 36 | Simulator Screenshot - iPhone 14 Pro - 2023-08-18 at 18 14 03 | Simulator Screenshot - iPhone 14 Pro - 2023-08-18 at 18 10 30 | Simulator Screenshot - iPhone 14 Pro - 2023-08-18 at 18 13 03 |
|  |  | Simulator Screenshot - iPhone 14 Pro - 2023-08-18 at 18 10 34 | Simulator Screenshot - iPhone 14 Pro - 2023-08-18 at 18 13 07 |

PR #917, opened 2023-09-04
refactor:  additional changes for react navigation 6.x

d4v3d4vE:
- moved styles to `drawerStyle` inside `screenOptions` of the Drawer in `DrawerNavigator.tsx`
  and refactored with orientation provider
- removed `drawerType` of the Drawer in `DrawerNavigator.tsx`,
  since the default for iOS is now set to `slide` and to `front` for other plattforms:
  https://reactnavigation.org/docs/drawer-navigator/#drawertype
- linted `AppStackNavigator.tsx`

SVA-1130

### DIFF
--- a/src/navigation/DrawerNavigator.tsx
+++ b/src/navigation/DrawerNavigator.tsx
@@ -1,14 +1,14 @@
 import { createDrawerNavigator } from '@react-navigation/drawer';
 import _isEmpty from 'lodash/isEmpty';
 import _reduce from 'lodash/reduce';
-import React, { useEffect, useState } from 'react';
-import { StyleSheet } from 'react-native';
+import React, { useContext, useEffect, useState } from 'react';
 
 import { LoadingSpinner } from '../components/LoadingSpinner';
 import { device, texts } from '../config';
 import { defaultStackConfig } from '../config/navigation';
 import { useStaticContent } from '../hooks';
 import { ScreenName } from '../types';
+import { OrientationContext } from '../OrientationProvider';
 
 import { getStackNavigator } from './AppStackNavigator';
 import { CustomDrawerContentComponent } from './CustomDrawerContentComponent';
@@ -83,6 +83,7 @@ const Drawer = createDrawerNavigator();
 
 export const DrawerNavigator = () => {
   const { loading, drawerRoutes } = useDrawerRoutes();
+  const { orientation } = useContext(OrientationContext);
 
   if (loading) return <LoadingSpinner loading />;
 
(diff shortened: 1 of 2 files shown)


ardasnturk (changes requested):
I thought this PR was only for the `react-navigation` package, but the code changes also include expo 49. I don't think this is the way it should be. Please update this PR only for `react-navigation`.

src/navigation/DrawerNavigator.tsx
```diff
@@ -96,20 +95,15 @@ export const DrawerNavigator = () => {
         />
       )}
       screenOptions={{
-        drawerPosition:"right",
-        headerShown: false
+        drawerPosition: 'right',
+        headerShown: false,
+        drawerStyle: {
+          width: device.width > device.height ? device.height * 0.8 : device.width * 0.8
+        }
```
ardasnturk, line 102: Why did you incorporate these styles into the component? Is there a special reason? Because I can't tell from the commit message.

d4v3d4vE: These additional styles were needed to keep the before used design with the new updates

d4v3d4vE: https://github.com/smart-village-solutions/smart-village-app-app/pull/901 in the drawer now bottom picture in this pr its visible that the drawer width is not as wide as it was before

src/navigation/DrawerNavigator.tsx
```diff
@@ -96,20 +95,15 @@ export const DrawerNavigator = () => {
...
       }}
-      drawerStyle={styles.drawerContainer}
-      drawerType={device.platform === 'ios' ? 'slide' : 'front'}
```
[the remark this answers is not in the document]
d4v3d4vE: yes its not needed anymore since it defaults to slide on iOS and front on other platforms.
https://reactnavigation.org/docs/drawer-navigator/#drawertype

src/navigation/DrawerNavigator.tsx
ardasnturk: I think the question I want to ask is not understood. Could this style be used as follows?

```
screenOptions={{
  drawerPosition: 'right',
  headerShown: false,
  drawerStyle: styles.drawerStyle
}}
...
  drawerStyle: {
    width: device.width > device.height ? device.height * 0.8 : device.width * 0.8
  }
});
```

d4v3d4vE: yes it can be used but I figured that the other solution is better since we don't use any other styles in this file, therefore we can save ourselves 4 additional lines of code which are practically not needed

donni106: i understand and think we can keep it in the render part.
but, i will sort props alphabetically and try to user our orientation provider to determine landscape/portrait instead of width height comparison here

donni106: 9d52c48

donni106 (approved):

ardasnturk (approved):

PR #920, opened 2023-09-13
feat: add a text list option with a large image as the first item

ardasnturk:
- added `TextListItem` to `listHooks` to show other items of the list
- added `textListItemWithBigCard` prop to `TextListItem` because it is a customised `listItem`
- added `rightImage` and `rightIcon` props for adding image to the right of `TextListItem`
- changed text types to new text types depending on the `textListItemWithBigCard` prop
- separated the `subtitle` into two different sections, above and below the title, to match the style
- added index control inside the switch section in `listHooks` to make the image of the item in the first index of the list bigger
- changed the fonts in `BigCardListItem` different from `CardList` component
- changed the image size in `BigCardListItem` different from `CardList` component
- separated `subtitle` into two different types of text to be placed above and below the `title`
- added `Divider` component for separation with text list
- added `TEXT_WITH_BIG_CARD_LIST` to consts to show and select the new list option in the `SettingsScreen`
- added `textWithBigCardList` to `texts.js` to show the name of the new list
- Added `TEXT_WITH_BIG_CARD_LIST` option to `listHooks` to render the list after the list is selected
- added some fonts required to be used in the component

MQGB-63

---

## Screenshots:

|SettingScreen|HomeScreen|IndexScreen|IndexScreen (poi)|IndexScreen (without image)|
|--|--|--|--|--|
Simulator Screenshot - iPhone 14 Pro Max - 2023-09-13 at 09 44 25 | image|Simulator Screenshot - iPhone 14 Pro Max - 2023-09-29 at 15 50 41|Simulator Screenshot - iPhone 14 Pro Max - 2023-09-29 at 15 49 32|Simulator Screenshot - iPhone 14 Pro Max - 2023-09-29 at 15 50 15


src/components/BigCardListItem.js
```diff
@@ -0,0 +1,111 @@
...
+import { Card, Divider } from 'react-native-elements';
+
+import { colors, consts, normalize } from '../config';
+import { imageHeight, imageWidth } from '../helpers';
+
+import { Image } from './Image';
+import { CategoryText, HeadlineText, RegularText } from './Text';
+import { Touchable } from './Touchable';
+
+export const BigCardListItem = memo(({ navigation, horizontal, item }) => {
```
d4v3d4vE, line 13: what does memo to in this part of the code?

ardasnturk: the `memo` in the code is a higher-order component from React that helps with performance optimization. It essentially memorizes the rendered output of the wrapped component and prevents unnecessary re-renders if the props haven't changed.

d4v3d4vE: thats cool, thank you for your explanation 😄

d4v3d4vE (approved):
work great and looks good!

src/components/Text.js
```diff
@@ -166,3 +166,34 @@ export const BoldText = styled(RegularText)`
...
+    props.small &&
+    css`
+      font-size: ${normalize(16)};
+      line-height: ${normalize(21)};
+    `};
+`;
+
+export const CategoryText = styled(RegularText)`
+  font-weight: 700;
+  font-family: bold;
```
donni106, line 191: if these should be bold, they can be built on top of `BoldText` instead of `RegularText`. and i guess 700 is not needed in the master repo. isn't it a release specific thing? the `BoldText` did not have any font-weight.

src/components/Text.js
```diff
@@ -166,3 +166,34 @@ export const BoldText = styled(RegularText)`
...
+export const HeadlineText = styled(RegularText)`
+  font-weight: 700;
+  font-family: bold;
+
+  ${(props) =>
+    props.medium &&
+    css`
+      font-size: ${normalize(18)};
+      line-height: ${normalize(23)};
+    `};
```
donni106, line 179: i think this is not needed. there are two cases i can see. medium or small. medium is something we do not have in our code base yet.
i suggest using the style of medium as general for `HeadlineText` and small as variation, where necessary.

src/config/consts.js
```diff
@@ -155,7 +155,8 @@ export const consts = {
   LIST_TYPES: {
     CARD_LIST: 'cardList',
     IMAGE_TEXT_LIST: 'imageTextList',
-    TEXT_LIST: 'textList'
+    TEXT_LIST: 'textList',
+    TEXT_WITH_BIG_CARD_LIST: 'textWithBigCardList'
```
donni106, line 159: following the scheme of `XYZ_LIST` here i suggest `CARD_TEXT_LIST`. we do not need a "big" because our card is our card, there is no big or small

src/components/TextListItem.tsx
```diff
@@ -34,87 +35,129 @@ type Props = {
   leftImage?: boolean | undefined;
   navigation: StackNavigationProp<Record<string, any>>;
   noSubtitle?: boolean | undefined;
+  rightImage?: boolean | undefined;
+  textListItemWithBigCard?: boolean | undefined;
```
donni106, line 39: we are inside `TextListItem`, i guess we do not have the name duplicated again in the prop. the other props are also fine without that prefix. and the card is the size of a card, there is no need for "big". i suggest `withCard` here as boolean prop to indicate, that the text list has a card or not.

src/components/BigCardListItem.js
donni106, line 13: following the same naming across everything related to the text list item with a card, i suggest `CardTextListItem`. so this naming runs through everywhere.

src/hooks/listHooks.js
```diff
@@ -64,6 +65,31 @@ export const useRenderItem = (query, navigation, options = {}) => {
...
+                ...item,
+                bottomDivider: true
+              }}
+              {...{
+                navigation,
+                noSubtitle: options.noSubtitle,
+                rightImage: true,
+                textListItemWithBigCard: true
+              }}
+            />
```
donni106, line 87: what are you thinking about this alternative from readability aspects?

```
<TextListItem
  item={{
    ...item,
    bottomDivider: true
  }}
  navigation={navigation}
  noSubtitle={options.noSubtitle}
  rightImage
  withCard
/>
```

src/hooks/listHooks.js
```diff
@@ -64,6 +65,31 @@ export const useRenderItem = (query, navigation, options = {}) => {
...
+          return (
+            <TextListItem
+              item={{
+                ...item,
+                bottomDivider: true
+              }}
+              {...{
+                navigation,
+                noSubtitle: options.noSubtitle,
+                rightImage: true,
```
donni106, line 84: why is the image on the right here and for text list on the left? do we want that difference in the master branch? its both a text list with an image, right? i see no reason to have an image once on the left and once on the right.

src/hooks/listHooks.js
```diff
@@ -64,6 +65,31 @@ export const useRenderItem = (query, navigation, options = {}) => {
...
+        if (index === 0) {
+          return (
+            <BigCardListItem navigation={navigation} horizontal={options.horizontal} item={item} />
+          );
+        } else {
+          return (
+            <TextListItem
+              item={{
+                ...item,
+                bottomDivider: true
```
donni106, line 79: is this ok to always `true`? compared to the existing text list item there is a logic for the bottom divider

src/components/BigCardListItem.js
```diff
@@ -0,0 +1,111 @@
...
+          {!!title && (
+            <HeadlineText medium style={{ marginTop: normalize(4) }}>
+              {horizontal ? (title.length > 60 ? title.substring(0, 60) + '...' : title) : title}
+            </HeadlineText>
+          )}
+          {!!subtitle && (
+            <RegularText small style={{ marginTop: normalize(6) }}>
+              {subtitle.split(' |')[0]}
+            </RegularText>
+          )}
```
donni106, line 45: i cannot understand wh we split the subtitle for the new list item type? this differs much from the existing ones, right? i guess this is something release related for a specific customer? so i suggest having the subtitles equal across the list items, but adjusted in that release

src/components/BigCardListItem.js
donni106, line 1: asking in general, couldn't we use our existing card list item in front of a text list instead of creating a new thing mixing both?

src/components/Text.js
ardasnturk: Yes, you're absolutely right. I also think CategoryText is too much for the main repository and I will remove it.

src/components/Text.js
ardasnturk: Yes. I realised this better in our update last week and I'm going to fix it.

src/config/consts.js
ardasnturk: Yeah, that makes more sense.

src/components/TextListItem.tsx
ardasnturk: Yes, we must avoid unnecessary name repetition.

src/components/BigCardListItem.js
ardasnturk: I think I'll remove this component. I think the `CardListItem` component will be enough.

src/hooks/listHooks.js
ardasnturk: Since the other props are also in this way, I used it in this way for code integrity, but I will fix it to increase readability.

src/components/BigCardListItem.js
ardasnturk: You're absolutely right. I'm going to return it to its normal format and reorganise it for release.

src/components/BigCardListItem.js
ardasnturk: Exactly! We don't need any extra components for this listItem. At the moment we can do the same with `CardListItem`. only style update will be required when it will be released.

ardasnturk: https://github.com/smart-village-solutions/smart-village-app-app/pull/920/commits/3a454a280b079b4a8d26d3f0edad7db5a24e7b56

src/components/BigCardListItem.js
ardasnturk: https://github.com/smart-village-solutions/smart-village-app-app/pull/920/commits/3a454a280b079b4a8d26d3f0edad7db5a24e7b56

src/components/Text.js
ardasnturk: https://github.com/smart-village-solutions/smart-village-app-app/pull/920/commits/24147097e638fbb4faab803fab473fc7bbe10503

src/components/Text.js
ardasnturk: https://github.com/smart-village-solutions/smart-village-app-app/pull/920/commits/24147097e638fbb4faab803fab473fc7bbe10503

src/config/consts.js
ardasnturk: https://github.com/smart-village-solutions/smart-village-app-app/pull/920/commits/ce9321414c47f430d32648b51629f9dd8d6c74ab

src/components/BigCardListItem.js
ardasnturk: https://github.com/smart-village-solutions/smart-village-app-app/pull/920/commits/3a454a280b079b4a8d26d3f0edad7db5a24e7b56

https://github.com/smart-village-solutions/smart-village-app-app/pull/920/commits/24147097e638fbb4faab803fab473fc7bbe10503

src/hooks/listHooks.js
ardasnturk: https://github.com/smart-village-solutions/smart-village-app-app/commit/74e7eb7f2b529145ccbb1f97a902aa1981ad5341

src/hooks/listHooks.js
ardasnturk: https://github.com/smart-village-solutions/smart-village-app-app/commit/74e7eb7f2b529145ccbb1f97a902aa1981ad5341

src/hooks/listHooks.js
ardasnturk: https://github.com/smart-village-solutions/smart-village-app-app/commit/74e7eb7f2b529145ccbb1f97a902aa1981ad5341

src/components/TextListItem.tsx
ardasnturk: https://github.com/smart-village-solutions/smart-village-app-app/commit/74e7eb7f2b529145ccbb1f97a902aa1981ad5341

donni106 (approved):

PR #922, opened 2023-09-18
feat(web urls): merge system dependent

donni106:
- added filtering on platform to exclude web urls for the foreign system

|before|after iOS|after Android|
|---|---|---|
|image|image|image|

MQGB-57


ardasnturk (approved):
Cool 🎉

PR #925, opened 2023-09-21
fix: add undefined control to url

ardasnturk:
- added control to prevent the app from crashing if the data is undefined

MQGB-57

Simulator Screenshot - iPhone 14 Pro Max - 2023-09-21 at 16 02 23


donni106 (approved):
thanks!

PR #926, opened 2023-09-25
refactor(map icons): use from `Icon` to be flexible

donni106:
- changed to render a map icon from our icon library instead of fixed location svg icons
- created `MapIcon` with default `'location'` as icon to be able to remove all occurrences across the code base where the location icon was set
- added `anchor` in `MapMarker` that was missing until now, to properly position icons on map with pin point at the bottom of location icons
  - removed icon anchor const exports as we do not need them individually
- added `OwnLocation` in our `Icon` library to be accessible with `Icon[_upperFirst(iconName)]`
- fixed type issues with `Icon[_upperFirst(iconName) as keyof typeof Icon]`

MQGB-34

### DIFF
--- a/src/components/map/Map.tsx
+++ b/src/components/map/Map.tsx
@@ -1,7 +1,7 @@
-import React, { useContext, useRef, useState } from 'react';
+import _upperFirst from 'lodash/upperFirst';
+import React, { useContext, useRef } from 'react';
 import { StyleProp, StyleSheet, TouchableOpacity, View, ViewStyle } from 'react-native';
 import MapView, { LatLng, MAP_TYPES, Marker, Polyline, Region, UrlTile } from 'react-native-maps';
-import { SvgXml } from 'react-native-svg';
 
 import { colors, device, Icon, normalize } from '../../config';
 import { imageHeight, imageWidth } from '../../helpers';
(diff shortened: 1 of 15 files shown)


src/components/map/Map.tsx
```diff
@@ -25,6 +25,18 @@ type Props = {
...
+  iconName = 'location'
+}: {
+  iconColor?: string;
+  iconName?: string;
+}) => {
+  const MarkerIcon = Icon[_upperFirst(iconName) as keyof typeof Icon];
+
+  return <MarkerIcon color={iconColor} size={MARKER_ICON_SIZE} />;
+};
+
```
ardasnturk, line 39: If the category does not have an icon, i.e. `iconName === null`, such an error occurs. Can we add an extra control here?


image
image

ardasnturk (changes requested):
Why is the name of the Branch in German?

donni106:
> Why is the name of the Branch in German?

Branch name is German because created directly from Jira, which is based on the ticket name.

src/components/map/Map.tsx
donni106: From which screen or component were you navigating? `'location'`. Where did you place the `console.log`? `iconName` should be string or undefined, to fall back to

ardasnturk: route = POI -> Bildung -> Kartenansicht 

I placed the log in the `MapIcon`.
image

donni106: I updated it to `iconName={marker.iconName ? marker.iconName : undefined}`

ardasnturk (approved):

d4v3d4vE (approved):

PR #927, opened 2023-09-25
fix: solve detail screen rerender problem

ardasnturk:
- fixed the `date` value with state, which causes the `DetailScreen` to re-render when the location information is active
- renamed `date` to `today` for clarity
- created today value in `TimeTables` with moment to ensure code integrity

MQGB-54

### DIFF
--- a/src/components/screens/TimeTables.tsx
+++ b/src/components/screens/TimeTables.tsx
@@ -1,4 +1,5 @@
 import _upperFirst from 'lodash/upperFirst';
+import moment from 'moment';
 import React, { Fragment } from 'react';
 import { StyleSheet, View } from 'react-native';
 import { Divider, ListItem } from 'react-native-elements';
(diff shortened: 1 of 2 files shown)


src/components/screens/TimeTables.tsx
```diff
@@ -28,15 +29,15 @@ export const TimeTables = ({
...
+  const today: string = moment().format('ddd, DD.MM.YYY');
 
   return (
     <>
       <SectionHeader title={texts.pointOfInterest.departureTimes} />
 
       <Wrapper style={styles.noPadding}>
         <BoldText small>
-          {texts.pointOfInterest.today} {momentFormat(today, 'ddd, DD.MM.YYYY')}
+          {texts.pointOfInterest.today} {today}
```
donni106, line 40: so much better to read in both places, `moment` on top and nothing but `today` down here.

donni106 (approved):

d4v3d4vE (approved):

PR #928, opened 2023-09-25
feat:(map icons): render two icons in some cases

donni106:
- if we have a given icon name, we want to render it individually on top of our location pin
  - except for `ownLocation` which already is a kind of location pin
- added styles to render the icon on top of the location pin smaller and positioned in the center top of the location pin with different color to be visible

|heartFilled|check|home|
|---|---|---|
|image|image|image|

MQGB-34

### DIFF
--- a/src/components/map/Map.tsx
+++ b/src/components/map/Map.tsx
@@ -27,14 +27,16 @@ const MARKER_ICON_SIZE = normalize(40);
 
 const MapIcon = ({
   iconColor,
-  iconName = 'location'
+  iconName = 'location',
+  iconSize = MARKER_ICON_SIZE
 }: {
   iconColor?: string;
   iconName?: string;
+  iconSize?: number;
 }) => {
   const MarkerIcon = Icon[_upperFirst(iconName) as keyof typeof Icon];
 
-  return <MarkerIcon color={iconColor} size={MARKER_ICON_SIZE} />;
+  return <MarkerIcon color={iconColor} size={iconSize} />;
 };
 
 export const Map = ({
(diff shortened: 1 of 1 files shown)


src/components/map/Map.tsx
```diff
@@ -126,21 +128,45 @@ export const Map = ({
...
+                      iconSize={MARKER_ICON_SIZE / 3.25}
+                    />
+                  </View>
+                </>
+              ) : (
+                <MapIcon
+                  iconColor={
+                    selectedMarker && marker.id === selectedMarker ? colors.accent : undefined
+                  }
+                  iconName={marker.iconName}
```
ardasnturk, line 164: We check `marker.iconName` above and check whether there is a value. This is the section where `marker.iconName` has no value. Should we add the `iconName` prop here?

ardasnturk (changes requested):

src/components/map/Map.tsx
donni106: here? there is `iconName={marker.iconName}`, which can be undefined to use `'location'`.

src/components/map/Map.tsx
```diff
@@ -27,14 +27,16 @@ const MARKER_ICON_SIZE = normalize(40);
...
 }: {
   iconColor?: string;
   iconName?: string;
+  iconSize?: number;
 }) => {
   const MarkerIcon = Icon[_upperFirst(iconName) as keyof typeof Icon];
 
-  return <MarkerIcon color={iconColor} size={MARKER_ICON_SIZE} />;
+  return <MarkerIcon color={iconColor} size={iconSize} />;
 };
```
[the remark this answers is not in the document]
donni106: jo, this will be done there, thanks

src/components/map/Map.tsx
donni106: should be done with https://github.com/smart-village-solutions/smart-village-app-app/pull/926#discussion_r1335790276

ardasnturk (approved):

d4v3d4vE (approved):

PR #938, opened 2023-10-02
feat(disturber): hide the close button on the main server

ardasnturk:
- added new `showButtonToClose` value to `homeDisturber` `staticContent` to show or not show the button to close the disturber

SVA-1161

### DIFF
--- a/src/components/Disturber.tsx
+++ b/src/components/Disturber.tsx
@@ -33,6 +33,7 @@ interface DataItem {
   description: string;
   id: number;
   picture?: { aspectRatio?: { HEIGHT: number; WIDTH: number }; uri: string };
+  showButtonToClose?: boolean;
   title: string;
 }
 
@@ -54,7 +55,7 @@ export const Disturber = ({ navigation, publicJsonFile }: Props) => {
   );
 
   const setDisturberComplete = () => {
-    setIsVisible((prev) => !prev);
+    setIsVisible(false);
     !!closestItem && addToStore(publicJsonFile, closestItem.id.toString());
   };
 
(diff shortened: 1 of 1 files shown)


d4v3d4vE (approved):

src/components/Disturber.tsx
```diff
@@ -39,6 +40,9 @@ interface DataItem {
 // eslint-disable-next-line complexity
 export const Disturber = ({ navigation, publicJsonFile }: Props) => {
   const [isVisible, setIsVisible] = useState(false);
+  const { globalSettings } = useContext(SettingsContext);
+  const { sections = {} } = globalSettings;
+  const { showDisturberHiddenButton = true } = sections;
```
donni106, line 45: please use a more readable name, like you would speak it, for example: `showButtonToHideDisturber`
quick translation to German with deepl:
image

donni106: maybe `close` would be better than `hide` as the other code uses the word close.

ardasnturk: Yes, I will fix it. You're right, I'll use `close` to make the code complete.

src/components/Disturber.tsx
```diff
@@ -83,9 +84,11 @@ export const Disturber = ({ navigation, publicJsonFile }: Props) => {
       <View style={[styles.containerRadius, { backgroundColor: closestItem.backgroundColor }]}>
         {!!closestItem && (
           <>
-            <TouchableOpacity onPress={setDisturberComplete} style={styles.closeButton}>
-              <Icon.Close color={colors.darkText} size={normalize(16)} />
-            </TouchableOpacity>
+            {closestItem.showButtonToClose && (
```
donni106, line 87: can be undefined here, right? i suggest a `true` fallback of the value, which can be overridden with `false` per configs.

ardasnturk: Yes, it can be It can be `undefined`. I set `true` by default. 

https://github.com/smart-village-solutions/smart-village-app-app/commit/6d16fa29134aa86fad0d96aab7a2c727e19a1592

src/components/Disturber.tsx
```diff
@@ -78,14 +79,18 @@ export const Disturber = ({ navigation, publicJsonFile }: Props) => {
 
   if (!isVisible || !closestItem) return null;
 
+  const { showButtonToClose = true } = closestItem;
```
donni106, line 82: with having destruction of `closestItem` here i suggest to add all of its used variables, like `backgroundColor` and so on

ardasnturk: https://github.com/smart-village-solutions/smart-village-app-app/commit/b9c4ceae1a0b0270e3498e20e1a157cb3b8d0d7c

src/components/Disturber.tsx
```diff
@@ -78,69 +79,73 @@ export const Disturber = ({ navigation, publicJsonFile }: Props) => {
...
-                      title={closestItem.button.title}
-                      onPress={() => {
-                        navigation.navigate(
-                          closestItem.button.navigationTo,
-                          closestItem.button.params
-                        );
-                      }}
-                    />
-                  )}
+                {!!button && !!button.title && !!button.navigationTo && !!button.params && (
```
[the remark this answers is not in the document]
ardasnturk: Yeah, I did that. 
https://github.com/smart-village-solutions/smart-village-app-app/commit/4fe814b2f46edaf0465d8089c7037ad45a15947f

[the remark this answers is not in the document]
donni106: whats the difference? there ist still that complex condition. i asked for a variable to handle that in the render context: `{showButton && (...)}`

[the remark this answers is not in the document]
ardasnturk: I didn't realise that's what you meant. I fixed it.
https://github.com/smart-village-solutions/smart-village-app-app/commit/fc9ca371b2a233022d2e4c99e7cb40454380a2e6

src/components/Disturber.tsx
```diff
@@ -78,69 +79,76 @@ export const Disturber = ({ navigation, publicJsonFile }: Props) => {
...
+  const {
+    backgroundColor,
+    button,
+    description,
+    headline,
+    picture,
+    showButtonToClose = true,
+    title
+  } = closestItem;
+  const { title: buttonTitle, navigationTo, params } = button;
```
donni106, line 91: if button is undefined, this could crash i guess.
can you check that please. because if button is not an object, destruction fails.

src/components/Disturber.tsx
```diff
@@ -78,69 +79,76 @@ export const Disturber = ({ navigation, publicJsonFile }: Props) => {
...
+
+  const showButton = !!button && !!buttonTitle && !!navigationTo && !!params;
```
donni106, line 93: if button was successfully destructed, it wouldn't be needed to check again.
i suggest removing the destruction and check every key of `button.` here and use `button.` as well in the render part?

PR #939, opened 2023-10-04
fix(volunteer): the problem of not creating events

ardasnturk:
- added `filterGroupDropDownData` to `useEffect` to ensure dropdown data is refreshed

HDVT-93

### DIFF
--- a/src/components/volunteer/VolunteerFormCalendar.tsx
+++ b/src/components/volunteer/VolunteerFormCalendar.tsx
@@ -165,7 +165,7 @@ export const VolunteerFormCalendar = ({
       setTimeout(() => {
         filterGroupDropDownData();
       }, 300);
-  }, [isRefetchingGroupsMy]);
+  }, [isRefetchingGroupsMy, filterGroupDropDownData]);
 
   const isFocused = useIsFocused();
 


src/components/volunteer/VolunteerFormCalendar.tsx
```diff
@@ -167,6 +167,10 @@ export const VolunteerFormCalendar = ({
       }, 300);
   }, [isRefetchingGroupsMy]);
 
+  useEffect(() => {
+    filterGroupDropDownData();
+  }, [filterGroupDropDownData]);
```
donni106, line 172: why is this necessary now?
and wouldn't adding the dependency to the previous effect have the same result?

ardasnturk: The `VolunteerFormCalendar` screen is logically similar to the `VolunteerFormConversation` screen. Both pages have a dropdown list. Through this list, users can select their groups. However, the dropdown list showing the groups on the `VolunteerFormCalendar` page cannot be loaded when the application is first opened. To solve this problem, I looked at the `VolunteerFormConversation` screen and added the `useEffect` to make the code similar to each other. This way the user can see the dropdown list when they go to the event creation page.

`VolunteerFormConversation`:
image

donni106: and what about the question about the previous effect? they are very similar. can we remove the existing one after adding the new one?

ardasnturk: Are you talking about this `useEffect`?
```
  useEffect(() => {
    !isRefetchingGroupsMy &&
      setTimeout(() => {
        filterGroupDropDownData();
      }, 300);
  }, [isRefetchingGroupsMy]);
```

I still think that this effect should remain. I think that if we delete it, the groups may not be refreshed properly.

donni106: both effects are running the same method, right? so at least we should combine them?

ardasnturk: I think we can update the code in this way.
```
  useEffect(() => {
    !isRefetchingGroupsMy &&
      setTimeout(() => {
        filterGroupDropDownData();
      }, 300);
  }, [isRefetchingGroupsMy, filterGroupDropDownData]);
```

donni106: i agree

donni106 (approved):

d4v3d4vE (approved):

src/components/volunteer/VolunteerFormCalendar.tsx
d4v3d4vE: thanks for this explanation/deconstruction you guy did here :)

PR #942, opened 2023-10-09
feature: add filter and pagination to OpeningTimesCard and fix event record detail screen 

ardasnturk:
- added filter to `openingHours` array to show date information in the appointment list from today
- added pagination so that if the appointment dates in the `OpeningTimesCard` component are more than 15, they do not take up much space on the screen
- added `filteredOpeningHoursFromToday` to filter the appointments from today and to show the pagination button according to the number of filtered elements and the data is loaded according to this filter
- added the number of data loaded with pagination as standard as 15
- added `moreData` state for pagination
- added slice method to `filteredOpeningHoursFromToday` array for pagination

SVA-1176

Also changed the `fetchPolicy` in `DetailScreen` to `cache-and-network` if the query is an event, to avoid loading data from cache when going from calendar view to DetailScreen.

### filter to show appointments from today:
|before|before (pullToRefresh)|after|
|--|--|--|
Simulator Screenshot - iPhone 14 Pro Max - 2023-10-09 at 11 54 38|Simulator Screenshot - iPhone 14 Pro Max - 2023-10-09 at 11 54 42|Simulator Screenshot - iPhone 14 Pro Max - 2023-10-09 at 11 54 02

### pagination:
|before|after|
|--|--|
Simulator Screenshot - iPhone 14 Pro Max - 2023-10-09 at 11 54 46|Simulator Screenshot - iPhone 14 Pro Max - 2023-10-09 at 11 54 05


src/components/screens/OpeningTimesCard.js
```diff
@@ -23,70 +26,94 @@ const DateBox = styled(TimeBox)`
...
-);
+export const OpeningTimesCard = ({ openingHours }) => {
+  const [moreData, setMoreData] = useState(1);
+
+  const loadMoreItems = () => {
+    setMoreData((prev) => prev + 1);
+  };
+
+  const filteredOpeningHoursFromToday =
+    openingHours?.filter((item) => {
```
donni106, line 37: i have changed to `openingHours?.filter` and added a `|| []` fallback, to be able to always do array operations on `filteredOpeningHoursFromToday`

donni106 (approved):

PR #945, opened 2023-10-12
fix: the bug of not navigating back from map view

ardasnturk:
- added `switchBetweenListAndMap` control to `useLayoutEffect` to fix the problem of not being able to go back in navigation when `top-filter` is on in POI list

SVA-1181

### DIFF
--- a/src/components/screens/Overviews.js
+++ b/src/components/screens/Overviews.js
@@ -236,7 +236,8 @@ export const Overviews = ({ navigation, route }) => {
     if (
       query === QUERY_TYPES.POINTS_OF_INTEREST &&
       showMap &&
-      initialFilter === FILTER_TYPES.LIST
+      initialFilter === FILTER_TYPES.LIST &&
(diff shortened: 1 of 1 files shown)


donni106 (approved):

src/components/screens/Overviews.js
```diff
@@ -236,7 +236,8 @@ export const Overviews = ({ navigation, route }) => {
     if (
       query === QUERY_TYPES.POINTS_OF_INTEREST &&
       showMap &&
-      initialFilter === FILTER_TYPES.LIST
+      initialFilter === FILTER_TYPES.LIST &&
+      switchBetweenListAndMap == SWITCH_BETWEEN_LIST_AND_MAP.BOTTOM_FLOATING_BUTTON
```
d4v3d4vE, line 240: what does the initialFilter and switchBetweenListAndMap do and where is this visible?

d4v3d4vE (approved):

src/components/screens/Overviews.js
donni106: what they "do" is the wrong question. both are variables holding a value at  the time of the condition. if these values match the ones in the condition, the condition gets true, otherwise false.

PR #947, opened 2023-10-12
fix(navigation): configs without `useContext` and refactorings

donni106:
the problem was, that changing settings context triggered a rerendering of the navigation as the config file relied on it. the refactoring made passing the navigation type on initialising the application without the need of reading from the settings context.

also there are some refactorings to unify the code and understand things better.

SVA-1130

### DIFF
--- a/src/components/map/LocationOverview.tsx
+++ b/src/components/map/LocationOverview.tsx
@@ -60,7 +60,7 @@ const mapToMapMarkers = (pointsOfInterest: any): MapMarker[] | undefined => {
 export const LocationOverview = ({ filterByOpeningTimes, navigation, queryVariables }: Props) => {
   const { isConnected, isMainserverUp } = useContext(NetworkContext);
   const { globalSettings } = useContext(SettingsContext);
-  const { navigation: navigationType = 'drawer' } = globalSettings;
+  const { navigation: navigationType } = globalSettings;
   const [selectedPointOfInterest, setSelectedPointOfInterest] = useState<string>();
   const fetchPolicy = graphqlFetchPolicy({ isConnected, isMainserverUp, refreshTime: undefined });
 
@@ -134,10 +134,7 @@ export const LocationOverview = ({ filterByOpeningTimes, navigation, queryVariab
       {selectedPointOfInterest && !detailsLoading && (
         <Wrapper
           small
-          style={[
-            styles.listItemContainer,
-            stylesWithProps({ navigation: navigationType }).position
-          ]}
+          style={[styles.listItemContainer, stylesWithProps({ navigationType }).position]}
         >
           <TextListItem
             item={{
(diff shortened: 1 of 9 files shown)


src/components/map/LocationOverview.tsx
```diff
@@ -180,10 +177,10 @@ const styles = StyleSheet.create({
...
 /* this works properly, we do not want that warning */
-const stylesWithProps = ({ navigation }: { navigation: string }) => {
+const stylesWithProps = ({ navigationType }: { navigationType: string }) => {
   return StyleSheet.create({
     position: {
-      bottom: navigation === 'tab' ? '4%' : '8%'
+      bottom: navigationType === 'tab' ? '4%' : '8%'
     }
   });
 };
```
ardasnturk, line 186: I have tried different navigation types. If no navigation value is specified in `globalSettins`, the following screenshot appears. Can you please check if this is the case with you? It is surprising that the preview looks like this even though we set the `navigationType` to `tab` by default.

|navigation = "tab"|navigation = undefined|navigation = "drawer"|
|--|--|--|
Simulator Screenshot - iPhone 14 Pro Max - 2023-10-13 at 09 41 45|Simulator Screenshot - iPhone 14 Pro Max - 2023-10-13 at 09 41 10|Simulator Screenshot - iPhone 14 Pro Max - 2023-10-13 at 09 42 08

src/components/settings/LocationSettings.js
```diff
@@ -53,8 +53,7 @@ export const LocationSettings = () => {
       Location.getForegroundPermissionsAsync().then((response) => {
         // if the system permission is granted, we can simply enable the sorting
         if (response.status === Location.PermissionStatus.GRANTED) {
-          const newSettings = { locationService: true };
-          setAndSyncLocationSettings(newSettings);
+          setAndSyncLocationSettings({ locationService: true });
```
d4v3d4vE, line 56: I'm curious what is the benefit of setting locationService to true here instead of getting it from newSettings?

src/navigation/Navigator.tsx
```diff
@@ -3,24 +3,28 @@ import React from 'react';
 import { StatusBar } from 'react-native';
 
 import { colors } from '../config';
-import { linkingConfig, navigatorConfig } from '../config/navigation';
```
d4v3d4vE, line 6: isn't the linkingConfig still needed? (Line 20)

src/index.js
```diff
@@ -42,10 +42,11 @@ const MainAppWithApolloProvider = () => {
...
+    hdvt: {},
+    navigation: 'tab',
     sections: {},
     settings: {},
-    widgets: [],
-    hdvt: {}
+    widgets: []
   });
   const [initialListTypesSettings, setInitialListTypesSettings] = useState({});
   const [initialLocationSettings, setInitialLocationSettings] = useState({});
```
d4v3d4vE, line 52: why is it needed to write all this again if its already present in the SettingsProvider.js file

d4v3d4vE (commented):
this are nice and interesting changes

src/components/settings/LocationSettings.js
donni106: reducing the variables. it was only used for the next line without any benefit.

src/navigation/Navigator.tsx
donni106: it is still there, from `const { navigatorConfig, linkingConfig } = navigationConfig(navigationType);`

src/index.js
donni106: it's a state variable here not related to the settings provider. at the end this will be passed to the settings provider. there should be a way to remove one or the other, but it doesn't hurt so far.

src/components/map/LocationOverview.tsx
donni106: 🤔 i will look into it

ardasnturk (approved):

PR #951, opened 2023-10-17
fix: navigation type error for volunteer

ardasnturk:
- added `globalSettings` where necessary for `navigationType` since the navigation structure of the application is now set over the server

SVA-1130

### DIFF
--- a/src/components/volunteer/VolunteerUser.tsx
+++ b/src/components/volunteer/VolunteerUser.tsx
@@ -1,9 +1,9 @@
 import { StackScreenProps } from '@react-navigation/stack';
-import React, { useCallback, useEffect, useState } from 'react';
+import React, { useCallback, useContext, useEffect, useState } from 'react';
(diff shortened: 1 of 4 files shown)


src/components/volunteer/VolunteerUser.tsx
```diff
@@ -28,6 +28,9 @@ export const VolunteerUser = ({
 }: { data: any } & StackScreenProps<any>) => {
   const [isMe, setIsMe] = useState<boolean>();
 
+  const { globalSettings } = useContext(SettingsContext);
+  const { navigation: navigationType } = globalSettings;
```
donni106, line 32: please move these two lines to the top in this component as well. the existing use state line can follow as third line.

ardasnturk: https://github.com/smart-village-solutions/smart-village-app-app/commit/7dd1b103cb0a5f045b70859f5356a5190615f456

PR #957, opened 2023-10-20
feature(carousel): add button

ardasnturk:
- added button prop to `ImagesCarousel` and `ImagesCarouselItem` so that `ImageButton` component can be shown in `ImageCarousel`
- removed the `position: 'absolute'` style from the `ImageMessage` component so that `ImageMessage` and `ImageButton` components do not overlap
- wrapped the content in the `Image` with `View` and added styling to make the `ImageMessage` component appear below the `Carousel`
- added `ImageButton` component to show in `homeCarousel`
- alphabetised props

SUE-3

## Test:
To try the new carousel, you need to change the version in app.json to 60. According to version 60, `globalSettings` and `homeCarousel` have been added on the main server.

## Screenshots:

|before|after without button|with button|
|--|--|--|
Simulator Screenshot - iPhone 14 Pro Max - 2023-10-20 at 14 19 16|Simulator Screenshot - iPhone 14 Pro Max - 2023-10-20 at 14 18 48|Simulator Screenshot - iPhone 14 Pro Max - 2023-10-20 at 14 18 37

### DIFF
--- a/src/components/Image.js
+++ b/src/components/Image.js
@@ -9,6 +9,7 @@ import { imageHeight, imageWidth } from '../helpers';
 import { useInterval } from '../hooks/TimeHooks';
 import { SettingsContext } from '../SettingsProvider';
 
+import { ImageButton } from './ImageButton';
 import { ImageMessage } from './ImageMessage';
 import { ImageRights } from './ImageRights';
 
@@ -32,6 +33,7 @@ export const Image = ({
   aspectRatio,
   resizeMode,
   borderRadius,
+  button,
   refreshInterval,
   source: sourceProp
 }) => {
@@ -109,16 +111,23 @@ export const Image = ({
         resizeMode={resizeMode}
         borderRadius={borderRadius}
       >
-        {!!message && <ImageMessage message={message} />}
-        {!!globalSettings?.showImageRights && !!sourceProp?.copyright && (
-          <ImageRights imageRights={sourceProp.copyright} />
-        )}
+        <View style={styles.contentContainerStyle}>
+          {!!message && <ImageMessage message={message} />}
+          {!!button && <ImageButton button={button} />}
+          {!!globalSettings?.showImageRights && !!sourceProp?.copyright && (
+            <ImageRights imageRights={sourceProp.copyright} />
+          )}
+        </View>
       </RNEImage>
     </View>
   );
 };
 
 const styles = StyleSheet.create({
+  contentContainerStyle: {
+    height: '100%',
+    justifyContent: 'flex-end'
+  },
   placeholderStyle: {
     backgroundColor: colors.transparent,
     flex: 1
@@ -145,6 +154,7 @@ const stylesForImage = (aspectRatio) => {
 Image.propTypes = {
   source: PropTypes.oneOfType([PropTypes.object, PropTypes.number]).isRequired,
   message: PropTypes.string,
+  button: PropTypes.object,
   containerStyle: PropTypes.object,
   style: PropTypes.oneOfType([PropTypes.object, PropTypes.array]),
   PlaceholderContent: PropTypes.object,
--- a/src/components/Image.js
+++ b/src/components/Image.js
@@ -26,16 +26,16 @@ const addQueryParam = (url, param) => {
 const NO_IMAGE = { uri: 'NO_IMAGE' };
 
 export const Image = ({
-  message,
-  style,
-  containerStyle,
-  PlaceholderContent,
   aspectRatio,
-  resizeMode,
   borderRadius,
   button,
+  containerStyle,
+  message,
+  PlaceholderContent,
   refreshInterval,
-  source: sourceProp
+  resizeMode,
+  source: sourceProp,
+  style
 }) => {
   const [source, setSource] = useState(null);
   const { globalSettings } = useContext(SettingsContext);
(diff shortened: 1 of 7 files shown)


src/components/Image.js
```diff
@@ -109,16 +111,23 @@ export const Image = ({
         resizeMode={resizeMode}
         borderRadius={borderRadius}
       >
-        {!!message && <ImageMessage message={message} />}
-        {!!globalSettings?.showImageRights && !!sourceProp?.copyright && (
-          <ImageRights imageRights={sourceProp.copyright} />
-        )}
+        <View style={styles.contentContainerStyle}>
```
donni106, line 114: i think that this view should only be rendered if one of its children will be presented

src/components/Image.js
```diff
@@ -109,16 +111,23 @@ export const Image = ({
...
+        </View>
       </RNEImage>
     </View>
   );
 };
 
 const styles = StyleSheet.create({
+  contentContainerStyle: {
+    height: '100%',
+    justifyContent: 'flex-end'
```
donni106, line 129: does this work in any case of possible children components?
100% means full height of the image?

src/components/ImageButton.tsx
```diff
@@ -0,0 +1,41 @@
...
+  const navigation = useNavigation();
+
+  if (!params || !routeName) {
+    return null;
+  }
+
+  return (
+    <Wrapper style={styles.noPadding}>
+      <Button
+        icon={icon || <Icon.Plus color={colors.lightestText} />}
```
donni106, line 28: what if there shouldn't be an icon at all? i suggest not having a fallback. if one wants an icon, it should be configured.

src/components/ImageButton.tsx
```diff
@@ -0,0 +1,41 @@
...
+
+type TImageButton = {
+  icon?: React.ReactElement;
+  params?: any;
+  routeName: string;
+  title?: string;
+};
+
+export const ImageButton = ({ button }: { button: TImageButton }) => {
+  const { icon, params, routeName, title = texts.sue.homeCarousel.button } = button;
```
donni106, line 18: i suggest not having a fallback title. we save some code here an in texts.js.
the button text should be passed in any case.

donni106 (changes requested):

src/components/Image.js
ardasnturk: I don't know exactly what you mean.

src/components/Image.js
ardasnturk: This style works correctly for all children components. 
Yes, the height value is directly proportional to the image. Covering 100% of the image's height

src/components/ImageButton.tsx
ardasnturk: Hmm. I'll change that. At the moment we have an icon in all cases. I'll make it updatable from the server whether we want an icon or not. 👍

src/components/ImageButton.tsx
ardasnturk: Yes, that's right. I'll change that.

src/components/Image.js
donni106: this view is rendered always, no matter if there are children or not. this could be changed.

ardasnturk: I don't think there would be a problem with an empty view appearing. Do you want me to add such a control?
```
{!!message && !!button && !!globalSettings?.showImageRights && !!sourceProp?.copyright && (
          <View style={styles.contentContainerStyle}>
...
)}
```

donni106: "if you do not see it, its fine" is not a proper reason. think about not rendering stuff we do not want to be processed and present, even if it is not visible. yes, you need a condition to render the view or not. but not with `&&` because there are cases, where just one of those should be presented, right?!

ardasnturk: Then I suggest such a check.


```
`const showChildren = !!message || !!button || (!!globalSettings?.showImageRights && !!sourceProp?.copyright);`
...
        {showChildren && (
          <View style={styles.contentContainerStyle}>
            {!!message && <ImageMessage message={message} />}
            {!!button && <ImageButton button={button} />}
            {!!globalSettings?.showImageRights && !!sourceProp?.copyright && (
              <ImageRights imageRights={sourceProp.copyright} />
            )}
          </View>
        )}
```

src/components/ImageButton.tsx
ardasnturk: https://github.com/smart-village-solutions/smart-village-app-app/pull/957/commits/8a32da27e4cd615ba5eae900256f46644863e75d

src/components/ImageButton.tsx
ardasnturk: https://github.com/smart-village-solutions/smart-village-app-app/pull/957/commits/8a32da27e4cd615ba5eae900256f46644863e75d

src/components/Image.js
donni106: ok. and what about adding a `const showImageRights = !!globalSettings?.showImageRights && !!sourceProp?.copyright` to avoid duplication of that condition?

src/components/ImageButton.tsx
```diff
@@ -0,0 +1,55 @@
...
+  iconName?: keyof typeof Icon;
+  params?: any;
+  routeName: string;
+  style?: {
+    big?: boolean;
+    disabled?: boolean;
+    iconColor?: string;
+    iconPosition?: 'left' | 'right';
+    invert?: boolean;
+    noFullWidth?: boolean;
```
donni106, line 20: `notFullWidth`

ardasnturk: https://github.com/smart-village-solutions/smart-village-app-app/pull/957/commits/cfba5a511de9c2f40e02271e375cc1e489e44f25

src/components/Image.js
ardasnturk: https://github.com/smart-village-solutions/smart-village-app-app/pull/957/commits/02549cf49e5f6e6b2ddd17cb0c79e4c48f4937b9

src/components/Image.js
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable complexity */
```
donni106, line 1: we never put this at the very top or am i wrong? it should be placed right infront of the code, that forces the linting issue

src/components/Image.js
```diff
@@ -109,16 +115,23 @@ export const Image = ({
...
         borderRadius={borderRadius}
       >
-        {!!message && <ImageMessage message={message} />}
-        {!!globalSettings?.showImageRights && !!sourceProp?.copyright && (
-          <ImageRights imageRights={sourceProp.copyright} />
+        {showChildren && (
+          <View style={styles.contentContainerStyle}>
+            {!!message && <ImageMessage message={message} />}
+            {!!button && <ImageButton button={button} />}
+            {!!showImageRights && <ImageRights imageRights={sourceProp.copyright} />}
```
donni106, line 122: like in `const showChildren = !!message || !!button || showImageRights;` you do not need the !! for the boolean here

src/components/Image.js
ardasnturk: Yes. At the same time, I reactivated the function where it had ended.
https://github.com/smart-village-solutions/smart-village-app-app/pull/957/commits/4879aaf1097c8856e567a9ae1ffd90d32d17d8cc

src/components/Image.js
ardasnturk: https://github.com/smart-village-solutions/smart-village-app-app/pull/957/commits/4879aaf1097c8856e567a9ae1ffd90d32d17d8cc

donni106 (approved):

PR #965, opened 2023-11-02
feature(sue): add the report done screen

ardasnturk:
- created a report done component to show the user that the report has been successfully sent after the reporting is finished

SUE-7


src/screens/SUE/SueReportScreen.tsx
```diff
@@ -199,10 +192,14 @@ export const SueReportScreen = ({ navigation }: { navigation: any }) => {
...
 
+  if (isDone) {
+    return <SueReportDone navigation={navigation} />;
+  }
+
   return (
     <>
       <SueReportProgress progress={data} currentProgress={currentPage + 1} />
-      <ScrollView ref={outerScrollViewRef}>
+      <SafeAreaViewFlex>
```
donni106, line 202: i don't think you need this here

src/screens/SUE/SueReportScreen.tsx
```diff
@@ -211,7 +208,7 @@ export const SueReportScreen = ({ navigation }: { navigation: any }) => {
           showsHorizontalScrollIndicator={false}
         >
           {data?.map((item: TProgress, index: number) => (
-            <View key={index} style={styles.contentContainer}>
+            <ScrollView key={index} contentContainerStyle={styles.contentContainer}>
```
donni106, line 211: is already part of `feature/SUE-7-report-a-case`

PR #968, opened 2023-11-06
feat: adjustments for map and card list

donni106:
- we introduced `overtitle` already with MeinQuartier so we should use it for SUE as well


src/components/CardListItem.js
donni106, line 1: ⚠️ we need to think of adjusting the global setting for SUE after releasing this or to prepare a versioned globalSettings for the new version number

ardasnturk (approved):

d4v3d4vE (approved):

PR #972, opened 2023-11-22
feat(disturber): add multiple images to disturber

ardasnturk:
- added `ImagesCarousel` component instead of `Image` component to show when more than one image is added to disturber
- added the ability to add a link directly from the image to disturber
- changed the duration of the cycle of the disturber carousel on the server
- added `isDisturber` prop to `ImagesCarousel` to not show pause button in carousel

SVAK-20

## Screenshots:

|first image with link|second image without link|third image without link|
|--|--|--|
Simulator Screenshot - iPhone 14 Pro - 2023-11-22 at 21 02 09|Simulator Screenshot - iPhone 14 Pro - 2023-11-22 at 21 02 11|Simulator Screenshot - iPhone 14 Pro - 2023-11-22 at 21 02 14

### DIFF
--- a/src/components/Disturber.tsx
+++ b/src/components/Disturber.tsx
@@ -1,17 +1,15 @@
 import { DrawerNavigationProp } from '@react-navigation/drawer';
 import React, { useEffect, useState } from 'react';
-import { FlatList, ScrollView, StyleSheet, TouchableOpacity, View } from 'react-native';
+import { StyleSheet, TouchableOpacity, View } from 'react-native';
 
-import { Icon, colors, device, normalize } from '../config';
-import { addToStore, findClosestItem, imageWidth, isActive, readFromStore } from '../helpers';
+import { Icon, colors, normalize } from '../config';
+import { addToStore, findClosestItem, isActive, readFromStore } from '../helpers';
 import { useHomeRefresh, useStaticContent } from '../hooks';
 
 import { Button } from './Button';
-import { Image } from './Image';
+import { ImagesCarousel } from './ImagesCarousel';
 import { BoldText, RegularText } from './Text';
 import { Wrapper, WrapperHorizontal } from './Wrapper';
-import { ImagesCarousel } from './ImagesCarousel';
-import Carousel from 'react-native-snap-carousel';
 
 type Props = {
   navigation: DrawerNavigationProp<any>;
@@ -112,7 +110,7 @@ export const Disturber = ({ navigation, publicJsonFile }: Props) => {
               </TouchableOpacity>
             )}
 
-            {!!pictures?.length && (
+            {pictures?.length ? (
               <View style={styles.carouselContainer}>
                 <ImagesCarousel
                   aspectRatio={aspectRatio || { HEIGHT: 0.7, WIDTH: 1 }}
@@ -123,6 +121,8 @@ export const Disturber = ({ navigation, publicJsonFile }: Props) => {
                   refreshTimeKey={`publicJsonFile-${publicJsonFile}`}
                 />
               </View>
+            ) : (
+              <View style={styles.withoutImageMarginTop} />
             )}
 
             <Wrapper style={styles.noPaddingBottom}>
@@ -198,6 +198,9 @@ const styles = StyleSheet.create({
     fontWeight: '700',
     lineHeight: normalize(16),
     textTransform: 'uppercase'
+  },
+  withoutImageMarginTop: {
+    marginTop: normalize(21)
   },
   noPaddingBottom: {
     paddingBottom: 0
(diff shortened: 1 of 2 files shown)


src/components/ImagesCarousel.js
```diff
@@ -133,13 +140,14 @@ export const ImagesCarousel = ({ data, navigation, refreshTimeKey, aspectRatio }
...
           autoplayDelay={0}
-          autoplayInterval={4000}
+          autoplayInterval={autoplayInterval || 4000}
           containerCustomStyle={styles.center}
           onScrollIndexChanged={setCarouselImageIndex}
         />
       )}
 
       {showSliderPauseButton &&
+        !isDisturber &&
```
donni106, line 150: please move this additional disturber check up to the `show: showSliderPauseButton = false,` line. in the rendering part we should only read if the slider pause button should be shown or not without asking ourself anything about disturbers.

src/components/Disturber.tsx
```diff
@@ -1,15 +1,17 @@
...
+import { Icon, colors, device, normalize } from '../config';
+import { addToStore, findClosestItem, imageWidth, isActive, readFromStore } from '../helpers';
 import { useHomeRefresh, useStaticContent } from '../hooks';
 
 import { Button } from './Button';
 import { Image } from './Image';
 import { BoldText, RegularText } from './Text';
 import { Wrapper, WrapperHorizontal } from './Wrapper';
+import { ImagesCarousel } from './ImagesCarousel';
+import Carousel from 'react-native-snap-carousel';
```
donni106, line 14: organize imports (manually?) to get this up.
or...is this even used?

src/components/Disturber.tsx
```diff
@@ -102,13 +112,17 @@ export const Disturber = ({ navigation, publicJsonFile }: Props) => {
...
+              <View style={styles.carouselContainer}>
+                <ImagesCarousel
+                  aspectRatio={aspectRatio || { HEIGHT: 0.7, WIDTH: 1 }}
+                  autoplayInterval={autoplayInterval}
+                  data={pictures}
+                  isDisturber
+                  navigation={navigation}
+                  refreshTimeKey={`publicJsonFile-${publicJsonFile}`}
+                />
+              </View>
```
donni106, line 125: if there is no image we should consider equal spacing on top like on bottom

image

src/components/Disturber.tsx
```diff
@@ -80,11 +88,13 @@ export const Disturber = ({ navigation, publicJsonFile }: Props) => {
...
+    autoplayInterval,
     backgroundColor,
     button,
     description,
     headline,
-    picture,
+    pictures,
     showButtonToClose = true,
     title
   } = closestItem;
```
d4v3d4vE, line 100: is there a specific reason why its written like this instead of below, or dosen't it make any difference?

const closestItem = {
    aspectRatio,
    autoplayInterval,
    backgroundColor,
    button,
    description,
    headline,
    picture,
    pictures,
    showButtonToClose = true,
    title
  };

src/components/Disturber.tsx
d4v3d4vE: In addition if possible I would reconsider using the 'big' button in line 158, since it design wise dosent really fit in when the text of the button is this much bigger than the other text of the disturber

d4v3d4vE: just checked again, in relation to the button of the news below it's also a big difference

src/components/ImagesCarousel.js
```diff
@@ -38,6 +38,8 @@ export const ImagesCarousel = ({
   const [isPaused, setIsPaused] = useState(false);
   const [carouselImageIndex, setCarouselImageIndex] = useState(0);
 
+  const shouldShowPauseButton = showSliderPauseButton && !isDisturber;
```
donni106, line 41: you can do this in line 33 i guess instead of creating a new const here

src/components/Disturber.tsx
ardasnturk: I didn't make any changes for the button because it was supposed to be done like this as a design in the first place. Later, the button can be updated when releasing according to the request.

src/components/ImagesCarousel.js
ardasnturk: https://github.com/smart-village-solutions/smart-village-app-app/pull/972/commits/6e8f47ee62bb5c24c8e94350732e8531ab8ef64b

src/components/Disturber.tsx
ardasnturk: https://github.com/smart-village-solutions/smart-village-app-app/pull/972/commits/fb9727fe6e376ed39e7259b0d68dd15527dd1fbc

src/components/ImagesCarousel.js
ardasnturk: I don't think I can integrate it here. Because if `showSliderPauseButton` returns true from the server, `isDisturber` doesn't matter. Or am I wrong?

src/components/Disturber.tsx
ardasnturk: If we write the code with a method like you said, we create an object called closestItem in the code. Since we need the values in closestItem, the code must be in this way. Also in this way we can update the code over the server.

PR #977, opened 2023-12-06
feat(sue): map marker depending on status

donni106:
- implemented complex conditions to colorize map markers depending on given values and active state
- needed to `parseListItemsFromQuery` in `mapToMapMarkers` to have `iconName` and `status` set in the object to map further for necessary icon values

SUE-4

image


ardasnturk (approved):
Everything looks right and works fine for me.

PR #979, opened 2023-12-06
feat(sue): reverse geocoding

donni106:
- renamed `homeNumber` to `houseNumber` as it is more common
- removed initial map marker when there is no address selected yet

SUE-7

https://github.com/smart-village-solutions/smart-village-app-app/assets/1942953/00aa1cb9-e07c-4cc8-8372-b4be06315203

### DIFF
--- a/src/components/SUE/report/SueReportLocation.tsx
+++ b/src/components/SUE/report/SueReportLocation.tsx
@@ -1,28 +1,68 @@
 import * as Location from 'expo-location';
-import React from 'react';
+import React, { useCallback } from 'react';
+import { UseFormSetValue } from 'react-hook-form';
 import { StyleSheet, View } from 'react-native';
 
 import { device, normalize, texts } from '../../../config';
 import { useLocationSettings, useSystemPermission } from '../../../hooks';
 import { MapMarker } from '../../../types';
 import { LoadingSpinner } from '../../LoadingSpinner';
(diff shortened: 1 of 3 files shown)


src/components/SUE/report/SueReportLocation.tsx
```diff
@@ -1,29 +1,69 @@
...
+    firstName: string;
+    houseNumber: string;
+    images: string;
+    lastName: string;
+    phone: string;
+    street: string;
+    termsOfService: boolean;
+    title: string;
+    zipCode: string;
+  }>;
```
ardasnturk, line 38: Do we need all these variables here? Can't we only add what we use in this component?

donni106: No, if i reduce them:

> Argument of type 'UseFormSetValue<{ city: string; description: string; email: string; firstName: string; houseNumber: string; images: string; lastName: string; phone: string; street: string; termsOfService: boolean; title: string; zipCode: string; }>' is not assignable to parameter of type 'UseFormSetValue<{ city: string; houseNumber: string; street: string; zipCode: string; }>'.

ardasnturk: Okay, then it stays. 👍

ardasnturk (approved):

PR #982, opened 2023-12-21
feat: update categories when report screen opens

ardasnturk:
- added refetch for services when the report screen is opened in order to show any changes made on the server actively in the application
- added `LoadingSpinner` to show data during loading

SUE-30

### DIFF
--- a/src/components/SUE/report/SueReportServices.tsx
+++ b/src/components/SUE/report/SueReportServices.tsx
@@ -1,10 +1,11 @@
-import React from 'react';
+import React, { useEffect, useState } from 'react';
 import { StyleSheet, TouchableOpacity, View } from 'react-native';
 import { useQuery } from 'react-query';
 
 import { colors, device, normalize } from '../../../config';
 import { imageHeight } from '../../../helpers';
 import { QUERY_TYPES, getQuery } from '../../../queries';
+import { LoadingSpinner } from '../../LoadingSpinner';
 import { BoldText } from '../../Text';
 
 export const SueReportServices = ({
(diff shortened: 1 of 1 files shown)


src/components/SUE/report/SueReportServices.tsx
```diff
@@ -14,7 +15,24 @@ export const SueReportServices = ({
   serviceCode: string;
   setServiceCode: any;
 }) => {
-  const { data } = useQuery([QUERY_TYPES.SUE.SERVICES], () => getQuery(QUERY_TYPES.SUE.SERVICES)());
+  const [loading, setLoading] = useState(true);
+
+  const { data, isLoading, refetch } = useQuery([QUERY_TYPES.SUE.SERVICES], () =>
+    getQuery(QUERY_TYPES.SUE.SERVICES)()
+  );
```
donni106, line 22: i guess it should be possible with passing a config with no cache time (0 ms) to always load new data? could this be possible to not need the workaround with refetching?

src/components/SUE/report/SueReportServices.tsx
```diff
@@ -14,7 +15,24 @@ export const SueReportServices = ({
...
+    await refetch();
+    setLoading(false);
+  };
+
+  if (isLoading || loading) {
+    return <LoadingSpinner loading />;
+  }
 
   return (
     <View style={styles.container}>
```
donni106, line 38: and what about pull to refresh? is this view a child of a scroll view? or could this be transformed into a scroll view?

PR #984, opened 2023-12-21
feat: store data on the report screen

ardasnturk:
- added the ability to write data to the device's memory with the `storeReportValues` function to prevent deletion of data when the user fills in a report screen, returns to the previous screen or closes and reopens the application
- added `setValue` in `readReportValuesFromStore` to add memorised data to inputs
- added delete button with `HeaderRight` to the right of the header to delete all saved data if any
- added `DeleteHeader` component to add delete button to the right of the header and added `withDelete` prop to `HeaderRight`
- added that if the report is sent successfully or if the user wants to delete the saved data, the `resetStoredValues` function is used to delete the saved data

SUE-28

## Screenshots:

|with Delete button|
|--|
image

### DIFF
--- a/src/screens/SUE/SueReportScreen.tsx
+++ b/src/screens/SUE/SueReportScreen.tsx
@@ -1,5 +1,7 @@
+import AsyncStorage from '@react-native-async-storage/async-storage';
+import { StackScreenProps } from '@react-navigation/stack';
 import * as Location from 'expo-location';
-import React, { useRef, useState } from 'react';
+import React, { useEffect, useLayoutEffect, useRef, useState } from 'react';
 import { UseFormSetValue, useForm } from 'react-hook-form';
 import { ActivityIndicator, Alert, ScrollView, StyleSheet } from 'react-native';
 import { Divider } from 'react-native-elements';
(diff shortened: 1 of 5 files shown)


src/screens/SUE/SueReportScreen.tsx
```diff
@@ -225,6 +285,39 @@ export const SueReportScreen = ({ navigation }: { navigation: any }) => {
...
+              route,
+              withDelete: true
+            }}
+          />
+        )
+      });
+    } else {
+      navigation.setOptions({
+        headerRight: () => null
+      });
```
donni106, line 317: is this else really necessary? what if there are no stored values and nothing wpuld be done?

src/screens/SUE/SueReportScreen.tsx
```diff
@@ -199,11 +213,57 @@ export const SueReportScreen = ({ navigation }: { navigation: any }) => {
...
+      setValue('email', storedValues.email);
+      setValue('firstName', storedValues.firstName);
+      setValue('houseNumber', storedValues.houseNumber);
+      setValue('images', storedValues.images);
+      setValue('lastName', storedValues.lastName);
+      setValue('phone', storedValues.phone);
+      setValue('street', storedValues.street);
+      setValue('termsOfService', storedValues.termsOfService);
+      setValue('title', storedValues.title);
+      setValue('zipCode', storedValues.zipCode);
```
donni106, line 242: wouldn't it be possible to go through all stored values and store them dynamically instead of defining each manually? the key is always the same for `setValue` as in `storedValues`

PR #987, opened 2024-01-03
feat: add voucher widget

ardasnturk:
- added widget for voucher to `HomeScreen`
- added `voucher` icon to show in widget

SVAK-35

## Test:

- change the connection in `secret.js` to the staging server so that the voucher query can work
  - this is because there is no database for the voucher on the development server

---

## TODO: 
widget currently redirects to the `VoucherHome` screen. This navigation needs to be updated after the `VoucherIndex` screen is created.

## Screenshots:

image


src/components/widgets/VoucherWidget.tsx
```diff
@@ -0,0 +1,43 @@
...
+  const { isConnected, isMainserverUp } = useContext(NetworkContext);
+
+  const fetchPolicy = graphqlFetchPolicy({ isConnected, isMainserverUp, refreshTime });
+
+  const { data, loading } = useQuery(getQuery(QUERY_TYPES.VOUCHERS), {
+    fetchPolicy,
+    skip: !refreshTime
+  });
+
+  // TODO: add redirection to Voucher IndexScreen
```
donni106, line 28: really? i which one will be a vouchers index? the one with categories or the screen with "latest vouchers"? we do not have a screen with all vouchers, right?!

donni106 (approved):

PR #990, opened 2024-01-09
feat: add voucher categories list

ardasnturk:
- added `GET_VOUCHERS_CATEGORIES` query to list only the category information of the voucher
- added `parseVouchersCategories` parser to `listItemParser` so that we can show the category list of the voucher in the format we want
- changed query in `parseListItemsFromQuery` to parse voucher category in `VoucherIndexScreen`
- added control in `ListHeaderComponent` to show only query value in `Vouchers`
- added count to `TextListItem` to show how many vouchers are in which category of the voucher on the list

SVAK-35

## Screenshots:

|VoucherIndexScreen `query === vouchers`|VoucherIndexScreen `query === vouchersCategories`|
|--|--|
Simulator Screenshot - iPhone 14 Pro - 2024-01-09 at 14 47 40|Simulator Screenshot - iPhone 14 Pro - 2024-01-09 at 14 47 43

### DIFF
--- a/src/components/TextListItem.tsx
+++ b/src/components/TextListItem.tsx
@@ -16,6 +16,7 @@ export type ItemData = {
   id: string;
   badge?: { value: string; textStyle: { color: string } };
   bottomDivider?: boolean;
+  count?: number;
   leftIcon?: React.ReactElement;
   onPress?: (navigation: any) => void;
   params: Record<string, unknown>;
@@ -48,6 +49,7 @@ export const TextListItem: NamedExoticComponent<Props> & {
   const {
     badge,
     bottomDivider,
+    count,
     leftIcon,
     onPress,
     params,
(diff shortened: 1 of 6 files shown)


src/components/TextListItem.tsx
```diff
@@ -146,7 +148,10 @@ export const TextListItem: NamedExoticComponent<Props> & {
         ) : undefined)}
 
       {!!navigation && !withCard && (
-        <Icon.ArrowRight color={colors.darkText} size={normalize(18)} />
+        <>
+          {!!count && <BoldText>({count})</BoldText>}
```
donni106, line 152: if we place the count on the right, it should not be surrounded by brackets

ardasnturk: Yes, you are right. I'll fix it. 👍

donni106 (approved):

src/components/TextListItem.tsx
donni106: no, i meant we can print 1 instead of (1)

ardasnturk: Ach sorry I misunderstood. I used `()` to make the application look the same as the category list. But you are right. If we show it on the right side, we don't need `()`.

PR #993, opened 2024-01-10
feat: integrate generic item query

ardasnturk:
- updated `vouchers` query with `genericItems` as `vouchers` query is no longer in use
- added controls to prevent the application from crashing if the data is undefined

SVAK-35

### DIFF
--- a/src/queries/vouchers.ts
+++ b/src/queries/vouchers.ts
@@ -5,12 +5,33 @@ import { VoucherLogin } from '../types';
 
 // TODO: memberId: 1 will be updated later. 1 will be replaced by the id of the logged in user
 export const GET_VOUCHERS = gql`
-  query Vouchers($limit: Int, $order: GenericItemOrder) {
-    vouchers(limit: $limit, order: $order) {
+  query GenericItems(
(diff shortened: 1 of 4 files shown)


src/queries/vouchers.ts
```diff
@@ -5,12 +5,34 @@ import { VoucherLogin } from '../types';
...
+    $genericType: String
+  ) {
+    genericItems(
+      ids: $ids
+      limit: $limit
+      skip: $offset
+      order: $order
+      dataProvider: $dataProvider
+      categoryId: $categoryId
+      genericType: $genericType
```
donni106, line 24: could this be fixed to `Voucher` instead of using a variable?

ardasnturk: Yes, that's absolutely right. Since this query is only Vouchers query, `Voucher` value can be used for `genericType`.

ardasnturk: https://github.com/smart-village-solutions/smart-village-app-app/commit/5ea2c6f1c749f78bbf8a1c1f43924dc2fee84051

PR #997, opened 2024-01-10
feat: add partner list

ardasnturk:
- added `vouchers` to query to see vouchers via `pointsOfInterest` query
- added control to `listItem` in `Overviews` to list only POIs with vouchers

SVAK-35

## Screenshots:

|IndexScreen with vouchers|IndexScreen map view|
|--|--|
Simulator Screenshot - iPhone 14 Pro - 2024-01-10 at 16 58 15|Simulator Screenshot - iPhone 14 Pro - 2024-01-10 at 16 58 19


src/queries/pointsOfInterest.js
```diff
@@ -78,6 +78,41 @@ export const GET_POINTS_OF_INTEREST = gql`
         dateTo
         useYear
       }
+      vouchers {
+        id
```
donni106, line 82: i guess id should be enough

donni106 (approved):

donni106 (approved):

PR #1004, opened 2024-01-26
fix: tile appearance

ardasnturk:
- added `childrenContainerStyle` instead of `containerStyle` to `Image` component in `ServiceTile` to increase the space between tiles
- added `childrenContainerStyle` prop to avoid style confusion in `Image` component

SVA-1256

## Screenshots:

|before|after|
|--|--|
Simulator Screenshot - iPhone 14 Pro Max - 2024-01-26 at 13 44 09|Simulator Screenshot - iPhone 14 Pro Max - 2024-01-26 at 13 44 00

### DIFF
--- a/src/components/Image.js
+++ b/src/components/Image.js
@@ -105,9 +105,7 @@ export const Image = ({
     <View>
       <RNEImage
         source={source}
-        childrenContainerStyle={
-          childrenContainerStyle || style || stylesForImage(aspectRatio).defaultStyle
-        }
+        childrenContainerStyle={childrenContainerStyle || stylesForImage(aspectRatio).defaultStyle}
         containerStyle={containerStyle}
         style={style}
         PlaceholderContent={PlaceholderContent}
--- a/src/components/Image.js
+++ b/src/components/Image.js
@@ -30,6 +30,7 @@ export const Image = ({
   aspectRatio,
   borderRadius,
   button,
+  childrenContainerStyle,
   containerStyle,
   message,
   PlaceholderContent,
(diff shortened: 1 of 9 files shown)


src/components/Image.js
```diff
@@ -104,7 +105,9 @@ export const Image = ({
     <View>
       <RNEImage
         source={source}
-        childrenContainerStyle={style || stylesForImage(aspectRatio).defaultStyle}
+        childrenContainerStyle={
+          childrenContainerStyle || style || stylesForImage(aspectRatio).defaultStyle
+        }
         containerStyle={containerStyle}
         style={style}
```
donni106, line 112: now `style` is passed in `childrenContainerStyle` and `style`.

in 4550329f16ccdef1de562c18775b5f594b9564b4 you have added `style` prop to `RNEImage` component to apply the style of the image more correctly. what did "more correctly" mean? 
in 4686b81529eaaa2b5f69b5baa642a454a7f5a328 you updated `style` prop in `Image` component to `childrenContainerStyle`.

the current version seems to be a weird mix of everything to me without knowing, what is applied where and why.
having a `childrenContainerStyle` with `childrenContainerStyle || style` and additionally a `style` with `style` is not comprehensible.

before the changes, there was one clear `style` with `style || stylesForImage(aspectRatio).defaultStyle}`.
please make sure, that passed styles for images are clear again without putting different styles in multiple places, because it seems to work.

ardasnturk: As far as I understand from the React-Native-Elements document, the Image component has 3 different styles. 

1st (style) is only the style of the image. 
2nd style (childrenContainerStyle) is the style of the child container that the image is in.
3. style (containerStyle) style of the container in which the whole component is wrapped.

These 3 different styles cause 3 different looks.

For clarity, I remove the `style` prop from `childrenContainerStyle` and update the `style` prop in all `Image` components we use with `childrenContainerStyle`.

ardasnturk: https://github.com/smart-village-solutions/smart-village-app-app/pull/1004/commits/d2087205b63ec66a07b38d86f0fcb00b95f2c031

donni106: this is more clear regarding the code! thanks.

can we be sure that all images render in the way they did before?

ardasnturk: When updating the prop of the image component, I checked the screens with the images and I can say that it works correctly. I will add screenshots.

donni106 (approved):

PR #1007, opened 2024-01-29
feat: take a photo with the camera

ardasnturk:
- added `useCaptureImage` to `imagePicker` to take a photo with the camera
- added alert to select image from gallery or camera
- added `useCaptureImage` to be able to run the camera
- added `imageFunction` prop to `imageSelect` function to avoid code repetition and made it work with the function sent with the prop
- updated `imageSelect` `onPress` in `Button` component with `secImageSourceAlert`

SUE-31

## How to test:
* [x] Android
* [x] iOS

## Screenshots:

|image source select alert|camera screen|image edit screen|show a photo taken from the camera|
|--|--|--|--|
IMG_6784|IMG_6785|IMG_6786|IMG_6787

### DIFF
--- a/src/config/texts.js
+++ b/src/config/texts.js
@@ -925,6 +925,14 @@ export const texts = {
         imageGreater10MBError: 'Das ausgewählte Bild darf maximal 10 MB groß sein.',
         imagesGreater30MBError:
           'Die ausgewählten Bilder dürfen insgesamt nicht größer als 30 MB sein.',
+        imageSelectAlert: {
+          camera: 'Kamera',
+          cancel: 'Abbrechen',
+          description:
+            'Möchten Sie ein Bild mit der Kamera aufnehmen oder aus der Galerie auswählen?',
+          gallery: 'Galerie',
+          title: 'Bildquelle auswählen'
+        },
         location: 'Bitte wählen Sie einen Ort auf der Karte aus.',
         serviceCode: 'Bitte wählen Sie aus, um welches Thema es in dem Bericht geht.',
         termsOfService: 'Bitte akzeptieren Sie die Datenschutzbestimmungen.',
--- a/src/hooks/imagePicker.ts
+++ b/src/hooks/imagePicker.ts
@@ -1,7 +1,9 @@
 import {
+  launchCameraAsync,
   launchImageLibraryAsync,
   MediaTypeOptions,
   PermissionStatus,
+  requestCameraPermissionsAsync,
   requestMediaLibraryPermissionsAsync
 } from 'expo-image-picker';
 import { useCallback, useState } from 'react';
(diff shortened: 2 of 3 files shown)


src/hooks/imagePicker.ts
```diff
@@ -45,3 +47,40 @@ export const useSelectImage = (
...
+    const { status } = await requestCameraPermissionsAsync();
+
+    if (status !== PermissionStatus.GRANTED) {
+      Alert.alert(texts.errors.image.title, texts.errors.image.body);
+      return;
+    }
+
+    // this allows for proper selecting and cropping to 1:1 images (and not videos)
+    // for more details about options see: https://docs.expo.dev/versions/latest/sdk/imagepicker/#imagepickermediatypeoptions
+    const result = await launchCameraAsync({
```
donni106, line 72: i cannot open a camera:

> Error: Camera not available on simulator

we had working camera in simulators before, what is different now?

src/config/texts.js
```diff
@@ -925,6 +925,13 @@ export const texts = {
         imageGreater10MBError: 'Das ausgewählte Bild darf maximal 10 MB groß sein.',
         imagesGreater30MBError:
           'Die ausgewählten Bilder dürfen insgesamt nicht größer als 30 MB sein.',
+        imageSelectAlert: {
+          camera: 'Kamera',
+          cancel: 'Abbrechen',
+          description: 'Möchten Sie das Bild aus der Kamera oder aus der Galerie auswählen?',
```
donni106, line 931: please change to: Möchten Sie ein Bild mit der Kamera aufnehmen oder aus der Galerie auswählen?

ardasnturk: https://github.com/smart-village-solutions/smart-village-app-app/commit/482c4d81085f4d491efb9b2067efc4929b96b4cf

src/hooks/imagePicker.ts
ardasnturk: I don't think there is a camera that works with the simulator. We need a real device to use the camera feature.

src/hooks/imagePicker.ts
```diff
@@ -45,3 +47,40 @@ export const useSelectImage = (
...
+      mediaTypes: mediaTypes ?? MediaTypeOptions.Images,
+      allowsEditing: allowsEditing ?? true,
```
donni106, line 74: i suggest fallback `false` here

src/hooks/imagePicker.ts
```diff
@@ -45,3 +47,40 @@ export const useSelectImage = (
...
+      aspect: aspect ?? [1, 1],
```
donni106, line 75: what about the fallback of 1, 1? can we please keep the original foto aspect ratio?

src/hooks/imagePicker.ts
ardasnturk: Since it is `true` in the `useSelectImage` function, I set it to `true` here.

image

src/hooks/imagePicker.ts
ardasnturk: If we remove the `aspect`, we have the original aspect ratio. But [1, 1] is also available for photos selected from the gallery. Should we update this too?

donni106: both should be identical, whether original format or 1:1

ardasnturk: https://github.com/smart-village-solutions/smart-village-app-app/pull/1007/commits/f9c2849385f73cb7ac9388da8f85a0675817a9aa

src/hooks/imagePicker.ts
ardasnturk: This screenshot is the section in `useSelectImage` where we set `allowsEditing`. Here I wanted to show that we set it to true before, but now I set it to false by default because we are not using the image editing feature.

https://github.com/smart-village-solutions/smart-village-app-app/pull/1007/commits/f9c2849385f73cb7ac9388da8f85a0675817a9aa

donni106 (approved):

PR #1009, opened 2024-01-29
feat: remove logo on the map

ardasnturk:
- added bottom value in `mapPadding` as twice the `height` of the device to remove the logo displayed on the map

SUE-33

Since the bottom value could never take a negative value, the `-normalize(30)` value was not working. For this reason, the bottom value was set to twice the height of the device and the logo was made invisible on the screen of the device.

## How to test:
* [x] Android portrait mode
* [x] Android landscape mode

## Screenshots:

|before|after|
|--|--|
Screenshot_20240129-142458|Screenshot_20240129-142440

### DIFF
--- a/src/components/map/Map.tsx
+++ b/src/components/map/Map.tsx
@@ -106,7 +106,7 @@ export const Map = ({
         mapPadding={{
           top: 0,
           right: 0,
-          bottom: -normalize(30),
+          bottom: normalize(device.height * 2),
           left: 0
         }}
         legalLabelInsets={{


src/components/map/Map.tsx
```diff
@@ -106,7 +106,7 @@ export const Map = ({
         mapPadding={{
           top: 0,
           right: 0,
-          bottom: -normalize(30),
+          bottom: normalize(device.height * 2),
```
donni106, line 109: have you checked in report location screen and in settings for location, where we do not have a full size map?

ardasnturk: Yes, I have checked and confirmed that the logo is not visible. I am sharing screenshots.

|report screen|settings screen|
|--|--|
Screenshot_20240129-144301|Screenshot_20240129-144310

donni106: but the link to some right infos on the bottom right is also gone now?

image

ardasnturk: This still visible, but only on the iOS platform. It was not visible on Android before, it was only visible on iOS.

donni106: i understand, all right

donni106 (approved):

PR #1010, opened 2024-01-29
feat(sue): use my location for report

donni106:
SUE-32

|||
|---|---|
|simulator_screenshot_CC95CE43-056E-4286-ACC1-7B9E18D861CA|Simulator Screen Shot - iPhone SE (3rd generation) - 2024-01-29 at 16 35 15|


src/components/SUE/report/SueReportLocation.tsx
```diff
@@ -92,13 +102,37 @@ export const SueReportLocation = ({
         locations={locations}
         mapCenterPosition={mapCenterPosition}
         mapStyle={styles.map}
+        isMyLocationButtonVisible
+        onMyLocationButtonPress={() =>
+          Alert.alert(texts.sue.report.alerts.hint, texts.sue.report.alerts.myLocation, [
+            {
+              text: texts.sue.report.alerts.no,
+              onPress: () => null
```
ardasnturk, line 110: Do we need an onPress that returns null?

src/components/SUE/report/SueReportLocation.tsx
```diff
@@ -92,13 +102,37 @@ export const SueReportLocation = ({
...
-
+          setUpdatedRegion(true);
+          setSelectedPosition(nativeEvent.coordinate);
           reverseGeocode(nativeEvent.coordinate);
         }}
+        updatedRegion={
+          selectedPosition && updatedRegion
+            ? { ...selectedPosition, latitudeDelta: 0.01, longitudeDelta: 0.01 }
+            : undefined
+        }
```
ardasnturk, line 135: This part of the code causes the map to be centred and the zoom to change for each pin added. Do we want this to happen? Maybe we can remove the `latitudeDelta` and `longitudeDelta` values?

src/components/SUE/report/SueReportLocation.tsx
```diff
@@ -92,13 +102,37 @@ export const SueReportLocation = ({
...
+              text: texts.sue.report.alerts.no,
+              onPress: () => null
+            },
+            {
+              text: texts.sue.report.alerts.yes,
+              onPress: () => {
+                const location = position || lastKnownPosition;
+
+                if (location) {
+                  setUpdatedRegion(true);
```
ardasnturk, line 118: I think `setUpdatedRegion(true)` is unnecessary here, because we already return `true` in the `geocode` callback. This value becomes true for every location change or text change. Since we have 4 inputs, it becomes true 4 times for each change. So I think it is unnecessary to use it here.

src/components/SUE/report/SueReportLocation.tsx
```diff
@@ -92,13 +102,37 @@ export const SueReportLocation = ({
...
+              }
+            }
+          ])
+        }
         onMapPress={({ nativeEvent }) => {
-          setSelectedPosition({
-            ...nativeEvent.coordinate
-          });
-
+          setUpdatedRegion(true);
```
ardasnturk, line 127: I think `setUpdatedRegion(true)` is also unnecessary here.

ardasnturk (changes requested):

src/components/SUE/report/SueReportLocation.tsx
donni106: please try with removing the deltas

src/components/SUE/report/SueReportLocation.tsx
donni106: please try without it

src/components/SUE/report/SueReportLocation.tsx
donni106: we only want to update the region if this is true. otherwise it would already be updated when mounting the screen and your current position would be centered. did you try it?

src/components/SUE/report/SueReportLocation.tsx
donni106: did you try it? we don't want to center to a region when launching the screen

src/components/SUE/report/SueReportLocation.tsx
ardasnturk: I tried it and the zoom didn't change. Because we already have a zoom value set before. So I'm deleting it.

|before choosing a location|after with detas value|after without detas value|
|--|--|--|
Simulator Screenshot - iPhone 14 Pro Max - 2024-01-30 at 12 08 21|Simulator Screenshot - iPhone 14 Pro Max - 2024-01-30 at 12 08 44|Simulator Screenshot - iPhone 14 Pro Max - 2024-01-30 at 12 08 59|

src/components/SUE/report/SueReportLocation.tsx
ardasnturk: I tried it and there was no change. I didn't notice any difference between it being `null` or not having `onPress`.

src/components/SUE/report/SueReportLocation.tsx
ardasnturk: I'm sorry but I don't think this is working correctly. I haven't made any code changes and it always shows me at the predetermined location. 

What exactly do we want, that the location shown on the map is our own location or the `defaultAlternativePosition` location? I didn't delete this code and tried it. When the map opens, I can't see my own location. I see whatever location information is in `defaultAlternativePosition`. Deleting the code doesn't change anything.

I log `updatedRegion`. I go to ReportScreen and I see that updatedRegion is `false`. It returns `true` only if a location is selected from the map or if we want to use our own location or if we enter any address in the inputs.

Simulator Screenshot - iPhone 14 Pro Max - 2024-01-30 at 12 20 48

ardasnturk: It's all right. It's my fault. We're actually thinking and wanting the same thing. We talked about it over Slack and clarified it.

src/components/SUE/report/SueReportLocation.tsx
ardasnturk: https://github.com/smart-village-solutions/smart-village-app-app/commit/937e0d0d7ff2eee80a4ea1a53ea2c175473bdc16

src/components/SUE/report/SueReportLocation.tsx
ardasnturk: Yes, I tried, and you're right, it doesn't work as desired.

ardasnturk (approved):

PR #1013, opened 2024-01-31
fix: app crashes when selecting location from map

ardasnturk:
- reverted `bottom` in `mapPadding` to fix bounce bug when selecting location on map
- added `OpenStreetMap` text only for android platform to cover the google logo that appears on android devices
- added `latitudeDelta` and `longitudeDelta` values to solve the delta error that causes the application to crash when the location is selected from the map on android devices

SUE-32

## Screenshots:

|before|after|
|--|--|
Screenshot (31 01 2024 11_30_13)|Screenshot (31 01 2024 10_30_54)

### DIFF
--- a/src/components/map/Map.tsx
+++ b/src/components/map/Map.tsx
@@ -7,6 +7,7 @@ import { colors, device, Icon, normalize } from '../../config';
 import { imageHeight, imageWidth } from '../../helpers';
 import { SettingsContext } from '../../SettingsProvider';
 import { MapMarker } from '../../types';
+import { RegularText } from '../Text';
 
 type Props = {
   geometryTourData?: LatLng[];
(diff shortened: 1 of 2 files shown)


donni106 (approved):

src/components/map/Map.tsx
```diff
@@ -220,6 +226,17 @@ const styles = StyleSheet.create({
...
+    alignItems: 'center',
+    backgroundColor: colors.surface,
+    bottom: 0,
+    height: normalize(30),
+    justifyContent: 'center',
+    left: normalize(13),
+    position: 'absolute',
+    width: normalize(90),
+    zIndex: 1
+  },
```
donni106, line 239: does not work in any case

image

i will update to not break the line

donni106: image

donni106 (approved):

PR #1015, opened 2024-02-06
chore: adjust packages

donni106:
- ran `npx expo-doctor` and `npx expo install --fix`
- kept @expo/vector-icons at 14 because of FontAwesome6

image


package.json
```diff
@@ -35,7 +35,7 @@
   },
   "dependencies": {
     "@apollo/react-hooks": "~3.1.5",
-    "@expo/vector-icons": "14.0.0",
+    "@expo/vector-icons": "^13.0.0",
```
ardasnturk, line 38: Does version 13 include the FontAwesome6 icon pack? If not, I would prefer to keep it in version 14.

donni106: ah interesting, so we will force to use the newer against expos suggestion, sure.

ardasnturk: Yeah, I'm sorry, but we're gonna have to do this. At least until we update to Expo 50.

ardasnturk (approved):

PR #1018, opened 2024-02-07
 fix: error in `HomeScreen` event list view

ardasnturk:
- added `isHomeScreen` prop to fix the `CardListItem` view error of the 2. index in the event list in `HomeScreen`

SVA-1248

## Screenshots:

|before `HomeScreen`|after `HomeScreen`|before event list|after event list|
|--|--|--|--|
Simulator Screenshot - iPhone 14 Pro Max - 2024-02-07 at 12 23 56|Simulator Screenshot - iPhone 14 Pro Max - 2024-02-07 at 12 23 33|Simulator Screenshot - iPhone 14 Pro Max - 2024-02-07 at 12 24 04|Simulator Screenshot - iPhone 14 Pro Max - 2024-02-07 at 12 23 45

### DIFF
--- a/src/hooks/listHooks.js
+++ b/src/hooks/listHooks.js
@@ -51,6 +51,7 @@ const EventSectionHeader = ({ item, navigation, options, query }) => (
  * @param {any} navigation
  * @param {{
  *          horizontal?: boolean;
+ *          isHomeScreen?: boolean
  *          noSubtitle?: boolean;
  *          openWebScreen?: () => void;
  *          queryVariables?: object
@@ -104,7 +105,11 @@ export const useRenderItem = (query, navigation, options = {}) => {
 
         // this logic was added because the 0th index of the `EventList` is `SectionHeader` and
         // the first content is the 1st index
-        if (index === 0 || (query === QUERY_TYPES.EVENT_RECORDS && index === 1)) {
+        // Also added `isHomeScreen` control to show a large image in the event list in `HomeScreen`
+        if (
+          index === 0 ||
+          (query === QUERY_TYPES.EVENT_RECORDS && index === 1 && !options.isHomeScreen)
+        ) {
           if (!item?.picture?.url) {
             return (
               <TextListItem
--- a/src/screens/HomeScreen.js
+++ b/src/screens/HomeScreen.js
@@ -147,6 +147,7 @@ const renderItem = ({ item }) => {
       {...{
         buttonTitle,
         fetchPolicy,
+        isHomeScreen: query === QUERY_TYPES.EVENT_RECORDS,
         navigate: () => navigation.navigate(NAVIGATION[navigate]),
         navigation,
         query,
--- a/src/components/DataListSection.tsx
+++ b/src/components/DataListSection.tsx
@@ -17,6 +17,7 @@ import { Touchable } from './Touchable';
 import { Wrapper } from './Wrapper';
 
 type Props = {
+  additionalData?: unknown[];
   buttonTitle?: string;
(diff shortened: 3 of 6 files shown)


src/screens/HomeScreen.js
```diff
@@ -147,6 +147,7 @@ const renderItem = ({ item }) => {
       {...{
         buttonTitle,
         fetchPolicy,
+        isHomeScreen: query === QUERY_TYPES.EVENT_RECORDS,
```
donni106, line 150: why is the prop called `isHomeScreen` when the check is if the query is event records?
all the components can be used on other screens as well, where we cannot have a prop `isHomeScreen` that we probably set to true or false.
can you think of a different naming solution?
what would you describe it to me in words? in which case do the prop needs to be true?

ardasnturk: I chose this option because this problem only occurs with the event list on `HomeScreen`, because there is a special case for the Event list in `listHook`.

How about changing the name to `isEventListOnHomeScreen`?

donni106: then it is still home screen. what is that "special case"? can it be named? which data is the reason?

ardasnturk: We have 2 different event listings. The 1st list is the list we list on the `EventList` screen. The data in this list starts at index 1. Because index 0 is `SectionHeader`.

The other event list is the one we show in `HomeScreen`. The data in this list starts at index 0.

We need to make the image of the first data bigger and the images of the other data smaller. Therefore, we need to consider 2 different situations in the event list.

In `listHook`, we solved the problem in `EventList` screen by adding filter according to index, but this time the list view in `HomeScreen` is broken. So we needed to add a special control for the event list in `HomeScreen`.

ardasnturk: You're absolutely right, by the way. I don't want to name a prop that is used everywhere as `HomeScreen` either, but unfortunately I haven't found a better solution.

donni106: > We have 2 different event listings. The 1st list is the list we list on the EventList screen. The data in this list starts at index 1. Because index 0 is SectionHeader.
>
> The other event list is the one we show in HomeScreen. The data in this list starts at index 0.

thanks for the explanation. this lets me think about `isIndexStartingAt0` or `isIndexStartingAt1` for example?

ardasnturk: Logical! I updated the code with `isIndexStartingAt0`.

ardasnturk: https://github.com/smart-village-solutions/smart-village-app-app/commit/7a5d494d9248f9cef1385dab1e4a0cc85376f69a

src/hooks/listHooks.js
```diff
@@ -102,9 +103,9 @@ export const useRenderItem = (query, navigation, options = {}) => {
           return <EventSectionHeader {...{ item, navigation, options, query }} />;
         }
 
-        // this logic was added because the 0th index of the `EventList` is `SectionHeader` and
-        // the first content is the 1st index
-        if (index === 0 || (query === QUERY_TYPES.EVENT_RECORDS && index === 1)) {
+        // in special lists such as `EventList`, since the data starts from the 1st index,
+        // `isIndexStartingAt1` and `index` control has been added
+        if (index === 0 || (index === 1 && options.isIndexStartingAt1)) {
```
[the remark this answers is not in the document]
ardasnturk: Adding `&& !options.isIndexStartingAt1` to the `index === 0` control will make the code better.

PR #1019, opened 2024-02-07
fix: unable to edit tile error

ardasnturk:
- created `UMLAUT_REGEX` to detect hopeless letters that cause the application to crash when editing tile view
- changed the umlaut letters in the texts used in `draggableId` to prevent the application from crashing when organising the order of the tiles

SVA-1291

### DIFF
--- a/src/components/screens/Service.tsx
+++ b/src/components/screens/Service.tsx
@@ -15,7 +15,25 @@ import { WrapperWrap } from '../Wrapper';
 import { DraggableGrid } from './DraggableGrid';
 import { ServiceTile, TServiceTile } from './ServiceTile';
 
-const { MATOMO_TRACKING } = consts;
+const { MATOMO_TRACKING, UMLAUT_REGEX } = consts;
+
+const umlautSwitcher = (text: string) => {
+  if (!text) return;
+
+  const umlautReplacements = {
+    ü: 'ue',
+    ä: 'ae',
+    ö: 'oe',
+    Ü: 'UE',
+    Ä: 'AE',
+    Ö: 'OE',
+    ß: 'ss'
+  };
+
+  const replacedText = text.replace(UMLAUT_REGEX, (match: string) => umlautReplacements[match]);
+
+  return replacedText;
+};
 
 export const Service = ({
   data,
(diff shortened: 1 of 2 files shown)


src/config/consts.js
```diff
@@ -77,6 +77,7 @@ export const consts = {
   URL_REGEX:
     /^https?:\/\/(?:www\.)?[-a-z0-9@:%._\+~#=]{1,256}\.[a-z0-9()]{1,6}\b(?:[-a-z0-9()@:%_\+.~#?&\/=]*)$/i,
   VIDEO_TYPE_REGEX: /\.(swf|avi|flv|mpg|rm|mov|wav|asf|3gp|mkv|rmvb|mp4)$/i,
+  UMLAUT_REGEX: /[\u00C0-\u017F]/g,
```
donni106, line 80: does this be exact the same as `/[üäöÜÄÖß]/g` or match more?

src/components/screens/Service.tsx
```diff
@@ -15,7 +15,34 @@ import { WrapperWrap } from '../Wrapper';
...
+      case 'Ö':
+        return 'OE';
+      case 'ß':
+        return 'ss';
+
+      default:
+        return match;
+    }
+  });
+  return replacedText;
```
donni106, line 44: i thought that there could be a smarter solution and asked chatgpt3.5:

```
const umlautReplacements = {
  'ü': 'ue',
  'ä': 'ae',
  'ö': 'oe',
  'Ü': 'UE',
  'Ä': 'AE',
  'Ö': 'OE',
  'ß': 'ss'
};

const replacedText = text.replace(UMLAUT_REGEX, match => umlautReplacements[match]);
```

what do you think? does that work and if yes, how do you feel about it?

src/config/consts.js
ardasnturk: Yes, practically the same.

src/components/screens/Service.tsx
ardasnturk: It looks much more beautiful! 👍

src/config/consts.js
donni106: i think my suggestion is more readable, would you like to change it?

ardasnturk: https://github.com/smart-village-solutions/smart-village-app-app/commit/630f877b1989fdb95f3832ea9b40ded03862c507

donni106 (approved):

PR #1021, opened 2024-02-12
feat: delete saved data when the application is reopened

ardasnturk:
- added `AsyncStorage.removeItem` to the `isFontLoaded` controller in `App.js` according to the `sueReportValues` key to ensure that the saved data is deleted when the application is reopened
- created `sueReportValues` constant to prevent code repetition

SUE-48

### DIFF
--- a/App.js
+++ b/App.js
@@ -1,3 +1,4 @@
+import AsyncStorage from '@react-native-async-storage/async-storage';
 import * as Font from 'expo-font';
 import * as SplashScreen from 'expo-splash-screen';
 import React, { useCallback, useEffect, useRef, useState } from 'react';
@@ -8,6 +9,7 @@ import * as Sentry from 'sentry-expo';
 import { MainApp } from './src';
 import { auth } from './src/auth';
 import { fontConfig, namespace, secrets } from './src/config';
+import { SUE_REPORT_VALUES } from './src/screens';
 
 const sentryApi = secrets[namespace].sentryApi;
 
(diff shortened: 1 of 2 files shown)


App.js
```diff
@@ -32,6 +34,10 @@ const App = () => {
     const subscription = AppState.addEventListener('change', (nextAppState) => {
       if (appState.current.match(/inactive|background/) && nextAppState === 'active') {
         auth();
+
+        // saved data in the sue module is deleted if the application returns to the foreground from
+        // the background or inactive state
+        AsyncStorage.removeItem(SUE_REPORT_VALUES);
```
donni106, line 40: i guess t should only happen if the app was closed entirely, not when just moved in the background. imagine you start filling the form, receive a short message, switches to messaging app and return after that. in that case the form data should be still present

ardasnturk: But we can't see if the application has been closed completely. Then I'll have to find another way.  🤷‍♂️

donni106 (changes requested):
It's not a solution for the problem, yet.

App.js
ardasnturk: I added the code to the section where we make SplashScreen hidden. I think it is much better in terms of code structure and logic.

https://github.com/smart-village-solutions/smart-village-app-app/commit/1d11ef5dd20fd0deb67b20f265b923b31b0d30f8

App.js
```diff
@@ -65,6 +67,10 @@ const App = () => {
       // we hide the splash screen once we know the root view has already
       // performed layout.
       await SplashScreen.hideAsync();
+
+      // saved data in the sue module is deleted if the application returns to the foreground from
+      // the background or inactive state
+      await AsyncStorage.removeItem(SUE_REPORT_VALUES);
```
donni106, line 73: i suggest placing this before hiding the splash screen to be sure that this process of deletion is done, before "starting" the app

donni106 (approved):

PR #1022, opened 2024-02-13
fix: app crashes in map view due to pagination

ardasnturk:
- removed the `limit` and `offset` required for pagination in list view because they are not required in map view and cause the app to crash

SUE-39

`limit` and `offset` values required for pagination have been removed from the api request in the map view. When `limit` and `offset` values are in `queryVariables`, a blank page is encountered when switching from map view to list view. In order to get the data in list view, it is necessary to pull the data from the api again with pull to refresh.


donni106 (approved):
👍 👍 👍

PR #1023, opened 2024-02-13
feat: add sending report loading screen

ardasnturk:
- added `isMutationLoading` to understand when sending the report
- added `loadingTitle` and `loadingSubtitle` that can be edited from the server to show when sending the report in the `SueReportDone` component
- added `isDone` and `isLoading` props to `SueReportDone` component to check if the report is in the status of being sent to the server
- updated the `style` prop in the `Image` component with `childrenContainerStyle` to fix the style distortion bug after the image is selected in `ImageSelector`
- apdated the name of the `SueReportDone` component to `SueReportSend` for clarity

SUE-38

## Screenshots:

|report send loading|
|--|
Simulator Screenshot - iPhone 14 Pro Max - 2024-02-13 at 13 01 24

### DIFF
--- a/src/screens/SUE/SueReportScreen.tsx
+++ b/src/screens/SUE/SueReportScreen.tsx
@@ -142,7 +142,7 @@ export const SueReportScreen = ({
     }
   });
 
-  const { mutateAsync, isLoading: isMutationLoading } = useMutation(postRequests);
+  const { mutateAsync } = useMutation(postRequests);
 
   const onSubmit = async (sueReportData: TReports) => {
     storeReportValues();
@@ -164,15 +164,21 @@ export const SueReportScreen = ({
     setIsLoading(true);
     mutateAsync(formData)
       .then(() => {
-        setIsDone(true);
-        resetStoredValues();
+        setTimeout(() => {
+          setIsDone(true);
+          resetStoredValues();
+        }, 1500);
       })
       .catch(() => {
         setIsLoading(false);
 
         return Alert.alert(texts.defectReport.alerts.hint, texts.defectReport.alerts.error);
       })
-      .finally(() => setIsLoading(false));
+      .finally(() => {
+        setTimeout(() => {
+          setIsLoading(false);
+        }, 1500);
+      });
   };
 
   /* eslint-disable complexity */
@@ -315,8 +321,8 @@ export const SueReportScreen = ({
     );
   }
 
-  if (isDone || isMutationLoading) {
-    return <SueReportSend navigation={navigation} isDone={isDone} isLoading={isMutationLoading} />;
+  if (isDone || isLoading) {
+    return <SueReportSend navigation={navigation} isDone={isDone} isLoading={isLoading} />;
   }
 
   return (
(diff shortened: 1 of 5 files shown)


src/screens/SUE/SueReportScreen.tsx
```diff
@@ -315,8 +315,8 @@ export const SueReportScreen = ({
     );
   }
 
-  if (isDone) {
-    return <SueReportDone navigation={navigation} />;
+  if (isDone || isMutationLoading) {
```
donni106, line 318: the state of `isLoading` can change very fast, which results in a bad user experience. we should at least wait 1.5 or 2 seconds showing the sending status before showing the done.

ardasnturk: Yes, that's a very good idea. For this I think we can update the state a little late with `setTimeout`.

donni106: but if we are uploading images, we do not want to have the additional timeout added, because the process is already running for a bit.

ardasnturk: But it's very difficult to control. Because if the image size is 100Kb, the report will still be sent very quickly. Maybe we can add size control 🤷‍♂️

ardasnturk: I added a 1.5 second timer.

Simulator Screen Recording - iPhone 15 Pro Max - 2024-02-13 at 18.15.33.mp4
https://github.com/smart-village-solutions/smart-village-app-app/pull/1023/commits/f3f860f5592136f19511931388343c6bdd677dd3

src/screens/SUE/SueReportScreen.tsx
```diff
@@ -164,15 +164,21 @@ export const SueReportScreen = ({
...
       })
       .catch(() => {
         setIsLoading(false);
 
         return Alert.alert(texts.defectReport.alerts.hint, texts.defectReport.alerts.error);
       })
-      .finally(() => setIsLoading(false));
+      .finally(() => {
+        setTimeout(() => {
+          setIsLoading(false);
```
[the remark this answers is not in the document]
ardasnturk: `catch` only works if an error occurs. `finally` always works. but I think it can be moved to `setIsLoading(false)`, `then` section. As a result, we set the `isLoading` value to `false` in `finally`. We can also do this in the `then` section. If an error occurs, `setIsDone` will never be `true`.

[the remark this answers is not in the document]
donni106: we want this to happen for `then` and `catch` right, and if we place it explicitly in `catch` we can move it to `then` as well an do not need the `finally`, or am i wrong?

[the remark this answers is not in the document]
ardasnturk: Yes, that's right. As I just said, it can be moved to the `then` section. 👍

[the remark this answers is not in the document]
ardasnturk: https://github.com/smart-village-solutions/smart-village-app-app/pull/1023/commits/35d639541e4ac609d683ea7f3a98741f68aa17ca

donni106 (approved):

PR #1025, opened 2024-02-13
fix: links open externally instead of the application

ardasnturk:
- added `useOpenWebScreen` to `DetailScreen` to open form links in bbbus service from web browser in application instead of browser

SVA-1299

## Test:

the navigation to follow in the app to test it:

Service Tab -> HTML - Buergerservice -> Bürger- und Unternehmenservice öffnen -> Gewerbe anmelden -> Gewerbe Anmeldung (Button)

### DIFF
--- a/src/screens/BB-BUS/DetailScreen.js
+++ b/src/screens/BB-BUS/DetailScreen.js
@@ -13,7 +13,7 @@ import { TextBlock } from '../../components/BB-BUS/TextBlock';
 import { FeedbackFooter } from '../../components/FeedbackFooter';
 import { LoadingSpinner } from '../../components/LoadingSpinner';
 import { colors, consts, device, normalize } from '../../config';
-import { matomoTrackingString, openLink } from '../../helpers';
+import { matomoTrackingString, openLink, rootRouteName } from '../../helpers';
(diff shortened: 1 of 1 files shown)


src/screens/BB-BUS/DetailScreen.js
```diff
@@ -157,7 +168,9 @@ export const DetailScreen = ({ route }) => {
         }
       >
         {!!forms?.length && (
-          <View style={styles.formContainer}>{forms.map((form) => renderForm(form))}</View>
+          <View style={styles.formContainer}>
+            {forms.map((form) => renderForm(headerTitle, form, rootRouteName))}
```
donni106, line 172: why do we need the `rootRouteName` here. it is an optional param.

ardasnturk:
```
  const rootRouteName = route.params?.rootRouteName ?? '';
```
Yes, it is not required, but we already have `rootRouteName` as param. That's why I added it.

donni106 (approved):

src/screens/BB-BUS/DetailScreen.js
donni106: ok thanks, keep it.

PR #1029, opened 2024-02-14
feat: get the service data from server to filter it

ardasnturk:
- added `servicesFilterData` to get services information from server and add it to filter dropdown
- updated the data of `servicesData` in `servicesFilterData` to fit the data coming from the server to the data type requested by the dropdown

SUE-14

## Screenshots:

|filter with services data|filter with services data|filtered list with `Wilder Müll` service|
|--|--|--|
Simulator Screenshot - iPhone 15 Pro Max - 2024-02-14 at 16 53 51|Simulator Screenshot - iPhone 15 Pro Max - 2024-02-14 at 16 53 54|Simulator Screenshot - iPhone 15 Pro Max - 2024-02-14 at 16 54 19

### DIFF
--- a/src/screens/SUE/SueListScreen.tsx
+++ b/src/screens/SUE/SueListScreen.tsx
@@ -55,19 +55,19 @@ export const SueListScreen = ({ navigation, route }: Props) => {
       })
   );
 
-  const { data: servicesData } = useQuery([QUERY_TYPES.SUE.SERVICES, {}], () =>
-    getQuery(QUERY_TYPES.SUE.SERVICES)({})
+  const { data: servicesData } = useQuery([QUERY_TYPES.SUE.SERVICES], () =>
+    getQuery(QUERY_TYPES.SUE.SERVICES)()
   );
 
-  const servicesFilterData = useMemo(() => {
+  const services = useMemo(() => {
     if (!servicesData?.length) return;
 
     return servicesData.map((item: any, index: number) => ({
-      selected: false,
-      value: item.serviceName,
       filterValue: item.serviceCode,
       id: index,
-      index
+      index,
+      selected: false,
+      value: item.serviceName
     }));
   }, [servicesData]);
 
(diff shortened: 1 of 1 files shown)


src/screens/SUE/SueListScreen.tsx
```diff
@@ -53,6 +55,22 @@ export const SueListScreen = ({ navigation, route }: Props) => {
...
   );
 
+  const { data: servicesData } = useQuery([QUERY_TYPES.SUE.SERVICES, {}], () =>
+    getQuery(QUERY_TYPES.SUE.SERVICES)({})
+  );
+
+  const servicesFilterData = useMemo(() => {
+    if (!servicesData?.length) return;
+
+    return servicesData.map((item: any, index: number) => ({
```
donni106, line 65: please sort them alphabetically on the `serviceName`

src/screens/SUE/SueListScreen.tsx
```diff
@@ -53,6 +55,22 @@ export const SueListScreen = ({ navigation, route }: Props) => {
       })
   );
 
+  const { data: servicesData } = useQuery([QUERY_TYPES.SUE.SERVICES, {}], () =>
+    getQuery(QUERY_TYPES.SUE.SERVICES)({})
+  );
+
+  const servicesFilterData = useMemo(() => {
```
donni106, line 62: what about naming this `services` like we have `statuses`?

src/screens/SUE/SueListScreen.tsx
```diff
@@ -53,6 +55,22 @@ export const SueListScreen = ({ navigation, route }: Props) => {
       })
   );
 
+  const { data: servicesData } = useQuery([QUERY_TYPES.SUE.SERVICES, {}], () =>
+    getQuery(QUERY_TYPES.SUE.SERVICES)({})
+  );
```
donni106, line 60: we do not need the `{}`s, or am I wrong? is this something "for the future"?

src/screens/SUE/SueListScreen.tsx
ardasnturk: Yeah, that makes sense. 👍

src/screens/SUE/SueListScreen.tsx
ardasnturk: No, nothing for the future. I added an empty object because I don't need `queryVariables` for services. But it can be removed.

donni106 (approved):

PR #1032, opened 2024-02-15
feat: improve inputs on the report screen

ardasnturk:
- updated the `behavior` value in `DefaultKeyboardAvoidingView` to `height` for android to have the same experience on iOS and Android
- wrapped the `Input` component with `forwardRef` since the reference value generated with `useRef` cannot be sent to the `Input` with prop
  - https://react.dev/reference/react/forwardRef
- added `ref` and `onSubmitEditing` props to automatically focus on the next input after pressing the done or submit button from the keyboard after writing to any input on the report screen

SUE-58

## Video

https://github.com/smart-village-solutions/smart-village-app-app/assets/11755668/bd3db55d-a96b-4a9a-923e-bdd2f11fa1c2

### DIFF
--- a/src/components/DefaultKeyboardAvoidingView.tsx
+++ b/src/components/DefaultKeyboardAvoidingView.tsx
@@ -29,7 +29,7 @@ export const DefaultKeyboardAvoidingView = ({ children }: { children: React.Reac
   return (
     <KeyboardAvoidingView
       enabled={device.platform === 'ios'}
-      behavior={device.platform === 'ios' ? 'padding' : 'height'}
+      behavior={device.platform === 'ios' ? 'padding' : undefined}
       keyboardVerticalOffset={
         device.platform === 'ios'
           ? getHeaderHeight(orientation) + statusBarHeight(orientation)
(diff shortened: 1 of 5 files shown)


donni106 (approved):

src/components/DefaultKeyboardAvoidingView.tsx
```diff
@@ -29,7 +29,7 @@ export const DefaultKeyboardAvoidingView = ({ children }: { children: React.Reac
   return (
     <KeyboardAvoidingView
       enabled={device.platform === 'ios'}
-      behavior={device.platform === 'ios' ? 'padding' : undefined}
+      behavior={device.platform === 'ios' ? 'padding' : 'height'}
```
donni106, line 32: does this change behaviors of other forms, not SUE?

ardasnturk: I updated the behaviour to undefined again for Android. I will make 'height' specially for SUE.

https://github.com/smart-village-solutions/smart-village-app-app/pull/1032/commits/85ed935a1d40cdca70e84db2cad132376ad1c7c2

PR #1033, opened 2024-02-16
refactor: update the design of filters

ardasnturk:
- added the `isFilter` prop to the `SueStatus` component since the status component has a special design for the filter and updated the style according to the `isFilter` prop
- added `invert` prop to the reset filters button in the filter component to avoid confusion for users
- updated size of search bar

SUE-14

## Screenshots:

|search bar & status filters|inverted button & selected status filter|
|--|--|
Simulator Screenshot - iPhone 15 Pro Max - 2024-02-16 at 09 54 26|Simulator Screenshot - iPhone 15 Pro Max - 2024-02-16 at 09 54 32

### DIFF
--- a/src/components/SUE/SueStatus.tsx
+++ b/src/components/SUE/SueStatus.tsx
@@ -3,18 +3,21 @@ import React, { useContext } from 'react';
 import { StyleSheet } from 'react-native';
 
 import { SettingsContext } from '../../SettingsProvider';
-import { Icon, normalize } from '../../config';
+import { Icon, colors, device, normalize } from '../../config';
 import { BoldText } from '../Text';
 import { WrapperRow } from '../Wrapper';
 
+/* eslint-disable complexity */
 export const SueStatus = ({
   disabled,
   iconName,
+  isFilter,
   small = false,
   status
 }: {
   disabled?: boolean;
   iconName?: string;
+  isFilter?: boolean;
   small?: boolean;
   status: string;
 }) => {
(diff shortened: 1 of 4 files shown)


src/components/SUE/SueStatus.tsx
```diff
@@ -28,8 +31,22 @@ export const SueStatus = ({
     statusTextColors = {}
   } = sueStatus;
 
-  const backgroundColor = disabled ? statusViewColors?.disabled : statusViewColors?.[status];
-  const textColor = disabled ? statusTextColors?.disabled : statusTextColors?.[status];
+  const backgroundColor =
+    isFilter && disabled
```
donni106, line 35: `disabled` is enough i guess? if it is disabled, it should be grayed out, no matter if is filter or not?

donni106: ah i see, the problem is that the colors are different?
so we would need more configs for colors for `isFilter`. no it is weird, that we have some colors configured and some not.
please create `statusViewColorsFilter` and `statusTextColorsFilter`, right?

src/components/SUE/SueStatus.tsx
```diff
@@ -71,6 +90,12 @@ const styles = StyleSheet.create({
     paddingHorizontal: normalize(14),
     paddingVertical: normalize(4)
   },
+  filterContainer: {
+    borderRadius: normalize(8),
+    borderWidth: normalize(1),
+    height: normalize(32),
+    width: device.width * 0.3
```
donni106, line 97: why * 0.3 here?

donni106: and how does it look like in landscape

src/components/SUE/SueStatus.tsx
ardasnturk: I hadn't thought of that since we don't provide customisation for the filter through the server, but it makes sense. I'm adding the following colour codes to the server.
```
      "statusViewColorsFilter": {
        "enabled": "#0f46181a",
        "disabled": "#FFFFFF"
      },
      "statusTextColorsFilter": {
        "enabled": "#141414",
        "disabled": "#A2A2A2"
      },
```

ardasnturk: https://github.com/smart-village-solutions/smart-village-app-app/pull/1033/commits/f5876c73b93ec7c25fc0a3aae93d663812b4a53f

src/components/SUE/SueStatus.tsx
ardasnturk: It looks like this. It's not perfect, but I think it's ideal for the test version.
image

donni106 (approved):

PR #1034, opened 2024-02-19
feat: start list from saved step

ardasnturk:
- added `currentProgress` to `storeReportValues` to determine which step we are on
- added `scrollTo` to `readReportValuesFromStore` to create the report screen at the saved step

SUE-48

### DIFF
--- a/src/screens/SUE/SueReportScreen.tsx
+++ b/src/screens/SUE/SueReportScreen.tsx
@@ -116,6 +116,7 @@ export const SueReportScreen = ({
   const [storedValues, setStoredValues] = useState<TReports>();
 
   const scrollViewRef = useRef(null);
+  const scrollViewContentRef = useRef(null);
 
   const {
     control,
@@ -148,6 +149,12 @@ export const SueReportScreen = ({
     storeReportValues();
 
     if (!sueReportData.termsOfService) {
+      scrollViewContentRef.current?.scrollTo({
+        x: 0,
+        y: device.height,
+        animated: true
+      });
+
       return Alert.alert(texts.sue.report.alerts.hint, texts.sue.report.alerts.termsOfService);
     }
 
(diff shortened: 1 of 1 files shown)


src/screens/SUE/SueReportScreen.tsx
```diff
@@ -219,14 +219,25 @@ export const SueReportScreen = ({
...
     const storedValues = await readFromStore(SUE_REPORT_VALUES);
 
     if (storedValues) {
       setStoredValues(storedValues);
+      setCurrentProgress(storedValues.currentProgress);
+      scrollViewRef?.current?.scrollTo({
+        x: device.width * storedValues.currentProgress,
+        y: 0,
+        animated: true
+      });
```
donni106, line 240: somehow this does not work efficiently, here two cases, where i start at one step less then i left before:

|||
|---|---|
|https://github.com/smart-village-solutions/smart-village-app-app/assets/1942953/0556ad18-3a85-4867-891c-83581c3fc7dc|https://github.com/smart-village-solutions/smart-village-app-app/assets/1942953/cfb1c770-0cef-4637-b7e2-3a17ecfa9176|

ardasnturk: I have searched for the problem, but I can't solve it exactly. I can describe the problem as follows:

Open application -> Go to ReportScreen -> Select category -> Click on Weiter button (data is saved) -> Go back to HomeScreen -> Go to ReportScreen again -> ReportScreen starts from step 2

There is no problem here. ScrollTo works correctly. But after that currentProgress is not saved correctly in AsyncStorage.

Write title -> Press Weiter button (data is saved) -> Return to HomeScreen -> Go back to ReportScreen -> ReportScreen starts from step 2 (expected to start from step 3)

currentProgress cannot be correctly saved to AsyncStorage. I am investigating why this might be happening.

donni106: we need to invest more tim ein this problem. for now, please remove it in this part of the project and keep it in a separate branch for later.
for v1 we will always start again from step 1

PR #1037, opened 2024-02-20
chore(sue): upgrade react-native-maps

d4v3d4vE:
- upgraded react-native maps from 1.7.1 to 1.10.3

SUE-45

For Testing:
- Node Version must be 18 or higher

### DIFF
--- a/package.json
+++ b/package.json
@@ -108,7 +108,7 @@
     "react-native-gesture-handler": "~2.12.0",
     "react-native-get-random-values": "~1.9.0",
     "react-native-gifted-chat": "^2.0.1",
-    "react-native-maps": "1.7.1",
+    "react-native-maps": "1.11.3",
     "react-native-markdown-display": "^7.0.0-alpha.2",
     "react-native-modal-dropdown": "https://github.com/donni106/react-native-modal-dropdown#feature/search-options",
     "react-native-progress": "^5.0.0",
--- a/yarn.lock
+++ b/yarn.lock
@@ -2633,10 +2633,10 @@
   dependencies:
     "@babel/types" "^7.20.7"
 
-"@types/geojson@^7946.0.10":
-  version "7946.0.10"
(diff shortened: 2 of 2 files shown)


donni106 (approved):

package.json
```diff
@@ -106,7 +106,7 @@
     "react-native-gesture-handler": "~2.12.0",
     "react-native-get-random-values": "~1.9.0",
     "react-native-gifted-chat": "^2.0.1",
-    "react-native-maps": "1.7.1",
+    "react-native-maps": "^1.10.3",
```
ardasnturk, line 109: I'm getting this error. What is the node version on your device?  We are currently using version 16.13.0 for the project, but it seems that the `react-native-maps` package node needs a version higher than 18. So I can't test it. 

image

ardasnturk (commented):
I think this branch should be merged with `epic/sue` instead of `master`.

donni106 (commented):
good points @ardasnturk thanks!
yes please change the base branch and write something about the node version in the PR description for reviewers @d4v3d4vE.

package.json
ardasnturk: if node needs to be updated, I think `.nvmrc` needs to be updated as well. As far as I can see, we have version `16.13.0` at the moment.

Which version of Node should we update to? Is `18.0.0` enough? Or do we want to use the latest version? As far as I know, the most current and stable version is `18.19.1`

donni106: @d4v3d4vE can you answer here please.

d4v3d4vE: I would suggest du use the latest stable version, I'm on node `v22.0.0` and have no issues. The current LTS is `v20.12.2`

donni106: [Expo says](https://docs.expo.dev/get-started/installation/):
> [Node.js LTS release](https://nodejs.org/en/) - Only Node.js LTS releases (even-numbered) are recommended.

donni106:
We will include this with the upcoming Expo SDK 50 upgrade.

PR #1043, opened 2024-02-27
feat: show inputs above the keyboard

ardasnturk:
- added the `headerHelper` to `DefaultKeyboardAvoidingView`, which is used in many parts of the code to avoid code duplication
- using `statusBarHeight` of `expo-constants` package because `react-native-status-bar-height` package cannot calculate status bar size of iPhone 14 Pro and higher models
  - https://github.com/ovr/react-native-status-bar-height
- removed `react-native-status-bar-height` package from `package.json` because it is not used
- added `getKeyboardHeight` helper to detect the size of the keyboard when it is opened
- added a `View` half the size of the keyboard, which is created only when the keyboard is opened, into `ScrollView` in order to make the inputs hidden behind the keyboard visible by scrolling the screen on android devices
- wrapped all content with `DefaultKeyboardAvoidingView` to make the buttons at the bottom of the screen appear above the keyboard when the keyboard is opened
  - this feature does not currently work for android
- added `Keyboard.dismiss()` to dismiss the keyboard if the continue or back button is pressed while the keyboard is open

SUE-58

please remember to run `yarn prebuild` as a package has been deleted.

## Screenshots:

iOS:
|SueReportDescription screen with button (keyboard active)|SueReportLocation screen with button (keyboard active) - focus on zip code|SueReportLocation Screen (keyboard deactive)|
|--|--|--|
Simulator Screenshot - iPhone 15 Pro Max - 2024-02-27 at 15 52 41|Simulator Screenshot - iPhone 15 Pro Max - 2024-02-27 at 15 52 49|Simulator Screenshot - iPhone 15 Pro Max - 2024-02-27 at 15 52 51

Android:
|SueReportLocation (keyboard deactive) - bottom of scroll view |SueReportLocation (keyboard active) - focus on city / bottom of scroll view|SueReportLocation (keyboard deactive) - view when the city input is clicked|landscape mode|
|--|--|--|--| 
Screenshot_20240227-155333|Screenshot_20240227-155337|Screenshot_20240227-155343|Screenshot_20240227-155442

### DIFF
--- a/src/hooks/keyboardHeight.ts
+++ b/src/hooks/keyboardHeight.ts
@@ -1,0 +1,22 @@
+import { useEffect, useState } from 'react';
+import { Keyboard, KeyboardEvent } from 'react-native';
+
+export const useKeyboardHeight = () => {
+  const [keyboardHeight, setKeyboardHeight] = useState(0);
+
+  useEffect(() => {
+    const onKeyboardChange = ({ endCoordinates }: KeyboardEvent) => {
+      setKeyboardHeight(endCoordinates?.height || 0);
+    };
+
+    const showSubscription = Keyboard.addListener('keyboardDidShow', onKeyboardChange);
+    const hideSubscription = Keyboard.addListener('keyboardDidHide', onKeyboardChange);
+
+    return () => {
+      showSubscription.remove();
+      hideSubscription.remove();
+    };
+  }, []);
+
+  return keyboardHeight;
+};
--- a/package.json
+++ b/package.json
@@ -117,7 +117,6 @@
     "react-native-safe-area-context": "4.6.3",
     "react-native-screens": "~3.22.0",
     "react-native-snap-carousel": "4.0.0-beta.6",
-    "react-native-status-bar-height": "^2.6.0",
     "react-native-svg": "13.9.0",
     "react-native-tab-view": "^3.5.2",
     "react-native-web": "~0.19.6",
--- a/src/components/DefaultKeyboardAvoidingView.tsx
+++ b/src/components/DefaultKeyboardAvoidingView.tsx
@@ -1,27 +1,9 @@
-import React from 'react';
-import { useContext } from 'react';
-import { KeyboardAvoidingView, Platform, PlatformIOSStatic, StyleSheet } from 'react-native';
-import { getStatusBarHeight } from 'react-native-status-bar-height';
+import React, { useContext } from 'react';
+import { KeyboardAvoidingView, StyleSheet } from 'react-native';
 
-import { device } from '../config';
 import { OrientationContext } from '../OrientationProvider';
-
-const getHeaderHeight = (orientation: string) =>
-  // Android always 56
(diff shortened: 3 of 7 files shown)


src/helpers/keyboardHelper.ts
```diff
@@ -0,0 +1,25 @@
+import { useEffect, useState } from 'react';
+import { Keyboard, KeyboardEvent } from 'react-native';
+
+export const getKeyboardHeight = () => {
```
donni106, line 4: this seems to be a hook and should be placed in hooks and should be named `useKey...`

ardasnturk: Yes, you are right. I've updated.

https://github.com/smart-village-solutions/smart-village-app-app/commit/e1405deffc70311423aee4034dfc441aee5c4eba

src/hooks/keyboardHeight.ts
```diff
@@ -0,0 +1,25 @@
+import { useEffect, useState } from 'react';
+import { Keyboard, KeyboardEvent } from 'react-native';
+
+export const useKeyboardHeight = () => {
+  const [keyboardHeight, setKeyboardHeight] = useState(0);
+
+  useEffect(() => {
+    function onKeyboardDidShow(e: KeyboardEvent) {
```
d4v3d4vE, line 8: what is `(e: KeyboardEvent)` here for and what does it do?

d4v3d4vE (approved):
Looks good!

src/hooks/keyboardHeight.ts
ardasnturk: The `(e: KeyboardEvent)` in the `onKeyboardDidShow` function parameter represents a Keyboard Event object in TypeScript. The `KeyboardEvent` is a type provided by the `react-native` library. It contains information about the keyboard, such as its coordinates and dimensions when it is shown. In this specific context, `e.endCoordinates.height` is used to extract the height of the keyboard from the event object. This height value is then set as the new keyboard `height`, allowing the application to respond dynamically to changes in the keyboard's visibility. The use of TypeScript helps ensure that the properties and methods of the e object are correctly accessed and prevents potential type-related errors during development.

d4v3d4vE:
@ardasnturk after some testing I found a issue that I cant open the keyboard on iPhone (15 Pro iOS 17.0) while it opens itself on Android without issues

d4v3d4vE:
> @ardasnturk after some testing I found a issue that I cant open the keyboard on my iPhone (15 Pro iOS 17.0) simulator while it opens itself on Android without issues

Same on iPhone (14 Pro iOS 16.4)

ardasnturk:
> @ardasnturk after some testing I found a issue that I cant open the keyboard on my iPhone (15 Pro iOS 17.0) simulator while it opens itself on Android without issues

Can you please try again? Because the test I ran didn't show any problems.
Simulator Screenshot - iPhone 15 Pro Max - 2024-02-28 at 15 51 07

PR #1044, opened 2024-02-28
feat: extend the waiting time when sending a report

ardasnturk:
- increased the waiting time of the report from 1.5 seconds to 3 seconds if there is no image
- added a control to avoid waiting an extra 3 seconds because images take a long time to load
- added width value of 100% to prevent the loading gif from being clipped

SUE-38

please remember to set the version in `app.json` to `60.0.0` during testing.

### DIFF
--- a/src/screens/SUE/SueReportScreen.tsx
+++ b/src/screens/SUE/SueReportScreen.tsx
@@ -172,11 +172,14 @@ export const SueReportScreen = ({
     setIsLoading(true);
     mutateAsync(formData)
       .then(() => {
-        setTimeout(() => {
-          setIsDone(true);
-          resetStoredValues();
-          setIsLoading(false);
-        }, 1500);
+        setTimeout(
+          () => {
+            setIsDone(true);
+            resetStoredValues();
+            setIsLoading(false);
+          },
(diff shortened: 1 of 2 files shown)


src/screens/SUE/SueReportScreen.tsx
```diff
@@ -172,11 +172,19 @@ export const SueReportScreen = ({
...
+
+          return;
+        }
+
         setTimeout(() => {
           setIsDone(true);
           resetStoredValues();
           setIsLoading(false);
-        }, 1500);
+        }, 3000);
```
donni106, line 187: what about avoiding the duplicate content of methods called when you add a condition of the time based on existing images? 3000 if no images and 0 if images, something like that?

ardasnturk: Definitely makes sense. I already found a bug in the code. If the image was not selected, I had to do `JSON.parse` first because `sueReport?.images?.length` was `'[]'` by default. I added this to the code and updated it.

https://github.com/smart-village-solutions/smart-village-app-app/commit/ce923a8b752c029c400c3a80046d55a917338232

PR #1048, opened 2024-03-04
fix: add error check to fix the problem that the report is not sent to api

ardasnturk:
- updated `mutateAsync` structure to `onError` and `onSuccess` to better analyse error reports from api
- added `status` check to `onSuccess` section because most of the errors in api are not displayed in `onError` section
- added the `termsOfService` control we created inside `onSubmit` to the `alertTextGeneratorForMissingData` function to make the code more understandable
- added control of values in `postRequests` to not send unfilled content in the form to api

SUE-63

### DIFF
--- a/src/queries/SUE/requests.tsx
+++ b/src/queries/SUE/requests.tsx
@@ -24,19 +24,20 @@ export const requests = async (queryVariables) => {
   });
 };
 
+/* eslint-disable complexity */
 export const postRequests = async (data: any) => {
   const formData = new FormData();
-  formData.append('address_string', data?.addressString);
-  formData.append('description', data?.description);
-  formData.append('email', data?.email);
-  formData.append('first_name', data?.firstName);
+  data?.addressString && formData.append('address_string', data.addressString);
+  data?.description && formData.append('description', data.description);
+  data?.email && formData.append('email', data.email);
+  data?.firstName && formData.append('first_name', data.firstName);
   formData.append('jurisdiction_id', jurisdictionId);
-  formData.append('last_name', data?.lastName);
-  formData.append('lat', data?.lat);
-  formData.append('long', data?.long);
-  formData.append('phone', data?.phone);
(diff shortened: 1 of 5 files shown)


d4v3d4vE (approved):

src/queries/SUE/requests.tsx
```diff
@@ -24,19 +24,20 @@ export const requests = async (queryVariables) => {
...
-  formData.append('long', data?.long);
-  formData.append('phone', data?.phone);
-  formData.append('service_code', data.serviceCode);
-  formData.append('title', data?.title);
+  data?.lastName && formData.append('last_name', data.lastName);
+  data?.lat && formData.append('lat', data.lat);
+  data?.long && formData.append('long', data.long);
+  data?.phone && formData.append('phone', data.phone);
+  data?.serviceCode && formData.append('service_code', data.serviceCode);
+  data?.title && formData.append('title', data.title);
```
donni106, line 40: can `data` be undefined here? isn't there always a `data` object passed in this method?

ardasnturk: In our previous control, we added a control against data being undefined. For example:
```
formData.append('title', data?.title);
```
 I used this control in the same way. After the changes made in the API, it started to not accept undefined data. So I decided to remove it from the form if there is no data.

donni106: ok, this way each line checks data. if you want to ensure that data is present what about a `if (data)` block?

ardasnturk: Yes, this would also make sense and would greatly reduce system requirements with a single control. 👍

donni106: @d4v3d4vE can you take over on this refactoring please?

donni106: done with a new commit, that will be cherry picked in master soon.

PR #1049, opened 2024-03-04
feat: add auto-fill inputs feature

ardasnturk:
- added the ability to automatically add the information saved in the iOS operating system to the inputs on the keyboard suggestions

SUE-65

## Screenshots:

I am not sharing screenshots as they contain personal information, but for the test go to the report page and click on the required inputs. If you have information saved in the ios system, it will automatically appear.


d4v3d4vE (approved):
Nice changes!

src/components/SUE/report/SueReportLocation.tsx
```diff
@@ -174,6 +175,7 @@ export const SueReportLocation = ({
           name="houseNumber"
           label={texts.sue.report.houseNumber}
           placeholder={texts.sue.report.houseNumber}
+          textContentType="off"
```
donni106, line 178: why was it necessary to set this to `off`?

src/components/SUE/report/SueReportLocation.tsx
```diff
@@ -162,6 +162,7 @@ export const SueReportLocation = ({
           name="street"
           label={texts.sue.report.street}
           placeholder={texts.sue.report.street}
+          textContentType="streetAddressLine1"
```
donni106, line 165: are there also `streetAddressLine2`, `2streetAddressLine3` and other? what does they stand for?

src/components/SUE/report/SueReportLocation.tsx
ardasnturk: Because here we do not get a correct suggestion from the iOS system. Therefore, I have closed the suggestion.

src/components/SUE/report/SueReportLocation.tsx
ardasnturk: There are only streetAddressLine1 and streetAddressLine2. In the iOS system, it depends on what information the user saves. For us, streetAddressLine1 was necessary. This way we could get the most accurate data.

It can be examined in more detail from the document here.: https://reactnative.dev/docs/textinput#textcontenttype-ios

PR #1050, opened 2024-03-04
feat: add style for landscape mode to photos

ardasnturk:
- added width style for the list to cover the entire screen when using the app in landscape mode while in list view

SUE-67

## Screenshots:

currently there is a problem on the API side, so the data is not being loaded. Therefore, I cannot upload before and after screenshots here. To test, switch to list view in the app and turn the phone sideways.

### DIFF
--- a/src/components/CardListItem.js
+++ b/src/components/CardListItem.js
@@ -156,14 +156,16 @@ const styles = StyleSheet.create({
     paddingTop: 0
   },
   sueContentContainer: {
-    borderWidth: 1,
     borderColor: colors.gray20,
-    borderRadius: normalize(8)
+    borderRadius: normalize(8),
+    borderWidth: 1,
+    width: '100%'
   },
   sueImageContainer: {
     alignSelf: 'auto',
     borderTopLeftRadius: normalize(8),
-    borderTopRightRadius: normalize(8)
+    borderTopRightRadius: normalize(8),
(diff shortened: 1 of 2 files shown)


d4v3d4vE (approved):
generally works great, shouldn't it also be applied to the images in the "news list component"?
Simulator Screenshot - iPhone 15 Pro - 2024-03-05 at 17 37 10

ardasnturk:
> generally works great, shouldn't it also be applied to the images in the "news list component"?

Please test for the correct branch. This PR is for SUE development, not for MQGB.

src/components/CardListItem.js
```diff
@@ -156,14 +156,16 @@ const styles = StyleSheet.create({
...
+    borderRadius: normalize(8),
+    borderWidth: 1,
+    width: '100%'
   },
   sueImageContainer: {
     alignSelf: 'auto',
     borderTopLeftRadius: normalize(8),
-    borderTopRightRadius: normalize(8)
+    borderTopRightRadius: normalize(8),
+    width: '100%'
```
donni106, line 168: doesn't this affect all apps, not only sue?

ardasnturk: Yes, it does. But I think this should already be for landscape mode, otherwise the pictures cannot be displayed correctly.

PR #1053, opened 2024-03-08
feat: disable orientation for the report screen

ardasnturk:
- added `PORTRAIT_UP` lock to `SueReportScreen` because calculating the sizes of different steps on the report screen and scrolling at the same rate is not possible on different sized devices
- added listener to automatically set the orientation of the screen to `DEFAULT` when the user leaves `SueReportScreen`
- updated the style so that the pictures in the `ImagesCarousel` we use in `HomeScreen` can also adjust themselves according to orientation

SUE-67

please set the version in app.json to 60.0.0 for testing.
## Screenshots:

|HomeScreen before|HomeScreen after|
|--|--|
Simulator Screenshot - iPhone 15 Pro Max - 2024-03-08 at 09 50 46|Simulator Screenshot - iPhone 15 Pro Max - 2024-03-08 at 09 50 25

### DIFF
--- a/src/components/ImagesCarousel.js
+++ b/src/components/ImagesCarousel.js
@@ -190,7 +190,9 @@ const styles = StyleSheet.create({
     alignSelf: 'center'
   },
   imageContainer: {
-    alignSelf: 'center'
+    alignItems: 'center',
+    alignSelf: 'center',
+    width: '100%'
   },
   pauseButton: {
     alignItems: 'center',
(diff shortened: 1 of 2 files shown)


d4v3d4vE (approved):
works great!

src/components/ImagesCarousel.js
```diff
@@ -190,7 +190,9 @@ const styles = StyleSheet.create({
     alignSelf: 'center'
   },
   imageContainer: {
-    alignSelf: 'center'
+    alignItems: 'center',
+    alignSelf: 'center',
+    width: '100%'
```
donni106, line 195: does this work for sliders in landscape with more than one image? there was always one centered image and the others images of a slider as "preview" on the left or right.

ardasnturk: Unfortunately I haven't thought about more than one picture. 😔 We'll see about that when we merge the `epic/SUE` branch with the master.

PR #1055, opened 2024-03-08
feat: add city and postcode limits to create a report

ardasnturk:
- added `city` and `zipCodes` limitations that can be updated via main-server when creating a report
- added `errorMessage` which can be set via main-server as there may be different alert messages for different cities

SUE-57

please set the version in app.json to 60.0.0 for testing.

The config added to main-server is as follows. For reports that do not match the postal codes or city in this config, the next step is prevented.
the limitation can be done only with the `zipCodes` array or only with the `city`.

If the `errorMessage` value is not added to the configuration, the following message is displayed by default.
`Leider kann die Sag's uns Einfach nur Standorte der  nutzen.

```
    "limitOfArea": {
      "zipCodes": [
        "85625"
      ],
      "city": "kiel",
      "errorMessage": "Leider kann die Sag's uns Einfach nur Standorte der Landeshaupstadt Kiel nutzen."
    }
```

## Screenshots:

|alert with false city and zipCode value|alert with false city value|alert with false zipCode value|with main-server config|
|--|--|--|--|
Simulator Screenshot - iPhone 15 Pro Max - 2024-03-08 at 10 47 16|Simulator Screenshot - iPhone 15 Pro Max - 2024-03-08 at 10 47 22|Simulator Screenshot - iPhone 15 Pro Max - 2024-03-08 at 10 49 09|Simulator Screenshot - iPhone 15 Pro Max - 2024-03-08 at 10 47 38

### DIFF
--- a/src/config/texts.js
+++ b/src/config/texts.js
@@ -959,6 +959,7 @@ export const texts = {
         limitOfArea: (city) =>
           `Leider kann die Sag's uns Einfach nur Standorte der ${city} nutzen.`,
         location: 'Bitte wählen Sie einen Ort auf der Karte aus.',
+        missingAnyInput: 'Bitte füllen Sie alle Pflichtfelder aus',
         no: 'Nein',
         myLocation: 'Möchten Sie Ihren aktuellen Standort verwenden?',
         serviceCode: 'Bitte wählen Sie aus, um welches Thema es in dem Bericht geht.',
--- a/src/config/texts.js
+++ b/src/config/texts.js
@@ -956,6 +956,8 @@ export const texts = {
           title: 'Bildquelle auswählen'
         },
         invalidMail: 'Die eingegebene E-Mail-Adresse ist nicht gültig.',
+        limitOfArea: (city) =>
+          `Leider kann die Sag's uns Einfach nur Standorte der ${city} nutzen.`,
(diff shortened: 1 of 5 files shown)


src/config/texts.js
```diff
@@ -956,7 +956,10 @@ export const texts = {
           title: 'Bildquelle auswählen'
         },
         invalidMail: 'Die eingegebene E-Mail-Adresse ist nicht gültig.',
+        limitOfArea: (city) =>
+          `Leider kann die Sag's uns Einfach nur Standorte der ${city} nutzen.`,
```
donni106, line 960: this text shouldn't be hardcoded to `Sag's uns Einfach`

donni106: no, needs to be discussed with others. if it is always the app name, we can use the config from app.json?

ardasnturk: Yes, we can definitely use it. That's what I thought. But I'll put it in my notes to ask at the next meeting.

ardasnturk: We talked about it and decided to leave it as it is. Because this text can actually be set on the mainserver.

donni106: Ok, i was not aware about the possibility to configure this text.

PR #1056, opened 2024-03-08
feat: set the requirement of inputs from the server

ardasnturk:
- updated `sueReportProgress` content to define required inputs via server
- added * next to the required inputs according to the data in `requiredInputs`
- added `isAnyInputMissing` control to `alertTextGeneratorForMissingData` function to prevent going to the next step without filling the required fields in each step

SUE-57

if you want to update the required inputs via the server, please go to the `sueReportProgress` content and add the `requiredInputs` array to each step. inside the `requiredInputs` array, type the names that match the names of the inputs in that step

For example:
```
[
  ...
  {
    "title": "Worum geht es?",
    "subtitle": "Nicht einmal von der allmächtigen Interpunktion werden die Blindtexte beherrscht.",
    "content": "description",
    "requiredInputs": [
      "title",
      "description"
    ]
  },
  ...
]
```


src/components/SUE/report/SueReportLocation.tsx
```diff
@@ -173,7 +176,9 @@ export const SueReportLocation = ({
       <Wrapper style={styles.noPaddingTop}>
         <Input
           name="houseNumber"
-          label={texts.sue.report.houseNumber}
+          label={`${texts.sue.report.houseNumber} ${
+            requiredInputs?.includes('houseNumber') ? '*' : ''
```
d4v3d4vE, line 180: what does this stand for:` ? '*' : ''`

ardasnturk: If `requiredInputs` includes the `houseNumber` value, we add `*` to the label. Because this field must be filled. If it is not required, we leave it blank.

PR #1058, opened 2024-03-14
feat: add update check

ardasnturk:
- added `react-native-version-check-expo` package to check if there is an updated app in the store
- added `useVersionCheck` hook to check for version updates
- added `bundleIdentifier` and `packageName` to make adjustments according to platform in `versionCheck` hook
- added `getLatestVersion` function from `VersionCheck` package to `versionCheck` hook to get version in store
- added `VersionCheck.getCurrentVersion()` function to get the version of the app on the device
- added alert to inform the user if there is a more updated version in the store
- added direct redirection to store urls separated by platform to alert
- added `versionCheckEnabled` control to settings section in `globalSettings` to use version check feature

SUE-77

## Test:

To test the feature, follow the procedure below:

- first select the branch and run the `yarn prebuild` command from the terminal
- then select an app from the release branch that is currently active in the app store and play store 
- replace the `bundleIdentifier` and `packageName` in the `versionCheck` hook with the values in the `app.json` of the app you selected from the release branch
For example:
```
const bundleIdentifier = appJson.expo.ios.bundleIdentifier; -> 'de.xx.x.smartvillage.app'
const packageName = appJson.expo.android.package; ->  -> 'de.xx.x.smartvillage.app'
```
- then make adjustments to get the `currentVersion` value from `appJson`, not from the store. The version of the app running in the simulator is always higher than the version in the store. That's why we won't see the alert. Now update the `currentVersion` as follows
```
const currentVersion = await VersionCheck.getCurrentVersion(); -> appJson.expo.version;
```
- Now we can run version control. go to app.json and change the version to a lower version than the one currently in the store.

For example, if the appStore version is 3.3.2, change the version in `app.json` to 3.3.1. Then the alert should appear. Please click on the link and check if you are redirected to the correct store page.

image

Also make sure that `versionCheckEnabled` in `globalSettings` is `true`. Please set it to `false` and run tests. This way we should not see the alert.

## Screenshots:

|alert|
|--|
Simulator Screenshot - iPhone 15 Pro Max - 2024-03-14 at 11 36 00


d4v3d4vE (approved):
Really nice addition and works great 🎉

PR #1059, opened 2024-03-14
feat(sue): list previous reports on the report screen

ardasnturk:
- added `query` to `SueReportLocation` to see open and edited reports in the map view shown in the location step on the report screen
- added `mapMarkers` to convert the data coming from the API via query into the correct format for the map
- exported from `SueMapScreen` to use the `mapToMarkers` function in `SueReportLocation`
- added location information created with `mapMarkers` to `locations` array
- updated query's `cacheTime` to 24 hours
- added `SueStatus` `enum` to see statuses in one place
- added `start_date` to `queryVariables` to get all data
- updated the query in `parseListItemsFromQuery` with `REQUESTS_WITH_SERVICE_REQUEST_ID` to keep the code in the same structure

SUE-75

## Test:

- To test, upgrade the version in `app.json` to `60.0.0`,
- Click the Jetzt Meldung machen button on the `HomeScreen`,
- Fill in the required information in the `ReportScreen` and proceed to step 3

## Screenshots:

|before (without selecting the location)|after (without selecting the location)|before (with selecting the location)|after (with selecting the location)|
|--|--|--|--|
Simulator Screenshot - iPhone 15 Pro Max - 2024-03-14 at 13 26 46|Simulator Screenshot - iPhone 15 Pro Max - 2024-03-14 at 13 26 26|Simulator Screenshot - iPhone 15 Pro Max - 2024-03-14 at 13 26 49|Simulator Screenshot - iPhone 15 Pro Max - 2024-03-14 at 13 26 30

### DIFF
--- a/src/components/SUE/report/SueReportLocation.tsx
+++ b/src/components/SUE/report/SueReportLocation.tsx
@@ -25,11 +25,9 @@ import { Map } from '../../map';
 import { getLocationMarker } from '../../settings';
 
 enum SueStatus {
-  CLOSED = 'TICKET_STATUS_CLOSED',
   IN_PROCESS = 'TICKET_STATUS_IN_PROCESS',
   INVALID = 'TICKET_STATUS_INVALID',
   OPEN = 'TICKET_STATUS_OPEN',
-  UNPROCESSED = 'TICKET_STATUS_UNPROCESSED',
   WAIT_REQUESTOR = 'TICKET_STATUS_WAIT_REQUESTOR',
   WAIT_THIRDPARTY = 'TICKET_STATUS_WAIT_THIRDPARTY'
 }
@@ -71,7 +69,7 @@ export const SueReportLocation = ({
 
   const queryVariables = {
     start_date: '1900-01-01T00:00:00+01:00',
-    status: `${SueStatus.UNPROCESSED},${SueStatus.OPEN},${SueStatus.IN_PROCESS},${SueStatus.WAIT_REQUESTOR},${SueStatus.WAIT_THIRDPARTY}`
+    status: Object.values(SueStatus).map((status) => status)
   };
 
   const { data, isLoading } = useQuery(
--- a/src/components/SUE/report/SueReportLocation.tsx
+++ b/src/components/SUE/report/SueReportLocation.tsx
@@ -23,6 +23,16 @@ import { Input } from '../../form';
 import { Map } from '../../map';
 import { getLocationMarker } from '../../settings';
 
+enum SueStatus {
+  CLOSED = 'TICKET_STATUS_CLOSED',
+  IN_PROCESS = 'TICKET_STATUS_IN_PROCESS',
+  INVALID = 'TICKET_STATUS_INVALID',
+  OPEN = 'TICKET_STATUS_OPEN',
+  UNPROCESSED = 'TICKET_STATUS_UNPROCESSED',
+  WAIT_REQUESTOR = 'TICKET_STATUS_WAIT_REQUESTOR',
+  WAIT_THIRDPARTY = 'TICKET_STATUS_WAIT_THIRDPARTY'
+}
+
 /* eslint-disable complexity */
 export const SueReportLocation = ({
   control,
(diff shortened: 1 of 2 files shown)


d4v3d4vE (approved):

src/components/SUE/report/SueReportLocation.tsx
```diff
@@ -48,6 +57,27 @@ export const SueReportLocation = ({
   const zipCodeInputRef = useRef();
   const cityInputRef = useRef();
 
+  const queryVariables = {
+    status:
+      'TICKET_STATUS_UNPROCESSED,TICKET_STATUS_OPEN,TICKET_STATUS_IN_PROCESS,TICKET_STATUS_WAIT_REQUESTOR,TICKET_STATUS_WAIT_THIRDPARTY'
+  };
```
donni106, line 63: doesn't we have some config for these? didn't we need this statuses in the normal map as well?

donni106: and what about `start_date: '1900-01-01T00:00:00+01:00'` here? not needed?

src/components/SUE/report/SueReportLocation.tsx
```diff
@@ -48,6 +57,27 @@ export const SueReportLocation = ({
...
+
+  const { data, isLoading } = useQuery([QUERY_TYPES.SUE.REQUESTS, queryVariables], () =>
+    getQuery(QUERY_TYPES.SUE.REQUESTS)(queryVariables)
+  );
+
+  const mapMarkers = useMemo(
+    () =>
+      mapToMapMarkers(
+        parseListItemsFromQuery(QUERY_TYPES.SUE.REQUESTS, data, undefined, {
```
donni106, line 72: why not `QUERY_TYPES.SUE.REQUESTS_WITH_SERVICE_REQUEST_ID` as for the markers on the map?

src/components/SUE/report/SueReportLocation.tsx
```diff
@@ -48,6 +57,27 @@ export const SueReportLocation = ({
...
   const cityInputRef = useRef();
 
+  const queryVariables = {
+    status:
+      'TICKET_STATUS_UNPROCESSED,TICKET_STATUS_OPEN,TICKET_STATUS_IN_PROCESS,TICKET_STATUS_WAIT_REQUESTOR,TICKET_STATUS_WAIT_THIRDPARTY'
+  };
+
+  const { data, isLoading } = useQuery([QUERY_TYPES.SUE.REQUESTS, queryVariables], () =>
+    getQuery(QUERY_TYPES.SUE.REQUESTS)(queryVariables)
+  );
```
donni106, line 67: how much does it affect performance getting all the data for markers on the screen?

src/components/SUE/report/SueReportLocation.tsx
ardasnturk: No. Unfortunately, we use the `appDesignSystem` on the main-server for this. Here the status types are defined. But adding it to the app. makes sense!
---
Good point! Yes, you're right. I should add `start_date`. Otherwise we can't access all the data.

src/components/SUE/report/SueReportLocation.tsx
ardasnturk: If I use `QUERY_TYPES.SUE.REQUESTS_WITH_SERVICE_REQUEST_ID`, only the marker for the report with a specific ID will be shown. We need to show markers for all data.

src/components/SUE/report/SueReportLocation.tsx
ardasnturk: That's hard to answer. The more data, the longer it will take. But since we add filtering according to certain statuses, we only get the filtered data from the API instead of the whole data. But of course there will be a time difference between 10 data and 1000 data. 🤷‍♂️

src/components/SUE/report/SueReportLocation.tsx
ardasnturk: Ow, I'm sorry. I misunderstood. It's just a parser. They both parse the same data. We can use the `REQUESTS_WITH_SERVICE_REQUEST_ID` query to parse.

src/components/SUE/report/SueReportLocation.tsx
donni106: So it is crucial to rely on react-queries cache here requesting the data. The cache time should have a certain time, not just five minutes, in order to not go through long request durations and waiting times in order to fill out the form.

src/components/SUE/report/SueReportLocation.tsx
donni106: Ok I thought of that, but wasn't sure. Yes, let's have it in the same way in both places.

src/components/SUE/report/SueReportLocation.tsx
ardasnturk: https://github.com/smart-village-solutions/smart-village-app-app/pull/1059/commits/6b58b969cec05a8eaa751b50e57bd9f033648eae

src/components/SUE/report/SueReportLocation.tsx
```diff
@@ -48,6 +67,31 @@ export const SueReportLocation = ({
   const zipCodeInputRef = useRef();
   const cityInputRef = useRef();
 
+  const queryVariables = {
+    start_date: '1900-01-01T00:00:00+01:00',
+    status: `${SueStatus.UNPROCESSED},${SueStatus.OPEN},${SueStatus.IN_PROCESS},${SueStatus.WAIT_REQUESTOR},${SueStatus.WAIT_THIRDPARTY}`
```
donni106, line 72: wouldn't it make more sense to have an object and iterate over it to place all the values here? it is not very practicable to create a string and manually put every key of the object here. how is this beneficial?

src/components/SUE/report/SueReportLocation.tsx
```diff
@@ -48,6 +67,31 @@ export const SueReportLocation = ({
...
+  };
+
+  const { data, isLoading } = useQuery(
+    [QUERY_TYPES.SUE.REQUESTS, queryVariables],
+    () => getQuery(QUERY_TYPES.SUE.REQUESTS)(queryVariables),
+    {
+      cacheTime: 1000 * 60 * 60 * 24 // 24 hours
```
donni106, line 79: can 24 hours be too long? what about 12 hours? or "until the end of the day" calculation?

did you try to send a report, make it visible per backend and create a new report? does the previous report be marked on the map? or is the cache object loaded without the previous entry?

ardasnturk: I think it would make more sense to do it by the end of the day. 👍 I'll update the code for this.

ardasnturk: I also tested. If a new report is shared, the newly created report is not displayed in `SueReportLocation`. So it works with cache. But if I close and open the application, I can see it.

https://github.com/smart-village-solutions/smart-village-app-app/pull/1059/commits/dd5effa061a261ed78b21ea6fbcb754f2f024336

src/components/SUE/report/SueReportLocation.tsx
ardasnturk: Yes, I think it makes more sense. It was not right to write with text. I updated the code with map function. 

https://github.com/smart-village-solutions/smart-village-app-app/pull/1059/commits/5d5f56afca33f876b5e8509dd22ab4bf64a81014

donni106 (approved):

PR #1062, opened 2024-03-25
feat(sue): update api key via server

ardasnturk:
- added `fetchSueEndpoints` function to `sueHelper` to update `apiKey` via `globalSettings`
- integrated `fetchSueEndpoints` into queries
- updated `SueReportScreen` because the `limitOfArea` object has been moved into the `sue` object

SUE-82

## For Test:

create a new object named `sue` inside the `settings` object in `globalSettings` for testing
create an object called `apiConfig` inside the `sue`
```
"settings": {
    "sue": {
      "apiConfig": {},
    }
  },
```

the `apiConfig` object may look as follows
```
"apiConfig": {
  "serverUrl": "xxx",
  "apiKey": "xxx",
  "jurisdictionId": "1"
}
```
```
"apiConfig": {
  "whichApi": "testApi",
  "testApi": {
    "serverUrl": "xxx",
    "apiKey": "xxx",
    "jurisdictionId": "2"
  },
  "productionApi": {
    "serverUrl": "xxx",
    "apiKey": "xxx",
    "jurisdictionId": "3"
  }
}
```
If more than one api is used and you need to switch between them, replace the text in the `whichApi` section with the name of the api object you want.

If only one api is to be used, you can write the necessary information directly into the `apiConfig`.

### DIFF
--- a/src/helpers/sueHelper.ts
+++ b/src/helpers/sueHelper.ts
@@ -1,24 +1,35 @@
-import * as appJson from '../../app.json';
-import { secrets } from '../config';
+import { storageHelper } from './storageHelper';
 
-const namespace = appJson.expo.slug as keyof typeof secrets;
-export const serverUrl = secrets[namespace]?.sue?.serverUrl;
-export const apiKey = secrets[namespace]?.sue?.apiKey;
-export const jurisdictionId = secrets[namespace]?.sue?.jurisdictionId;
+export const fetchSueEndpoints = async (serviceRequestId?: number) => {
+  const globalSettings = await storageHelper.globalSettings();
+  const { apiConfig = {} } = globalSettings?.settings?.sue || {};
 
-export const sueFetchObj = {
-  method: 'GET',
-  headers: {
-    accept: 'application/json',
-    api_key: apiKey
-  }
+  const { apiKey, jurisdictionId, serverUrl } = apiConfig[apiConfig?.whichApi] || apiConfig;
+
+  const sueFetchObj = {
+    method: 'GET',
+    headers: {
+      accept: 'application/json',
+      api_key: apiKey
+    }
+  };
+
+  const suePostRequest = `${serverUrl}/requests`;
+  const suePrioritiesUrl = `${serverUrl}/priorities?jurisdiction_id=${jurisdictionId}`;
+  const sueRequestsUrl = `${serverUrl}/requests?jurisdiction_id=${jurisdictionId}`;
+  const sueRequestsUrlWithServiceId = `${serverUrl}/requests/${serviceRequestId}?jurisdiction_id=${jurisdictionId}`;
+  const sueServicesUrl = `${serverUrl}/services?jurisdiction_id=${jurisdictionId}`;
+  const sueStatusesUrl = `${serverUrl}/statuses?jurisdiction_id=${jurisdictionId}`;
+
+  return {
+    apiKey,
+    jurisdictionId,
+    sueFetchObj,
+    suePostRequest,
+    suePrioritiesUrl,
+    sueRequestsUrl,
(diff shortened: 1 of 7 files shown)


src/helpers/sueHelper.ts
```diff
@@ -1,24 +1,40 @@
...
+export const fetchSueEndpoints = async (serviceRequestId?: number) => {
+  const globalSettings = await storageHelper.globalSettings();
+  const { apiConfig = {} } = globalSettings?.settings?.sue || {};
 
-export const sueFetchObj = {
-  method: 'GET',
-  headers: {
-    accept: 'application/json',
-    api_key: apiKey
+  if (!apiConfig) {
```
donni106, line 7: is this possible? with the `{}` fallbacks i guess it can never be falsey/undefined. so do we need this check here? and what if undefined is returned here instead of an object, does the other codes work well?

src/helpers/sueHelper.ts
```diff
@@ -1,24 +1,40 @@
...
+  };
+
+  const suePrioritiesUrl = `${serverUrl}/priorities?jurisdiction_id=${jurisdictionId}`;
+  const sueRequestsUrl = `${serverUrl}/requests?jurisdiction_id=${jurisdictionId}`;
+  const sueRequestsUrlWithServiceId = `${serverUrl}/requests/${serviceRequestId}?jurisdiction_id=${jurisdictionId}`;
+  const sueServicesUrl = `${serverUrl}/services?jurisdiction_id=${jurisdictionId}`;
+  const sueStatusesUrl = `${serverUrl}/statuses?jurisdiction_id=${jurisdictionId}`;
 
-export const suePostRequest = `${serverUrl}/requests`;
+  const suePostRequest = `${serverUrl}/requests`;
```
donni106, line 27: what about ordering them in the same way as in the return object?

src/queries/SUE/requestsWithServiceRequestId.tsx
```diff
@@ -1,12 +1,12 @@
...
 import _mapKeys from 'lodash/mapKeys';
 
-import { sueFetchObj, sueRequestsUrlWithServiceId } from '../../helpers';
+import { sueApi, sueRequestsUrlWithServiceId } from '../../helpers';
 
 export const requestsWithServiceRequestId = async (serviceRequestId: number) => {
-  const response = await (
-    await fetch(`${sueRequestsUrlWithServiceId(serviceRequestId)}`, sueFetchObj)
-  ).json();
+  const { sueFetchObj = {}, sueRequestsUrlWithServiceId = '' } = await sueApi(serviceRequestId);
```
donni106, line 7: not `fetchSueEndpoints` here as well?

src/helpers/sueHelper.ts
ardasnturk: Yes, you are right. Even without this check, the application doesn't crash and works thanks to `{}` fallbacks. Since there is no data exchange, Network request failed error occurs and `EmptyMessage` component is displayed.

src/helpers/sueHelper.ts
ardasnturk: I wanted to sort alphabetically, but I didn't sort the variables. I will sort the variables alphabetically.

src/queries/SUE/requestsWithServiceRequestId.tsx
ardasnturk: Yeah, that's right, I missed that.

ardasnturk: https://github.com/smart-village-solutions/smart-village-app-app/commit/1a6a8373aec308a3a1f1eeb83763b55eb027ee6f

donni106 (approved):

PR #1063, opened 2024-03-26
feat: expand `PermanentFilterProvider` with `mowasRegionalKeys`

ardasnturk:
- updated `PermanentFilterProviderValues` type to include separate state and dispatch variables for `dataProviderIds` and `mowasRegionalKeys`
- updated `PermanentFilterContext` to reflect the changes in `PermanentFilterProviderValues`
- extended `PermanentFilterProvider` to use `mowasRegionalKeysReducer` and added logic to load and dispatch `mowasRegionalKeys` alongside `dataProviderIds`
- refactored the `usePermanentFilter` hook to return `excludeDataProviderIds` and `excludeMowasRegionalKeys`
- updated `PermanentFilterSettings` to integrate the change to the `usePermanentFilter` hook
- integrated changes to the `usePermanentFilter` hook into `Overviews` and `HomeScreen`
- added `excludeMowasRegionalKeys` `queryVariables` to `newsItems` query
- added `excludeMowasRegionalKeys` to `queryVariables` in `Overviews` and `HomeScreen`

SVAK-37

Since the `newsItems` query does not have the `excludeDataProviderIds` `queryVariables`, the data is not currently visible. The data will be displayed after the query has this `queryVariables`.

## Test:

Please attach the following log to `HomeScreen` for testing.
```
console.warn({ excludeDataProviderIds, excludeMowasRegionalKeys });
```
Then go to the `SettingsScreen` and change the switchers in the Datenquellen or MoWaS-Regionen section. When you refresh the app, you will see the following text on the log screen.
```
{"excludeDataProviderIds": ["5"], "excludeMowasRegionalKeys": ["120630008008"]}
```

### DIFF
--- a/src/PermanentFilterProvider.tsx
+++ b/src/PermanentFilterProvider.tsx
@@ -2,26 +2,46 @@ import { noop } from 'lodash';
 import React, { createContext, useCallback, useEffect, useReducer } from 'react';
 
 import { readFromStore } from './helpers';
-import { DATA_PROVIDER_FILTER_KEY, permanentFilterReducer } from './reducers';
+import {
+  DATA_PROVIDER_FILTER_KEY,
+  MOWAS_REGIONAL_KEYS,
+  MowasFilterAction,
+  MowasFilterReducerAction,
+  mowasRegionalKeysReducer,
+  permanentFilterReducer
+} from './reducers';
 import { FilterAction, FilterReducerAction } from './types';
 
 type PermanentFilterProviderValues = {
-  state: string[];
-  dispatch: React.Dispatch<FilterReducerAction>;
+  dataProviderState: string[];
+  dataProviderDispatch: React.Dispatch<FilterReducerAction>;
+  mowasRegionalKeysState: string[];
+  mowasRegionalKeysDispatch: React.Dispatch<MowasFilterReducerAction>;
 };
 
 export const PermanentFilterContext = createContext<PermanentFilterProviderValues>({
-  dispatch: noop,
-  state: []
+  dataProviderState: [],
+  dataProviderDispatch: noop,
+  mowasRegionalKeysState: [],
+  mowasRegionalKeysDispatch: noop
 });
 
 export const PermanentFilterProvider = ({ children }: { children?: React.ReactNode }) => {
-  const [state, dispatch] = useReducer(permanentFilterReducer, []);
+  const [dataProviderState, dataProviderDispatch] = useReducer(permanentFilterReducer, []);
+  const [mowasRegionalKeysState, mowasRegionalKeysDispatch] = useReducer(
+    mowasRegionalKeysReducer,
+    []
+  );
 
   const loadFilters = useCallback(async () => {
     const dataProviderIds = ((await readFromStore(DATA_PROVIDER_FILTER_KEY)) ?? []) as string[];
+    const mowasRegionalKeys = ((await readFromStore(MOWAS_REGIONAL_KEYS)) ?? []) as string[];
 
-    dispatch({ type: FilterAction.OverwriteDataProviders, payload: dataProviderIds });
+    dataProviderDispatch({ type: FilterAction.OverwriteDataProviders, payload: dataProviderIds });
+    mowasRegionalKeysDispatch({
+      type: MowasFilterAction.OverwriteMowasRegionalKeys,
+      payload: mowasRegionalKeys
+    });
   }, []);
 
   useEffect(() => {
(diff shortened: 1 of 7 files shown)


src/components/settings/PermanentFilterSettings.tsx
```diff
@@ -20,7 +20,8 @@ export const PermanentFilterSettings = () => {
 
   const fetchPolicy = graphqlFetchPolicy({ isConnected, isMainserverUp, refreshTime });
 
-  const { state: excludedDataProviderIds, dispatch } = usePermanentFilter();
+  const { excludeDataProviderIds: excludedDataProviderIds, dataProviderDispatch: dispatch } =
```
donni106, line 23: `excludeDataProviderIds` should be `dataProviderState` or am I wrong?

src/hooks/permanentFilter.ts
```diff
@@ -2,4 +2,21 @@ import { useContext } from 'react';
...
+export const usePermanentFilter = () => {
+  const {
+    dataProviderState,
+    dataProviderDispatch,
+    mowasRegionalKeysState,
+    mowasRegionalKeysDispatch
+  } = useContext(PermanentFilterContext);
+
+  const excludeDataProviderIds = dataProviderState;
+  const excludeMowasRegionalKeys = mowasRegionalKeysState;
```
donni106, line 14: why are you renaming these?

src/PermanentFilterProvider.tsx
```diff
@@ -2,34 +2,61 @@ import { noop } from 'lodash';
...
+  mowasRegionalKeysDispatch: React.Dispatch<MowasFilterReducerAction>;
 };
 
 export const PermanentFilterContext = createContext<PermanentFilterProviderValues>({
-  dispatch: noop,
-  state: []
+  dataProviderState: [],
+  dataProviderDispatch: noop,
+  mowasRegionalKeysState: [],
+  mowasRegionalKeysDispatch: noop
```
donni106, line 26: alphabetical order would be dispatch before state, right?

src/hooks/permanentFilter.ts
ardasnturk: Because according to the ticket, we have to use it as follows. For this reason, I have decided to change the name.

```
  const { excludeDataProviderIds, excludeMowasRegionalKeys } = usePermanentFilter();
```

src/PermanentFilterProvider.tsx
ardasnturk: Yes, you are right. I can alphabetise here. 👍

src/components/settings/PermanentFilterSettings.tsx
ardasnturk: I answered here. I think that'll be enough.

https://github.com/smart-village-solutions/smart-village-app-app/pull/1063#discussion_r1548134400

src/PermanentFilterProvider.tsx
ardasnturk: https://github.com/smart-village-solutions/smart-village-app-app/pull/1063/commits/9114e12a4846fe105183b67e08910216b6e8e4ee

src/components/settings/PermanentFilterSettings.tsx
donni106: wouldn't it be more readable to refactor the code to `const { excludeDataProviderIds, dataProviderDispatch }` instead of renaming things. so the code needs to be adjusted to not use `excludedDataProviderIds` and `dispatch` anymore. no one knows what `dispatch` is related to.

src/hooks/permanentFilter.ts
donni106: hm ok, but then we do not need to create to variables. you can rename in the return directly, right?!

ardasnturk: Yes, that's right. We don't need an extra variable for this. 👍

donni106 (approved):

src/queries/newsItems.js
```diff
@@ -8,6 +8,7 @@ export const GET_NEWS_ITEMS = gql`
     $dataProvider: String
     $dataProviderId: ID
     $excludeDataProviderIds: [ID]
+    $excludeMowasRegionalKeys: [ID]
```
donni106, line 11: please change this to `[String]`

ardasnturk: https://github.com/smart-village-solutions/smart-village-app-app/pull/1063/commits/f0667f51f883a7478cddd9ef80edd61b736d3f45

PR #1064, opened 2024-03-27
feat(voucher): add qr code scan screen

ardasnturk:
- added `detail` to `screens` object in `config.ts` to redirect qr codes read with system camera to `DetailScreen` in the app
- added `VoucherScannerScreen` to be able to scan qr from within the app
- added `parseQrCode` function to get the data we need for routing from the data generated when qr is scanned
  - the link that should be in the qr is like this -> `smart-village-app://detail?query=pointOfInterest&id=41147`
- added query value as `pointOfInterest` as default, although there is no query value in the data
- added alert to notify the user if the data in the qr does not contain an `id` and added state control to prevent the alert from being rendered infinitely

SVAK-40

## Test:

For the test, go to the VoucherScannerScreen and show the following QR to the camera.

image

After the QR is recognised you will be automatically redirected to the `DetailScreen` of the POI with id `41147`

## Screenshots:

|VoucherScannerScreen without QR|VoucherScannerScreen with QR|
|--|--|
IMG_7010|IMG_7011

### DIFF
--- a/src/config/texts.js
+++ b/src/config/texts.js
@@ -824,7 +824,9 @@ export const texts = {
     },
     voucher: {
       home: 'TreueClub',
-      index: 'Neue Angebote'
+      index: 'Neue Angebote',
+      partner: 'Kooperationspartner',
+      qr: 'Code scannen'
     },
     wasteCollection: 'Abfallkalender',
     weather: 'Wetter'
@@ -1171,6 +1173,12 @@ export const texts = {
     offersCategories: 'Angebote Kategorien',
     result: 'Ergebnis',
     results: 'Ergebnisse',
+    scannerScreen: {
+      errorBody:
+        'Beim Scannen des Codes ist ein Fehler aufgetreten. Bitte versuchen Sie es erneut.',
+      errorButton: 'Erneut versuchen',
+      errorTitle: 'Fehler'
+    },
     secret: 'Kundennummer'
   },
   wasteCalendar: {
--- a/src/config/texts.js
+++ b/src/config/texts.js
@@ -1140,6 +1140,8 @@ export const texts = {
       close: 'Schließen',
       daily: 'pro Tag',
       desiredQuantity: 'Gewünschte Anzahl',
+      emptyMessage:
+        'Der Inhalt kann nicht geladen werden. Bitte überprüfen Sie Ihre Internetverbindung.',
       frequency: (maxPerPerson, frequency) =>
         `${maxPerPerson}x pro Person ${texts.voucher.detailScreen[frequency]} einlösbar`,
       limit: (availableQuantity, maxQuantity) =>
@@ -1174,8 +1176,7 @@ export const texts = {
     result: 'Ergebnis',
     results: 'Ergebnisse',
     scannerScreen: {
-      errorBody:
-        'Beim Scannen des Codes ist ein Fehler aufgetreten. Bitte versuchen Sie es erneut.',
+      errorBody: 'Beim Scannen des Codes ist ein Fehler aufgetreten. Bitte erneut versuchen.',
       errorButton: 'Erneut versuchen',
       errorTitle: 'Fehler'
     },
--- a/src/screens/DetailScreen.js
+++ b/src/screens/DetailScreen.js
@@ -4,6 +4,7 @@ import { Query } from 'react-apollo';
 import { ActivityIndicator, RefreshControl, ScrollView } from 'react-native';
 
 import {
+  EmptyMessage,
   EventRecord,
   LoadingContainer,
   NewsItem,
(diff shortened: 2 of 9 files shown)


src/config/texts.js
```diff
@@ -1171,6 +1173,12 @@ export const texts = {
     offersCategories: 'Angebote Kategorien',
     result: 'Ergebnis',
     results: 'Ergebnisse',
+    scannerScreen: {
+      errorBody:
+        'Beim Scannen des Codes ist ein Fehler aufgetreten. Bitte versuchen Sie es erneut.',
```
donni106, line 1178: Bitte versuchen Sie es erneut. -> Bitte erneut versuchen.

src/screens/DetailScreen.js
```diff
@@ -83,7 +83,8 @@ const useRootRouteByCategory = (details, navigation) => {
 export const DetailScreen = ({ navigation, route }) => {
   const { isConnected, isMainserverUp } = useContext(NetworkContext);
   const query = route.params?.query ?? '';
-  const queryVariables = route.params?.queryVariables ?? {};
+  const detailId = route.params?.id ?? undefined;
```
donni106, line 86: is `id` already in use here, so you have chosen `deatilId`?
there is no `?? undefined;` needed. if there is no `route.params?.id` it is undefined.

src/screens/Voucher/VoucherScannerScreen.tsx
```diff
@@ -0,0 +1,69 @@
...
+import { texts } from '../../config';
+import { ScreenName } from '../../types';
+import { QUERY_TYPES } from '../../queries';
+
+const parseQrCode = (
+  data: string
+): { query: string; queryVariables: { id: number | undefined } } => {
+  const result = Linking.parse(data);
+
+  if (result.scheme === appJson.expo.scheme && result.hostname === 'detail') {
```
donni106, line 18: can we make the hostnames a constant somewhere in the configs section as we have more than one now? we should not let this grow hardcoded in several places.

src/screens/Voucher/VoucherScannerScreen.tsx
```diff
@@ -0,0 +1,69 @@
...
+      query,
+      title: texts.screenTitles.voucher.partner
+    });
+  };
+
+  return (
+    <SafeAreaViewFlex>
+      {isScanning ? (
+        <Camera onBarCodeScanned={handleBarCodeScanned} style={styles.scanner} />
+      ) : null}
```
donni106, line 59: in which case will this be `null` and what will we see on the screen? is this reachable?

src/screens/DetailScreen.js
ardasnturk: I thought `detailId` was a better nomenclature, but since we are already in `DetailScreen`, id nomenclature can also be used.
Yes, that's right. I will update this. 👍

src/screens/Voucher/VoucherScannerScreen.tsx
ardasnturk: Yes, that's right. For this I will create a new section for `hostname` in `consts` and pull it from there. Like:
```
  HOST_NAMES: {
    DETAIL: 'detail',
    ENCOUNTER: 'encounter'
  },
```

src/screens/Voucher/VoucherScannerScreen.tsx
```diff
@@ -0,0 +1,69 @@
...
+  const result = Linking.parse(data);
+
+  if (result.scheme === appJson.expo.scheme && result.hostname === 'detail') {
+    const id = result.queryParams?.id as number | undefined;
+    const query = result.queryParams?.query?.toString() || '';
+
+    return { query, queryVariables: { id } };
+  }
+
+  return { query: QUERY_TYPES.POINT_OF_INTEREST, queryVariables: { id: undefined } };
```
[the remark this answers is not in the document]
ardasnturk: In this case, if the `id` value is `undefined`, we show an alert on the screen and say that the QR code needs to be scanned again. In fact, we are not redirecting to an empty screen. But if there is no data for this id in `DetailScreen`, we can use the `EmptyMessage` component. At the moment we are just showing a blank screen. 

```
 if (!Component) return null;
```

src/screens/Voucher/VoucherScannerScreen.tsx
ardasnturk: https://github.com/smart-village-solutions/smart-village-app-app/pull/1064/commits/590d4d69adc0850e894de679de501a1ebf77f9d5

src/screens/Voucher/VoucherScannerScreen.tsx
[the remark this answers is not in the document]
ardasnturk: https://github.com/smart-village-solutions/smart-village-app-app/pull/1064/commits/8aa9eef4fb385a167f678f03f3ff106ecf50eded

donni106 (approved):

PR #1070, opened 2024-04-05
chore: compile in the latest xcode version for iOS

ardasnturk:
- updated xcode version in `eas.json` to latest to eliminate Apple's warning when a new build is created

SVA-1305

### DIFF
--- a/eas.json
+++ b/eas.json
@@ -6,7 +6,10 @@
     "production": {
       "node": "16.13.0",
       "yarn": "1.22.19",
-      "channel": "production"
+      "channel": "production",
+      "ios": {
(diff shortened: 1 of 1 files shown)


eas.json
```diff
@@ -6,7 +6,10 @@
     "production": {
       "node": "16.13.0",
       "yarn": "1.22.19",
-      "channel": "production"
+      "channel": "production",
+      "ios": {
+        "image": "latest"
```
donni106, line 11: can you find out, which image will be used by default, if not `latest`?

donni106 (approved):

PR #1072, opened 2024-04-11
feat: add voucher list to poi detail screen

ardasnturk:
- added `|| data` to `listItemParser` to fix the error that occurs in POI query because the data is not in the `genericItem` object and the direct data is an array
- added `EmptyMessage` component to display the `quota` data in case it is null
- added `FlatList` to show the coupon list if there is voucher data in `DetailScreen`
- added `parseListItemsFromQuery` to convert voucher data to the format requested by `VoucherListItem`
- added the feature that if a POI has more than one voucher, instead of showing all of them, it shows 2 at the beginning and then shows 5 more by pressing the button at the bottom of the list

## Test:
Please change the server connection as the voucher feature is not available on the development server. 
Then follow this path:
Service Tab -> Stadtwerke Tile -> TreueClub Tile -> Kooperations-partner Tile -> Select any POI

## Screenshots:

|POI `DetailScreen`|vouchers list in `DetailScreen` with mehr button|vouchers list|voucher detail screen|
|--|--|--|--|
Simulator Screenshot - iPhone 15 Pro Max - 2024-04-11 at 10 19 36|Simulator Screenshot - iPhone 15 Pro Max - 2024-04-11 at 10 19 40|Simulator Screenshot - iPhone 15 Pro Max - 2024-04-11 at 10 19 46|Simulator Screenshot - iPhone 15 Pro Max - 2024-04-11 at 10 19 50

### DIFF
--- a/src/helpers/parser/listItemParser.js
+++ b/src/helpers/parser/listItemParser.js
@@ -495,7 +495,7 @@ export const parseListItemsFromQuery = (query, data, titleDetail, options = {}) 
     case QUERY_TYPES.VOLUNTEER.PROFILE:
       return parseVolunteers(data, query, skipLastDivider);
     case QUERY_TYPES.VOUCHERS:
-      return parseVouchers(data[QUERY_TYPES.GENERIC_ITEMS], skipLastDivider);
+      return parseVouchers(data[QUERY_TYPES.GENERIC_ITEMS] || data, skipLastDivider);
     case QUERY_TYPES.VOUCHERS_CATEGORIES:
       return parseVouchersCategories(data[QUERY_TYPES.GENERIC_ITEMS], skipLastDivider);
     case QUERY_TYPES.CONSUL.DEBATES:
--- a/src/helpers/parser/listItemParser.js
+++ b/src/helpers/parser/listItemParser.js
@@ -495,7 +495,7 @@ export const parseListItemsFromQuery = (query, data, titleDetail, options = {}) 
     case QUERY_TYPES.VOLUNTEER.PROFILE:
       return parseVolunteers(data, query, skipLastDivider);
     case QUERY_TYPES.VOUCHERS:
-      return parseVouchers(data[QUERY_TYPES.GENERIC_ITEMS] || data, skipLastDivider);
+      return parseVouchers(data[QUERY_TYPES.GENERIC_ITEMS], skipLastDivider);
     case QUERY_TYPES.VOUCHERS_CATEGORIES:
       return parseVouchersCategories(data[QUERY_TYPES.GENERIC_ITEMS], skipLastDivider);
     case QUERY_TYPES.CONSUL.DEBATES:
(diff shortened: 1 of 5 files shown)


src/screens/Voucher/VoucherDetailScreen.tsx
```diff
@@ -43,7 +45,12 @@ export const VoucherDetailScreen = ({ route }: StackScreenProps<any>) => {
     return <LoadingSpinner loading />;
   }
 
-  const { contentBlocks, discountType, quota, id } = data[QUERY_TYPES.GENERIC_ITEM];
+  const { contentBlocks, discountType, quota, id } = data[QUERY_TYPES.GENERIC_ITEM] || data;
```
donni106, line 48: is `|| data` really needed here?

src/helpers/parser/listItemParser.js
```diff
@@ -495,7 +495,7 @@ export const parseListItemsFromQuery = (query, data, titleDetail, options = {})
     case QUERY_TYPES.VOLUNTEER.PROFILE:
       return parseVolunteers(data, query, skipLastDivider);
     case QUERY_TYPES.VOUCHERS:
-      return parseVouchers(data[QUERY_TYPES.GENERIC_ITEMS], skipLastDivider);
+      return parseVouchers(data[QUERY_TYPES.GENERIC_ITEMS] || data, skipLastDivider);
```
donni106, line 498: is `|| data` really needed her? or could it be passed as `{ [QUERY_TYPES.GENERIC_ITEMS]: data }` in here, to not need the condition?

ardasnturk: You need it here. If the data is not in the `genericItems` object, I think `{ [QUERY_TYPES.GENERIC_ITEMS]: data }` will give an error.

ardasnturk: But then what about the data inside the genericItem object? Then won't genericItem.genericItem = data?

donni106: my suggestion should only be applied where data is passed directly without `genericItem:` to ensure, that this key exists in any case.

ardasnturk: Aha. This is what you are talking about. Yes. We can delete the logic in `listItemParser`.

```
  const voucherListItems = useMemo(() => {
    return parseListItemsFromQuery(
      QUERY_TYPES.VOUCHERS,
      { [QUERY_TYPES.GENERIC_ITEMS]: vouchers },
      undefined,
      {
        withDate: false
      }
    );
  }, [vouchers]);
```

src/screens/Voucher/VoucherDetailScreen.tsx
ardasnturk: Yes. You're right, this control can be removed.

src/helpers/parser/listItemParser.js
ardasnturk: https://github.com/smart-village-solutions/smart-village-app-app/pull/1072/commits/b7eead536633a9e4c6b0a47ace778ba0ece94787

donni106 (approved):

PR #1079, opened 2024-04-11
fix: fetch more data error

ardasnturk:
- updated the `query` in the `fetchMoreData` function with `genericItems` because the data from the server is in the `genericItems` object
- added `_filter` by `categories` to `count` value because we can't list vouchers without `categories`
- added `GET_VOUCHERS` query to `FETCH_MORE_QUERIES` to run voucher query with `getFetchMoreQuery`

SVAK-35


src/screens/Voucher/VoucherIndexScreen.tsx
```diff
@@ -118,7 +126,7 @@ export const VoucherIndexScreen = ({ navigation, route }: StackScreenProps<any>)
     return <LoadingSpinner loading />;
   }
 
-  const count = listItems?.length;
+  const count = _filter(listItems, ({ categories }) => categories.length > 0).length;
```
donni106, line 129: what about `listItems.filter({ categories } => !!categories.length).length`? i do not think that we need lodash here and on the other side we often not check length with > 0 but with boolean !!

ardasnturk: There are also two different usages in the code. Both do the same thing. loadsh has no extra advantage. I can change the code. 

Yes you are right, I will update the category control.

ardasnturk: https://github.com/smart-village-solutions/smart-village-app-app/commit/7d0a8ceffce23adcff5deea4e69324d54058c226

src/screens/Voucher/VoucherIndexScreen.tsx
```diff
@@ -1,5 +1,6 @@
 import { StackScreenProps } from '@react-navigation/stack';
 import _uniqBy from 'lodash/uniqBy';
+import _filter from 'lodash/filter';
```
donni106, line 3: can be removed now, right?

ardasnturk: yeah! 👍 😄

donni106 (approved):

PR #1081, opened 2024-04-16
feat: add redeemed vouchers query

ardasnturk:
- added `Vouchers` query to access the data of the vouchers used by the user
- added new case in `listItemParser` to convert the data from `Vouchers` query to the desired format
- added `data[QUERY_TYPES.VOUCHERS]` because the data from the server is in the vouchers object
- added new case to `ListComponent` for the list to be `VoucherList`

SVAK-35


src/queries/index.js
```diff
@@ -75,6 +80,7 @@ export const getQuery = (query, filterOptions = {}) => {
     [QUERY_TYPES.POINTS_OF_INTEREST_AND_TOURS]: GET_POINTS_OF_INTEREST_AND_TOURS,
     [QUERY_TYPES.PUBLIC_HTML_FILE]: GET_PUBLIC_HTML_FILE,
     [QUERY_TYPES.PUBLIC_JSON_FILE]: GET_PUBLIC_JSON_FILE,
+    [QUERY_TYPES.REDEEMED_VOUCHERS]: GET_REDEEMED_VOUCHERS,
```
donni106, line 83: please change it to ´VOUCHERS_REDEEMED` to have all the vouchers at one place

ardasnturk: https://github.com/smart-village-solutions/smart-village-app-app/commit/c4ed62afe1d53ab40d8f4fad62d781cb09e0613d

src/queries/index.js
```diff
@@ -79,6 +84,7 @@ export const getQuery = (query, filterOptions = {}) => {
     [QUERY_TYPES.WASTE_STREET]: WASTE_STREET,
     [QUERY_TYPES.VOUCHER]: GET_VOUCHER,
     [QUERY_TYPES.VOUCHERS_CATEGORIES]: GET_VOUCHERS_CATEGORIES,
+    [QUERY_TYPES.VOUCHERS_REDEEMED]: GET_REDEEMED_VOUCHERS,
```
donni106, line 87: it would have make sense to change `GET_REDEEMED_VOUCHERS` as well, or what do you say?

ardasnturk: Yes, it will be more logical with this change.  👍

ardasnturk: https://github.com/smart-village-solutions/smart-village-app-app/commit/7b92f2e633df56952015b99c46f92a330551cb21

donni106 (approved):

PR #1082, opened 2024-04-16
feat: add availability for voucher

ardasnturk:
- updated the queries according to the type variables made on the server
- added checking how many vouchers a user can use based on `availableQuantityForMember` instead of `maxPerPerson`
- added availability checks of the voucher according to the response from the server
- updated button's title according to availability

SVAK-35

## Screenshots:

|redeemable voucher|redeemed voucher|unavailable voucher|
|--|--|--|
Simulator Screenshot - iPhone SE (3rd generation) - 2024-04-16 at 19 12 40|Simulator Screenshot - iPhone SE (3rd generation) - 2024-04-16 at 19 12 31|Simulator Screenshot - iPhone SE (3rd generation) - 2024-04-16 at 19 12 43

### DIFF
--- a/src/components/vouchers/VoucherRedeem.tsx
+++ b/src/components/vouchers/VoucherRedeem.tsx
@@ -17,6 +17,7 @@ import { Wrapper, WrapperRow, WrapperVertical } from '../Wrapper';
 
 const defaultTime = 15 * 60; // 15 * 60 sec.
 
+/* eslint-disable complexity */
 export const VoucherRedeem = ({ quota, voucherId }: { quota: TQuota; voucherId: string }) => {
   const { isLoggedIn, memberId } = useVoucher();
   const [isVisible, setIsVisible] = useState(false);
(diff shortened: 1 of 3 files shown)


src/components/vouchers/VoucherRedeem.tsx
```diff
@@ -25,10 +26,31 @@ export const VoucherRedeem = ({ quota, voucherId }: { quota: TQuota; voucherId:
...
+        return;
+      }
+
+      return;
+    }
+
+    if (availableQuantityForMember === 0) {
+      setIsRedeemedVoucher(true);
+
+      return;
```
donni106, line 51: effects should not return anything. returns are used for cleanup functions.
in our case, we do not have the cleanup so there shouldn't be returns. please rework it with ifs and elses

donni106: chatgpt is refactoring this to

```
useEffect(() => {
  const hasAvailableQuantity = availableQuantity !== 0;
  const hasNoAvailableQuantityForMember = availableQuantityForMember === 0;

  setIsAvailableVoucher(hasAvailableQuantity);
  
  if (hasNoAvailableQuantityForMember) {
    setIsRedeemedVoucher(true);
  }
}, [availableQuantity, availableQuantityForMember]);
```

can you give it a try or find a mixed solution?

ardasnturk: I tested it and it works. 👍
https://github.com/smart-village-solutions/smart-village-app-app/commit/c6956ae520ab0fe55a1ed53cf90f60bfb9f25bbd

donni106 (approved):

PR #1099, opened 2024-05-06
feat: add html content above and below tile screen

ardasnturk:
- added `useStaticContent` to `ServiceTiles.js` to get htmlContent from server
- added the `htmlContent.forAbove` content as content with the `html` prop in `ServiceTiles` and merged them in the `contentForAbove` constant and turned them into a single html content
- added `contentForBelow` html to be shown at the bottom of `ServiceTiles`
- added the name in `useStaticContent` to `'${staticJsonName}Content'` so that this feature works on all staticContent screens

SVAK-85

## Test:

To use this feature, create a JSON on the main-server with the same staticContent name but with 'Content' appended to the end.
For example: 
staticContent = "homeService"
for staticContent html = "homeServiceContent"

Make sure that `homeServiceContent` is in JSON format.

The content of `homeServiceContent` can be as follows.

```
{
  "forAbove": "html",
  "forBelow": "html"
}
```

## Screenshots:

|htmlContent?.forAbove|htmlContent?.forBelow|html|
|--|--|--|
Simulator Screenshot - iPhone 15 Pro Max - 2024-05-06 at 11 29 34|Simulator Screenshot - iPhone 15 Pro Max - 2024-05-06 at 11 29 38|Simulator Screenshot - iPhone 15 Pro Max - 2024-05-06 at 11 29 45

### DIFF
--- a/src/components/screens/ServiceTiles.js
+++ b/src/components/screens/ServiceTiles.js
@@ -2,15 +2,15 @@ import PropTypes from 'prop-types';
 import React, { useCallback, useContext, useState } from 'react';
 import { RefreshControl, ScrollView, StyleSheet, View } from 'react-native';
 
+import { NetworkContext } from '../../NetworkProvider';
 import { colors, normalize } from '../../config';
 import { useStaticContent, useVolunteerRefresh } from '../../hooks';
-import { NetworkContext } from '../../NetworkProvider';
 import { HtmlView } from '../HtmlView';
 import { Image } from '../Image';
 import { LoadingSpinner } from '../LoadingSpinner';
 import { SafeAreaViewFlex } from '../SafeAreaViewFlex';
-import { Wrapper } from '../Wrapper';
 import { SectionHeader } from '../SectionHeader';
+import { Wrapper } from '../Wrapper';
 
 import { Service } from './Service';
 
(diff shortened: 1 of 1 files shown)


src/components/screens/ServiceTiles.js
```diff
@@ -70,15 +85,21 @@ export const ServiceTiles = ({
...
+            {!!contentForAbove && (
               <Wrapper>
-                <HtmlView html={html} />
+                <HtmlView html={contentForAbove} />
               </Wrapper>
             )}
 
             <View style={styles.padding}>
-              {!error && <Service data={data} staticJsonName={staticJsonName} />}
+              {(!error || !htmlError) && <Service data={data} staticJsonName={staticJsonName} />}
```
donni106, line 95: how is `htmlError` important here? isn't it just about the first query with `staticJsonName`? so just `error` is still the condition to check?
otherwise i guess this would need a `&&` so in case of no errors at all we want to render, right?

ardasnturk: you are right, actually we don't need `htmlError` here.

ardasnturk: https://github.com/smart-village-solutions/smart-village-app-app/commit/4a14506f26a9acce9158aeae6f03598b5af9de05

donni106 (approved):

PR #1101, opened 2024-05-06
fix: pictures not showing up problem for android

ardasnturk:
- added new control to fix the problem of images not appearing through cache only in development mode and on android platform

SVA-1317

## Screenshots:

|before|after|
|--|--|
Screenshot (06 05 2024 14_04_01)|Screenshot (06 05 2024 14_01_40)

### DIFF
--- a/src/components/Image.js
+++ b/src/components/Image.js
@@ -76,6 +76,13 @@ export const Image = ({
         break effect;
       }
 
+      // we needed this control to solve the problem of image not loading on android devices in development environment
+      if (__DEV__ && device.platform === 'android') {
+        setSource({ uri: sourceProp.uri });
+
+        break effect;
+      }
+
       sourceProp.uri
         ? CacheManager.get(sourceProp.uri)
             .getPath()


src/components/Image.js
```diff
@@ -80,7 +80,11 @@ export const Image = ({
         ? CacheManager.get(sourceProp.uri)
             .getPath()
             .then((path) => {
-              mounted && setSource({ uri: path ?? NO_IMAGE.uri });
+              mounted &&
+                setSource({
+                  uri:
+                    __DEV__ && device.platform === 'android' ? sourceProp.uri : path ?? NO_IMAGE.uri
```
donni106, line 86: in case of `__DEV__ && device.platform === 'android'` we can directly return `sourceProp.uri` without going through `CacheManager.get(sourceProp.uri)`, right?!

donni106 (changes requested):

src/components/Image.js
ardasnturk: Yes we can. But I didn't want to add an extra if check for this, so I wanted to do it here. But I will update the code to increase the understandability of the code.

ardasnturk: https://github.com/smart-village-solutions/smart-village-app-app/commit/9aeaeab544c1cbede401fa2322a0fcb70c077ebf

donni106: it is not just about understandability but about the async process (get image from cache) we do not need to run. we can save this effort.

donni106 (approved):

src/components/Image.js
```diff
@@ -76,6 +76,12 @@ export const Image = ({
         break effect;
       }
 
+      if (__DEV__ && device.platform === 'android') {
+        setSource({ uri: sourceProp.uri });
+
+        break effect;
```
donni106, line 82: please add a comment like in the prior condition

ardasnturk: https://github.com/smart-village-solutions/smart-village-app-app/commit/3fb7816d8ba405edd24837aee5e3466695dd3679

d4v3d4vE (commented):
really nice and efficient solution

PR #1103, opened 2024-05-07
feat: add title on pins

ardasnturk:
- added the `Callout` component of the `react-native-maps` package to add title on the pins in the report screen
- added `calloutTextEnabled` prop to enable the `Callout` feature on the pin
- added condition to `onMapPress` prop in `sueReportLocation` to ensure that the location is not selected after clicking the pin on the map

SUE-75

## Screenshots:

|before|after|
|--|--|
Simulator Screenshot - iPhone 15 Pro - 2024-05-07 at 16 11 49|Simulator Screenshot - iPhone 15 Pro - 2024-05-07 at 16 11 46

### DIFF
--- a/src/components/map/Map.tsx
+++ b/src/components/map/Map.tsx
@@ -2,7 +2,7 @@ import _upperFirst from 'lodash/upperFirst';
 import React, { useContext, useRef } from 'react';
 import { StyleProp, StyleSheet, TouchableOpacity, View, ViewStyle } from 'react-native';
 import MapView from 'react-native-map-clustering';
-import { LatLng, MAP_TYPES, Marker, Polyline, Region, UrlTile } from 'react-native-maps';
+import { Callout, LatLng, MAP_TYPES, Marker, Polyline, Region, UrlTile } from 'react-native-maps';
 
 import { colors, device, Icon, normalize } from '../../config';
 import { imageHeight, imageWidth } from '../../helpers';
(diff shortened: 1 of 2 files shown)


src/components/map/Map.tsx
```diff
@@ -191,6 +193,7 @@ export const Map = ({
         )}
         {locations?.map((marker, index) => {
           const isActiveMarker = selectedMarker && marker.id === selectedMarker;
+          const calloutText = marker.title || undefined;
```
donni106, line 196: `|| undefined` is not necessary. `marker.title` returns `undefined` if it is not defined.

donni106 (approved):

src/components/map/Map.tsx
```diff
@@ -246,6 +249,14 @@ export const Map = ({
                   iconName={marker.iconName ? marker.iconName : undefined}
                 />
               )}
+
+              {calloutTextEnabled && (
+                <Callout style={styles.callout}>
+                  <RegularText smallest center>
+                    {calloutText}
```
donni106, line 256: what about directly `marker.title` instead of creating the variable `calloutText`?

ardasnturk: Yes, very logical. I've made this change.
https://github.com/smart-village-solutions/smart-village-app-app/commit/219b63c72de3f54270f773026c7b5fa3318b29c8

donni106: but now you have added an additional check, which needs !!

PR #1104, opened 2024-05-08
feat: add the check whether the voucher is redeemed locally

ardasnturk:
- added `localRedeemedVoucherCheck` function to ensure that if a user has redeemed a voucher without an internet connection, they cannot redeem the same voucher again and updated the button to disabled if the voucher has been redeemed
- saved coupon information in the device's memory in case of a problem with the internet connection has been moved to the catch section
- moved `deviceToken` information out of the `try` blog as needed when saving voucher information
- added `redeemLocalVouchers` function to ensure that the vouchers stored in the device are sent to the server and used online in case of internet connection
- added resetting of the `voucherTransactions` array saved on the device if the exchange of information with the server has been successful

SVAK-90

## Test:

To test, create a fake error in try-cache before mutation. This error will save the coupon information in the device's memory. Then open `VoucherHomeScreen` and send the vouchers saved in the device's memory to the server. You can then see these vouchers in the list of redeemed vouchers.

```
 throw new Error('Fake error!');
```

### DIFF
--- a/src/components/vouchers/VoucherRedeem.tsx
+++ b/src/components/vouchers/VoucherRedeem.tsx
@@ -33,6 +33,18 @@ export const VoucherRedeem = ({ quota, voucherId }: { quota: TQuota; voucherId: 
 
   const { availableQuantityForMember = 0, availableQuantity = 0 } = quota;
 
+  const localRedeemedVoucherCheck = async () => {
+    const voucherTransactions = (await readFromStore(VOUCHER_TRANSACTIONS)) || [];
+
+    if (voucherTransactions.length) {
+      const isRedeemed = voucherTransactions.some(
+        (transaction) => transaction.voucherId === voucherId && transaction.memberId === memberId
+      );
+
+      setIsRedeemedVoucher(isRedeemed);
+    }
+  };
+
   useEffect(() => {
     const hasAvailableQuantity = availableQuantity !== 0;
     const hasNoAvailableQuantityForMember = availableQuantityForMember === 0;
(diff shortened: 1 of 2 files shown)


src/components/vouchers/VoucherRedeem.tsx
```diff
@@ -42,7 +42,21 @@ export const VoucherRedeem = ({ quota, voucherId }: { quota: TQuota; voucherId:
...
+    const voucherTransactions = (await readFromStore(VOUCHER_TRANSACTIONS)) || [];
+
+    if (voucherTransactions.length) {
+      const isRedeemed = voucherTransactions.some(
+        (transaction) => transaction.voucherId === voucherId && transaction.memberId === memberId
+      );
+
+      setIsRedeemedVoucher(isRedeemed);
+    }
+  };
```
donni106, line 59: please move this method declaration before the effect to have it defined before used. this makes the flow of reading the file more appropriate.

src/components/vouchers/VoucherRedeem.tsx
```diff
@@ -67,15 +81,15 @@ export const VoucherRedeem = ({ quota, voucherId }: { quota: TQuota; voucherId:
...
-        addToStore(VOUCHER_DEVICE_TOKEN, deviceToken);
-      }
+    if (!deviceToken) {
+      deviceToken = uuid();
+      addToStore(VOUCHER_DEVICE_TOKEN, deviceToken);
+    }
 
-      redeemQuotaOfVoucher({
+    try {
+      await redeemQuotaOfVoucher({
```
donni106, line 92: why was the `await` added here?

src/components/vouchers/VoucherRedeem.tsx
```diff
@@ -84,6 +98,9 @@ export const VoucherRedeem = ({ quota, voucherId }: { quota: TQuota; voucherId:
         }
       });
 
+      setIsRedeemedVoucher(true);
+      setIsRedeemingVoucher(true);
```
donni106, line 102: can we use `finally` for these tow to not duplicate them in `try` and `catch`?

donni106 (changes requested):

src/components/vouchers/VoucherRedeem.tsx
ardasnturk: Yeah, you're absolutely right. 🤩

src/components/vouchers/VoucherRedeem.tsx
ardasnturk: I think it is an async function, so I thought it would make more sense to use it with `await`.

donni106: but are you awaiting something in order to go on with the result? i don't think so. this can be called and done, i guess.

ardasnturk: so far it's worked without `await`. I'll delete it and we'll see what happens.

donni106: sure, the method is still called ;)

ardasnturk: https://github.com/smart-village-solutions/smart-village-app-app/pull/1104/commits/888657a5fc0230d15449fb430c961ae348791255

PR #1105, opened 2024-05-08
feat: use vouchers in offline mode

ardasnturk:
- removed the check in `useEffect` to fix the issue where the `isLoggedIn` value was `false` because the `logInCallback` function was not executed again when the connection was lost, even though the user's token was registered on the device and logged in
- updated `addToStore` with `AsyncStorage.removeItem` to fix `memberId` not being deleted in `voucherHelper`

SVAK-90

### DIFF
--- a/src/helpers/voucherHelper.ts
+++ b/src/helpers/voucherHelper.ts
@@ -1,7 +1,6 @@
-import AsyncStorage from '@react-native-async-storage/async-storage';
 import * as SecureStore from 'expo-secure-store';
 
-import { addToStore, readFromStore } from './storageHelper';
+import { addToStore, readFromStore, removeFromStore } from './storageHelper';
 
 const VOUCHER_AUTH_TOKEN = 'VOUCHER_AUTH_TOKEN';
 export const VOUCHER_MEMBER_ID = 'VOUCHER_MEMBER_ID';
@@ -36,7 +35,7 @@ export const storeVoucherMemberId = (memberId?: string) => {
   if (memberId) {
     addToStore(VOUCHER_MEMBER_ID, memberId);
   } else {
-    AsyncStorage.removeItem(VOUCHER_MEMBER_ID);
+    removeFromStore(VOUCHER_MEMBER_ID);
   }
 };
 
@@ -50,7 +49,7 @@ export const voucherMemberId = async () => {
     memberId = await readFromStore(VOUCHER_MEMBER_ID);
   } catch {
     // Token deleted here so that it can be recreated
-    AsyncStorage.removeItem(VOUCHER_MEMBER_ID);
+    removeFromStore(VOUCHER_MEMBER_ID);
   }
 
   return memberId;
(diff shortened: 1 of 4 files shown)


src/helpers/voucherHelper.ts
```diff
@@ -36,7 +36,7 @@ export const storeVoucherMemberId = (memberId?: string) => {
   if (memberId) {
     addToStore(VOUCHER_MEMBER_ID, memberId);
   } else {
-    addToStore(VOUCHER_MEMBER_ID);
+    AsyncStorage.removeItem(VOUCHER_MEMBER_ID);
```
donni106, line 39: please create the missing `removeFromStore` method in our helper and use it here.

src/hooks/voucherHooks.ts
```diff
@@ -35,12 +35,7 @@ export const useVoucher = (): {
   }, []);
 
   useEffect(() => {
-    if (isConnected && isMainserverUp && !isLoggedIn) {
-      logInCallback();
-    } else {
-      setIsLoading(false);
-      setIsError(true);
-    }
+    logInCallback();
```
donni106, line 38: isn't `setIsLoading(false);`  missing now for some cases?

src/hooks/voucherHooks.ts
```diff
@@ -35,12 +35,7 @@ export const useVoucher = (): {
...
   }, [isConnected, isMainserverUp, logInCallback]);
```
donni106, line 39: do we still need both `isConnected, isMainserverUp` dependencies, even if we do not use its values?

donni106 (changes requested):

src/hooks/voucherHooks.ts
ardasnturk: There is `setIsLoading(false)` statement in `loginCallback`. Therefore, I do not think it is missing.

src/hooks/voucherHooks.ts
ardasnturk: I think it would be correct to re-run the function in case of a disconnection. If any error still occurs, the `isError` statement will be `true`.

src/helpers/voucherHelper.ts
ardasnturk: https://github.com/smart-village-solutions/smart-village-app-app/pull/1105/commits/36fc25dc31a63b93d2a0c9bd25c2393cafcd95ac

src/hooks/voucherHooks.ts
donni106: ok. and therefore my question again, do we need both dependencies?

ardasnturk: Now I understand. Only 1 of them will be enough. I think `isConnected` will be enough and `loginCallback`

ardasnturk: https://github.com/smart-village-solutions/smart-village-app-app/pull/1105/commits/451e27e255b6e901bbf03e76d3dbe7a1b70cc1df

donni106 (approved):

PR #1108, opened 2024-05-14
feat: add other offers section to detail screen

ardasnturk:
- added `genericItems` query with `ids` variables to `VoucherDetailScreen` to show current coupons list
- added the `id` of the vouchers to the `pointOfInterest` query in `vouchers.ts` to get the `ids` variables
- added filtering with `id` to not list the voucher shown in DetailScreen
- coded the listing codes in the same way as POI detail screen

SVAK-87

## Screenshots:

|voucher detail screen|voucher detail screen actual list|
|--|--|
Simulator Screenshot - iPhone 15 Pro Max - 2024-05-14 at 16 46 55|Simulator Screenshot - iPhone 15 Pro Max - 2024-05-14 at 16 46 58

### DIFF
--- a/src/screens/Voucher/VoucherDetailScreen.tsx
+++ b/src/screens/Voucher/VoucherDetailScreen.tsx
@@ -192,7 +192,7 @@ export const VoucherDetailScreen = ({ navigation, route }: StackScreenProps<any>
 
       {!!voucherListItems?.length && (
         <>
-          <SectionHeader title={texts.pointOfInterest.vouchers} />
+          <SectionHeader title={texts.pointOfInterest.vouchersMore} />
           <FlatList
             data={voucherListItems.slice(0, loadedVoucherDataCount)}
             renderItem={({ item }) => <VoucherListItem item={item} navigation={navigation} />}
--- a/src/screens/Voucher/VoucherDetailScreen.tsx
+++ b/src/screens/Voucher/VoucherDetailScreen.tsx
@@ -6,7 +6,7 @@ import { RefreshControl, ScrollView, StyleSheet } from 'react-native';
 import { NetworkContext } from '../../NetworkProvider';
 import {
   BoldText,
-  DataProviderButton,
+  Button,
   Discount,
   HtmlView,
   ImageSection,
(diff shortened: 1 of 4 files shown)


src/screens/Voucher/VoucherDetailScreen.tsx
```diff
@@ -40,10 +46,35 @@ export const VoucherDetailScreen = ({ navigation, route }: StackScreenProps<any>
...
+      QUERY_TYPES.VOUCHERS,
+      { [QUERY_TYPES.GENERIC_ITEMS]: vouchers },
+      undefined,
+      {
+        withDate: false
+      }
+    );
+
+    return parsedListItem?.filter((item: TVoucherItem) => item?.id !== queryVariables?.id);
+  }, [poiVouchersData]);
```
donni106, line 71: this could be refactored to querying generic items with certain ids as variables, which would return the voucher data directly. additionally the filter method for the current voucher is not necessary, as it will be excluded at querying the server. the voucher generic item ids needs to be passed instead of the `poiId`

donni106 (changes requested):

src/screens/Voucher/VoucherDetailScreen.tsx
ardasnturk: Can you please check the code again? I tried a different method and I also used `ids` in this method. I think it is easier and more understandable. I'll close this conversation for now.

src/screens/Voucher/VoucherDetailScreen.tsx
```diff
@@ -63,7 +56,39 @@ export const VoucherDetailScreen = ({ navigation, route }: StackScreenProps<any>
     quota,
     subtitle,
     title
-  } = data[QUERY_TYPES.GENERIC_ITEM];
+  } = data?.[QUERY_TYPES.GENERIC_ITEM] ?? {};
+
+  const ids = pointOfInterest?.vouchers
+    .map((voucher: TVoucherItem) => voucher.id)
+    .filter((id: string) => id !== queryVariables?.id);
```
donni106, line 63: please add `?` for `map` and `filter` as we do not know if `.vouchers` always return an array.

donni106: what about using `reduce` to have a single iteration instead of two for `map` and `filter`?

```
const ids = pointOfInterest?.vouchers?.reduce((acc, voucher) => {
    if (voucher.id !== queryVariables?.id) {
        acc.push(voucher.id);
    }
    return acc;
}, []);
```

src/screens/Voucher/VoucherDetailScreen.tsx
```diff
@@ -164,6 +189,28 @@ export const VoucherDetailScreen = ({ navigation, route }: StackScreenProps<any>
           </Wrapper>
         </>
       )}
+
+      {!!voucherListItems?.length && (
+        <>
+          <SectionHeader title={texts.pointOfInterest.vouchersMore} />
```
donni106, line 195: i have updated this to

image

src/screens/Voucher/VoucherDetailScreen.tsx
ardasnturk: Yeah, it's better. I'll update this.

ardasnturk: https://github.com/smart-village-solutions/smart-village-app-app/pull/1108/commits/19532129b36ff2bcce22bbe2a8c2e8ad90e3f93b

donni106 (approved):

PR #1113, opened 2024-05-16
feat(voucher): send saved redeemed vouchers to the server

ardasnturk:
- developed the `useRedeemLocalVouchers` hook and added it to `HomeScreen` so that if there is a coupon saved on the device, the user can redeem coupons online no matter which screen the user is on in the whole app

SVAK-35

### DIFF
--- a/src/hooks/voucherHooks.ts
+++ b/src/hooks/voucherHooks.ts
@@ -1,7 +1,10 @@
 import { useCallback, useContext, useEffect, useState } from 'react';
+import { useMutation } from 'react-apollo';
 
-import { voucherAuthToken, voucherMemberId } from '../helpers/voucherHelper';
 import { NetworkContext } from '../NetworkProvider';
+import { addToStore, readFromStore } from '../helpers';
+import { VOUCHER_TRANSACTIONS, voucherAuthToken, voucherMemberId } from '../helpers/voucherHelper';
+import { REDEEM_QUOTA_OF_VOUCHER } from '../queries/vouchers';
 
 export const useVoucher = (): {
   refresh: () => Promise<void>;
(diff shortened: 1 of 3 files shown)


src/hooks/voucherHooks.ts
```diff
@@ -46,3 +49,60 @@ export const useVoucher = (): {
...
+  }, []);
+
+  useEffect(() => {
+    if (isConnected && isMainserverUp) {
+      redeemLocalVouchers();
+    }
+  }, [isConnected, isMainserverUp, redeemLocalVouchers]);
+
+  return {
+    refresh: redeemLocalVouchers,
```
donni106, line 104: why is this named `refresh`?

src/hooks/voucherHooks.ts
```diff
@@ -46,3 +49,60 @@ export const useVoucher = (): {
...
+    isLoading,
+    isError
+  };
```
donni106, line 107: do we need to return anything? i see that we are not using any of the values, right?!

src/hooks/voucherHooks.ts
```diff
@@ -46,3 +49,60 @@ export const useVoucher = (): {
...
+    } finally {
+      setIsLoading(false);
+    }
+  }, []);
+
+  useEffect(() => {
+    if (isConnected && isMainserverUp) {
+      redeemLocalVouchers();
+    }
+  }, [isConnected, isMainserverUp, redeemLocalVouchers]);
```
donni106, line 101: `isMainserverUp` needed again?

src/hooks/voucherHooks.ts
```diff
@@ -46,3 +49,60 @@ export const useVoucher = (): {
...
+                memberId
+              }
+            });
+          } catch (e) {
+            console.error(e);
+            error = true;
+            break;
+          }
+        }
+        if (!error) {
```
donni106, line 85: please add a space before the condition starts

donni106 (changes requested):

src/hooks/voucherHooks.ts
ardasnturk: https://github.com/smart-village-solutions/smart-village-app-app/commit/f2a3dfc7af2f00510b8f4a9ccd27aeb503edd0de

src/hooks/voucherHooks.ts
ardasnturk: We don't really need it. So I'm deleting it.

src/hooks/voucherHooks.ts
ardasnturk: Yeah, we don't need any turns. I'm deleting the returns.

src/hooks/voucherHooks.ts
ardasnturk: There's no reason. Because I copy and paste.

donni106 (approved):

PR #1114, opened 2024-05-17
fix: fetch more vouchers redeemed

ardasnturk:
- added `VOUCHERS_REDEEMED` control to `fetchMoreData` to list all redeemed coupons
- added filtering according to the ids of the data that the list has to prevent data repetition
- added `ListEmptyComponent` prop to `VoucherList` to show when data is categorised for display in the list

SVAK-35


src/components/screens/PointOfInterest.js
```diff
@@ -76,10 +76,11 @@ export const PointOfInterest = ({ data, hideMap, navigation, route }) => {
   const voucherListItems = useMemo(() => {
     return parseListItemsFromQuery(
       QUERY_TYPES.VOUCHERS,
-      { [QUERY_TYPES.GENERIC_ITEMS]: vouchers },
+      { [QUERY_TYPES.VOUCHERS]: vouchers },
       undefined,
       {
-        withDate: false
+        withDate: false,
+        queryKey: QUERY_TYPES.VOUCHERS
```
donni106, line 83: this should be `QUERY_TYPES.GENERIC_ITEMS` i guess, compared to `query == QUERY_TYPES.VOUCHERS_REDEEMED ? QUERY_TYPES.VOUCHERS : QUERY_TYPES.GENERIC_ITEMS`

src/components/vouchers/VoucherList.tsx
```diff
@@ -115,13 +125,15 @@ export const VoucherList = ({
       estimatedItemSize={MAX_INITIAL_NUM_TO_RENDER}
       keyExtractor={keyExtractor}
       ListFooterComponent={() => {
-        if (data?.length >= MAX_INITIAL_NUM_TO_RENDER) {
+        if (sectionedData?.length >= MAX_INITIAL_NUM_TO_RENDER) {
```
donni106, line 128: this just works because of FlashList right? otherwise with SectionList we could not count the object

donni106 (changes requested):

src/components/screens/PointOfInterest.js
ardasnturk: I think `QUERY_TYPES.VOUCHERS` is correct. Because this is not a `genericItems` query. The data is in the `vouchers` object inside the POI query.  If we update `queryKey` to `genericItems`, it will be `undefined`. Because there is no data inside the `genericItems` object.

donni106: ah! inside POI object, yes, true, thanks!

src/components/vouchers/VoucherList.tsx
donni106: but in sectioned data the section items are included in the length, which distorts the data count?

donni106: i reverted to data to count the real data amount

PR #1118, opened 2024-05-21
feat: save images captured with the camera to the gallery

ardasnturk:
- added `expo-media-library` package needed to save images to gallery
- added `saveImage` prop to `useCaptureImage` hook so that captured images can be saved to gallery
- added the feature that if the `saveImage` prop is `true`, the picture taken is added to the album with the same name according to the albums previously created or on the device, or a new album is created and the picture is added to the new album
- added album name to be created in gallery as app name

SUE-79

A new app build needs to be created due to native dependencies.  
`yarn ios -d` or `yarn android` commands can be used.

### DIFF
--- a/src/hooks/imagePicker.ts
+++ b/src/hooks/imagePicker.ts
@@ -6,10 +6,40 @@ import {
   requestCameraPermissionsAsync,
   requestMediaLibraryPermissionsAsync
 } from 'expo-image-picker';
+import {
+  addAssetsToAlbumAsync,
+  createAlbumAsync,
+  createAssetAsync,
+  getAlbumAsync,
+  requestPermissionsAsync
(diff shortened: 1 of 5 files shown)


src/hooks/imagePicker.ts
```diff
@@ -77,10 +86,38 @@ export const useCaptureImage = (
...
+
+      if (saveImage) {
+        await saveImageToGallery(uri);
+      }
+
       return result.assets[0];
     }
   }, [onChange]);
 
+  const saveImageToGallery = async (uri: string) => {
```
donni106, line 100: this does not need to be part of the hook and would be nice to be defined before used for better readability and understanding.

donni106 (approved):

PR #1121, opened 2024-05-22
feat(sue): connect with the area service

ardasnturk:
- added `apiKey` and `serverUrl` information to add connection with field service can be updated via main-server as `requests` request
- added `zipCode` state due to late change of `getValue('zipCode')` and no data from area service
- added rules that `limitOfCity` must be true and `zipCode` must be 5 digits to activate an area service request
- added rule to show `errorMessage` alert that can be updated via main server in case the `limitOfCity` value that can be updated via main server does not match with area service
- updated area service link with `id` instead of `getByPostalCode`
- added comparing postcode data from area service with the postcodes of the location selected from the map
- added errors in the `reverseGeocode` function to catch the error message
- updated `onMapPress` and `onMyLocationButtonPress` props with then-catch structure to catch errors added to `reverseGeocode` function
- added `areaServiceLoading` to fix the problem of the warning that you are outside the area that appears when clicking on the desired area on the map because the data in the area service is not loaded correctly
- extended the control in the Map component to fix the problem of empty `calloutText` when clicking on the marker in the marked area
- extended the control within `onMapPress` in `SueReportLocation` to fix the location not found warning that appears when `calloutText` is pressed

SUE-57

### DIFF
--- a/src/screens/SUE/SueReportScreen.tsx
+++ b/src/screens/SUE/SueReportScreen.tsx
@@ -4,11 +4,12 @@ import * as Location from 'expo-location';
 import * as ScreenOrientation from 'expo-screen-orientation';
 import parsePhoneNumber from 'libphonenumber-js';
 import React, { useContext, useEffect, useLayoutEffect, useRef, useState } from 'react';
-import { useForm, UseFormGetValues, UseFormSetValue } from 'react-hook-form';
+import { UseFormGetValues, UseFormSetValue, useForm } from 'react-hook-form';
 import { ActivityIndicator, Alert, Keyboard, ScrollView, StyleSheet, View } from 'react-native';
 import { Divider } from 'react-native-elements';
-import { useMutation } from 'react-query';
+import { useMutation, useQuery } from 'react-query';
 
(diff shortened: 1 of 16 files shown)


src/screens/SUE/SueReportScreen.tsx
```diff
@@ -279,10 +306,7 @@ export const SueReportScreen = ({
             return texts.sue.report.alerts.zipCode;
           }
 
-          if (
-            !!limitOfCity &&
-            limitOfCity.toLocaleLowerCase() !== getValues().city.toLocaleLowerCase()
-          ) {
+          if (!!limitOfCity && !areaServiceData?.postalCodes?.includes(getValues('zipCode'))) {
```
donni106, line 309: can we ignore `limitOfCity` here, as it controls the query with `enabled`? so if `limitOfCity` is false, `areaServiceData?` isn't defined.

ardasnturk: Yes, that's right. If the `limitOfCity` value is not set, the query will not work. So the control here can be removed.

donni106 (approved):

PR #1124, opened 2024-05-31
fix: apply patch to react-native-maps to fix street name overlap on iOS 16+

d4v3d4vE:
Added a patch to react-native-maps package to resolve an issue where street names overlap custom map overlays on iOS 16 and above.

Thanks to [@cixio](https://github.com/cixio) for the starting solution to use `MKOverlayLevelAboveLabels` https://github.com/react-native-maps/react-native-maps/pull/4549/commits/49bbdf693d9ddcd052fec2e5bec90349cc04210a

**Changes:**
- Updated AirMap.m to use `MKOverlayLevelAboveLabels` for overlays.
- Added patch-package and postinstall script to package.json to ensure patch is applied automatically.
- Included the generated patch file in the patches directory.

This ensures consistent behavior across all developer environments.

SUE-43

Usage:
- To see the wanted results, make sure to run `yarn prebuild` and rebuild the app

**Additional information:**
The +1.7.1 in the patch file name refers to the version of the react-native-maps package that was used at the time the patch was created. If the package is updated, the patch file name might change, and it might also be necessary to review and adjust the patch itself to ensure it continues to work.

**Checking the Patch:**
- Verify if the patch is still necessary and compatible. Navigate to the file the patch refers to (AirMap.m) and check if the structure of the file has changed.

**Creating a New Patch (if necessary):**
- If the patch is still necessary but the file has changed, make the necessary changes and create a new patch:
`npx patch-package react-native-maps`

- This will create a new patch file with the current version of the package, e.g., react-native-maps+1.8.0.patch.

### DIFF
--- a/package.json
+++ b/package.json
@@ -19,7 +19,8 @@
     "lint": "./node_modules/.bin/eslint __tests__ src --ext .js,.jsx,.ts,.tsx",
     "test": "./node_modules/.bin/jest",
     "test:debug": "node --inspect-brk node_modules/jest/bin/jest.js --runInBand",
-    "test:coverage": "node_modules/.bin/jest --coverage"
+    "test:coverage": "node_modules/.bin/jest --coverage",
+    "postinstall": "patch-package"
   },
   "jest": {
     "preset": "jest-expo",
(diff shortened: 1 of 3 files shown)


donni106 (approved):

ardasnturk:
So should we add a new step to create a build from now on? If we want to create a build, how should we do it? Can you explain step by step? Is just `yarn prebuild` enough or do we also need to run the `postinstall` script?

d4v3d4vE:
> So should we add a new step to create a build from now on? If we want to create a build, how should we do it? Can you explain step by step? Is just `yarn prebuild` enough or do we also need to run the `postinstall` script?

Its integrated in our current build process, with the use of yarn prebuild the script executes automatically and applies the patch to the `react-native-maps` package

ardasnturk (approved):

PR #1131, opened 2024-06-06
feat: integrate app with config api

ardasnturk:
- added `ConfigurationsProvider` to control configurations in the whole app
- added `configurations.tsx` to provide connection with config api
- added all requests to `configuration.tsx` as there are 3 different config requests
- returned in the desired format to better distribute the config data within the app and to make it more understandable
- added control to the config api to run only if the sue module is active, since this config is only specific to the sue module
- added `configurations` file to add standard sue configuration to provider and added default objects in `sueDefaultConfigurations`
- added to `ConfigurationsProvider` to `appDesignSystem` config, which is required to update the app's appearance
- added `appDesignSystem` config as a separate JSON on main-server
- updated `appDesignSystem` config previously called from `SettingsContext` with `ConfigurationsContext`
- added `sueReportProgress` content created on main-server to configurations context
- added new functions to `storageHelper` to save config to device memory
- added the `appDesignSystem` property to the config context from within `globalSettings` instead of a separate `staticContent`
- added props to update the `Map` component with data from the config api
- added `configuration` prop to distribute data from config context to components on report screen
- added configuration data to control the size of images that can be uploaded
- added privacyPolicy value from config data to `SueReportUser` component
- fixed an error with the map on the report screen
- added `maxFileUploads` control to `ImageSelector` to make the maximum number of images that can be uploaded adjustable with config
- added `allowedAttachmentTypes` control to check whether the image format is in the desired format
- updated the format of the required inputs from the config api to the desired format with the `sueProgressWithRequiredInputs` function because the format of the required inputs does not match the structure of the code
- combined the `sueProgress` created by main-server with the required inputs sent by config api and set to `sueProgressWithConfig` state

SUE-97

## Test:

If you are not in Halle or do not have a picture taken in Halle, open areaService.tsx and update the `id` as follows to be able to upload pictures.

```
const id = geoMap.areas[0].id || apiConfig.areaService.id;
```
to
```
const id = "18116"
```

This `id` belongs to Kiel.


src/config/configurations/index.js
donni106, line 1: we should not have a configurations named folder inside of config. this is too much duplication.
also the file names do not need the suffix configurations i guess.
i suggest:

```
- config
  - sue
    - defaults.ts
    - index.ts
  - appDesignSystem
    - defaults.ts
    - index.ts
  - index.ts
  - texts.ts
```

this can grow with future extensions.

src/config/configurations/sueDefaultConfigurations.ts
```diff
@@ -0,0 +1,74 @@
...
+  jurisdictionId: '',
+  serverUrl: '',
+  testApi: { apiKey: '', jurisdictionId: '', serverUrl: '' },
+  whichApi: ''
+};
+
+const geoMapDefaultConfig = {
+  areas: [],
+  center: {},
+  clisterTreshold: 2,
```
donni106, line 13: what is clister? or a typo for cluster?

src/components/consul/selectors/ImageSelector.js
```diff
@@ -171,7 +180,9 @@ export const ImageSelector = ({
           <Input {...item} control={control} hidden name={name} value={JSON.parse(value)} />
 
           <Button
-            disabled={values?.length >= 5}
+            disabled={
+              values?.length >= (parseInt(configuration?.limitation?.maxFileUploads?.value) ?? 5)
```
donni106, line 184: we have defaults in the code, right. so the hardcoded `5` should come from that defaults i guess.

src/components/SUE/report/SueReportUser.tsx
```diff
@@ -104,7 +106,10 @@ export const SueReportUser = ({
             <Checkbox
               checked={!!value}
               onPress={() => onChange(!value)}
-              link="https://smart-village.app/datenschutzerklaerung/"
+              link={
+                configuration?.limitation?.privacyPolicy?.value ||
+                'https://smart-village.app/datenschutzerklaerung/'
```
donni106, line 111: we have defaults in the code, right. so the hardcoded url should come from that defaults i guess.

src/components/SUE/report/SueReportLocation.tsx
```diff
@@ -255,7 +254,7 @@ export const SueReportLocation = ({
               setSelectedPosition(nativeEvent.coordinate);
 
               try {
-                await reverseGeocode(nativeEvent.coordinate);
+                await handleGeocode(nativeEvent.coordinate);
```
donni106, line 257: is this change intended here with this PR?

src/config/configurations/sueDefaultConfigurations.ts
```diff
@@ -0,0 +1,74 @@
...
+  {
+    content: '',
+    requiredInputs: [],
+    subtitle: '',
+    title: ''
+  }
+];
+
+export const defaultSueAppConfig = {
+  apiConfig: defaultApiConfig,
```
donni106, line 68: like the other this would fit more naming it `apiDefaultConfig`?

src/config/configurations/sueDefaultConfigurations.ts
```diff
@@ -0,0 +1,74 @@
...
+  geoMap: geoMapDefaultConfig,
+  limitation: limitationDefaultConfig,
+  limitOfArea: defaultLimitOfAreaConfig,
+  requiredFields: defaultRequiredFieldsConfig,
+  sueProgress: defaultSueProgressConfig
```
donni106, line 73: what about shorting the variable names like: apiDefaults, geoMapDefaults, limitationDefaults, ...?

src/config/texts.js
```diff
@@ -945,6 +945,7 @@ export const texts = {
           title: 'Eingabe löschen'
         },
         hint: 'Hinweis',
+        imageType: 'Bitte laden Sie formatkonforme Inhalte hoch.',
```
donni106, line 948: 'Der verwendete Dateityp wird nicht unterstützt.'

src/queries/SUE/areaService.tsx
```diff
@@ -1,10 +1,13 @@
...
-  const globalSettings = await storageHelper.globalSettings();
-  const { apiConfig = {} } = globalSettings?.settings?.sue || {};
+  const configurations = await storageHelper.configurations();
+  const { sueConfig = {} } = configurations;
+  const { apiConfig = {}, geoMap = {} } = sueConfig;
 
-  const { apiKey, serverUrl, id } = apiConfig.areaService || {};
+  const { apiKey, serverUrl } = apiConfig.areaService || {};
+
+  const id = geoMap.areas[0].id || apiConfig.areaService.id;
```
donni106, line 10: do you think we could need some `?` to secure the objects accesses?

src/queries/SUE/configurations.tsx
```diff
@@ -0,0 +1,43 @@
...
+  } = await fetchSueEndpoints();
+
+  const configPromise = await (await fetch(`${sueConfigurationsUrl}`, sueFetchObj)).json();
+  const geoMapPromise = await (await fetch(`${sueGeoMapConfigurationUrl}`, sueFetchObj)).json();
+  const requiredFieldPromise = await (
+    await fetch(`${sueContactRequiredFieldConfigurationUrl}`, sueFetchObj)
+  ).json();
+
+  return Promise.all([configPromise, geoMapPromise, requiredFieldPromise])
+    .then(([configResponse, geoMapResponse, requiredFieldResponse]) => {
```
donni106, line 21: what about some refactoring here. i suggest something more readable without that much awaits chained. we can have some more code to optimize understandment.

---

a ChatGPT summary:

Here's a refactored version of your code that optimizes readability and maintains efficiency by handling the fetch requests in parallel without chaining unnecessary `await` statements on the `.json()` calls:

```javascript
async function fetchConfigurationData() {
  // Use Promise.all to handle parallel fetch requests
  const urls = [
    sueConfigurationsUrl,
    sueGeoMapConfigurationUrl,
    sueContactRequiredFieldConfigurationUrl
  ];

  // Map over the URLs and fetch each configuration
  const fetchPromises = urls.map(url => fetch(url, sueFetchObj));

  try {
    // Wait for all fetch requests to resolve
    const responses = await Promise.all(fetchPromises);

    // Convert each response to JSON
    const jsonDataPromises = responses.map(response => response.json());

    // Resolve all JSON conversion promises
    const [configResponse, geoMapResponse, requiredFieldResponse] = await Promise.all(jsonDataPromises);

    // Handle the resolved data as needed
    return { configResponse, geoMapResponse, requiredFieldResponse };
  } catch (error) {
    // Handle errors, such as network issues or invalid JSON
    console.error('Failed to fetch configurations:', error);
    throw error;  // Rethrow or handle error as needed
  }
}
```

#### Changes and Improvements:
1. **Parallel Fetches:** This refactor ensures that the initial `fetch` requests are sent out in parallel, without awaiting each one sequentially.
2. **Promise Handling:** The `.json()` method is called on each response after all fetch requests have resolved, which maintains parallelism in processing responses as well.
3. **Error Handling:** The try-catch block catches errors from either the fetch or JSON parsing phase and handles them appropriately, making the function more robust.
4. **Clean Structure:** Using functions and separating concerns makes the code more modular and easier to read or modify.

---

please have a closer look and try different solutions as well to find the one that fits well for us.

src/screens/SUE/SueReportScreen.tsx
```diff
@@ -27,12 +27,47 @@ import {
...
+      requiredInputs[field] = fields[section][field];
+    }
+  }
+
+  return progress.map((item) => {
+    if (!item.requiredInputs) {
+      item.requiredInputs = [];
+    }
+
+    item.requiredInputs = item.requiredInputs.filter((key) => requiredInputs[key] !== false);
```
donni106, line 59: what about reducing this to:

```
item.requiredInputs = (item.requiredInputs || []).filter((key) => requiredInputs[key] !== false);
```

src/screens/SUE/SueReportScreen.tsx
```diff
@@ -288,6 +348,8 @@ export const SueReportScreen = ({
           const images = JSON.parse(getValues().images);
 
           let totalSize = 0;
+          const totalSizeLimit = parseInt(limitation?.maxAttachmentSize?.value) || 31457280;
```
donni106, line 351: we have defaults in the code, right. so the hardcoded `31457280` should come from that defaults i guess.

src/ConfigurationsProvider.tsx
```diff
@@ -0,0 +1,51 @@
...
+  const { globalSettings } = useContext(SettingsContext);
+  const { settings, appDesignSystem = {} } = globalSettings;
+  const { sue = {} } = settings || {};
+
+  const [configurations, setConfigurations] = useState(defaultConfiguration);
+
+  const { data: sueConfigData } = useQuery(
+    [QUERY_TYPES.SUE.CONFIGURATIONS],
+    () => getQuery(QUERY_TYPES.SUE.CONFIGURATIONS)(),
+    { enabled: !!sue }
```
donni106, line 27: check this:

image

i think you need to check more deeply, if `sue` data is given

src/config/configurations/sueDefaultConfigurations.ts
ardasnturk: It should be `clusterThreshold`. It's a typo.

src/components/consul/selectors/ImageSelector.js
ardasnturk: Not until now. But I added `MAX_FILE_UPLOADS`.

ardasnturk: Or are you talking about the default data in `sueDefaultConfigurations`?

src/components/SUE/report/SueReportUser.tsx
ardasnturk: Are you talking about the default data in `sueDefaultConfigurations`?

src/components/SUE/report/SueReportLocation.tsx
ardasnturk: I have seen this error and wanted to correct it.

src/components/SUE/report/SueReportUser.tsx
donni106: should be there, or am i wrong? it is the default and fallback for sue here?
do you think it does not fit to the defaults?

src/components/SUE/report/SueReportLocation.tsx
donni106: ok yes, I already thought so

src/components/SUE/report/SueReportUser.tsx
ardasnturk: No, that's right. But I wonder if the default will not be overwritten if there is no data? For example, `privacyPolicy.value = "link"` by default. But if no data comes from the api, won't this default be overwritten? I thought it would be. 😔

src/config/configurations/sueDefaultConfigurations.ts
ardasnturk: Yes, it makes sense. Since they are already config objects, they do not need to be called config again. 👍

src/queries/SUE/areaService.tsx
ardasnturk: Yes, that's right. I forgot about this safety. 👍

src/screens/SUE/SueReportScreen.tsx
ardasnturk: Yes, it makes sense. Then we don't need the if check above.

src/ConfigurationsProvider.tsx
ardasnturk: How about a control like that?
`!!Object.keys(sue).length;`

src/components/SUE/report/SueReportUser.tsx
donni106: ok, in this case, we can think about it in a later process when integrating more and more configurations.

src/config/configurations/index.js
ardasnturk: https://github.com/smart-village-solutions/smart-village-app-app/pull/1131/commits/c15b9ba52d90b78daa9a86dbd37b75355393aad4

src/config/configurations/sueDefaultConfigurations.ts
ardasnturk: https://github.com/smart-village-solutions/smart-village-app-app/pull/1131/commits/4b4abafe9ad39a3435adf6e113da7b9ec15e7920

src/queries/SUE/configurations.tsx
ardasnturk: This code doesn't do exactly what we want. But it is much more understandable. So I merged the two codes together and came up with a code that is more understandable and does the same thing.

https://github.com/smart-village-solutions/smart-village-app-app/pull/1131/commits/8bed36e034809bec34bcfb61029f917f0feabfa4

donni106 (approved):

PR #1133, opened 2024-06-10
feat: add category details texts under tiles

ardasnturk:
- updated `scrollViewContentRef` to `[]` instead of `null` to create a separate scroll view ref for each step
- added `onContentSizeChange` prop to calculate the content size at each step and added `handleContentSizeChange` helper
- updated the `scrollTo` function to automatically scroll the scroll view according to the size of the content at each step
- added `useEffect` to scroll the scroll view at that step in the content size when the `serviceCode` changes
- updated the `serviceCode` state to `service` because the description and title data of the service is needed
- added the category description to `SueReportScreen` if a category is selected

SUE-84

## Screenshots:

|before|after|
|--|--|
Simulator Screenshot - iPhone 15 Pro Max - 2024-06-10 at 15 29 57|Simulator Screenshot - iPhone 15 Pro Max - 2024-06-10 at 15 29 48

### DIFF
--- a/src/screens/SUE/SueReportScreen.tsx
+++ b/src/screens/SUE/SueReportScreen.tsx
@@ -11,10 +11,12 @@ import { useMutation, useQuery } from 'react-query';
 
 import { ConfigurationsContext } from '../../ConfigurationsProvider';
 import {
+  BoldText,
   Button,
   DefaultKeyboardAvoidingView,
   HeaderRight,
   LoadingContainer,
+  RegularText,
   SafeAreaViewFlex,
   SueReportDescription,
   SueReportLocation,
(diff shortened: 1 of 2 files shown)


src/screens/SUE/SueReportScreen.tsx
```diff
@@ -650,5 +687,8 @@ const styles = StyleSheet.create({
   },
   contentContainer: {
     width: device.width
+  },
+  noPaddingTop: {
+    paddingTop: 0
```
donni106, line 692: this makes the divider to close to the buttons on the next screens.

image

we need the padding like before.

donni106, line 692: and regarding the divider with open description, it should be full width to correctly divide the screen parts. left and right parts are missing of the divider line:

image

ardasnturk: That's right! I missed that. I've updated it now. Now there is `noPaddingBottom` instead of `noPaddingTop`. So it will appear correctly after step 2. 👍

ardasnturk: divider updated to the width of the screen 👍

donni106 (approved):

PR #1135, opened 2024-06-11
feat(sue): make feedback form adjustable from main server

ardasnturk:
- connected to the `showFeedbackSection` value in `ConfigurationsContext` to make the feedback section shown after sending a report adjustable from main-server

SUE-78

## Tests:

To test, set the if condition in line 437 of `SueReporScreen` to `true`.

`if (isDone || isLoading) {...` -> `if (true) {...`

Then set the condition on line 98 in the `SueReportSend` component to `true`.

`{!isLoading && isDone && (...` -> `{true && (...`

Then, when you go to the report sending screen in the app, you will see the feedback form.

### DIFF
--- a/src/config/sue/default.ts
+++ b/src/config/sue/default.ts
@@ -55,6 +55,8 @@ const requiredFieldsDefaults = {
   }
 };
 
+const showFeedbackSectionDefaults = false;
+
 const sueProgressDefaults = [
   {
     content: '',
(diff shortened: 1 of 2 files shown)


src/config/sue/default.ts
```diff
@@ -55,6 +55,8 @@ const requiredFieldsDefaults = {
   }
 };
 
+const showFeedbackSection = false;
```
donni106, line 58: this shouldn't be an extra variable just to hold the value false here, right?
can be `showFeedbackSection: false` directly in line 76

donni106, line 58: should we name it `showFeedbackSectionDefault` to match the others?

ardasnturk: Yes, that's right. 👍

ardasnturk: https://github.com/smart-village-solutions/smart-village-app-app/commit/150eb625b1f2e7b261cae293fe76b05a6c4aa6e3

PR #1137, opened 2024-06-12
feat(sue): add current location to full screen map view

ardasnturk:
- added button to show current location in full screen map view 

SUE-44

image

### DIFF
--- a/src/screens/SUE/SueMapScreen.tsx
+++ b/src/screens/SUE/SueMapScreen.tsx
@@ -1,5 +1,6 @@
 import { RouteProp } from '@react-navigation/native';
 import { StackNavigationProp } from '@react-navigation/stack';
+import * as Location from 'expo-location';
 import _upperFirst from 'lodash/upperFirst';
 import React, { useContext, useMemo, useState } from 'react';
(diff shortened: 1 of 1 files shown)


src/screens/SUE/SueMapScreen.tsx
```diff
@@ -135,7 +149,18 @@ export const SueMapScreen = ({ navigation, route }: Props) => {
...
         minZoom={geoMap?.minZoom}
+        onMyLocationButtonPress={() => {
+          const location = position || lastKnownPosition;
+
+          setUpdatedRegion(true);
+          setCurrentPosition(location?.coords);
+
+          setTimeout(() => {
+            setUpdatedRegion(false);
+          }, 300);
```
donni106, line 162: why is a timeout needed? and why 300 and not 100 like a few lines beneath?

ardasnturk: Because once the user has displayed his own location on the map, the location on the map is not updated the second time the button is pressed. Therefore, the value of the `updateRegion` state had to change.

There's no particular reason for 300. I updated it to 100.

donni106 (approved):

PR #1139, opened 2024-06-12
feat(sue): add warning when address inputs are filled from the image

ardasnturk:
SUE-100

## Test:
You can only get this warning if you add an image of the specified location in the config. Otherwise this feature will not work. 

To do this:
Go to areaService.tsx and update the id to 18116.
Then select a photo taken in Kiel from the gallery. If the address information is loaded correctly, you will not receive an error warning. Press the Continue button and see the new warning.

## Screenshots:

image

### DIFF
--- a/src/screens/SUE/SueReportScreen.tsx
+++ b/src/screens/SUE/SueReportScreen.tsx
@@ -374,6 +374,10 @@ export const SueReportScreen = ({
         if (isAnyInputMissing) {
           return texts.sue.report.alerts.missingAnyInput;
         }
+
+        if (selectedPosition) {
+          Alert.alert(texts.sue.report.alerts.hint, texts.sue.report.alerts.imageLocation);
+        }
         break;
       case 2:
         if (getValues().houseNumber && !getValues().street) {
--- a/src/components/consul/selectors/ImageSelector.js
+++ b/src/components/consul/selectors/ImageSelector.js
@@ -163,6 +163,8 @@ export const ImageSelector = ({
         coordinateCheck.setSelectedPosition(location);
         coordinateCheck.setUpdateRegionFromImage(true);
       } catch (error) {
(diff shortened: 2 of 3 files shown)


donni106:
I cannot reach that part with that image you have used. I receive:

image

Can you provide an image or setup to use for testing please.

src/screens/SUE/SueReportScreen.tsx
```diff
@@ -374,6 +374,10 @@ export const SueReportScreen = ({
         if (isAnyInputMissing) {
           return texts.sue.report.alerts.missingAnyInput;
         }
+
+        if (selectedPosition) {
+          Alert.alert(texts.sue.report.alerts.hint, texts.sue.report.alerts.imageLocation);
```
donni106, line 379: in my case the alert came too late when i switched to the next step:

image

ardasnturk: What exactly do you mean when you say it came late? You pressed the button and were directed to the next step. How many seconds afterwards did you get the warning? The fact that the alert comes after the continue button is pressed does not mean that it comes late. But if you receive the alert after 3-4 seconds, then there is a problem.

donni106: in your screenshot the alert cam on step 2, where i upload the image. this is what i expect.
in my case i needed to go to the next step to get the alert, which is confusing.

ardasnturk: Yes, but the ticket says it's supposed to be like this. If you want to see this warning when the image is uploaded, we will see this warning every time the image is selected. I think it's more confusing this way.

donni106: all right

donni106 (approved):

PR #1141, opened 2024-06-13
fix: the problem that the button in the feedback section does not appear

ardasnturk:
- added a `View` up to the height of the keyboard to fix the problem of the button staying behind the keyboard when the keyboard is opened on the android platform

SUE-78

## Screenshots:

|before|after|
|--|--|
Screenshot_20240613-090921|Screenshot_20240613-090948

### DIFF
--- a/src/components/SUE/report/SueReportSend.tsx
+++ b/src/components/SUE/report/SueReportSend.tsx
@@ -6,7 +6,8 @@ import { Alert, Keyboard, ScrollView, StyleSheet, View } from 'react-native';
 import { Divider, Rating } from 'react-native-elements';
 
 import { ConfigurationsContext } from '../../../ConfigurationsProvider';
-import { colors, normalize, texts } from '../../../config';
+import { colors, device, normalize, texts } from '../../../config';
+import { useKeyboardHeight } from '../../../hooks';
 import { QUERY_TYPES, createQuery } from '../../../queries';
 import { ScreenName } from '../../../types';
 import { Button } from '../../Button';
@@ -37,6 +38,7 @@ export const SueReportSend = ({
   const { title: loadingTitle = '', subtitle: loadingSubtitle = '' } = reportSendLoading;
   const { title: doneTitle = '', subtitle: doneSubtitle = '' } = reportSendDone;
 
+  const keyboardHeight = useKeyboardHeight();
   const scrollViewRef = useRef(null);
 
   const { control, reset, handleSubmit } = useForm({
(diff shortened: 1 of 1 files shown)


src/components/SUE/report/SueReportSend.tsx
```diff
@@ -58,15 +67,26 @@ export const SueReportSend = ({
...
       dataType: 'json',
       dataSource: 'form',
-      content: JSON.stringify(createAppUserContentNewData)
+      content: JSON.stringify({
+        action: 'notify_and_destroy',
+        appVersion: appJson.expo.version,
+        created_At: moment().format('YYYY-MM-DD HH:mm:ss'),
+        message: createAppUserContentNewData.message,
+        rating: createAppUserContentNewData.ratingCount
+      })
```
donni106, line 76: what about `...createAppUserContentNewData, other data`? this also needs to name `ratingCount` to `rating`, right!? it seems strange to have a different name in here, we can use the name from the api

src/components/SUE/report/SueReportSend.tsx
```diff
@@ -158,6 +178,10 @@ export const SueReportSend = ({
                     }
                   />
                 </View>
+
+                {device.platform === 'android' && (
+                  <View style={{ height: normalize(keyboardHeight) * 0.8 }} />
```
donni106, line 183: why 0.8?

src/components/SUE/report/SueReportSend.tsx
ardasnturk: Yes, that's right. I will update. 👍

src/components/SUE/report/SueReportSend.tsx
ardasnturk: I found 0.8 by experimenting. It gives the right height for the keyboard.

donni106 (approved):

src/components/SUE/report/SueReportSend.tsx
```diff
@@ -32,16 +35,22 @@ export const SueReportSend = ({
...
+  const [showFeedbackSection, setShowFeedbackSection] = useState(feedbackSection);
 
+  const keyboardHeight = useKeyboardHeight();
   const scrollViewRef = useRef(null);
 
   const { control, reset, handleSubmit } = useForm({
     defaultValues: {
       message: '',
-      ratingCount: 4
+      rating: 4
```
donni106, line 53: 4 should be default? let's have a global setting for this in `sueReportScreen` like `showFeedbackSection`!

ardasnturk: yes it makes sense. For this I made the `defaultRating` value adjustable from `globalSettings`. I gave the value 4 as default.

PR #1144, opened 2024-06-17
feat(sue): update keys according to config api

ardasnturk:
- updated the keys used in the app to update the inputs with configApi
  - `firstName` -> `name`
  - `lastName` -> `familyName`
  - `zipCode` -> `postalCode`
- updated types to simplify the code

SUE-97

### DIFF
--- a/src/screens/SUE/SueReportScreen.tsx
+++ b/src/screens/SUE/SueReportScreen.tsx
@@ -26,13 +26,15 @@ import {
   SueReportUser,
   Wrapper
 } from '../../components';
-import { colors, device, normalize, texts } from '../../config';
+import { colors, consts, device, normalize, texts } from '../../config';
 import { addToStore, formatSize, readFromStore } from '../../helpers';
 import { useKeyboardHeight } from '../../hooks';
 import { QUERY_TYPES, getQuery } from '../../queries';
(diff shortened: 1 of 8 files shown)


src/screens/SUE/SueReportScreen.tsx
```diff
@@ -305,14 +306,16 @@ export const SueReportScreen = ({
...
       !!sueReportData.city
     ) {
-      addressString = `${sueReportData.street}; ${sueReportData.houseNumber}; ${sueReportData.zipCode}; ${sueReportData.city}`;
+      addressString = `${sueReportData.street}; ${sueReportData.houseNumber}; ${sueReportData.postalCode}; ${sueReportData.city}`;
     }
 
     const formData = {
       addressString,
+      firstName: sueReportData?.name,
+      lastName: sueReportData?.familyName,
```
donni106, line 318: is `...sueReportData` still needed now? this adds values to the form data we do not need, or am I wrong?

ardasnturk: There is still a lot of data in `sueReportData` that we need. For example `email` or `title`

donni106: ah i see, ok

donni106 (approved):

PR #1145, opened 2024-06-18
feat(sue): set address inputs according to `geoMap` config

ardasnturk:
- added required information of address inputs to `requiredInputs` according to `locationStreetIsRequired` and `locationIsRequired` values

SUE-57

## Screenshots:

||||
|--|--|--|
imageimageimage|imageimageimage|imageimageimage

### DIFF
--- a/src/screens/SUE/SueReportScreen.tsx
+++ b/src/screens/SUE/SueReportScreen.tsx
@@ -26,13 +26,15 @@ import {
   SueReportUser,
   Wrapper
 } from '../../components';
-import { colors, device, normalize, texts } from '../../config';
+import { colors, consts, device, normalize, texts } from '../../config';
 import { addToStore, formatSize, readFromStore } from '../../helpers';
 import { useKeyboardHeight } from '../../hooks';
 import { QUERY_TYPES, getQuery } from '../../queries';
 import { postRequests } from '../../queries/SUE';
 
 export const SUE_REPORT_VALUES = 'sueReportValues';
+
+const { INPUT_KEYS } = consts;
 
 type TRequiredFields = {
   [key: string]: {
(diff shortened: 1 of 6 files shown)


src/screens/SUE/SueReportScreen.tsx
```diff
@@ -41,17 +41,27 @@ type TRequiredFields = {
...
   }
 
+  if (geoMap?.locationIsRequired) {
+    requiredInputs['city'] = true;
+    requiredInputs['postalCode'] = true;
+  }
+
+  if (geoMap?.locationStreetIsRequired) {
+    requiredInputs['street'] = true;
+  }
```
donni106, line 63: what about

```
requiredInputs['city'] = !!geoMap?.locationIsRequired;
requiredInputs['postalCode'] = !!geoMap?.locationIsRequired;
requiredInputs['street'] = !!geoMap?.locationStreetIsRequired;
```

donni106 (changes requested):

src/screens/SUE/SueReportScreen.tsx
ardasnturk: In requiredInputs add `city`, `postalCode` and `street` and set their values to `true`. Otherwise this data will not be in the object. This also works with the `requiredInputs.city` logic. Now I thought, we use these texts in different parts of the code. Maybe we should create `const` to update these texts in one place.

```
SUE.INPUTS.CITY = 'city'
SUE.INPUTS.POSTALCODE = 'postalCode'
SUE.INPUTS.STREET = 'street'
```

donni106: i don't understand how this relates to my suggestion of refactoring?

donni106: Sure, this can be an additional improvement to avoid the "hardcoded" strings in multiple places.
(POSTAL_CODE not POSTALCODE)

ardasnturk: https://github.com/smart-village-solutions/smart-village-app-app/pull/1147

ardasnturk: > requiredInputs['city'] = !!geoMap?.locationIsRequired;
> requiredInputs['postalCode'] = !!geoMap?.locationIsRequired;
> requiredInputs['street'] = !!geoMap?.locationStreetIsRequired;

That's exactly the right point of view. 👍

ardasnturk: https://github.com/smart-village-solutions/smart-village-app-app/commit/80a3dd450ed7e72972b8a37f2dc61734b19cea64

donni106 (approved):

PR #1147, opened 2024-06-19
feat(sue): update sue input keys

ardasnturk:
- added `INPUT_KEYS` constant to update the input keys used in many different parts of the code in one place

SUE-57

### DIFF
--- a/src/hooks/reverseGeocode.ts
+++ b/src/hooks/reverseGeocode.ts
@@ -11,7 +11,7 @@ export const useReverseGeocode = () => {
       position,
       setValue
     }: {
-      areaServiceData: { postalCodes: string[] } | undefined;
+      areaServiceData?: { postalCodes?: string[] };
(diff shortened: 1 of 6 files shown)


src/hooks/reverseGeocode.ts
```diff
@@ -11,7 +11,7 @@ export const useReverseGeocode = () => {
       position,
       setValue
     }: {
-      areaServiceData: { postalCodes: string[] } | undefined;
+      areaServiceData?: { postalCodes?: string[] };
```
donni106, line 14: does not relate to this commit, right?

donni106 (approved):

src/hooks/reverseGeocode.ts
ardasnturk: No. That was a minor type update. That's why I didn't add it to the commit.

PR #1152, opened 2024-07-02
feat(sue): improve the ability to get coordinate information from a photo

ardasnturk:
- added controls to the `onDeleteImage` function to delete the information shown in the address step when an image with coordinate information is deleted
- added `showCoordinatesFromImageAlert` control to not show the received coordinates from image alert every time
- added `from` prop to `imageSelect` to be able to know through which function the image was selected
- added location data from the `exif` to the `location` object if the image contains coordinate information
- added the device's location information to the `location` object since there is no coordinate information in the exif when the image is taken from the camera
- added control to skip the location check if the coordinate information is not correctly contained in the location object
- updated `onPress` functions to resolve `imageSelect = undefined` error detected in other components

SUE-76

### DIFF
--- a/src/components/consul/selectors/ImageSelector.js
+++ b/src/components/consul/selectors/ImageSelector.js
@@ -29,6 +29,11 @@ import { WrapperRow, WrapperVertical } from '../../Wrapper';
 
 const { IMAGE_SELECTOR_TYPES, IMAGE_TYPE_REGEX, URL_REGEX } = consts;
 
+const IMAGE_FROM = {
+  CAMERA: 'Camera',
+  GALLERY: 'Gallery'
+};
+
 const deleteImageAlert = (onPress) =>
   Alert.alert(
     texts.consul.startNew.deleteAttributesAlertTitle,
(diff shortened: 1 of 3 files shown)


src/components/consul/selectors/ImageSelector.js
```diff
@@ -103,6 +108,26 @@ export const ImageSelector = ({
       }
     }
 
+    if (selectorType === IMAGE_SELECTOR_TYPES.SUE) {
+      const { GPSLatitude, GPSLongitude } = imagesAttributes[index].exif || {};
```
donni106, line 112: is safe that `imagesAttributes[index]` always exists? otherwise please `imagesAttributes?.[index]?.exif`

src/components/consul/selectors/ImageSelector.js
```diff
@@ -103,6 +108,26 @@ export const ImageSelector = ({
...
+      const { selectedPosition, setValue = () => {} } = coordinateCheck || {};
+
+      if (
+        !!GPSLatitude &&
+        !!GPSLongitude &&
+        selectedPosition.latitude === GPSLatitude &&
+        selectedPosition.longitude === GPSLongitude
```
donni106, line 119: why are there the two checks of presence before the === checks?

ardasnturk: Actually unnecessary. There is no need to double check here. The === check will be enough. 👍

ardasnturk: https://github.com/smart-village-solutions/smart-village-app-app/commit/b94d017584b4fcd1ad1be4b4f781619dbe741f1a

src/components/consul/selectors/ImageSelector.js
ardasnturk: https://github.com/smart-village-solutions/smart-village-app-app/commit/b94d017584b4fcd1ad1be4b4f781619dbe741f1a

donni106 (approved):

PR #1154, opened 2024-07-03
refactor(map): change map & clustering package

d4v3d4vE:
- added `maplibre-react-native` through `yarn add maplibre-react-native`
https://github.com/maplibre/maplibre-react-native
- updated `app.json` according to `maplibre-react-native`
- adjusted the code accordingly to keep all old map functionalities with the new map package `'@maplibre/maplibre-react-native'`
- implemented a calculation mechanism to always show a map with multiple markers with the perfect positioning & zoom settings to display all markers on the map
- ran `yarn remove react-native-map-cluster` to remove the package
- ran `yarn add supercluster` to add the package
- implemented new more efficient clustering via `supercluster`
- added `translucentAccent` to `colors.js` for the cluster design
- sorted imports
- formatted code
- removed prop `isMultipleMarkersMap`, since we calculate the exact camera positioning to show all markers on the map over the marker placements on the map to ensure always the perfect map view with all set markers
- added old cluster design with a adjustable hitbox, set to 80
- added the functionality to disable clustering, clustering is enabled by default
- added back prop functionality to all used props to ensure the same map behaviour as before
- added `View` to `LocationSettings.js` to ensure a reliable map behaviour
- added new `LoadingSinnerMap.tsx` for a compatible loading spinner overlay as the map is loading
- removed translucentAccent from `colors.js` as it's not needed anymore
- adjusted map styling all over the projects used map components to ensure a consistent map behaviour
- added `normalize()` throughout the files I worked in to ensure a consistent design
- updated styles in `Map.tsx` to prevent the app from crashing on some screens with maps
- added `containerStyle` prop to LoadingContainer
- deleted because `LoadingSpinnerMap` was replaced with `LoadingSpinner`
- added collapsable structure in `LocationSettings` again
- added map style to `SueMapScreen` and `SueMapViewScreen` to get full screen map view
- added `onMarkerPress` and `selectedMarker` to `SueMapViewScreen` and `SueReportLocation` to show callout text when marker is pressed
- removed `setTimout`s because they caused unnecessary waiting while loading the map
- unused props removed
- added blank lines to the code to increase readability
- corrected sorting errors
- corrected some minor spelling errors
- added `setSelectedMarker(undefined)` to `onMapPress` function in `SueReportLocation` and `SueMapViewScreen` to remove callout text from the screen
- added `calculateBoundsToFitAllMarkers` helper to prevent the Map.tsx component from growing larger
- updated the name of the `calculateBounds` helper to `calculateBoundsToFitAllMarkers` for clarity
- added comment line to understand values in `bounds` array
- increased the `animationDuration` value from 0 to 1000 to solve the problem of the map updating itself according to the desired location after a certain period of time
- updated or deleted unnecessary codes

To test this changes run `yarn prebuild` and `yarn android -d` for Android or `yarn ios -d` for iOS DevBuild installations.

SVA-1365

<table>
  <tr>
    <td>Bild 1</td>
    <td>Bild 2</td>
    <td>Bild 3</td>
    <td>Bild 4</td>
    <td>Bild 5</td>
  </tr>
  <tr>
    <td>Bild 6</td>
    <td>Bild 7</td>
    <td>Bild 8</td>
    <td>Bild 9</td>
    <td>Bild 10</td>
  </tr>
</table>


src/components/map/Map.tsx
```diff
@@ -157,138 +107,219 @@ export const Map = ({
...
-        clusterColor={colors.primary}
-        clusterFontFamily="regular"
-        clusteringEnabled={clusteringEnabled}
-        radius={clusterDistance}
-        renderCluster={renderCluster}
-        initialRegion={initialRegion}
-        region={updatedRegion}
-        mapType={device.platform === 'android' ? MAP_TYPES.NONE : MAP_TYPES.STANDARD}
+      <LoadingSpinner loading={isLoading} />
+      <MapLibreGL.MapView
```
ardasnturk, line 180: How about importing `MapView` directly when importing the `MapLibreGL` package instead of `MapLibreGL.MapView`?

```
import { MapView } from '@maplibre/maplibre-react-native';
```

src/components/map/Map.tsx
```diff
@@ -157,138 +107,219 @@ export const Map = ({
...
-      <MapView
-        clusterColor={colors.primary}
-        clusterFontFamily="regular"
-        clusteringEnabled={clusteringEnabled}
-        radius={clusterDistance}
-        renderCluster={renderCluster}
-        initialRegion={initialRegion}
-        region={updatedRegion}
-        mapType={device.platform === 'android' ? MAP_TYPES.NONE : MAP_TYPES.STANDARD}
+      <LoadingSpinner loading={isLoading} />
```
ardasnturk, line 179: How about using it the same way we use it in other parts of the code to preserve the structure in the code?

With an if check before return. For example:

```
  if (isLoading) {
    return <LoadingSpinner loading/>;
  }
```

src/components/map/Map.tsx
```diff
@@ -157,138 +107,219 @@ export const Map = ({
...
+      <MapLibreGL.MapView 
+        style={styles.map}
+        styleJSON='https://tileserver-gl.smart-village.app/styles/osm-liberty/style.json'
         onPress={onMapPress}
```
ardasnturk, line 183: Does the `{nativeEvent}` return we get with `onPress` have the same values as the `onPress` we used before?

src/components/map/Map.tsx
```diff
@@ -157,138 +107,219 @@ export const Map = ({
...
+            setZoomLevel(newZoomLevel);
+          }
+        }}  
+        showUserLocation={showsUserLocationSetting}
       >
-        <UrlTile
-          urlTemplate="https://tileserver-gl.smart-village.app/styles/osm-liberty/{z}/{x}/{y}.png"
-          shouldReplaceMapContent={device.platform === 'ios'}
+        <MapLibreGL.Camera
+          ref={cameraRef}
```
ardasnturk, line 194: Why do we have a camera with `MapLibreGL`?

src/components/map/Map.tsx
```diff
@@ -157,138 +107,219 @@ export const Map = ({
...
+                coordinate={[longitude, latitude]}
+              >
+                <View style={styles.clusterContainer}>
+                  <View style={styles.cluster}>
+                    <Text style={styles.clusterText}>{pointCount}</Text>
+                  </View>
+                </View>
+              </MapLibreGL.PointAnnotation>
+            );
+          }
```
ardasnturk, line 214: Is the new cluster structure we are using here fully compatible with the previous cluster? As far as I can see from the screenshots, the design is not exactly the same.

src/components/map/Map.tsx
```diff
@@ -157,138 +107,219 @@ export const Map = ({
...
+
+          const marker = locations[cluster.properties.index];
```
ardasnturk, line 216: What happens if the `marker` value is `undefined`?

src/components/map/Map.tsx
```diff
@@ -304,18 +335,28 @@ export const Map = ({
 
 const styles = StyleSheet.create({
   callout: {
-    width: normalize(120)
+    width: normalize(120),
+    backgroundColor: colors.surface,
+    padding: normalize(5),
+    borderRadius: normalize(10),
+    alignItems: 'center'
```
ardasnturk, line 342: alphabetical sorting

src/components/map/Map.tsx
```diff
@@ -304,18 +335,28 @@ export const Map = ({
...
     alignItems: 'center',
     justifyContent: 'center',
-    position: 'absolute'
   },
-  clusterMarker: {
+  cluster: {
+    width: normalize(40),
+    height: normalize(40),
+    borderRadius: normalize(20),
+    backgroundColor: colors.translucentAccent,
```
ardasnturk, line 352: alphabetical sorting

src/components/map/Map.tsx
```diff
@@ -365,7 +406,18 @@ const styles = StyleSheet.create({
...
+    alignItems: 'center',
+    justifyContent: 'center',
+  },
+  activeMarkerContainer: {
+    // Define your styles for active marker here
+  },
+  map: {
+    width: '100%',
+    height: '100%',
+  },
```
ardasnturk, line 420: alphabetical sorting

ardasnturk, line 420: What does `map` style contribute to us?

src/components/map/Map.tsx
```diff
@@ -50,61 +51,15 @@ const MapIcon = ({
...
   calloutTextEnabled = false,
-  clusterDistance,
+  clusterDistance = 50,
   clusteringEnabled = false,
   geometryTourData,
   isMaximizeButtonVisible = false,
   isMultipleMarkersMap = false,
   isMyLocationButtonVisible = false,
-  locations,
+  locations = [],
```
ardasnturk, line 62: I see more than one unused prop here. Are these props here to be used later?

ardasnturk (changes requested):

src/components/map/Map.tsx
d4v3d4vE: done https://github.com/smart-village-solutions/smart-village-app-app/pull/1154/commits/9afc3d250b644b57ef04987bebbd63ccfb60221c

src/components/map/Map.tsx
d4v3d4vE: done https://github.com/smart-village-solutions/smart-village-app-app/pull/1154/commits/9afc3d250b644b57ef04987bebbd63ccfb60221c

src/components/map/Map.tsx
d4v3d4vE: done https://github.com/smart-village-solutions/smart-village-app-app/pull/1154/commits/9afc3d250b644b57ef04987bebbd63ccfb60221c

src/components/map/Map.tsx
d4v3d4vE: you're right, not needed anymore https://github.com/smart-village-solutions/smart-village-app-app/pull/1154/commits/7d15a3d5f2af482f630518fdc922da1732420e58

src/components/map/Map.tsx
d4v3d4vE: fills the available container so this styling could be removed from all other files where the map component is used, since the styling is reoccurring

src/components/map/Map.tsx
```diff
@@ -34,7 +35,7 @@ type Props = {
 };
 
 const MARKER_ICON_SIZE = normalize(40);
-const CIRCLE_SIZES = [60, 50, 40, 30];
+MapLibreGL.setAccessToken(null);
```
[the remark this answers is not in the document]
d4v3d4vE: not needed anymore https://github.com/smart-village-solutions/smart-village-app-app/pull/1154/commits/7d15a3d5f2af482f630518fdc922da1732420e58#diff-4c6f3a32c46718d48a9634ef0ae21287e206c925ecbf614505b3e266a01bc6ea

src/components/LoadingSpinner.tsx
```diff
@@ -1,18 +1,50 @@
...
+//   loading?: boolean;
+// };
+
+// export const LoadingSpinner = ({ loading }: Props) => {
+//   return loading ? (
+//     <LoadingContainer>
+//       <ActivityIndicator color={colors.refreshControl} />
+//     </LoadingContainer>
+//   ) : null;
+// };
```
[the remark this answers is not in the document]
d4v3d4vE: so it could be used again as it is in the newest commit https://github.com/smart-village-solutions/smart-village-app-app/pull/1154/commits/7d15a3d5f2af482f630518fdc922da1732420e58#diff-4c6f3a32c46718d48a9634ef0ae21287e206c925ecbf614505b3e266a01bc6ea

src/components/LoadingSpinner.tsx
```diff
@@ -1,18 +1,50 @@
...
+
+const styles = StyleSheet.create({
+  loadingContainer: {
+    ...StyleSheet.absoluteFillObject,
+    backgroundColor: 'rgba(255, 255, 255, 0.7)',
+    justifyContent: 'center',
+    alignItems: 'center',
+    zIndex: 10,
+  },
+});
```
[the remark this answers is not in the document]
d4v3d4vE: made a new LoadingSpinnerMap, map specific designed spinner.
this spinner is back to "normal" as it was before. https://github.com/smart-village-solutions/smart-village-app-app/pull/1154/commits/7d15a3d5f2af482f630518fdc922da1732420e58#diff-4c6f3a32c46718d48a9634ef0ae21287e206c925ecbf614505b3e266a01bc6ea

package.json
```diff
@@ -37,6 +37,7 @@
   "dependencies": {
     "@apollo/react-hooks": "~3.1.5",
     "@expo/vector-icons": "14.0.0",
+    "@maplibre/maplibre-react-native": "^10.0.0-alpha.5",
```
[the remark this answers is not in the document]
d4v3d4vE: Alpha is adjusted for the newest iOS Versions and has no concerning issues, its basically stable.

app.json
```diff
@@ -93,7 +93,8 @@
       "expo-font",
       "expo-localization",
       "expo-secure-store",
-      "./config-plugins/withAndroidMailQueriesAndWhatsappPackage"
+      "./config-plugins/withAndroidMailQueriesAndWhatsappPackage",
+      "@maplibre/maplibre-react-native"
```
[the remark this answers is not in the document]
d4v3d4vE: done https://github.com/smart-village-solutions/smart-village-app-app/pull/1154/commits/61d4b1b9ff4022c2cb2b5f28ae45a1d4364b49df

src/components/map/Map.tsx
d4v3d4vE: then we see the map with no markers

src/components/map/Map.tsx
d4v3d4vE: yes, implemented all functionalities and adjusted the design in the newest update https://github.com/smart-village-solutions/smart-village-app-app/pull/1154/commits/7d15a3d5f2af482f630518fdc922da1732420e58

src/components/map/Map.tsx
d4v3d4vE: Camera is just the packages name for the functionality to show the desired map frame.
for exp. as you open the map you see immediately Berlin at zoom lvl 10 if the so called camera is set properly, instead of opening the map and seeing the whole world at zoom lvl 0

src/components/map/Map.tsx
d4v3d4vE: yes, adjusted in the newest version https://github.com/smart-village-solutions/smart-village-app-app/pull/1154/commits/7d15a3d5f2af482f630518fdc922da1732420e58

src/components/map/Map.tsx
d4v3d4vE: imo its more efficient to use it this way since it requires less code and the best code is no code therefore I'd leave it as it is
https://github.com/smart-village-solutions/smart-village-app-app/pull/1154/commits/7d15a3d5f2af482f630518fdc922da1732420e58

src/components/map/Map.tsx
d4v3d4vE: I didn't see the benefit and therefore did the best practice like in MapLibre's examples, and left it as it is in the newest update https://github.com/smart-village-solutions/smart-village-app-app/pull/1154/commits/7d15a3d5f2af482f630518fdc922da1732420e58

src/components/map/Map.tsx
```diff
@@ -157,138 +107,219 @@ export const Map = ({
...
+          setTimeout(() => {
+            setIsLoading(false);
+          }, 4000); // Timeout in Millisekunden anpassen je nach Bedarf
+          setIsInitialFit(false); // Verhindert zukünftige automatische Anpassungen
+        }
+      } else {
+        const [minLng, minLat] = coordinates.reduce((prev, curr) => [Math.min(prev[0], curr[0]), Math.min(prev[1], curr[1])], [Infinity, Infinity]);
+        const [maxLng, maxLat] = coordinates.reduce((prev, curr) => [Math.max(prev[0], curr[0]), Math.max(prev[1], curr[1])], [-Infinity, -Infinity]);
+        const deltaLng = (maxLng - minLng) * 0.1; // Add 10% padding to the bounds
+        const deltaLat = (maxLat - minLat) * 0.1; // Add 10% padding to the bounds
```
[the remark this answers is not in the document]
d4v3d4vE: it's part of the map frame calculation if we have multiple markers on the map to always show the perfect frame with all markers visible, and the `* 0.1` adds as the comment says 10% additional padding on the outer markers so they don't lay on the devices screen edges and the map view/camera is adjusted so they are in the frame not on the edge of it
its just a view thing, marker positioning stays correctly the same https://github.com/smart-village-solutions/smart-village-app-app/pull/1154/commits/7d15a3d5f2af482f630518fdc922da1732420e58

src/components/map/Map.tsx
```diff
@@ -157,138 +107,219 @@ export const Map = ({
...
+            animationDuration: 0,
+          });
+          setTimeout(() => {
+            setIsLoading(false);
+          }, 4000); // Timeout in Millisekunden anpassen je nach Bedarf
+          setIsInitialFit(false); // Verhindert zukünftige automatische Anpassungen
+        }
+      } else {
+        const [minLng, minLat] = coordinates.reduce((prev, curr) => [Math.min(prev[0], curr[0]), Math.min(prev[1], curr[1])], [Infinity, Infinity]);
+        const [maxLng, maxLat] = coordinates.reduce((prev, curr) => [Math.max(prev[0], curr[0]), Math.max(prev[1], curr[1])], [-Infinity, -Infinity]);
```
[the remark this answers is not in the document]
d4v3d4vE: then we should see a map with no markers https://github.com/smart-village-solutions/smart-village-app-app/pull/1154/commits/7d15a3d5f2af482f630518fdc922da1732420e58

src/components/map/Map.tsx
```diff
@@ -157,138 +107,219 @@ export const Map = ({
...
+        const deltaLat = (maxLat - minLat) * 0.1; // Add 10% padding to the bounds
+        if (cameraRef.current) {
+          cameraRef.current.fitBounds(
+            [minLng - deltaLng, minLat - deltaLat],
+            [maxLng + deltaLng, maxLat + deltaLat],
+            0 // animationDuration in ms
+          );
+          setTimeout(() => {
+            setIsLoading(false);
+          }, 1000); // Timeout in Millisekunden anpassen je nach Bedarf
```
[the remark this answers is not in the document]
d4v3d4vE: the timeouts are for the time the map is loading as its calculating the marker positioning and adjusting the camera accordingly, measured the timeouts on various devices and took the best amount to achieve the best result over different plattform
https://github.com/smart-village-solutions/smart-village-app-app/pull/1154/commits/7d15a3d5f2af482f630518fdc922da1732420e58

src/components/map/Map.tsx
```diff
@@ -157,138 +107,219 @@ export const Map = ({
...
+      if (locations.length === 1) {
+        if (cameraRef.current) {
+          cameraRef.current.setCamera({
+            centerCoordinate: coordinates[0],
+            zoomLevel: 15,
+            animationDuration: 0,
+          });
+          setTimeout(() => {
+            setIsLoading(false);
+          }, 4000); // Timeout in Millisekunden anpassen je nach Bedarf
```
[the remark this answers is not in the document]
d4v3d4vE: the timeouts are for the time the map is loading as its calculating the marker positioning and adjusting the camera accordingly, measured the timeouts on various devices and took the best amount to achieve the best result over different platforms
https://github.com/smart-village-solutions/smart-village-app-app/commit/7d15a3d5f2af482f630518fdc922da1732420e58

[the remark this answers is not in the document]
d4v3d4vE: should I remove the comment line?

src/components/map/Map.tsx
```diff
@@ -113,33 +68,28 @@ export const Map = ({
...
-  // LATITUDE_DELTA handles the zoom, see: https://github.com/react-native-maps/react-native-maps/issues/2129#issuecomment-457056572
-  const LATITUDE_DELTA = zoom || 0.0922;
-  // example for longitude delta: https://github.com/react-native-maps/react-native-maps/blob/0.30.x/example/examples/DisplayLatLng.js#L18
-  const LONGITUDE_DELTA = LATITUDE_DELTA * (device.width / (device.height / 2));
+  const showsUserLocationSetting =
+    locationSettings?.locationService ?? otherProps.showsUserLocation ?? showsUserLocation;
 
   // center of Germany
-  let initialRegion: Region = {
+  let initialRegion: { latitude: number; longitude: number } = {
```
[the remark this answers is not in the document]
d4v3d4vE: figured this way its better defined with the newest code enhancements 
https://github.com/smart-village-solutions/smart-village-app-app/commit/7d15a3d5f2af482f630518fdc922da1732420e58

src/components/map/Map.tsx
```diff
@@ -113,33 +68,28 @@ export const Map = ({
...
   ...otherProps
 }: Props) => {
   const { globalSettings } = useContext(SettingsContext);
   const { settings = {} } = globalSettings;
-  const { zoomLevelForMaps = {}, locationService = {} } = settings;
   const { locationSettings } = useLocationSettings();
+  const refForMapView = useRef<MapLibreGL.MapView>(null);
+  const cameraRef = useRef<MapLibreGL.Camera>(null);
+  const [clusters, setClusters] = useState<any[]>([]);
+  const [zoomLevel, setZoomLevel] = useState<number>(14);
```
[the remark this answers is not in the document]
d4v3d4vE: The default value for `zoomLevel` is set to 14, for practical reasons to have a reasonable default setting for the map display. A zoom level of 14 typically shows city areas in a detailed view, providing a good balance between too much and too little detail. This value can later be adjusted based on the specific requirements of the application or user interaction.

src/components/map/Map.tsx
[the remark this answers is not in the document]
d4v3d4vE: Yes, having a default value of `locations = []` is sensible. This ensures that the `locations` prop is always an array, even if it is not provided by the parent component. This prevents potential errors when accessing `locations`, especially with array operations like `.map()`.

src/components/map/Map.tsx
```diff
@@ -157,138 +107,219 @@ export const Map = ({
...
+      cameraRef.current.setCamera({
+        centerCoordinate: [
+          locationSettings.currentLocation.latitude,
+          locationSettings.currentLocation.longitude
+        ],
+        zoomLevel: 15,
+        animationDuration: 1000
+      });
+    }
+  };
```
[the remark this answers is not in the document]
d4v3d4vE: Yes, it appears that the `currentLocation` or a corresponding value is included in the `locationSettings`. The `showsUserLocationSetting` constant is derived from `locationSettings?.locationService ?? otherProps.showsUserLocation ?? showsUserLocation`. This implies that `locationSettings` contains information about the location service, which influences the `showsUserLocation` value.

package.json
```diff
@@ -111,7 +112,6 @@
     "react-native-gesture-handler": "~2.16.1",
     "react-native-get-random-values": "~1.11.0",
     "react-native-gifted-chat": "^2.0.1",
-    "react-native-map-clustering": "^3.4.2",
     "react-native-maps": "1.14.0",
```
donni106, line 115: shouldn't this be removed now?

src/components/settings/LocationSettings.js
```diff
@@ -100,14 +100,16 @@ export const LocationSettings = () => {
...
+        <View style={styles.mapContainer}>
+          <Map
+            locations={locations}
+            onMapPress={({ nativeEvent }) => {
+              setSelectedPosition({
+                ...nativeEvent.coordinate
+              });
+            }}
+          />
+        </View>
```
donni106, line 112: opening the area shows me a too small map, is this intended?


<br/><br/>

and trying to drag around does not work. it behaves strange and does not react correctly on my interactions (not very visible on the video, as my touches are not included):

https://github.com/smart-village-solutions/smart-village-app-app/assets/1942953/57b58a20-1efc-4571-af84-56d33a8421c2

src/components/LoadingSpinner.tsx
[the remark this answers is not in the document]
donni106: place adjust the PR text as well...there was still info about a changed loading spinner for me to read.

src/components/LoadingSpinnerMap.tsx
```diff
@@ -0,0 +1,28 @@
+import React from 'react';
+import { ActivityIndicator, StyleSheet, View } from 'react-native';
+
+import { colors } from '../config';
+
+type Props = {
+  loading?: boolean;
+};
+
+export const LoadingSpinnerMap = ({ loading }: Props) => {
```
donni106, line 10: what is the idea behind this component? what needs to be waited for? and what is the benefit of "seeing through the loading component"?

package.json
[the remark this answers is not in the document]
donni106: do you have a link source for this?

src/components/map/Map.tsx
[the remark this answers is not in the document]
donni106: > This value can later be adjusted based on the specific requirements of the application or user interaction.

what means later and specific requirements? who will be responsible for that?

src/components/map/Map.tsx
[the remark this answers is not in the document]
donni106: when clicking on the commit hash link this brings me to a lot of changes. what do you mean exactly? and what was wrong with `Region`?

src/components/map/Map.tsx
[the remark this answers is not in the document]
donni106: remove it, how could it help having something like this in the code? for whom?

src/components/map/Map.tsx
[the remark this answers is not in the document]
donni106: please remove such comments everywhere

src/components/map/Map.tsx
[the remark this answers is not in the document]
donni106: the comment translates the code, which is not helpful. if commented, please give a reason why this code is needed.

and again, why is the commit hash linked here?

src/components/map/Map.tsx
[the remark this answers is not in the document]
donni106: where does `currentLocation` come from? i cannot find anything about it in the code.

src/components/map/Map.tsx
donni106: please link to examples when speaking about examples.

isn't there more imported as necessary with `import MapLibreGL from '@maplibre/maplibre-react-native';` compared to picking the ones we need per `{ ... } from '@maplibre/maplibre-react-native';`?

src/components/map/LocationOverview.tsx
```diff
@@ -155,19 +154,18 @@ const styles = StyleSheet.create({
...
     shadowColor: colors.shadowRgba,
     shadowOffset: {
-      height: 5,
+      height: normalize(5),
       width: 0
     },
-    shadowOpacity: 0.5,
-    shadowRadius: 3
+    shadowOpacity: normalize(0.5),
+    shadowRadius: normalize(3)
```
donni106, line 168: > added normalize() throughout the files I worked in to ensure a consistent design

what was not consistent here? can you share a comparison please of what is more accurate now

src/components/settings/LocationSettings.js
```diff
@@ -149,5 +151,9 @@ export const LocationSettings = () => {
 const styles = StyleSheet.create({
   containerStyle: {
     marginBottom: normalize(21)
+  },
+  mapContainer: {
+    height: normalize(250),
+    width: '100%'
```
donni106, line 157: is this width really needed? i tried without it and nothing changed.
the problem of rendering the map in the small width needs to be tackled. the width here seems to not have any impact.

src/screens/SUE/SueDetailScreen.tsx
```diff
@@ -142,7 +142,6 @@ export const SueDetailScreen = ({ navigation, route }: StackScreenProps<any>) =>
               <Map
                 isMaximizeButtonVisible
                 locations={[{ position: { latitude, longitude } }]}
-                mapStyle={styles.map}
                 onMaximizeButtonPress={() =>
                   navigation.navigate(ScreenName.MapView, {
                     locations: [{ position: { latitude, longitude } }]
```
donni106, line 147: there are overlaying "buttons" now, which needs to be avoided

image

src/components/map/Map.tsx
```diff
@@ -357,14 +489,25 @@ const styles = StyleSheet.create({
...
     width: normalize(48),
     zIndex: 1
+  },
+  markerContainer: {
+    alignItems: 'center',
+    justifyContent: 'center'
+  },
+  activeMarkerContainer: {
+    // Define your styles for active marker here
+  },
```
donni106, line 507: can be removed i guess?

src/components/map/Map.tsx
```diff
@@ -317,11 +441,19 @@ const styles = StyleSheet.create({
...
+    height: normalize(HITBOX_SIZE),
+    justifyContent: 'center',
+    width: normalize(HITBOX_SIZE)
+  },
   container: {
     alignItems: 'center',
     backgroundColor: colors.surface,
     flex: 1,
-    justifyContent: 'center'
+    height: normalize(200),
```
donni106, line 454: why 200 here and 250 in settings? how come the difference? shouldn't we have small maps always the same height?

src/components/map/Map.tsx
```diff
@@ -34,7 +34,8 @@ type Props = {
 };
 
 const MARKER_ICON_SIZE = normalize(40);
-const CIRCLE_SIZES = [60, 50, 40, 30];
+const CIRCLE_SIZES = [normalize(60), normalize(50), normalize(40), normalize(30)];
+const HITBOX_SIZE = normalize(80); // Größe der unsichtbaren Hitbox
```
donni106, line 38: no German please. and the comment can be removed, as nobody needs a explanation for hitbox size i guess

src/components/map/Map.tsx
```diff
@@ -157,131 +144,264 @@ export const Map = ({
...
+              }
+            }}
+          >
+            <MapLibreGL.LineLayer
+              id="linelayer1"
+              style={{ lineColor: colors.primary, lineWidth: 2 }}
+            />
+          </MapLibreGL.ShapeSource>
+        )}
+        {showsUserLocationSetting && <MapLibreGL.UserLocation visible={true} onUpdate={() => {}} />}
```
donni106, line 403: you do not need to pass `true` hardcoded explicitly, use just `visible` instead

src/components/map/Map.tsx
```diff
@@ -157,131 +144,264 @@ export const Map = ({
...
+                coordinates: geometryTourData.map((point) => [point.longitude, point.latitude])
+              }
+            }}
+          >
+            <MapLibreGL.LineLayer
+              id="linelayer1"
+              style={{ lineColor: colors.primary, lineWidth: 2 }}
+            />
+          </MapLibreGL.ShapeSource>
+        )}
```
donni106, line 402: can you share a screenshot of this as well please

src/components/map/Map.tsx
```diff
@@ -157,131 +144,264 @@ export const Map = ({
...
-                        backgroundColor: isActiveMarker
-                          ? marker.iconColor
-                          : marker.iconBackgroundColor
-                      }
-                    ]}
-                  >
+              <View
+                style={[
+                  styles.markerContainer,
+                  isActiveMarker ? styles.activeMarkerContainer : undefined
```
donni106, line 320: having a styles array you can do `isActiveMarker && styles.activeMarkerContainer` without having the explicit else with undefined in this case

src/components/map/Map.tsx
```diff
@@ -157,131 +144,264 @@ export const Map = ({
...
           const serviceName = truncateText(marker.serviceName);
           const title = truncateText(marker.title);
 
           return (
-            <Marker
-              centerOffset={marker.iconAnchor || { x: 0, y: -(MARKER_ICON_SIZE / 2) }}
-              coordinate={marker.position}
-              identifier={marker.id}
+            <MapLibreGL.PointAnnotation
+              id={`${index}-${marker.id}`}
```
donni106, line 312: what is the `id` for?

src/components/map/Map.tsx
```diff
@@ -50,61 +51,51 @@ const MapIcon = ({
...
-    <Marker
+    <MapLibreGL.PointAnnotation
       key={`cluster-${id}`}
-      coordinate={{
-        longitude: geometry.coordinates[0],
-        latitude: geometry.coordinates[1]
-      }}
-      style={styles.clusterMarker}
-      onPress={onPress}
+      id={`cluster-${id}`}
```
donni106, line 61: what is the `id` for?

src/components/map/Map.tsx
```diff
@@ -157,131 +144,264 @@ export const Map = ({
...
+
+        {clusters.map((cluster, index) => {
+          const [longitude, latitude] = cluster.geometry.coordinates;
+          const { cluster: isCluster, point_count: pointCount } = cluster.properties;
+
+          if (clusteringEnabled && isCluster) {
+            return renderCluster({
+              clusterColor: colors.primary,
+              geometry: { coordinates: [longitude, latitude] },
+              id: `cluster-${index}`,
```
donni106, line 299: setting this to `cluster-3` results in `cluser-cluster-3` in the method, right? i guess it is not intended?

src/components/map/Map.tsx
```diff
@@ -157,131 +144,264 @@ export const Map = ({
...
-            strokeColor={colors.primary}
-            zIndex={1}
-          />
-        )}
-        {locations?.map((marker, index) => {
+        <MapLibreGL.Camera ref={cameraRef} minZoomLevel={minZoom} />
+
+        {clusters.map((cluster, index) => {
+          const [longitude, latitude] = cluster.geometry.coordinates;
+          const { cluster: isCluster, point_count: pointCount } = cluster.properties;
```
donni106, line 293: `pointCount` is only used to pass it again to `properties` as `point_count`. i see no reason to create a new variable in camel case just to be passed again to the one in snake case.

i suggest

```
const { cluster: isCluster, point_count } = cluster.properties;

...

properties: { point_count }
```

src/components/map/Map.tsx
```diff
@@ -157,131 +144,264 @@ export const Map = ({
...
-          left: 0
-        }}
-        legalLabelInsets={{
-          top: 0,
-          right: 0,
-          bottom: normalize(10),
-          left: 0
+        onRegionDidChange={(region) => {
+          const newZoomLevel = Math.round(region.properties.zoomLevel);
+          if (!isNaN(newZoomLevel)) {
```
donni106, line 283: this checks if `newZoomLevel` is a number? this reads very hard "is not not a number"...
i suggest checking the type of the variable on "number" like explained at https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/isNaN

> [Number.isNaN()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/isNaN) is a more reliable way to test whether a value is the number value NaN or not. Alternatively, the expression x !== x can be used, and neither of the solutions is subject to the false positives that make the global isNaN() unreliable. To test if a value is a number, use [typeof x === "number"](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/typeof).

src/components/map/Map.tsx
```diff
@@ -157,131 +144,264 @@ export const Map = ({
...
-        clusteringEnabled={clusteringEnabled}
-        radius={clusterDistance}
-        renderCluster={renderCluster}
-        initialRegion={initialRegion}
-        region={updatedRegion}
-        mapType={device.platform === 'android' ? MAP_TYPES.NONE : MAP_TYPES.STANDARD}
-        onPress={onMapPress}
+      <LoadingSpinnerMap loading={isLoading} />
+      <MapLibreGL.MapView
+        style={mapStyle || styles.map}
```
donni106, line 277: could there be cases where both styles should be used? like passing some additional styles per `mapStyle` to `styles.map` in some cases? therefore we could change to array with `mapStyle` last.

src/components/map/Map.tsx
```diff
@@ -157,131 +144,264 @@ export const Map = ({
...
+          centerCoordinate: [longitude, latitude],
+          zoomLevel: zoom,
+          animationDuration: 1000
+        });
+      }
+
+      // Wait for the animation to finish
+      setTimeout(() => {
+        setIsAnimating(false);
+      }, 1000);
```
donni106, line 268: please create a const vor the duration and use it in both places. otherwise someone somebody could change one of those without the other, because not knowing about the dependency.

and please elaborate the comment a bit, saying which animation instead of "the", and why?!

src/components/map/Map.tsx
```diff
@@ -157,131 +144,264 @@ export const Map = ({
...
+    }
+  };
+
+  const handleClusterPress = useCallback(
+    (cluster) => {
+      if (isAnimating) return; // Skip if animation is already in progress
+
+      setIsAnimating(true);
+      const [longitude, latitude] = cluster.geometry.coordinates;
+      const zoom = Math.min(zoomLevel + 2, 20); // Increase the zoom level by 2 or set to max zoom level (20)
```
donni106, line 256: what happens when pressing a cluster? it zooms the map by 2 and stops? or does it increase until markers are visible?

src/components/map/Map.tsx
```diff
@@ -157,131 +144,264 @@ export const Map = ({
...
+          if (clusteringEnabled) {
+            setTimeout(() => {
+              setIsLoading(false);
+            }, 1000); // Timeout in Millisekunden anpassen je nach Bedarf
+            setIsInitialFit(false); // Verhindert zukünftige automatische Anpassungen
+          } else {
+            setTimeout(() => {
+              setIsLoading(false);
+            }, 4000); // Timeout in Millisekunden anpassen je nach Bedarf
+            setIsInitialFit(false); // Verhindert zukünftige automatische Anpassungen
```
donni106, line 220: no German comments please

donni106, line 220: no German comments please

src/components/map/Map.tsx
```diff
@@ -157,131 +144,264 @@ export const Map = ({
...
+        const [minLng, minLat] = coordinates.reduce(
+          (prev, curr) => [Math.min(prev[0], curr[0]), Math.min(prev[1], curr[1])],
+          [Infinity, Infinity]
+        );
+        const [maxLng, maxLat] = coordinates.reduce(
+          (prev, curr) => [Math.max(prev[0], curr[0]), Math.max(prev[1], curr[1])],
+          [-Infinity, -Infinity]
+        );
+        const deltaLng = (maxLng - minLng) * 0.1; // Add 10% padding to the bounds
+        const deltaLat = (maxLat - minLat) * 0.1; // Add 10% padding to the bounds
```
donni106, line 204: please move this heavy maths in a separate function outside of the component with a speaking name and some comment.
this will lead to more readability inside the component.

src/components/map/Map.tsx
```diff
@@ -157,131 +144,264 @@ export const Map = ({
...
+          properties: { cluster: false, id: location.id, index },
+          geometry: {
+            type: 'Point',
+            coordinates: [location.position.longitude, location.position.latitude]
+          }
+        }))
+      );
+    }
+
+    // Berechne die Begrenzungskoordinaten für alle Marker
```
donni106, line 179: no German comments please.

src/components/map/Map.tsx
```diff
@@ -157,131 +144,264 @@ export const Map = ({
...
+        type: 'Feature',
+        properties: { cluster: false, id: location.id, index },
+        geometry: {
+          type: 'Point',
+          coordinates: [location.position.longitude, location.position.latitude]
+        }
+      }));
+
+      index.load(points);
+      const bounds = [-180, -85, 180, 85];
```
donni106, line 164: how come these exact values?

src/components/map/Map.tsx
```diff
@@ -157,131 +144,264 @@ export const Map = ({
...
+      setClusters(index.getClusters(bounds, zoomLevel));
+    } else {
+      setClusters(
+        locations.map((location, index) => ({
+          type: 'Feature',
+          properties: { cluster: false, id: location.id, index },
+          geometry: {
+            type: 'Point',
+            coordinates: [location.position.longitude, location.position.latitude]
+          }
```
donni106, line 174: this is the same code a few lines before, right? i suggest moving it in a function and use it in both places.

src/components/map/Map.tsx
```diff
@@ -157,131 +144,264 @@ export const Map = ({
     };
   }
 
+  useEffect(() => {
+    if (clusteringEnabled) {
+      const index = new Supercluster({
+        radius: clusterDistance,
+        maxZoom: 20
```
donni106, line 151: ha! the 20 is also used in multiple places, right? so please create a const for it and use it everywhere.

src/components/map/Map.tsx
```diff
@@ -113,33 +104,29 @@ export const Map = ({
...
 }: Props) => {
   const { globalSettings } = useContext(SettingsContext);
   const { settings = {} } = globalSettings;
-  const { zoomLevelForMaps = {}, locationService = {} } = settings;
   const { locationSettings } = useLocationSettings();
+  const refForMapView = useRef<MapLibreGL.MapView>(null);
+  const cameraRef = useRef<MapLibreGL.Camera>(null);
+  const [clusters, setClusters] = useState<any[]>([]);
+  const [zoomLevel, setZoomLevel] = useState<number>(14);
+  const [isInitialFit, setIsInitialFit] = useState<boolean>(true);
```
donni106, line 119: what is initial fit for?

src/components/map/Map.tsx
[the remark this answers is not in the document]
donni106: and please create a const for the 14 and use it everywhere.

src/components/LoadingSpinner.tsx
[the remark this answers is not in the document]
d4v3d4vE: already done that, I probably also wouldn't have noticed it in this big PR text block.

> added new LoadingSinnerMap.tsx for a compatible loading spinner overlay as the map is loading

package.json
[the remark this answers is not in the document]
d4v3d4vE: https://github.com/maplibre/maplibre-react-native/issues/402#issuecomment-2163513467

src/components/LoadingSpinner.tsx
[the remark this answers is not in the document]
donni106: i read that. but the other info was still there.

src/components/map/Map.tsx
d4v3d4vE: I followed the documentation and and examples to ensure we are using the library as intended
image
https://github.com/maplibre/maplibre-react-native/blob/main/example/src/App.js

If it would reduce the bundle size and therefore loading times is not clear, could have an impact and also could not.

src/components/map/Map.tsx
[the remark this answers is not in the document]
d4v3d4vE: this code is needed for the correct positioning of the camera so all markers are properly visible.

src/components/map/Map.tsx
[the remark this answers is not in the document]
d4v3d4vE: it's linked cause the code is outdated. 

you are right, `reduce` would fail for undefined. I'm going to implement a check for that case.

src/components/map/Map.tsx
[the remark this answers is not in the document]
d4v3d4vE: will be removed in the upcoming commit

src/components/map/Map.tsx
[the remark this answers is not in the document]
d4v3d4vE: will be removed in the upcoming commit

src/components/map/Map.tsx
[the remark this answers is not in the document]
d4v3d4vE: this change ensures that `initialRegion` has exactly the expected fields `latitude` and `longitude`, without additional or unexpected properties.

src/components/map/Map.tsx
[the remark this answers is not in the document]
d4v3d4vE: > what means later and specific requirements? who will be responsible for that?
if some customer might want different settings for example. we of course if its the customers wish for a different setting until they get the ability to set it themselves.

> and please create a const for the 14 and use it everywhere.
change is coming in the upcoming commit.

package.json
d4v3d4vE: you're right, thanks
will be removed in the upcoming commit

src/components/map/LocationOverview.tsx
d4v3d4vE: we use `normalize` throughout the app and I just added it where it wasn't used before for a consistent styling over all devices. 
we should add `normalize` every time we work on some files where it's not used but should be.

src/components/map/Map.tsx
d4v3d4vE: would also say so, will be removed in the upcoming commit.

src/components/map/Map.tsx
d4v3d4vE: removed in the upcoming commit.

src/components/map/Map.tsx
d4v3d4vE: thanks, updated in the upcoming commit.

src/components/map/Map.tsx
d4v3d4vE: since we don't even need `activeMarkerContainer` anymore I can remove the check completely.

src/components/map/Map.tsx
d4v3d4vE: as the map opens with a set camera, wich it does when markers are available, the initialfit of the camera gets overwritten so the user is able to navigate through the map freely.
without initialfit the camera stays set, so as you open a map with markers and try to move the mapview gets set back to the initial camera view the map was initiated in.

src/components/map/Map.tsx
d4v3d4vE: its only used once, as the 15 is, just 14 is used twice. should it nevertheless be a const?

src/components/map/Map.tsx
d4v3d4vE: true, thank you. will be adjusted in the upcoming commit.

src/components/map/Map.tsx
d4v3d4vE: the values represent the maximum possible latitude and longitude values for the earth's coordinate system, defining a rectangular area that covers the entire world.

but you're right, in our case it's simply not needed and might also improve performance slightly.
In the upcoming commit I've changed it to `const bounds = [2, 46, 18, 56];` to cover germany and parts of countries around. 
- Westernmost longitude: 2° E (covering parts of France and the Netherlands)
- Easternmost longitude: 18° E (covering parts of Poland and the Czech Republic)
- Northernmost latitude: 56° N (covering parts of Denmark)
- Southernmost latitude: 46° N (covering parts of Switzerland and Austria)

src/components/map/Map.tsx
d4v3d4vE: will be removed in the upcoming commit.

src/components/map/Map.tsx
d4v3d4vE: will be removed in the upcoming commit.

src/components/map/Map.tsx
```diff
@@ -157,131 +144,264 @@ export const Map = ({
...
+      }
+    }
+  }, [locations, clusterDistance, zoomLevel, isInitialFit, clusteringEnabled]);
+
+  useEffect(() => {
+    if (updatedRegion && cameraRef.current) {
+      cameraRef.current.setCamera({
+        centerCoordinate: [updatedRegion.longitude, updatedRegion.latitude],
+        zoomLevel: 14,
+        animationDuration: 1000
```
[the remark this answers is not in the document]
d4v3d4vE: will be done in the upcoming commit.

src/components/map/Map.tsx
d4v3d4vE: it zooms in by 2 and stops, should it zoom in until the clustered markers are visible?

src/components/map/Map.tsx
d4v3d4vE: thank you, will be done in the upcoming commit.

src/components/map/Map.tsx
d4v3d4vE: thank you for the tip, will be adjusted in the upcoming commit.

src/components/map/Map.tsx
d4v3d4vE: thanks, will be adjusted in the upcoming commit.

src/components/map/Map.tsx
d4v3d4vE: usually yes, but if you check the badbelzig app there is also a difference in the set alternative location screen and the small map screen, therefore I chose those two sizes for the same design.

src/screens/SUE/SueDetailScreen.tsx
d4v3d4vE: should we place the "maximize map button" in the top right corner instead the bottom right one?

src/components/settings/LocationSettings.js
d4v3d4vE: you are right, it's not needed and will be removed in the upcoming commit.

src/components/map/Map.tsx
d4v3d4vE: you are right, this was not intended.
it will be changed in the upcoming commit to: 
`id: index`

src/components/map/Map.tsx
d4v3d4vE: The id attribute in the <MapLibreGL.PointAnnotation> component is used to uniquely identify each annotation (point or cluster) on the map. This unique identifier is crucial for various reasons:

1. **Interaction Handling**: The id allows you to handle user interactions with specific annotations, such as selecting, deselecting, or clicking on a point. When an annotation is selected, the map can identify which annotation was interacted with based on its id.

2. **Performance Optimization**: By using unique identifiers, the map library can efficiently manage and update only the annotations that have changed, rather than re-rendering all annotations. This can lead to performance improvements, especially when dealing with a large number of points.

3. **Data Association**: The id helps in associating the annotations with the underlying data points. For example, if you need to display additional information about a specific location when it is clicked, you can use the id to fetch the relevant data.

src/components/map/Map.tsx
d4v3d4vE: The id attribute in the <MapLibreGL.PointAnnotation> component is used to uniquely identify each annotation (point or cluster) on the map. This unique identifier is crucial for various reasons:

1. **Interaction Handling**: The id allows you to handle user interactions with specific annotations, such as selecting, deselecting, or clicking on a point. When an annotation is selected, the map can identify which annotation was interacted with based on its id.

2. **Performance Optimization**: By using unique identifiers, the map library can efficiently manage and update only the annotations that have changed, rather than re-rendering all annotations. This can lead to performance improvements, especially when dealing with a large number of points.

3. **Data Association**: The id helps in associating the annotations with the underlying data points. For example, if you need to display additional information about a specific location when it is clicked, you can use the id to fetch the relevant data.

src/components/settings/LocationSettings.js
d4v3d4vE: it should show the whole map, there is some kind of bug caused by the collapsible option, if we'd display the map component by default without the collapsible for example this wouldn't happen

src/components/LoadingSpinnerMap.tsx
d4v3d4vE: we need to wait for the map to finish loading, loading the markers, clusters and the correct mapview regarding the clusters and markers set on the map.

the see through is mainly for dev reasons so we can check the loading times exactly once we have this new map feature in our dev build to test it on various devices. 
it's also psychologically beneficiary for us cause the user is willing to wait a second longer if they see that something is happening wanting to see the result of the finished map instead of just waiting for the plain loadingspinner to stop spinning.

src/components/map/Map.tsx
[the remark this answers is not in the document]
d4v3d4vE: you're right, forgot to remove it after I tested something.
will be removed in the upcoming commit.

src/components/map/Map.tsx
d4v3d4vE: thanks, it will be changed in the upcoming commit to: 

`if (typeof newZoomLevel === 'number' && !isNaN(newZoomLevel)) {...`

src/components/map/Map.tsx
d4v3d4vE: true, thanks for your suggestion.

will be implemented this way in the upcoming commit.

src/components/map/Map.tsx
d4v3d4vE: Simulator Screenshot - iPhone 15 Pro - 2024-07-12 at 13 38 23
testing wise I've hardcoded some random coordinates and ist works properly displaying a route.

package.json
```diff
@@ -111,8 +112,6 @@
     "react-native-gesture-handler": "~2.16.1",
     "react-native-get-random-values": "~1.11.0",
     "react-native-gifted-chat": "^2.0.1",
-    "react-native-map-clustering": "^3.4.2",
-    "react-native-maps": "1.14.0",
```
donni106, line 115: with removing this package, the following occurs. please adjust it:

```
$ patch-package
patch-package 8.0.0
Applying patches...
Error: Patch file found for package react-native-maps which is not present at node_modules/react-native-maps
---
patch-package finished with 1 error(s).
```

d4v3d4vE: adjusted it:
```
$ patch-package
patch-package 8.0.0
Applying patches...
No patch files found
```

should I remove the `patch-package 8.0.0` completely?

donni106: if we do not use/need it, yes.

d4v3d4vE: removed it entirely.

src/components/LoadingSpinner.tsx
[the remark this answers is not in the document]
d4v3d4vE: okay, removed.

src/components/map/Map.tsx
donni106: yes please all of those values distributed across the file to be able to see them at one spot.

src/components/map/Map.tsx
donni106: this was the last maps function, so i think so, yes. where does the "zoom in by 2" comes from? does other maps do it this way?

src/components/map/Map.tsx
donni106: please reduce the size of the screenshot next time or per html image integration here directly. it is hard to have a look with this too big image in the discussions.

src/components/map/Map.tsx
donni106: sagsunseinfach is not a topic for the main code base. so please align them.

src/components/map/Map.tsx
d4v3d4vE: will be adjusted in the upcoming commit.

src/components/map/Map.tsx
d4v3d4vE: okay, will be set to 200 for both screens in the upcoming commit.

src/screens/SUE/SueDetailScreen.tsx
d4v3d4vE: should we use it like this?
<image src="https://github.com/user-attachments/assets/b2f39106-69d0-4018-92bc-e90109c51edf" width="250" />

ur put it in the bottom left corner with the old styling? dosen't seem like we use the bottom left corner for anything.
<image src="https://github.com/user-attachments/assets/6993d10f-e53d-4853-929b-b609c8ef082a" width="250" />

src/components/map/Map.tsx
d4v3d4vE: "zoom in by 2" was an idea of mine to handle it and every map handles it differently. 
in the upcoming commit I've changed it to the behaviour we are used to to zoom in until the cluster & markers update/change

src/screens/SUE/SueDetailScreen.tsx
donni106: first! some actions right others left is no option. additionally positioning left is bad ux for most users which are right handed

src/components/settings/LocationSettings.js
donni106: ok remove the collapsible and add a bold text in front of the text with the title of the button, that is used now to open the area.
the button itself and the "abbrechen" when opened can be removed with this as well.

src/screens/SUE/SueDetailScreen.tsx
d4v3d4vE: applied the fist change.

src/components/settings/LocationSettings.js
d4v3d4vE: now the rendering issue caused by the collapsible is fixed but there is still this issue where the pin does move but the map stays static and once we close the map and reopen it it all of a sudden works perfect.
just cant wrap my head around it maybe you'll have a fix for this one.

https://github.com/user-attachments/assets/8d2449ee-226f-476d-b7c8-fc01389c76a1

donni106: the video does not show the requested change with removing the collapsible and the other things mentioned before. so what to check?

d4v3d4vE: there is no collapsible anymore, just similar behaviour handled completely differently.

donni106: i see. this was not discussed :(
but now keep it, we do not have time to get into it again.

@ardasnturk please check the problem @d4v3d4vE is asking help for.

donni106: another point is, that when closing the area and opening again, the whole map gets reinitialized and loads again for some time. this should not happen. once loaded it needs to stay loaded.

ardasnturk: I solved the problem and added the collabsable structure again.

https://github.com/smart-village-solutions/smart-village-app-app/pull/1154/commits/a465ce7e87004b3810f9245288d27ad9d5deb404

src/components/LoadingSpinnerMap.tsx
ardasnturk: I updated the `LoadingContainer` component by adding `containerStyle` prop and removed this component. I don't think we need an extra component. 

I also shortened the wait times. As far as I noticed, this loading indicator disappears after 4 seconds of setTimout. We had an extra 4 seconds of waiting time every time the map view was opened. The 4 seconds does not check if the map is loaded or not.

src/components/map/Map.tsx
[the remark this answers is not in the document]
ardasnturk: I see that we no longer have that function, so I'm closing this conversation.

src/components/map/Map.tsx
```diff
@@ -157,141 +179,256 @@ export const Map = ({
...
+                          iconColor={
+                            marker.iconColor
+                              ? isActiveMarker
+                                ? colors.surface
+                                : marker.iconColor
+                              : colors.surface
+                          }
+                          iconName={marker.iconName}
+                          iconSize={MARKER_ICON_SIZE / 3.25}
+                        />
```
ardasnturk, line 387: I have noticed that if there is a pin after the map is loaded, the map updates itself late or we have to move it to update the map. I haven't found a solution so far, but I found where the problem is coming from.

The problem is caused by the custom marker we developed. In Map.tsx, if we make 332 to 397 of the code a comment line, the map loads perfectly.  Otherwise the map first shows the whole world and places the pin in the center of the world. If we wait for a certain time or move the map we can see the exact location.

|beginning|after moving the map|
|--|--|
Simulator Screenshot - iPhone 15 Pro Max - 2024-07-22 at 17 30 40|Simulator Screenshot - iPhone 15 Pro Max - 2024-07-22 at 17 30 45

ardasnturk (changes requested):
I have noticed that if there is a pin after the map is loaded, the map updates itself late or we have to move it to update the map. I haven't found a solution so far, but I found where the problem is coming from.

The problem is caused by the custom marker we developed. In Map.tsx, if we make 332 to 397 of the code a comment line, the map loads perfectly.  Otherwise the map first shows the whole world and places the pin in the center of the world. If we wait for a certain time or move the map we can see the exact location.

Translated with www.DeepL.com/Translator (free version)

src/components/map/Map.tsx
```diff
@@ -157,141 +179,256 @@ export const Map = ({
...
+                        {title}
+                      </RegularText>
+                    )}
+                  </View>
+                )}
+
+                <View style={styles.markerContainer}>
+                  {!!marker.iconName &&
+                  marker.iconName != 'ownLocation' &&
+                  marker.iconName != 'location' ? (
```
ardasnturk, line 353: I have noticed that if there is a pin after the map is loaded, the map updates itself late or we have to move it to update the map. I haven't found a solution so far, but I found where the problem is coming from.

The problem is caused by the custom marker we developed. In Map.tsx, if we make 332 to 397 of the code a comment line, the map loads perfectly.  Otherwise the map first shows the whole world and places the pin in the center of the world. If we wait for a certain time or move the map we can see the exact location.

|beginning|after moving the map|
|--|--|
Simulator Screenshot - iPhone 15 Pro Max - 2024-07-22 at 17 30 40|Simulator Screenshot - iPhone 15 Pro Max - 2024-07-22 at 17 30 45

ardasnturk (changes requested):

src/components/map/Map.tsx
donni106: and what was the outcome? any performance improvements?

src/components/map/Map.tsx
donni106: both? the hard to understand double negation is still present for what reason?

src/components/settings/LocationSettings.js
donni106: nice, and when closing/opening the area the map stay loaded where it was before.

package.json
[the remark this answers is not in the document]
donni106: there is a [10.0.0-alpha.10](https://github.com/maplibre/maplibre-react-native/releases/tag/10.0.0-alpha.10) already. please update it and quick test the screens @d4v3d4vE

src/components/settings/LocationSettings.js
```diff
@@ -93,41 +89,43 @@ export const LocationSettings = () => {
...
+    });
+  };
+
+  const handleSave = () => {
+    if (selectedPosition) {
+      setAndSyncLocationSettings({
+        alternativePosition: geoLocationToLocationObject(selectedPosition)
+      });
+    }
+    setSelectedPosition(undefined);
```
donni106, line 104: why is this set to undefined if there is a selected position a line before?

src/screens/SUE/SueMapViewScreen.tsx
```diff
@@ -63,6 +65,7 @@ export const SueMapViewScreen = ({ route }: { route: any }) => {
             setSelectedPosition(undefined);
           }
         }}
+        onMarkerPress={(id) => setSelectedMarker(id)}
```
donni106, line 68: `onMarkerPress={setSelectedMarker}` should be fine here, right?

src/components/map/Map.tsx
```diff
@@ -50,61 +69,69 @@ const MapIcon = ({
...
+              {point_count}
+            </RegularText>
+          </View>
+        ))}
+      </View>
+    </PointAnnotation>
   );
 };
 
 /* eslint-disable complexity */
```
donni106, line 108: should stay in front of `export const Map = ({`

src/components/map/Map.tsx
```diff
@@ -50,61 +69,69 @@ const MapIcon = ({
...
+// calculate the bounds of the map to fit all markers
+const calculateBounds = (coordinates) => {
```
donni106, line 110: let's rename the method to `calculateBoundsToFitAllMarkers` instead of having the need to place that comment

donni106: the whole method should be moved to a helper file instead of enlarging this file here.

src/components/map/Map.tsx
donni106: ok strange or do i get that wrong? which scenario can exist where this is intended?

src/components/map/Map.tsx
```diff
@@ -157,141 +179,256 @@ export const Map = ({
...
+    }
+  }));
+
+  useEffect(() => {
+    if (clusteringEnabled) {
+      const index = new Supercluster({
+        radius: clusterDistance,
+        maxZoom: MAX_ZOOM_LEVEL
+      });
+      const bounds = [2, 46, 18, 56];
```
donni106, line 198: please add the comment about the bound numbers, which they are representing, directly in the code here

src/components/settings/LocationSettings.js
ardasnturk: This code is copied and pasted into the `handleSave` function. The previous version of the code is protected.

src/screens/SUE/SueMapViewScreen.tsx
ardasnturk: Yes 🙌

src/components/map/Map.tsx
ardasnturk: With the new commit this file has been freed from complexity. So I will remove this line. 👍

src/components/map/Map.tsx
```diff
@@ -157,141 +179,256 @@ export const Map = ({
...
+            0
+          );
+
+          if (clusteringEnabled) {
+            setIsLoading(false);
+            setIsInitialFit(false);
+          } else {
+            setIsLoading(false);
+            setIsInitialFit(false);
+          }
```
donni106, line 237: ääh, no difference here? if this is intended, then please remove the if-else, otherwise correct the values.

src/components/map/Map.tsx
donni106: maybe animation duration needs to be > 0 to make something happen automatically?

ardasnturk: Yes, that's exactly it! An excellent point of view. Problem solved! 🥳

donni106: woooow 😃

src/components/map/Map.tsx
ardasnturk: https://github.com/smart-village-solutions/smart-village-app-app/pull/1154/commits/800969674da23ee9825df39f73ddb943008dbe34

src/components/map/Map.tsx
ardasnturk: https://github.com/smart-village-solutions/smart-village-app-app/pull/1154/commits/800969674da23ee9825df39f73ddb943008dbe34

src/components/map/Map.tsx
ardasnturk: https://github.com/smart-village-solutions/smart-village-app-app/pull/1154/commits/800969674da23ee9825df39f73ddb943008dbe34

src/components/map/Map.tsx
ardasnturk: https://github.com/smart-village-solutions/smart-village-app-app/pull/1154/commits/800969674da23ee9825df39f73ddb943008dbe34

src/screens/SUE/SueMapViewScreen.tsx
ardasnturk: https://github.com/smart-village-solutions/smart-village-app-app/pull/1154/commits/800969674da23ee9825df39f73ddb943008dbe34

src/helpers/index.js
```diff
@@ -5,6 +5,7 @@ export * from './parser';
 export * from './accessibilityListeners';
 export * from './addressHelper';
 export * from './bookmarkHelper';
+export * from './calculateBoundsToFitAllMarkers.ts';
```
donni106, line 8: without .ts

src/helpers/calculateBoundsToFitAllMarkers.ts
```diff
@@ -0,0 +1,17 @@
+// calculate the bounds of the map to fit all markers
```
donni106, line 1: comment not needed with new method name

src/helpers/index.js
ardasnturk: Autocomplete 🤦‍♂️

src/components/settings/LocationSettings.js
d4v3d4vE: very nice fix!

src/components/map/Map.tsx
d4v3d4vE: works better this way, true
the map still isn't fully loaded after the `animationDuration` but I think that should be acceptable this way

src/components/map/Map.tsx
d4v3d4vE: should it just be: `if (!isNaN(newZoomLevel)) {...` ?

donni106: no, this was you first approach which caused the discussion, right?!

donni106: instead of adding `typeof newZoomLevel === 'number'` you should have exchanged it

src/components/map/Map.tsx
donni106: is the bounds the reason for fading other parts of the map that are outside? you can see what i mean when zoomed fully out seeing the whole world.

src/components/settings/LocationSettings.js
donni106: @d4v3d4vE do you know?

package.json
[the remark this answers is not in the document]
d4v3d4vE: seems as good as the alpha.5 we are currently using.
I'm pushing the commit.

src/components/settings/LocationSettings.js
d4v3d4vE: ChatGPT's explanation:
> The line setSelectedPosition(undefined); is setting selectedPosition to undefined to reset or clear the selected position after the user saves or aborts the operation. This is a common practice to ensure that the state is reset, preventing any residual state from interfering with subsequent operations.


personally I couldn't explain it on the fly

src/components/map/Map.tsx
d4v3d4vE: that was how the app behaved with the code until I introduced initial fit, so with `initialFit` after the camera gets set we can navigate freely without getting set back to the cameras initial view if we try to navigate/swipe throughout the map

src/components/map/Map.tsx
d4v3d4vE: exchanged for what? I don't get what you want me to do rn.

src/components/map/Map.tsx
d4v3d4vE: no, the bounds are just there for clustering, it just takes in the mapmarker inside the bounds into account for clustering. 
that fade out you mean has todo with our `.mbtiles` map file cause we haven't got the data to display the whole world.

src/components/map/Map.tsx
donni106: ok strange? which scenario can exist where this is intended when users move and it jumps back?

src/components/map/Map.tsx
donni106: exchange the difficult !NaN... with `typeof newZoomLevel === 'number'`... `if (typeof newZoomLevel === 'number')...`

src/components/map/Map.tsx
d4v3d4vE: maybe a scenario where the mapmarker should always be in the main focus and not in a draggable map, but since it's not our case we got the solution with setting and removing the initial fit

src/components/settings/LocationSettings.js
donni106: so it is used first for alternative position and cleared afterwards, right? maybe the collapsible is closed with unsetting selected position? what happens if we do not set it to undefined?

d4v3d4vE: yes, you are correct. the selectedPosition is used to update the alternativePosition in the location settings and is then cleared to reset the state

the collapsible component's visibility is controlled by showMap, which is set to false immediately after clearing the selected position

but I also just checked manually various cases and I couldn't finde any error wich should appear when we are not setting `setSelectedPosition(undefined);`, therefore I'd say that we can get rid of it

donni106:
MapLibre integration will be coming in a completely overhauled version, planned for milestone 4.2.0.

PR #1158, opened 2024-07-04
feat(sue): improve performance problems

ardasnturk:
- updated components with `useMemo` hook to avoid performance problem and reduce re-render issue
- stopped the `Carousel` component according to the isFocused hook because it was constantly running in the background, causing unnecessary re-rendering and badly affecting performance

SUE-76

### DIFF
--- a/src/ConfigurationsProvider.tsx
+++ b/src/ConfigurationsProvider.tsx
@@ -1,4 +1,4 @@
-import React, { ReactNode, createContext, useContext, useEffect, useState } from 'react';
+import React, { ReactNode, createContext, useContext, useEffect, useMemo, useState } from 'react';
 import { useQuery } from 'react-query';
 
 import { SettingsContext } from './SettingsProvider';
@@ -48,15 +48,18 @@ export const ConfigurationsProvider = ({ children }: { children?: ReactNode }) =
     type: 'json'
   });
 
-  useEffect(() => {
+  const mergedConfig = useMemo(() => {
     const config = {
       appDesignSystem,
       sueConfig: { ...sue, ...sueConfigData, sueProgress }
     };
+    return mergeDefaultConfiguration({ ...defaultConfiguration }, config);
+  }, [appDesignSystem, sue, sueConfigData, sueProgress]);
 
-    setConfigurations((prev) => mergeDefaultConfiguration(prev, config));
-    storageHelper.setConfigurations(config);
-  }, [sueConfigData]);
+  useEffect(() => {
+    setConfigurations(mergedConfig);
+    storageHelper.setConfigurations(mergedConfig);
+  }, [mergedConfig]);
 
   return (
     <ConfigurationsContext.Provider value={configurations}>
--- a/src/ConfigurationsProvider.tsx
+++ b/src/ConfigurationsProvider.tsx
@@ -53,6 +53,7 @@ export const ConfigurationsProvider = ({ children }: { children?: ReactNode }) =
       appDesignSystem,
       sueConfig: { ...sue, ...sueConfigData, sueProgress }
     };
+
     return mergeDefaultConfiguration({ ...defaultConfiguration }, config);
   }, [appDesignSystem, sue, sueConfigData, sueProgress]);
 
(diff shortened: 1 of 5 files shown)


src/screens/SUE/SueReportScreen.tsx
```diff
@@ -393,7 +369,18 @@ export const SueReportScreen = ({
...
+          let images;
+
+          try {
+            images = JSON.parse(getValues(INPUT_KEYS.SUE.IMAGES));
+          } catch (error) {
+            console.error('Invalid JSON in images:', error);
+            return;
+          }
+
+          if (!images.length) {
```
donni106, line 381: `undefined.length` will crash. please add a ? or make images a [] by default

ardasnturk: `images` has the default value `'[]'`

```
...
  } = useForm({
    mode: 'onBlur',
    defaultValues: {
      ...
      [INPUT_KEYS.SUE.IMAGES]: '[]',
      ...
    }
  });
```

donni106: please check few lines before...`let images;`...which is undefined initially

src/components/ImagesCarousel.js
```diff
@@ -12,6 +12,7 @@ import { OrientationContext } from '../OrientationProvider';
 import { getQuery } from '../queries';
 import { SettingsContext } from '../SettingsProvider';
 
+import { useIsFocused } from '@react-navigation/native';
```
donni106, line 15: order imports?

src/components/SUE/report/SueReportDescription.tsx
```diff
@@ -11,6 +11,8 @@ import { Input } from '../../form';
 
 const { IMAGE_SELECTOR_TYPES, IMAGE_SELECTOR_ERROR_TYPES, INPUT_KEYS } = consts;
 
+const MemoizedImageSelector = React.memo((props) => <ImageSelector {...props} />);
```
donni106, line 14: please import memo and use it directly instead of from React like in the other file src/screens/SUE/SueReportScreen.tsx

src/components/SUE/report/SueReportServices.tsx
```diff
@@ -26,34 +41,35 @@ export const SueReportServices = ({
...
     setLoading(false);
-  };
+  }, [refetch]);
+
+  const handleSetService = useCallback(
+    (item) => {
+      setService(item);
+    },
+    [setService]
+  );
```
donni106, line 54: how should this help? i guess it can be reverted

src/components/ImagesCarousel.js
```diff
@@ -39,6 +40,13 @@ export const ImagesCarousel = ({
   const [carouselImageIndex, setCarouselImageIndex] = useState(0);
 
   const carouselRef = useRef();
+  const isFocused = useIsFocused();
+
+  if (isFocused) {
+    carouselRef.current?.startAutoplay();
+  } else {
+    carouselRef.current?.stopAutoplay();
+  }
```
donni106, line 49: ❤️ this!

src/ConfigurationsProvider.tsx
```diff
@@ -48,15 +48,18 @@ export const ConfigurationsProvider = ({ children }: { children?: ReactNode }) =
...
+    return mergeDefaultConfiguration({ ...defaultConfiguration }, config);
+  }, [appDesignSystem, sue, sueConfigData, sueProgress]);
 
-    setConfigurations((prev) => mergeDefaultConfiguration(prev, config));
-    storageHelper.setConfigurations(config);
-  }, [sueConfigData]);
+  useEffect(() => {
+    setConfigurations(mergedConfig);
+    storageHelper.setConfigurations(mergedConfig);
+  }, [mergedConfig]);
```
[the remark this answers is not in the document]
ardasnturk: I don't know how to measure it. 😔🤷‍♂️

src/screens/SUE/SueReportScreen.tsx
ardasnturk: If the `images` value is `undefined`, we may get an error here. You are right. I will add a check.

donni106 (approved):

PR #1161, opened 2024-07-09
fix(sue): remove throwing errors where not needed

donni106:
We do not need them anymore, right? Or did I miss something?

SUE-76

### DIFF
--- a/src/components/SUE/report/SueReportLocation.tsx
+++ b/src/components/SUE/report/SueReportLocation.tsx
@@ -226,6 +226,7 @@ export const SueReportLocation = ({
       } catch (error) {
         setSelectedPosition(undefined);
         Alert.alert(texts.sue.report.alerts.hint, error.message);
+        return { error: error.message };
       }
     }
   };
--- a/src/components/SUE/report/SueReportLocation.tsx
+++ b/src/components/SUE/report/SueReportLocation.tsx
@@ -177,18 +177,13 @@ export const SueReportLocation = ({
     }
   }, []);
 
-  const handleGeocode = async (position: { latitude: number; longitude: number }) => {
-    try {
-      await reverseGeocode({
-        areaServiceData,
-        errorMessage,
-        position,
-        setValue
-      });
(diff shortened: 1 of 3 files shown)


src/components/SUE/report/SueReportLocation.tsx
```diff
@@ -175,18 +175,13 @@ export const SueReportLocation = ({
...
-      await reverseGeocode({
-        areaServiceData,
-        errorMessage,
-        position,
-        setValue
-      });
-    } catch (error) {
-      throw new Error(error.message);
-    }
-  };
```
ardasnturk, line 189: Since we are not generating errors in this section, the marker in the full screen map view on the report screen is not removed from the screen if it is not at the desired location. This is because the cache section in `SueMapViewScreen` is not working. I prefer to return the error instead of generating an error in the cache section. This way we can add a control according to the return in `SueMapViewScreen`.

|At the moment|should be|
|--|--|
Simulator Screenshot - iPhone 15 Pro Max - 2024-07-18 at 14 54 00|Simulator Screenshot - iPhone 15 Pro Max - 2024-07-18 at 14 54 06

SueReportLocation:
```
      try {
        await handleGeocode(nativeEvent.coordinate);
      } catch (error) {
        setSelectedPosition(undefined);
        Alert.alert(texts.sue.report.alerts.hint, error.message);
        return { error: error.message };
      }
```

SueMapViewScreen:
```
            const mapPress = await onMapPress({ nativeEvent });

            if (!!mapPress?.error) {
              setSelectedPosition(undefined);
            }
```

ardasnturk (changes requested):

src/components/SUE/report/SueReportLocation.tsx
donni106: ok, seems to be neat

donni106: ok, seems to be neat. can you adjust it please.

PR #1166, opened 2024-07-26
fix: bugs in map package

ardasnturk:
- added `clusteringEnabled` prop as `false` to solve the problem of collapsable closing when the map component in `LocationSettings` clicks
  somewhere other than the coordinates inside the bounds array
- added `anchor` prop to `PointAnnotation` component to fix the problem that the marker is not positioned
  correctly after clicking on the map
- added `zIndex` style to `markerContainer` to fix the problem of markers not displaying correctly for android
- added `undefined` value to `onMarkerPress` function when zoom value is changed in `onRegionDidChange` prop to
  fix the bug that callout text continues to be shown when zoomIn or zoomOut on the map
- added `onDidFinishLoadingMap` prop to `MapView` to prevent the map from being shown without loading and updated `isLoading` to `false`
- added style to `Collapsible` component to solve the problem of collapsible buttons overlapping after loading the map on the settings screen
- improved code readability

SVA-1399

### DIFF
--- a/src/components/map/Map.tsx
+++ b/src/components/map/Map.tsx
@@ -55,7 +55,7 @@ type TCluster = {
 
 const CIRCLE_SIZES = [normalize(60), normalize(50), normalize(40), normalize(30)];
 const DEFAULT_ZOOM_LEVEL = 14;
-const HIT_BOX_SIZE = normalize(80);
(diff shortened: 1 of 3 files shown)


donni106:
I guess this can be closed regarding the upcoming 4.2.0 with MapLibre @ardasnturk?

PR #1172, opened 2024-07-30
fix(map): cluster problem

ardasnturk:
- added `ShapeSource` component instead of `PointAnnotation` component to solve clustering bug in android
- Added `clustersData` in `renderCluster` to create the desired data format of the `shape` in `ShapeSource`
- added value 1000 to the `fitBounds` function to show the cluster in an animated way
- updated with `parseInt` inside `handleClusterPress` since `cluster.id` is string in android

SVA-1399

fix_me: point_count will be added inside the cluster

|before|after|
|--|--|
Screenshot_1722344429|Screenshot_1722344393


src/components/map/Map.tsx
```diff
@@ -74,40 +76,40 @@ const MapIcon = ({
...
-const renderCluster = (cluster: TCluster) => {
-  const { clusterColor: backgroundColor, geometry, id, onPress, properties = {} } = cluster;
-  const { point_count } = properties;
+const renderCluster = ({ markers, onPress }: { markers: any[]; onPress: () => void }) => {
+  let clustersData = {
+    features: [],
+    type: 'FeatureCollection'
+  };
+
+  markers.map((item) => {
```
donni106, line 85: you want to loop here, right, not map some value and return a new array? so please change to a loop over `markers` to update the values in `clustersData.features`

src/components/map/Map.tsx
```diff
@@ -209,7 +211,8 @@ export const Map = ({
           cameraRef.current.fitBounds(
             [minLng - deltaLng, minLat - deltaLat],
             [maxLng + deltaLng, maxLat + deltaLat],
-            0
+            0,
+            1000
```
donni106, line 215: what is the fourth value here that is set with 1000 now?

src/components/map/Map.tsx
```diff
@@ -74,40 +76,40 @@ const MapIcon = ({
...
 
   return (
-    <PointAnnotation
-      coordinate={geometry.coordinates}
-      id={`cluster-${id}`}
-      key={`cluster-${id}`}
-      onSelected={onPress}
-      style={styles.hitBox}
+    <ShapeSource
+      hitbox={{ width: HIT_BOX_SIZE, height: HIT_BOX_SIZE }}
```
donni106, line 93: why do we need to override the default of 44x44?

src/components/map/Map.tsx
donni106: image

ardasnturk: I added this to the commit.

- added value 1000 to the fitBounds function to show the cluster in an animated way

src/components/map/Map.tsx
ardasnturk: Because our hitbox size is not 44x44 but instead `normalize(50)`

src/components/map/Map.tsx
donni106: i know. but this does not clearly say what the value stands for exactly

donni106:
Stop this for now, clustering still not work properly, especially the display of the point counts.

donni106:
I guess this can be closed regarding the upcoming 4.2.0 with MapLibre @ardasnturk?

PR #1173, opened 2024-07-31
fix(config): app not loading problem

ardasnturk:
- added control to `useEffect` to fix the problem of entering an infinite loop when the app first starts if there is no sue object in `globalSettings`
- added skip to prevent loading of `sueReportProgress` staticContent if sue object is not in `globalSettings`

SVA-1403

## How to test:

I added a new `globalSettings` of version 4.1.0 to main-server. This version does not contain sue object. You can try by changing the version in `app.json` to 4.1.0.

### DIFF
--- a/src/ConfigurationsProvider.tsx
+++ b/src/ConfigurationsProvider.tsx
@@ -45,16 +45,19 @@ export const ConfigurationsProvider = ({ children }: { children?: ReactNode }) =
   const { data: sueProgress } = useStaticContent({
     refreshTimeKey: 'publicJsonFile-sueReportProgress',
     name: 'sueReportProgress',
-    type: 'json'
+    type: 'json',
+    skip: !Object.keys(sue).length
   });
 
   const mergedConfig = useMemo(() => {
-    const config = {
+    if (!Object.keys(sue).length) {
+      return defaultConfiguration;
+    }
+
+    return mergeDefaultConfiguration(defaultConfiguration, {
       appDesignSystem,
       sueConfig: { ...sue, ...sueConfigData, sueProgress }
-    };
-
-    return mergeDefaultConfiguration({ ...defaultConfiguration }, config);
+    });
   }, [appDesignSystem, sue, sueConfigData, sueProgress]);
 
   useEffect(() => {


src/ConfigurationsProvider.tsx
```diff
@@ -58,8 +59,10 @@ export const ConfigurationsProvider = ({ children }: { children?: ReactNode }) =
   }, [appDesignSystem, sue, sueConfigData, sueProgress]);
 
   useEffect(() => {
-    setConfigurations(mergedConfig);
-    storageHelper.setConfigurations(mergedConfig);
+    if (JSON.stringify(configurations) !== JSON.stringify(mergedConfig)) {
```
donni106, line 62: can you explain this in real words? its hard for me. if there is no SUE, why would we need t do something with merged configs? if there ist no SUE, we should not habe something merged but only the default configs.
with skipping requests, we receive no data. with having no data we can have a check in `mergedConfig` to return only default configs.
and if this i the case, the effect does not need to set new configs.
you know what i mean!?

ardasnturk: This control sets the config if there is a difference between the default config and mergedConfig. If they are the same then we use the default config. I don't want to bind this control to sue because in the future we want to be able to use config not only for sue but also for other features.

ardasnturk: https://github.com/smart-village-solutions/smart-village-app-app/commit/9b73a39d97fe47fee665fac87344a4a56b2f55a3

src/ConfigurationsProvider.tsx
```diff
@@ -45,16 +45,22 @@ export const ConfigurationsProvider = ({ children }: { children?: ReactNode }) =
...
-      appDesignSystem,
-      sueConfig: { ...sue, ...sueConfigData, sueProgress }
-    };
+    if (!Object.keys(sue).length) {
+      return defaultConfiguration;
+    }
 
-    return mergeDefaultConfiguration({ ...defaultConfiguration }, config);
+    return mergeDefaultConfiguration(
+      { ...defaultConfiguration },
```
donni106, line 58: why spreading the object into a new object? you can use `defaultConfiguration` instead `{ ...defaultConfiguration }`

ardasnturk: Yes, that's right. I updated the code.

https://github.com/smart-village-solutions/smart-village-app-app/commit/0d2e01bebc02a31441157626adf2da4c2d49d134

PR #1179, opened 2024-08-13
feat: integrate adaptive icon

ardasnturk:
- added adaptive-icon to use different icons for ios and android
- added `adaptiveIcon` to `app.json` to add adaptive icon feature for android
  - https://docs.expo.dev/versions/latest/config/app/#adaptiveicon
- added foregroundImage inside adaptiveIcon to show foreground image in adaptive icon
- added `backgroundColor` inside `adaptiveIcon` to update the background color of the icon

SVAK-117

## Screenshots:

|before|after|
|--|--|
SCR-20240813-iofc|image


app.json
```diff
@@ -37,6 +37,10 @@
       }
     },
     "android": {
+      "adaptiveIcon": {
+        "foregroundImage": "./assets/adaptive-icon.png",
+        "backgroundColor": "#0F243F"
```
donni106, line 42: just setting the background color was not enough? did you try?

ardasnturk: I haven't tried it, but I don't think it will be enough, because the icon we are currently using doesn't have a safe zone. If we don't use a different icon for `foregroundImage`, the icon will look like below.

image

Though Int/Dev may not be the right place to test this. Because there is a difference between the icon developed for municipalities and the icon developed for int/dev.

Municipalities' icons: have background color
Int/Dev icon: transparent.

I just tried it with a municipality's app and it works.

image

But it still doesn't mean it will work for every municipality. For some municipalities the icons may need to be updated.

ardasnturk: https://github.com/smart-village-solutions/smart-village-app-app/commit/d7441f20580fb412d7f0b85c47ff2c1c79ca1880

donni106: all right, i understand fully now. but this means, that we should keep the foreground image as it will be necessary for most of the customers!
you could make the int/dev logo a bit smaller in the visible area to not get the edges cut i guess. give it a try 🤞

ardasnturk: Adjust the dimensions as in the documentation and the icon looks much more accurate.

image

ardasnturk: OK. I've added foregroundImage again. After this change, I thought we needed it too.

https://github.com/smart-village-solutions/smart-village-app-app/commit/093cb036be76d83412d729b45a527ea906c85df2

donni106: whuuu nice!

donni106 (approved):

PR #1185, opened 2024-09-19
fix(tabler): icon problem on android

ardasnturk:
- added `tabler-icons-react-native` package to fix the tabler icon family's spacing issue for Android
- added `IconComponent` to the `NamedIcon` function so that the code can work properly with different icon families
- removed from the code because there is no need for `Tabler` icon family as font
- moved the `Icon.tsx` file back into the `config` folder since the tabler font was deleted and there was no need for the extra `icons` folder

SVASD-6

## Screenshots:

|before|after|
|--|--|
screenshot-2024-09-19_10 32 22 02|screenshot-2024-09-19_10 31 53 072

### DIFF
--- a/src/config/Icon.tsx
+++ b/src/config/Icon.tsx
@@ -1,6 +1,9 @@
+import _camelCase from 'lodash/camelCase';
+import _upperFirst from 'lodash/upperFirst';
 import React, { ComponentProps } from 'react';
 import { StyleProp, View, ViewStyle } from 'react-native';
 import { SvgXml } from 'react-native-svg';
+import * as Tabler from 'tabler-icons-react-native';
 
 import {
   addImage,
@@ -42,6 +45,7 @@ import {
   send,
   service,
   share,
+  sueBroom,
   sueCheckSmall,
   sueClock,
   sueClockSmall,
@@ -55,13 +59,11 @@ import {
   verifiedBadge,
   visible,
   voucher
-} from '../../icons';
-import { sueBroom } from '../../icons/SUE/broom';
+} from '../icons';
 
-import { colors } from './../colors';
-import { device } from './../device';
-import { normalize } from './../normalize';
-import { Tabler } from './Tabler';
+import { colors } from './colors';
+import { device } from './device';
+import { normalize } from './normalize';
 
 export type IconProps = {
   accessibilityLabel?: string;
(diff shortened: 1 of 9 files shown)


src/config/Icon.tsx
```diff
@@ -115,12 +118,19 @@ const NamedIcon = ({
...
+  name?: ComponentProps<typeof IconSet>['name'];
+  stroke?: ComponentProps<number>;
+}) => {
+  const TablerName = ('Icon' + _upperFirst(_camelCase(name))) as keyof typeof Tabler;
+  const IconComponent = Tabler?.[TablerName] || IconSet;
+
   return (
     <View accessibilityLabel={accessibilityLabel} style={style} hitSlop={getHitSlops(size)}>
-      <IconSet name={name} size={size} color={color} style={iconStyle} />
+      <IconComponent name={name} size={size} color={color} style={iconStyle} stroke={stroke} />
```
donni106, line 133: so let's summarize, we pass a `name` to `NamedIcon` without knowing if it is from `IconSet` or `Tabler`? `name` is marked as optional with ?, wich is not possible, right? we need a name! and the type is not of `IconSet` anymore but from `IconSet` or `Tabler`, so i guess the types are not final right now.

what happens, if we pass a `name` that exists for `Tabler` and `IconSet` and we want it from `IconSet`? with the current logic, `Tabler` would be used first.

i think we need a more detailed separation or test for the two possible icon types, before setting the one used as `IconComponent`.

ardasnturk: Okay, you're right. How about adding a control like this before the return? That way we can check which `IconSet` we are using and create the `IconComponent` accordingly.


```
  let IconComponent: any;

  if (IconSet === Tabler) {
    const TablerName = ('Icon' + _upperFirst(_camelCase(name))) as TablerIconName;
    IconComponent = Tabler[TablerName] || Tabler['IconQuestionMark'];
  } else {
    IconComponent = IconSet;
  }

  return (
    <View accessibilityLabel={accessibilityLabel} style={style} hitSlop={getHitSlops(size)}>
      <IconComponent name={name} size={size} color={color} style={iconStyle} stroke={stroke} />
    </View>
  );
```

donni106: the code makes much more sense this way, yes! so we clearly have one type of icon set and not a mix of two. great!
please change it and i will re-check.

ardasnturk: https://github.com/smart-village-solutions/smart-village-app-app/commit/83142b71005d5e1a0824d8e710af96c5f6d18453

donni106 (approved):

PR #1186, opened 2024-09-19
fix: the problem with the area service request

ardasnturk:
- updated `areaService` to access zip codes via `rs` since not all zip codes are available via `id`
- added `postalCodes` object in the return section to get all postal codes obtained via getByRs in the desired format
- added `_uniq` to remove repeated zip codes from the array

SUE-133

Log:
```
 {"postalCodes": ["24011", "24012", "24013", "24014", "24015", "24016", "24017", "24018", "24019", "24020", "24021", "24022", "24023", "24024", "24025", "24026", "24027", "24028", "24029", "24030", "24031", "24032", "24033", "24034", "24035", "24036", "24037", "24038", "24039", "24040", "24041", "24042", "24043", "24044", "24046", "24047", "24048", "24049", "24062", "24063", "24065", "24103", "24105", "24106", "24107", "24109", "24111", "24113", "24114", "24116", "24118", "24143", "24145", "24146", "24147", "24148", "24149", "24154", "24159", "24170", "24171"]}
```

### DIFF
--- a/src/queries/SUE/areaService.tsx
+++ b/src/queries/SUE/areaService.tsx
@@ -1,3 +1,5 @@
+import _uniq from 'lodash/uniq';
+
 import { storageHelper } from '../../helpers';
 
 export const areaService = async () => {
@@ -17,7 +19,15 @@ export const areaService = async () => {
     }
   };
 
-  return await (
-    await fetch(`${serverUrl}/${id}?selectAttributes=postalCodes`, areaServiceFetchObj)
+  const rs = (
+    await (await fetch(`${serverUrl}/${id}?selectAttributes=rs`, areaServiceFetchObj)).json()
+  )?.rs;
+
+  const postalCodes = await (
+    await fetch(`${serverUrl}/getByRs?rs=${rs}*&selectAttributes=postalCodes`, areaServiceFetchObj)
   ).json();
+
+  return {
+    postalCodes: _uniq(postalCodes?.values?.flatMap((item: any) => item.postalCodes || []))
+  };
 };


src/queries/SUE/areaService.tsx
```diff
@@ -17,7 +19,21 @@ export const areaService = async () => {
...
+      `${serverUrl}/getByRs?rs=${getRS?.rs}*&selectAttributes=postalCodes`,
+      areaServiceFetchObj
+    )
   ).json();
+
+  return {
+    postalCodes: _uniqBy(
+      postalCodes?.values?.flatMap((item: any) => item.postalCodes || []),
+      (code) => code
+    )
```
donni106, line 37: lodashs _uniq(Array) should be enough here

src/queries/SUE/areaService.tsx
```diff
@@ -17,7 +19,21 @@ export const areaService = async () => {
...
 
-  return await (
-    await fetch(`${serverUrl}/${id}?selectAttributes=postalCodes`, areaServiceFetchObj)
+  const getRS = await (
+    await fetch(`${serverUrl}/${id}?selectAttributes=rs`, areaServiceFetchObj)
+  ).json();
+
+  const postalCodes = await (
+    await fetch(
+      `${serverUrl}/getByRs?rs=${getRS?.rs}*&selectAttributes=postalCodes`,
```
donni106, line 28: what about

```
const rs = (await (
    await fetch(`${serverUrl}/${id}?selectAttributes=rs`, areaServiceFetchObj)
  ).json())?.rs;

...
`${serverUrl}/getByRs?rs=${rs}
...
```

should work, right?

ardasnturk: I tried this first but it didn't work so I had to change the code to `getRS?.rs`. 🤷‍♂️

donni106:
```
const rs = (await (await fetch(`${serverUrl}/${id}?selectAttributes=rs`, areaServiceFetchObj)).json()).rs;
```

src/queries/SUE/areaService.tsx
ardasnturk: But we don't have just one array. We have an array of different zip codes inside an array. So I'm not sure if it will work correctly with just _uniq(Array). The array we use uniq on here looks like the one below.

here I think that for `_uniqBy` to work correctly we need to give an array with only postal codes in it. For example= [2312,3123,42323...]. For this reason we use `flatMap` to format `postalCodes.values` as an array containing only postal codes.

```
{
  "values": [
    {},
    {
      "postalCodes": [
        "39539",
        "39615"
      ]
    },
    {
      "postalCodes": [
        "39596"
      ]
    },
    {
      "postalCodes": [
        "39596"
      ]
    },
    {
      "postalCodes": [
        "39606"
      ]
    },
    {
      "postalCodes": [
        "39596"
      ]
    },
    {
      "postalCodes": [
        "39596"
      ]
    },
    {
      "postalCodes": [
        "39579"
      ]
    },
    {
      "postalCodes": [
        "39596"
      ]
    }
  ],
  "offset": 0,
  "totalCount": 9,
  "count": 9
}
```

src/queries/SUE/areaService.tsx
ardasnturk: Okay, interesting. Maybe I did something wrong, but it's working now. Thank you 👍

src/queries/SUE/areaService.tsx
donni106: sure, flat map is needed. but uniqBy shuld be able to be changed to uniq

ardasnturk: Ow mann! I didn't understand what you meant by that. I'm sorry. I will learn the difference between uniyBy and uniq. 👍

ardasnturk: https://github.com/smart-village-solutions/smart-village-app-app/commit/9ee1866a9a7aae19d1adf5afa86c0aae73a082ac

donni106 (approved):

PR #1188, opened 2024-09-23
feat(filter): create filter overlay

ardasnturk:
|index screen|overlay filter|collapsible filter|
|--|--|--|
Simulator Screenshot - iPhone 16 Pro Max - 2024-09-23 at 13 42 54|Simulator Screenshot - iPhone 16 Pro Max - 2024-10-01 at 17 56 27|Simulator Screenshot - iPhone 16 Pro Max - 2024-09-23 at 13 43 11

|current location alert|filtered IndexScreen|unfiltered IndexScreen|
|--|--|--|
Simulator Screenshot - iPhone 16 Pro Max - 2024-10-01 at 17 56 20|Simulator Screenshot - iPhone 16 Pro Max - 2024-10-01 at 17 56 23|Simulator Screenshot - iPhone 16 Pro Max - 2024-10-01 at 17 56 32

SVAK-128


src/PermanentFilterProvider.tsx
```diff
@@ -40,19 +41,22 @@ export const PermanentFilterProvider = ({ children }: { children?: React.ReactNo
...
     []
   );
-  const [filterState, filterDispatch] = useReducer(overlayFilterReducer, []);
+  const [resourceFilterState, resourceFilterDispatch] = useReducer(resourceFilterReducer, []);
 
   const loadFilters = useCallback(async () => {
     const dataProviderIds = ((await readFromStore(DATA_PROVIDER_FILTER_KEY)) ?? []) as string[];
     const mowasRegionalKeys = ((await readFromStore(MOWAS_REGIONAL_KEYS)) ?? []) as string[];
-    const overlayFilter = ((await storageHelper.filter()) ?? []) as string[];
+    const resourceFilter = ((await readFromStore(RESOURCE_FILTER_KEY)) ?? []) as string[];
```
donni106, line 49: we save multiple resource filters, right? so i suggest naming all the consts in plural

src/components/filter/Filter.tsx
```diff
@@ -1,69 +1,42 @@
...
 };
 
-export const Filter = ({ filterTypes, initialFilters, setQueryVariables }: Props) => {
-  const [filters, setFilters] = useState(initialFilters);
+export const Filter = ({
+  filterTypes,
+  initialFilters,
+  isOverlay = false,
+  setQueryVariables,
+  widthSearch = false
```
donni106, line 30: with

src/components/filter/Filter.tsx
```diff
@@ -73,7 +46,7 @@ export const Filter = ({ filterTypes, initialFilters, setQueryVariables }: Props
     filters.initial_start_date && !Object.keys(_omit(filters, Object.keys(initialFilters))).length;
 
   return (
-    <>
+    <Wrapper style={widthSearch && styles.noPadding}>
```
donni106, line 49: if you have conditions for `style` please use an array, as otherwise it could return a boolean value which is not correct for the prop

src/components/filter/Filter.tsx
```diff
@@ -91,59 +64,82 @@ export const Filter = ({ filterTypes, initialFilters, setQueryVariables }: Props
...
+          <Header
+            backgroundColor={colors.transparent}
+            centerComponent={{
+              text: texts.filter.filter,
+              style: { color: colors.darkText, fontSize: normalize(18), fontWeight: '700' }
+            }}
+            rightComponent={{
+              icon: 'close',
+              color: colors.darkText,
+              onPress: () => setIsCollapsed(!isCollapsed)
```
donni106, line 83: what about Androids look?

src/components/filter/DropdownFilter.tsx
```diff
@@ -28,10 +25,13 @@ export const DropdownFilter = ({
...
-    initiallySelectedItem,
-    ...data
-  ]);
+  const [dropdownData, setDropdownData] = useState<DropdownProps[]>(() => {
+    if (data[0]?.value === '- Alle -') {
+      return [...data];
+    }
+
+    return [initiallySelectedItem, ...data];
+  });
```
donni106, line 34: why is this change necessary?

src/components/filter/Filter.tsx
```diff
@@ -156,6 +152,12 @@ const styles = StyleSheet.create({
   icon: {
     paddingLeft: normalize(8)
   },
+  noPadding: {
+    paddingTop: 0,
+    paddingBottom: 0,
+    paddingLeft: 0,
+    paddingRight: 0
```
donni106, line 159: `padding: 0` wasn't fine?

src/components/filter/SliderFilter.tsx
```diff
@@ -0,0 +1,59 @@
...
+  label?: string;
+  onValueChange: (value: number) => void;
+  thumbStyle?: ViewStyle;
+  value: number;
+};
+
+export const SliderFilter = ({ label, onValueChange, thumbStyle, value, ...props }: Props) => {
+  return (
+    <>
+      <Label>{label}</Label>
```
donni106, line 20: `{!!label && <Label>{label}</Label>}`?

src/components/filter/SliderFilter.tsx
```diff
@@ -0,0 +1,59 @@
...
+      <WrapperRow spaceBetween style={styles.alignItemsCenter}>
+        <Slider
+          maximumValue={100}
+          onValueChange={onValueChange}
+          step={10}
```
donni106, line 25: the max value and steps should be define by the values we receive from the server configs, right?

src/helpers/filterTypesHelper.ts
```diff
@@ -0,0 +1,168 @@
...
+        )?.map((entry: any) => ({
+          ...entry,
+          filterValue: entry.value
+        }));
+        filterType.label = getLabel(key);
+        filterType.name = QUERY_VARIABLES_ATTRIBUTES.LOCATION;
+        filterType.placeholder = getPlaceholder(key, query);
+        break;
+      default:
+        break;
```
donni106, line 161: couldn't these two be default instead of having them in each case?

```
filterType.label = getLabel(key);
filterType.placeholder = getPlaceholder(key, query);
```

src/components/filter/Filter.tsx
```diff
@@ -91,59 +64,82 @@ export const Filter = ({ filterTypes, initialFilters, setQueryVariables }: Props
...
+            <WrapperRow center spaceAround>
+              <Button
+                disabled={!!isNoFilterSet}
+                invert
+                notFullWidth
+                onPress={() => {
+                  setIsCollapsed(!isCollapsed);
+
+                  setTimeout(() => {
+                    setFilters(initialFilters);
```
donni106, line 101: testing the reset, the slider value remains as is selected it before

src/components/screens/EventRecords.js
```diff
@@ -257,6 +260,17 @@ export const EventRecords = ({ navigation, route }) => {
     showVolunteerEvents
   ]);
 
+  const filterTypes = useMemo(() => {
+    return filterTypesHelper({
+      data,
```
donni106, line 265: image

i think this is because in filter types helper `usePermanentFilter` is called, so the `useMemo` here warns

src/components/screens/Overviews.js
```diff
@@ -349,6 +363,15 @@ export const Overviews = ({ navigation, route }) => {
...
             ListHeaderComponent={
               <>
+                <Filter
+                  filterTypes={filterTypes}
+                  initialFilters={queryVariables}
+                  isOverlay
+                  setQueryVariables={setQueryVariables}
+                />
+
+                <Divider />
```
donni106, line 373: i think this should be placed where the filter by opening time is placed, to be not part of the scrollable list but always reachable

src/components/filter/Filter.tsx
ardasnturk: I tried but unfortunately it doesn't work. There is a padding:normalize(14) value for the wrapper. I cannot overwrite it.

src/helpers/filterTypesHelper.ts
ardasnturk: Yes, that makes sense! 👍

src/components/filter/DropdownFilter.tsx
ardasnturk: Because there is no placeholder data in the dropdown array we created for Sue, so we had to add it here. But the array we created for resourceFilter contains the placeholder data. So we needed an extra control.

src/components/filter/SliderFilter.tsx
ardasnturk: Yes, that's right

src/components/filter/Filter.tsx
ardasnturk: I think it looks good.

image

src/components/filter/DropdownFilter.tsx
donni106: i cannot understand it. what has this code to do with `sue`? can you please give examples for `data` and `initiallySelectedItem` in different cases.

src/components/filter/Filter.tsx
donni106: did you inspected the element? sometimes there are `paddingTop`, `paddingBottom`, `paddingLeft` and `paddingRight` set by something else, which cannot be overridden with just `padding`, so the explicit override with each value is needed. is this the case here?

src/components/filter/SliderFilter.tsx
donni106: i see values with 1, 11, 21, ..., 71, ... is this correct?

Simulator Screen Recording - iPhone 15 Plus - 2024-09-27 at 16 38 11

and: it is very laggy using the slider! does it work smoothly for you?

src/components/filter/Filter.tsx
donni106: it's still the case, right? the slider values is not being reset:

Simulator Screen Recording - iPhone 15 Plus - 2024-09-27 at 16 40 29

src/components/filter/Filter.tsx
```diff
@@ -1,148 +1,155 @@
...
+            animationType="slide"
+            onRequestClose={() => setIsCollapsed(!isCollapsed)}
+            presentationStyle="pageSheet"
+            visible={!isCollapsed}
+          >
+            <Header
+              backgroundColor={colors.transparent}
+              centerComponent={{
+                text: texts.filter.filter,
+                style: { color: colors.darkText, fontSize: normalize(18), fontWeight: '700' }
```
donni106, line 81: i think this needs to be more accurate with our normal header styles from `screenOptions`:

```
headerTitleStyle: {
    color: colors.lightestText,
    fontFamily: device.platform === 'ios' ? 'bold' : 'regular',
    fontSize: normalize(20),
    fontWeight: '400',
    lineHeight: normalize(29)
  },
```

src/components/filter/Filter.tsx
```diff
@@ -1,148 +1,155 @@
...
+              }}
+              rightComponent={{
+                icon: 'close',
```
donni106, line 84: which icon set is used here?

src/components/filter/FilterComponent.tsx
```diff
@@ -0,0 +1,111 @@
...
+                    value: !filters[item.name]
+                  })
+                );
+              }}
+              title={item.placeholder}
+              checkedColor={colors.accent}
+              checkedIcon="check-square-o"
+              uncheckedColor={colors.darkText}
+              uncheckedIcon="square-o"
+              containerStyle={styles.checkboxContainerStyle}
```
donni106, line 74: i have adjusted to a square as in all other places across the app

src/components/filter/SliderFilter.tsx
donni106: i found out, that every sliding re-renders all the filters instead of just the slider, you can see it in the resulting console warnings here:

Simulator Screen Recording - iPhone 15 Plus - 2024-09-27 at 18 38 47

this needs to be tackled.

src/components/filter/Filter.tsx
```diff
@@ -1,153 +1,163 @@
...
+                  disabled={!!isNoFilterSet}
+                  notFullWidth
+                  onPress={() => {
+                    setIsCollapsed(!isCollapsed);
+                    setQueryVariables((prev) => ({ search: prev.search || '', ...filters }));
+                  }}
+                  title={texts.filter.filter}
                 />
-              )}
+              </WrapperRow>
```
donni106, line 119: i have adjusted the layout a bit according to the mockups

Image Description

src/components/filter/Filter.tsx
ardasnturk: ionicon

src/components/filter/Filter.tsx
ardasnturk: I solved the problem.

src/components/filter/SliderFilter.tsx
ardasnturk: I solved the problem.

src/components/filter/Filter.tsx
ardasnturk: this part of the code has been reverted

src/components/filter/DropdownFilter.tsx
ardasnturk: this part of the code has been reverted

src/components/filter/Filter.tsx
ardasnturk: It is possible to overwrite the `padding` by using only the small prop for the `padding` in the wrapper component. I added `View` instead of `Wrapper` component and updated the style.

src/helpers/filterTypesHelper.ts
```diff
@@ -0,0 +1,196 @@
...
+  }
+
+  if (filtersByQuery.config.dateStart && filtersByQuery.config.dateEnd) {
+    filtersByQuery.config.dateStart = {
+      dateStart: filtersByQuery.config.dateStart,
+      dateEnd: filtersByQuery.config.dateEnd,
+      type: FILTER_TYPES.DATE
+    };
+
+    delete filtersByQuery.config.dateEnd;
```
donni106, line 106: why is this line needed?

src/components/filter/DateFilter.tsx
```diff
@@ -88,7 +119,7 @@ export const DateFilter = ({ containerStyle, data, filters, setFilters }: Props)
 
   return (
     <>
-      <Label>{texts.filter.date}</Label>
+      <Label bold>{texts.filter.date}</Label>
       <WrapperRow spaceBetween>
         {data.map((item) => {
           useEffect(() => {
```
donni106, line 125: in this effect i found the following code

```
const endTimeOfDay = item.name === 'start_date' ? 'T00:00:00+01:00' : 'T22:59:59+01:00';
```

if 22:59+1 = 23:59 than 00:00+01 = 01:00, isn't it? why did we do that? shouldn't it be 00:00+00 and 23:59+00 to have the full time range of a day?

ardasnturk: The reason we did this is that we needed to add this data in the date filter we developed for SUE as per api requirements.

If we select a date, it automatically adds, for example, `dateT14:23:43+01:93`. If we added this value, we could not filter correctly.

donni106: and what do you say about

> 00:00+01 = 01:00

?
doesn't we want from 00?

PR #1189, opened 2024-09-27
fix: default prop error

ardasnturk:
SVA-1429

### DIFF
--- a/src/components/BB-BUS/IndexFilter.js
+++ b/src/components/BB-BUS/IndexFilter.js
@@ -36,14 +36,14 @@ const initialAZFilterData = alphabetWithUmlauts.map((value, index) => ({
 /* *** */
 
 export const IndexFilter = ({
-  selectedFilter,
-  results,
-  listItems,
-  setListItems,
   areaId,
-  setAreaId,
   areas,
-  loading
+  listItems,
+  loading,
+  results,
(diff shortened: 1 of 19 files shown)


src/components/BB-BUS/IndexFilter.js
```diff
@@ -225,16 +225,12 @@ const styles = StyleSheet.create({
...
-  listItems: PropTypes.array.isRequired,
-  setListItems: PropTypes.func.isRequired,
   areaId: PropTypes.string.isRequired,
-  setAreaId: PropTypes.func.isRequired,
   areas: PropTypes.array,
-  loading: PropTypes.bool.isRequired
-};
-
-IndexFilter.defaultProps = {
-  communities: []
```
donni106, line 239: was this before `areas`? and should `areas` prop be an empty array now per default?

donni106 (approved):

src/components/BB-BUS/IndexFilter.js
donni106: and areas does not need a default?

ardasnturk: I don't think we need it because so far we haven't created it for a default data `areas`. So far I haven't encountered any errors. 🤷‍♂️

PR #1190, opened 2024-09-27
fix: key error & key extractor error

ardasnturk:
SVA-1429

### DIFF
--- a/src/components/CardListItem.js
+++ b/src/components/CardListItem.js
@@ -39,23 +39,24 @@ const renderCardContent = (item, horizontal, sue) => {
           borderRadius={sue ? 0 : imageBorderRadius}
           childrenContainerStyle={stylesWithProps({ aspectRatio, horizontal }).image}
           containerStyle={[styles.imageContainer, styles.sueImageContainer, imageStyle]}
+          key={picture.url}
           source={{ uri: picture.url }}
         />
       ),
     subtitle: () =>
       !!subtitle && (
-        <RegularText small style={[generalStyle, subtitleStyle]}>
+        <RegularText key={subtitle} small style={[generalStyle, subtitleStyle]}>
           {subtitle}
         </RegularText>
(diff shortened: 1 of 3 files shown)


src/screens/AppIntroScreen.tsx
```diff
@@ -9,7 +9,7 @@ import { useStaticContent } from '../hooks';
 import { parseIntroSlides } from '../jsonValidation';
 import { AppIntroSlide } from '../types';
 
-const keyExtractor = (slide: AppIntroSlide) => slide.title + slide.text;
+const keyExtractor = (slide: AppIntroSlide, index: number) => `index${index}-text${slide.text}`;
```
donni106, line 12: why have you decided to go with just text and remove title from key?

ardasnturk: We use it this way in many places in the code and to keep the same structure. I removed it because I don't think we need title if we have id+text.

src/components/CardListItem.js
```diff
@@ -39,23 +39,24 @@ const renderCardContent = (item, horizontal, sue) => {
...
     title: () =>
       !!title && (
-        <BoldText style={[generalStyle, titleStyle]}>
+        <BoldText key={title} style={[generalStyle, titleStyle]}>
           {horizontal ? (title.length > 60 ? title.substring(0, 60) + '...' : title) : title}
         </BoldText>
       ),
     topTitle: () => (
-      <RegularText small style={[!!topTitleStyle && topTitleStyle]}>
+      <RegularText key={topTitle} small style={[!!topTitleStyle && topTitleStyle]}>
```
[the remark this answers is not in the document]
ardasnturk: this contains the elements we added to the sequenceMap array. Since this is an array format, each element needs a key. Otherwise we get this error.

image

donni106 (approved):

PR #1195, opened 2024-10-07
feat(ar): update viro package

ardasnturk:
- updated to async await format because the `isARSupportedOnDevice` function has been updated
- removed `height` and `width` values for `ViroQuad` because they were causing the app to crash

SVA-1306

Currently, the Viro package does not support Expo 51, which causes the app to crash on the iOS platform. 
To test on Android platform, please remove the `expo-av` package in `package.json` and change the `Video` component in `Chat.js `to a comment line. this is necessary due to the incompatibility between the` expo-av` package and Viro.


src/components/augmentedReality/AugmentedReality.js
```diff
@@ -50,18 +50,24 @@ export const AugmentedReality = ({
...
       // if Viro is not integrated, we need to catch the error for `isARSupportedOnDevice of null`
-      console.warn(error);
+      console.error(error);
+    }
+  };
+
+  useEffect(() => {
+    const { settings = {} } = globalSettings;
+
+    if (!!settings.ar.tourId) {
```
donni106, line 69: why is `tourId` added here? do we always have to set a tour id?

src/components/augmentedReality/AugmentedReality.js
```diff
@@ -50,18 +50,24 @@ export const AugmentedReality = ({
   const [isModalVisible, setIsModalVisible] = useState(false);
   const selectedFilterId = filter.find((entry) => entry.selected)?.id;
 
-  useEffect(() => {
-    const { settings = {} } = globalSettings;
-
+  const supportAr = async () => {
```
donni106, line 53: should be a `useCallback` i guess to avoid running multiple times various re-rendering?

src/components/augmentedReality/AugmentedRealityView.js
```diff
@@ -2,6 +2,7 @@ import {
   Viro3DObject,
   ViroAmbientLight,
   ViroARImageMarker,
+  ViroARPlane,
```
donni106, line 5: is this used?

src/components/augmentedReality/AugmentedReality.js
```diff
@@ -50,18 +50,24 @@ export const AugmentedReality = ({
...
-        );
+      const isARSupported = (await isARSupportedOnDevice())?.isARSupported;
+
+      if (isARSupported) {
+        setIsARSupported(true);
+      }
     } catch (error) {
       // if Viro is not integrated, we need to catch the error for `isARSupportedOnDevice of null`
-      console.warn(error);
+      console.error(error);
```
donni106, line 62: why did you switch to `error` instead of `warn`? wouldn't this crash the app entirely?

src/components/augmentedReality/AugmentedReality.js
donni106: or couldn't this be placed directly in the `useEffect`?

src/components/augmentedReality/AugmentedReality.js
ardasnturk: Yes, if you have an app that supports AR, you need to add `tourId` to the `ar` object in `globalSettings`. I think it would be better to add `ar?.tourId` control here just in case.

src/components/augmentedReality/AugmentedReality.js
ardasnturk: I prefer to write it as in `SettingsScreen.js` to have the same structure.

src/components/augmentedReality/AugmentedRealityView.js
ardasnturk: No. I'm deleting it.

src/components/augmentedReality/AugmentedReality.js
ardasnturk: No, this is just a log type. If we need to show an error in the `catch` section, I prefer to show it via `error` instead of `warn`.

src/components/augmentedReality/AugmentedReality.js
donni106: then we should also add `?` to `Component = <AugmentedReality id={settings.ar.tourId} onSettingsScreen />;`?

src/components/augmentedReality/AugmentedReality.js
```diff
@@ -53,16 +53,20 @@ export const AugmentedReality = ({
...
+        if (isARSupported) {
+          setIsARSupported(true);
+        }
+      } catch (error) {
+        // if Viro is not integrated, we need to catch the error for `isARSupportedOnDevice of null`
+        console.error(error);
+      }
+    };
+
+    !!settings.ar?.isARSupported && supportAr();
```
donni106, line 69: there is no `settings.ar?.isARSupported`! you need to remove it. the check with `supportAr()` is enough, right?

src/components/augmentedReality/AugmentedReality.js
ardasnturk: yes you're right

donni106 (approved):

PR #1196, opened 2024-10-08
feat(sue): reduce loading time

ardasnturk:
- integrated `locations` endpoint to reduce long loading time in map view
- added `useInfiniteQuery` for pagination instead of `useQuery`
- added `limit` and `offset` values to query to use pagination
- updated `SORT_BY` options according to server requirements
- removed codes to sort listItem locally since there are sorting options on the server
- added `cacheTime` until end of day
- deleted `RegularText` added for pagination test
- reduced unnecessary codes

SUE-144

### DIFF
--- a/src/queries/SUE/location.tsx
+++ b/src/queries/SUE/location.tsx
@@ -27,8 +27,7 @@ export const locations = async (queryVariables) => {
     resolve(
       response.map((item: any) => {
         const camelCasedItem = _mapKeys(item, (value, key) => _camelCase(key));
-        const statusCode = item.status_code;
-        const status = StatusMapping[statusCode as keyof typeof StatusMapping];
+        const status = StatusMapping[item.status_code as keyof typeof StatusMapping];
 
         return { ...camelCasedItem, status };
       })
--- a/src/queries/SUE/location.tsx
+++ b/src/queries/SUE/location.tsx
@@ -1,0 +1,37 @@
+import _camelCase from 'lodash/camelCase';
+import _mapKeys from 'lodash/mapKeys';
+
+import { fetchSueEndpoints } from '../../helpers';
+
+/* eslint-disable @typescript-eslint/no-duplicate-enum-values */
+enum StatusMapping {
+  TICKET_STATUS_CLOSED = 'Geschlossen',
+  TICKET_STATUS_INVALID = 'Geschlossen',
+  TICKET_STATUS_IN_PROCESS = 'In Bearbeitung',
+  TICKET_STATUS_WAIT_REQUESTOR = 'In Bearbeitung',
+  TICKET_STATUS_WAIT_THIRDPARTY = 'In Bearbeitung',
+  TICKET_STATUS_UNPROCESSED = 'Offen',
+  TICKET_STATUS_OPEN = 'Offen'
+}
+/* eslint-enable @typescript-eslint/no-duplicate-enum-values */
+
+export const locations = async (queryVariables) => {
+  const queryParams = new URLSearchParams(queryVariables);
+  const { sueFetchObj = {}, sueLocationsUrl = '' } = await fetchSueEndpoints();
+
+  const response = await (
+    await fetch(`${sueLocationsUrl}?${queryParams.toString()}`, sueFetchObj)
+  ).json();
+
+  return new Promise((resolve) => {
+    resolve(
+      response.map((item: any) => {
+        const camelCasedItem = _mapKeys(item, (value, key) => _camelCase(key));
+        const statusCode = item.status_code;
+        const status = StatusMapping[statusCode as keyof typeof StatusMapping];
+
+        return { ...camelCasedItem, status };
+      })
+    );
+  });
+};
--- a/src/screens/SUE/SueListScreen.tsx
+++ b/src/screens/SUE/SueListScreen.tsx
@@ -2,11 +2,10 @@
 import { RouteProp } from '@react-navigation/native';
 import { StackNavigationProp } from '@react-navigation/stack';
 import _filter from 'lodash/filter';
-import _reverse from 'lodash/reverse';
-import _sortBy from 'lodash/sortBy';
+import moment from 'moment';
 import React, { useCallback, useContext, useEffect, useMemo, useState } from 'react';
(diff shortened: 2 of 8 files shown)


src/queries/SUE/location.tsx
```diff
@@ -0,0 +1,37 @@
...
+  const response = await (
+    await fetch(`${sueLocationsUrl}?${queryParams.toString()}`, sueFetchObj)
+  ).json();
+
+  return new Promise((resolve) => {
+    resolve(
+      response.map((item: any) => {
+        const camelCasedItem = _mapKeys(item, (value, key) => _camelCase(key));
+        const statusCode = item.status_code;
+        const status = StatusMapping[statusCode as keyof typeof StatusMapping];
```
donni106, line 31: could be reduced to one line as `statusCode` is never used elsewhere

src/queries/SUE/location.tsx
```diff
@@ -0,0 +1,37 @@
...
+export const locations = async (queryVariables) => {
+  const queryParams = new URLSearchParams(queryVariables);
+  const { sueFetchObj = {}, sueLocationsUrl = '' } = await fetchSueEndpoints();
+
+  const response = await (
+    await fetch(`${sueLocationsUrl}?${queryParams.toString()}`, sueFetchObj)
+  ).json();
+
+  return new Promise((resolve) => {
+    resolve(
```
donni106, line 27: why is this new promise and resolve needed?

src/screens/SUE/SueListScreen.tsx
```diff
@@ -29,10 +29,10 @@ import { QUERY_TYPES, getQuery } from '../../queries';
...
 
 const SORT_BY = {
-  REQUESTED_DATE_TIME: 'requestedDatetime',
-  STATUS: 'status',
-  TITLE: 'title',
-  UPDATED_DATE_TIME: 'updatedDatetime'
+  REQUESTED_DATE_TIME: 'requested_datetime DESC',
+  STATUS: 'status DESC',
+  TITLE: 'title ASC',
+  UPDATED_DATE_TIME: 'updated_datetime DESC'
```
donni106, line 35: why is everything DESC but title is ASC?

src/screens/SUE/SueListScreen.tsx
```diff
@@ -80,29 +80,51 @@ export const SueListScreen = ({ navigation, route }: Props) => {
...
+    ({ pageParam = 0 }) =>
       getQuery(query)({
         ...queryVariables,
-        start_date: queryVariables.start_date || queryVariables.initial_start_date
-      })
+        limit,
+        offset: pageParam,
+        start_date: queryVariables.start_date || queryVariables.initial_start_date,
+        sort_attribute: sorting
+      }),
```
donni106, line 107: please reduce duplication with setting

```
const [queryVariables, setQueryVariables] = useState({ 
  ...initialQueryVariables, limit,
  offset: 0,
  start_date: queryVariables.start_date || queryVariables.initial_start_date,
  sort_attribute: sorting 
});
```

and update just `offset: pageParam` down here

src/screens/SUE/SueListScreen.tsx
```diff
@@ -155,12 +168,26 @@ export const SueListScreen = ({ navigation, route }: Props) => {
...
     isConnected && (await refetch());
     setRefreshing(false);
   };
 
+  const fetchMoreData = useCallback(async () => {
+    if (hasNextPage) {
+      return await fetchNextPage();
+    }
+
+    return {};
```
donni106, line 188: why returning empty object here?

src/screens/SUE/SueListScreen.tsx
```diff
@@ -80,29 +80,51 @@ export const SueListScreen = ({ navigation, route }: Props) => {
...
+      }),
+    {
+      getNextPageParam: (lastPage, allPages) => {
+        if (lastPage.length < limit) {
+          return undefined;
+        }
+
+        return allPages.length * limit;
+      },
+      cacheTime: 1000 * 60 * 60 * 24 // 1 day in milliseconds
```
donni106, line 116: please cache always until the end of day and not for 24 hours

src/queries/SUE/location.tsx
ardasnturk: Yes, that's right. 👍

src/queries/SUE/location.tsx
ardasnturk: I used the same method here to ensure that it has the same structure as the other queries (for example `requests`). I don't remember exactly why this is necessary at the moment.

src/screens/SUE/SueListScreen.tsx
ardasnturk: Because this is the correct sorting by the server. If title DESC is set, it is sorted in reversed.

src/queries/SUE/location.tsx
donni106: ah yes i see

src/screens/SUE/SueListScreen.tsx
```diff
@@ -155,12 +168,26 @@ export const SueListScreen = ({ navigation, route }: Props) => {
     return parsedListItem;
   }, [data, query, queryVariables]);
 
+  useEffect(() => {
+    if (queryVariables.sortBy) {
+      setSorting(queryVariables.sortBy);
+    }
+  }, [queryVariables]);
```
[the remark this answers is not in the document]
ardasnturk: Yeah, that's right, you're right. We don't need an extra state. I didn't see that.

src/screens/SUE/SueListScreen.tsx
ardasnturk: When ASC is sorted for status, completed reports are sorted first. But with the last commit I sent, I changed DESC to ASC because in the previous version, the completed reports are sorted first.

src/screens/SUE/SueListScreen.tsx
ardasnturk: I reduced the code repetition a bit, but the `sort_attribute` variable should remain as it is.

donni106 (approved):

PR #1198, opened 2024-10-10
fix: the problem of personalized tiles not being sorted correctly

ardasnturk:
- updated `sortTitles` values in `personalizedTiles` with `umlautSwitcher` to bring mismatched titles into the same format to fix a bug where personalized tiles were not sorted correctly after saving
- removed `replace` from `id` in `DraggableGrid` because `replace` was added at the end of the `umlautSwitcher` function

SVA-1291

Before update:
|before save|update|after save|
|--|--|--|
Simulator Screenshot - iPhone 16 Pro Max - 2024-10-10 at 16 57 06|Simulator Screenshot - iPhone 16 Pro Max - 2024-10-10 at 16 57 19|Simulator Screenshot - iPhone 16 Pro Max - 2024-10-10 at 16 57 22

After update:
|before save|update|after save|
|--|--|--|
Simulator Screenshot - iPhone 16 Pro Max - 2024-10-10 at 17 00 35|Simulator Screenshot - iPhone 16 Pro Max - 2024-10-10 at 17 00 53|Simulator Screenshot - iPhone 16 Pro Max - 2024-10-10 at 17 01 41

### DIFF
--- a/src/components/screens/Service.tsx
+++ b/src/components/screens/Service.tsx
@@ -17,7 +17,7 @@ import { ServiceTile, TServiceTile } from './ServiceTile';
 
 const { MATOMO_TRACKING, UMLAUT_REGEX } = consts;
 
-const umlautSwitcher = (text: string) => {
+export const umlautSwitcher = (text: string) => {
   if (!text) return;
 
   const umlautReplacements = {
@@ -30,7 +30,9 @@ const umlautSwitcher = (text: string) => {
     ß: 'ss'
   };
 
-  const replacedText = text.replace(UMLAUT_REGEX, (match: string) => umlautReplacements[match]);
+  const replacedText = text
+    .replace(UMLAUT_REGEX, (match: string) => umlautReplacements[match])
+    ?.replace('​', '');
 
   return replacedText;
 };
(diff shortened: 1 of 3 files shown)


src/components/screens/Service.tsx
```diff
@@ -32,7 +32,7 @@ const umlautSwitcher = (text: string) => {
 
   const replacedText = text.replace(UMLAUT_REGEX, (match: string) => umlautReplacements[match]);
 
-  return replacedText;
+  return replacedText?.replace('​', '');
```
donni106, line 35: please add `?.replace('​', '')` directly to the line before when creating `replacedText`. thi reads much more accurate i guess.

donni106, line 35: why do we replace `''` with `''`? what trick is this?

ardasnturk: I don't know the reason for this either. There was this code in the development about 2 years ago and since I don't know exactly why, I thought I'd add it here. 🤷‍♂️

ardasnturk: https://github.com/smart-village-solutions/smart-village-app-app/commit/b224795fb9c4703537927397763d18a133ea3a6d

donni106: no hints in the commit message from back then?

ardasnturk: Aha. I found it.

image

donni106: oh yeah! that's how we like it.
so did you copy the code or typed newly? needs to be copied to get the correct signs, right!?

donni106 (approved):

PR #1199, opened 2024-10-11
feat: add `autoplayInterval` to slider

ardasnturk:
- added `autoplayInterval` to give more customization options to the slider
- added options to customize the slider with the `sliderSettings` object added inside the `globalSettings.settings` object
  - autoplayInterval: this value is used to set how many milliseconds the slider will scroll to the next image
  - refreshInterval: allows the image in the slider to be displayed without being cached and refreshes the image after the milliseconds entered here

EP-21

```
{
 ...
  "sliderSettings": {
    "autoplayInterval": 3000,
    "refreshInterval": 3000
  }
},
...
```

before

https://github.com/user-attachments/assets/312db700-8d65-4f78-a878-76b15f6259a7

after

https://github.com/user-attachments/assets/7b6719a6-d59b-4138-8c7b-c5693c33a261

### DIFF
--- a/src/components/ImagesCarousel.js
+++ b/src/components/ImagesCarousel.js
@@ -28,7 +28,7 @@ export const ImagesCarousel = ({
   const { isConnected, isMainserverUp } = useContext(NetworkContext);
   const { globalSettings } = useContext(SettingsContext);
   const { settings = {} } = globalSettings;
-  const { sliderPauseButton = {} } = settings;
+  const { sliderPauseButton = {}, sliderSettings = {} } = settings;
   const {
     horizontalPosition = 'right',
     show: showSliderPauseButton = false,
@@ -62,7 +62,7 @@ export const ImagesCarousel = ({
   const itemWidth = imageWidth();
 
   const renderItem = useCallback(
-    ({ item }) => {
+    ({ item, refreshInterval }) => {
       const { routeName: name, params } = item.picture || {};
 
       // params are available, but missing `shareContent` and `details`
@@ -104,7 +104,7 @@ export const ImagesCarousel = ({
                   containerStyle={styles.imageContainer}
                   message={item.message}
                   navigation={navigation}
-                  refreshInterval={item.refreshInterval}
+                  refreshInterval={item.refreshInterval || refreshInterval}
                   source={item.picture}
                 />
               );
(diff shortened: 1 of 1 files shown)


src/components/ImagesCarousel.js
```diff
@@ -130,8 +130,10 @@ export const ImagesCarousel = ({
 
   if (!data || data.length === 0) return null;
 
+  const sliderSettings = data[0]?.sliderSettings || {};
+
   // filter data for present items and items with active date/time periods
-  const carouselData = data.filter((item) => item && isActive(item));
+  const carouselData = data.filter((item) => item && isActive(item) && !item.sliderSettings);
```
donni106, line 136: the comment is now missing information about the newly added condition

src/components/ImagesCarousel.js
```diff
@@ -130,8 +130,10 @@ export const ImagesCarousel = ({
 
   if (!data || data.length === 0) return null;
 
+  const sliderSettings = data[0]?.sliderSettings || {};
```
donni106, line 133: `sliderSettings`? in your example in the PR description there ist `settings`.
please explain the new "first element of data" logic more explicitly and how this affect the `carouselData`

src/components/ImagesCarousel.js
ardasnturk: I updated the commit.

src/components/ImagesCarousel.js
ardasnturk: I updated the code but forgot to send a commit. The code is up to date now.

src/components/ImagesCarousel.js
donni106: i don't get it. i read that the carousel data is only the items, that has no `settings` entry? so what about the items, that have `settings` explicitly to set a refresh interval for example?

donni106: ah ok, i see that each item itself normally do not have a `settings` but the configs directly, like

```
[
  {
    "picture": {
      "uri": "https://cataas.com/cat"
    },
    "message": "<h1 style=\"color: #ffffff; margin-bottom: 0\">CatCäm</h1>",
    "dates": [
      {
        "dateStart": "2021-01-01"
      }
    ],
    "refreshInterval": 5000
  }
]
```

ardasnturk: `autoplayInterval` is the value that allows the slider to scroll the images after the desired time.

Ok, ich habe den Code aktualisiert und `sliderSettings` zu `globalSettings.settings` hinzugefügt, so dass `homeCarousel` wie bisher verwendet werden kann.

donni106 (approved):

d4v3d4vE (approved):

PR #1202, opened 2024-10-15
feat: remove key warnings

ardasnturk:
- added a general `keyExtractor` to eliminate the key warning in `CardListItem`
- added `draggableKey` prop to remove key warning on personalized tiles

SVA-1429

### DIFF
--- a/src/components/screens/Service.tsx
+++ b/src/components/screens/Service.tsx
@@ -76,10 +76,11 @@ export const Service = ({
     (item: TServiceTile, index: number) => (
       <ServiceTile
         draggableId={umlautSwitcher(item.title) || umlautSwitcher(item.accessibilityLabel)}
+        draggableKey={`item${item.title || item.accessibilityLabel}-index${index}`}
         hasDiagonalGradientBackground={hasDiagonalGradientBackground}
         isEditMode={isEditMode}
         item={item}
-        key={index + (item.title || item.accessibilityLabel)}
+        key={`item${item.title || item.accessibilityLabel}-index${index}`}
         onToggleVisibility={onToggleVisibility}
         tileSizeFactor={tileSizeFactor}
       />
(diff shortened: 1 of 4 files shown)


src/components/screens/Service.tsx
```diff
@@ -76,6 +76,7 @@ export const Service = ({
     (item: TServiceTile, index: number) => (
       <ServiceTile
         draggableId={umlautSwitcher(item.title) || umlautSwitcher(item.accessibilityLabel)}
+        draggableKey={index + (item.title || item.accessibilityLabel)}
```
donni106, line 79: let's use `item${item}-index${index}` here as well with `title` or `accessibilityLabel` as `item`?

ardasnturk: Yes, that's right. I didn't want to change it just because the `key` prop is used in this format.

ardasnturk: https://github.com/smart-village-solutions/smart-village-app-app/commit/6090ffdf266c658530e01696e576e549ef9a5065

src/components/screens/Service.tsx
```diff
@@ -76,10 +76,11 @@ export const Service = ({
     (item: TServiceTile, index: number) => (
       <ServiceTile
         draggableId={umlautSwitcher(item.title) || umlautSwitcher(item.accessibilityLabel)}
+        draggableKey={`item${item.title || item.accessibilityLabel}-index${index}`}
         hasDiagonalGradientBackground={hasDiagonalGradientBackground}
         isEditMode={isEditMode}
         item={item}
-        key={index + (item.title || item.accessibilityLabel)}
+        key={`item${item.title || item.accessibilityLabel}-index${index}`}
```
donni106, line 83: why do we need both keys?

ardasnturk: Because we really use it for the key property, it is not a specialized prop. We cannot catch the key prop with `child.props.key` that we use in the `DraggableItem` component in `DraggableGrid`. Instead of key prop, we need a key with the same value as key (not necessarily the same, but uniq) but with a different prop name. This way we can catch the key for `DraggableItem` and use it. Otherwise `child.props.key` is always `undefined`.

donni106 (approved):

d4v3d4vE (commented):
nice changes

PR #1205, opened 2024-10-16
feat(event records): reduce queried fields and new query for count

donni106:
this should decrease the work load when requesting data from the server to the minimum needed amount of data.

SVAK-135


src/queries/eventRecords.js
```diff
@@ -1,11 +1,9 @@
 import gql from 'graphql-tag';
 
-const defaultFields = `
+const defaultFragment = `
```
donni106, line 3: renamed to be conform with the other fragment

src/queries/eventRecords.js
```diff
@@ -138,7 +114,44 @@ export const GET_EVENT_RECORDS = gql`
...
+      locations: $locations
+      skip: $offset
+      order: $order
+      categoryId: $categoryId
+      dateRange: $dateRange
+      dataProvider: $dataProvider
+      dataProviderId: $dataProviderId
+    ) {
+      id
+      ...dateFields
```
ardasnturk, line 153: Do we need `dateFields`? since this query only shows the count of current events, I think the `id` value is enough.

src/components/widgets/EventWidget.tsx
```diff
@@ -1,49 +1,59 @@
...
+            : QUERY_TYPES.EVENT_RECORDS_COUNT
+        ),
+        queryVariables
+      );
+    },
     {
-      fetchPolicy,
-      variables: queryVariables,
-      skip: !refreshTime
+      keepPreviousData: true
```
ardasnturk, line 56: What exactly does `keepPreviousData` do? How long does it keep the data? When new data comes in, does it overwrite it or add it to the old data?

donni106: ha good part. this is something useful for pagination but not needed here. it can be eliminated i guess.
for pagination it can be used to keep data from a fetch available while new data is being requested.

src/queries/eventRecords.js
ardasnturk: Ach okay. I didn't think about the server side. 👍

d4v3d4vE (commented):
> in addition we need react-query instead of apollo to make the new query with date fields succeed

why didn't it work before with react-apollo, what was the reason for the switch to react-query?

donni106:
seems that react-apollo has problem with fragment in the new query

d4v3d4vE:
> seems that react-apollo has problem with fragment in the new query

ah okay. I haven't heard of fragments before and just took a glimpse into the react-query docs about fragments. 
fragments are really cool.

PR #1206, opened 2024-10-17
fix: home section query with react-query

donni106:
- react-apollo is not able to handle the new `date` field, which we forgot to use for home section queries

SVASD-10

### DIFF
--- a/src/components/HomeSection.tsx
+++ b/src/components/HomeSection.tsx
@@ -1,9 +1,10 @@
 import { StackNavigationProp } from '@react-navigation/stack';
 import React from 'react';
-import { useQuery } from 'react-apollo';
+import { useQuery } from 'react-query';
 
 import { useHomeRefresh, useVolunteerData } from '../hooks';
 import { getQuery, QUERY_TYPES } from '../queries';
+import { ReactQueryClient } from '../ReactQueryClient';
 
 import { DataListSection } from './DataListSection';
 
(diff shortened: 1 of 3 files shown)


src/components/HomeSection.tsx
```diff
@@ -40,14 +40,17 @@ export const HomeSection = ({
...
+    [query, queryVariables],
+    async () => {
+      const client = await ReactQueryClient();
+
+      return await client.request(getQuery(query), queryVariables);
+    },
+    {
+      keepPreviousData: true
+    }
+  );
```
ardasnturk, line 53: How about creating a generic client provider instead? Because we're going to need it. I think creating a new provider for react-query along with react-apollo will make future integrations easier.

donni106: yes you are right, this will need a refactoring somewhen, when we completely switch to react query. for now we keep it like this with some duplications.

src/components/HomeSection.tsx
```diff
@@ -40,14 +40,17 @@ export const HomeSection = ({
...
-  });
+  const { data, isLoading, refetch } = useQuery(
+    [query, queryVariables],
+    async () => {
+      const client = await ReactQueryClient();
+
+      return await client.request(getQuery(query), queryVariables);
+    },
+    {
+      keepPreviousData: true
```
donni106, line 51: i will also remove this here, see https://github.com/smart-village-solutions/smart-village-app-app/pull/1205#discussion_r1808830666

ardasnturk (approved):

PR #1212, opened 2024-11-04
feat(feedback): add text info for required fields

donni106:
Simulator Screenshot - iPhone 16 - 2024-11-04 at 17 53 07

SVASD-13

### DIFF
--- a/src/config/texts.js
+++ b/src/config/texts.js
@@ -1,5 +1,4 @@
 import appJson from '../../app.json';
-import { search } from '../helpers';
 
 export const texts = {
   accessibilityLabels: {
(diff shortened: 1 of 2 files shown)


src/config/texts.js
```diff
@@ -442,7 +441,8 @@ export const texts = {
       email: 'E-Mail',
       phone: 'Telefon',
       message: 'Ihre Mitteilung',
-      checkbox: 'Ich bin mit dem Speichern meiner Daten einverstanden.'
+      checkbox: 'Ich bin mit dem Speichern meiner Daten einverstanden.',
+      requiredFields: '* Pflichtfelder'
```
ardasnturk, line 445: How about alphabetical sorting?

ardasnturk (approved):

src/config/texts.js
donni106: i do not wanted to mess around here with just adding one new key. and the new key is sorted correctly :)

PR #1214, opened 2024-11-20
feat: add eslint/no-cycle rule

d4v3d4vE:
- added to prevent the formation of [barrel files](https://tkdodo.eu/blog/please-stop-using-barrel-files)

eslint doc: https://github.com/import-js/eslint-plugin-import/blob/main/docs/rules/no-cycle.md#importno-cycle

### DIFF
--- a/.eslintrc.yml
+++ b/.eslintrc.yml
@@ -55,6 +55,9 @@ rules:
     - always
   react/display-name:
     - off
+  import/no-cycle:
+    - error
+    - maxDepth: 1
   import/order:
     - error
     - newlines-between: always


donni106:
Important note that I have found in the docs:

> This rule is comparatively computationally expensive. If you are pressed for lint time, or don't think you have an issue with dependency cycles, you may not want this rule enabled.

https://github.com/import-js/eslint-plugin-import/blob/09476d7dac1ab36668283f9626f85e2223652b37/docs/rules/no-cycle.md#when-not-to-use-it

Does this effect/interest us?

ardasnturk:
> Important note that I have found in the docs:
> 
> > This rule is comparatively computationally expensive. If you are pressed for lint time, or don't think you have an issue with dependency cycles, you may not want this rule enabled.
> 
> https://github.com/import-js/eslint-plugin-import/blob/09476d7dac1ab36668283f9626f85e2223652b37/docs/rules/no-cycle.md#when-not-to-use-it
> 
> Does this effect/interest us?

No, I don't think we have such a problem at the moment. Therefore, this is not important to us.

PR #1215, opened 2024-11-26
feat: add link that opens from browser

ardasnturk:
- added the ability to automatically open the browser if the `WebScreen` in the app is redirected with the `isExternal` param
- updated `MultiButtonWithSubQuery` to add `isExternal` as param when redirecting via multi button screen

SVASD-15

changes in main-server:
- redirect to browser:

```
  {
    "title": "YT",
    "routeName": "Web",
    "params": {
      "webUrl": "https://www.youtube.com/",
      "isExternal": true
    }
  }
```

- redirect to WebScreen in the app:

```
  {
    "title": "YT",
    "routeName": "Web",
    "params": {
      "webUrl": "https://www.youtube.com/"
    }
  }
```

### DIFF
--- a/src/components/MultiButtonWithSubQuery.tsx
+++ b/src/components/MultiButtonWithSubQuery.tsx
@@ -17,6 +17,7 @@ export const navigateWithSubQuery = ({
   title?: string;
   params?:
     | {
+        isExternal?: boolean;
         routeName: string;
         webUrl: string;
         paramsForButton?: {
@@ -55,6 +56,7 @@ export const navigateWithSubQuery = ({
     // if `params` is an object and does not contain the `paramsForButton` object,
     // it contains a `routeName` and a `webUrl`
     return navigation.push(params.routeName, {
+      isExternal: params.isExternal,
       rootRouteName,
       title,
       webUrl: params.webUrl
@@ -68,6 +70,7 @@ export const navigateWithSubQuery = ({
   // will be ignored
   return navigation.push(subQuery.routeName, {
     ...subParams,
+    isExternal: subQuery.isExternal,
     rootRouteName,
     title,
     webUrl: subQuery.webUrl
--- a/src/screens/WebScreen.js
+++ b/src/screens/WebScreen.js
@@ -1,12 +1,13 @@
 import { noop } from 'lodash';
 import PropTypes from 'prop-types';
 import React, { useContext, useEffect } from 'react';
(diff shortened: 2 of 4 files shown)


src/components/MultiButtonWithSubQuery.tsx
```diff
@@ -57,7 +58,8 @@ export const navigateWithSubQuery = ({
     return navigation.push(params.routeName, {
       rootRouteName,
       title,
-      webUrl: params.webUrl
+      webUrl: params.webUrl,
+      isExternal: params.isExternal
```
donni106, line 62: alphabetical as well?

src/components/MultiButtonWithSubQuery.tsx
```diff
@@ -70,7 +72,8 @@ export const navigateWithSubQuery = ({
     ...subParams,
     rootRouteName,
     title,
-    webUrl: subQuery.webUrl
+    webUrl: subQuery.webUrl,
+    isExternal: subQuery.isExternal
```
donni106, line 76: alphabetical as well?

src/screens/WebScreen.js
```diff
@@ -17,13 +18,21 @@ export const WebScreen = ({ navigation, route }) => {
...
     isConnected && webUrl && trackScreenViewAsync(`${MATOMO_TRACKING.SCREEN_VIEW.WEB} / ${webUrl}`);
   }, [webUrl]);
 
+  useEffect(() => {
+    if (isExternal) {
+      openLink(webUrl);
+      navigation.goBack();
+    }
+  }, []);
+
```
donni106, line 35: i think we should add `isExternal` here `if (!webUrl) return null;` to ensure, that nothing is tried to be rendered for these cases. i do not know how react behaves before navigating. with rendering `null` we know, that nothing happens that we do not want, right?!

ardasnturk: Yes, that's right. I will update the code as follows. 
Because as you said, the page will be rendered until goBack() runs, maybe even for a short time. I think this will work with an invisible speed, but it will still be an unnecessary process.

```
  if (!webUrl && isExternal) return null;
```

donni106: no, it should be an or with ||. you will probably never have no web url and is external at the same time

ardasnturk: Yes, sorry. You're right 👍

donni106 (approved):

PR #1226, opened 2025-01-07
feat(tab): configureable tab navigation

ardasnturk:
tab navigation feature that can be configured via main-server has been added

To test this feature:

- Create a JSON type content named `tabNavigation` on main-server:
- Update the JSON content as follows or add the desired navigation structure.
```
{
  "activeTintColor": "#0f4618",
  "activeBackgroundColor": "#fff",
  "inactiveTintColor": "#107821",
  "inactiveBackgroundColor": "#fff",
  "tabConfigs": [
    "Home",
    "Service",
    "Events",
    "Index",
    "About",
    {
      "label": "Custom",
      "screen": "VolunteerHome",
      "iconName": "Volunteer",
      "iconSize": 30,
      "accessibilityLabel": "Volunteer",
      "iconStyle": {
        "marginTop": 20
      },
      "iconLandscapeStyle": {
        "marginTop": 20
      }
    }
  ]
}
```
- Delete the app's cache and reload it

## Screenshots
|before|after|
|--|--|
Simulator Screenshot - iPhone 16 Pro Max - 2025-01-07 at 12 22 36|Simulator Screenshot - iPhone 16 Pro Max - 2025-01-07 at 12 21 44

SVA-1531

### DIFF
--- a/src/components/OrientationAwareIcon.tsx
+++ b/src/components/OrientationAwareIcon.tsx
@@ -1,11 +1,12 @@
 import React, { useContext } from 'react';
 import { StyleProp, StyleSheet, ViewStyle } from 'react-native';
 
-import { consts, device, IconProps, normalize } from '../config';
+import { consts, device, Icon as IconComponent, IconProps, normalize } from '../config';
 import { OrientationContext } from '../OrientationProvider';
 
 type Props = IconProps & {
   Icon: (props: IconProps) => JSX.Element;
+  iconName: keyof typeof IconComponent;
   landscapeIconStyle?: StyleProp<ViewStyle>;
   landscapeStyle?: StyleProp<ViewStyle>;
   size?: number;
@@ -17,6 +18,7 @@ export const OrientationAwareIcon = (props: Props) => {
   const { orientation, dimensions } = useContext(OrientationContext);
   const {
     Icon,
+    iconName,
     iconStyle,
     landscapeIconStyle,
     landscapeStyle,
@@ -24,19 +26,37 @@ export const OrientationAwareIcon = (props: Props) => {
     style
   } = props;
 
+  const SelectedIcon = Icon || IconComponent.NamedIcon;
+
   const needLandscapeStyle =
     orientation === 'landscape' || dimensions.width > consts.DIMENSIONS.FULL_SCREEN_MAX_WIDTH;
 
   // we need adjustments only on iOS, so otherwise just return the item with the usual props
   // TODO: is this still needed with Expo 48?
   if (!needLandscapeStyle || !(device.platform === 'ios')) {
-    return <Icon {...props} />;
+    if (SelectedIcon === IconComponent.NamedIcon) {
+      return <SelectedIcon {...props} name={iconName} />;
+    }
+
+    return <SelectedIcon {...props} />;
   }
 
   const adjustedSize = size * LANDSCAPE_ADJUSTMENT_FACTOR;
 
+  if (SelectedIcon === IconComponent.NamedIcon) {
+    return (
+      <SelectedIcon
+        {...props}
+        name={iconName}
+        iconStyle={[iconStyle, styles.marginIcon, { width: adjustedSize }, landscapeIconStyle]}
+        size={adjustedSize}
+        style={[style, landscapeStyle]}
+      />
+    );
+  }
+
   return (
-    <Icon
+    <SelectedIcon
       {...props}
       iconStyle={[iconStyle, styles.marginIcon, { width: adjustedSize }, landscapeIconStyle]}
       size={adjustedSize}
--- a/src/components/OrientationAwareIcon.tsx
+++ b/src/components/OrientationAwareIcon.tsx
@@ -27,42 +27,32 @@ export const OrientationAwareIcon = (props: Props) => {
   } = props;
 
   const SelectedIcon = Icon || IconComponent.NamedIcon;
+  const isNamedIcon = SelectedIcon === IconComponent.NamedIcon;
 
   const needLandscapeStyle =
     orientation === 'landscape' || dimensions.width > consts.DIMENSIONS.FULL_SCREEN_MAX_WIDTH;
 
-  // we need adjustments only on iOS, so otherwise just return the item with the usual props
-  // TODO: is this still needed with Expo 48?
-  if (!needLandscapeStyle || !(device.platform === 'ios')) {
-    if (SelectedIcon === IconComponent.NamedIcon) {
-      return <SelectedIcon {...props} name={iconName} />;
+  const getIconProps = () => {
+    const baseProps = {
+      ...props,
+      ...(isNamedIcon ? { name: iconName } : {})
+    };
+
+    if (device.platform === 'android' || !needLandscapeStyle) {
+      return baseProps;
     }
 
-    return <SelectedIcon {...props} />;
-  }
+    const adjustedSize = size * LANDSCAPE_ADJUSTMENT_FACTOR;
 
-  const adjustedSize = size * LANDSCAPE_ADJUSTMENT_FACTOR;
+    return {
+      ...baseProps,
+      iconStyle: [iconStyle, styles.marginIcon, { width: adjustedSize }, landscapeIconStyle],
+      size: adjustedSize,
+      style: [style, landscapeStyle]
+    };
+  };
 
-  if (SelectedIcon === IconComponent.NamedIcon) {
-    return (
-      <SelectedIcon
-        {...props}
-        name={iconName}
-        iconStyle={[iconStyle, styles.marginIcon, { width: adjustedSize }, landscapeIconStyle]}
-        size={adjustedSize}
-        style={[style, landscapeStyle]}
-      />
-    );
-  }
-
-  return (
-    <SelectedIcon
-      {...props}
-      iconStyle={[iconStyle, styles.marginIcon, { width: adjustedSize }, landscapeIconStyle]}
-      size={adjustedSize}
-      style={[style, landscapeStyle]}
-    />
-  );
+  return <SelectedIcon {...getIconProps()} />;
 };
 
 const styles = StyleSheet.create({
--- a/src/config/navigation/defaultStackConfig.tsx
+++ b/src/config/navigation/defaultStackConfig.tsx
@@ -89,89 +89,108 @@ const { MATOMO_TRACKING } = consts;
 
 export const defaultStackConfig = ({
   initialRouteName,
+  initialParams,
   isDrawer
 }: {
   initialRouteName: ScreenName;
+  initialParams?: Record<string, any>;
   isDrawer: boolean;
 }): StackConfig => ({
   initialRouteName,
   screenOptions: getScreenOptions({ withDrawer: isDrawer }),
   screenConfigs: [
     {
+      initialParams,
       routeName: ScreenName.About,
       screenComponent: AboutScreen,
       screenOptions: { title: texts.screenTitles.about }
     },
     {
+      initialParams,
       routeName: ScreenName.ARInfo,
       screenComponent: ARInfoScreen,
       screenOptions: { title: texts.augmentedReality.arInfoScreen.header }
     },
     {
+      initialParams,
       routeName: ScreenName.ARShow,
       screenComponent: ARShowScreen,
       screenOptions: { headerShown: false }
     },
     {
+      initialParams,
       routeName: ScreenName.ArtworkDetail,
       screenComponent: ArtworkDetailScreen,
       screenOptions: { title: texts.augmentedReality.artworkDetailScreen.header }
     },
     {
+      initialParams,
       routeName: ScreenName.BBBUSIndex,
       screenComponent: BBBUSIndexScreen
     },
     {
+      initialParams,
       routeName: ScreenName.BBBUSDetail,
       screenComponent: BBBUSDetailScreen,
       screenOptions: getScreenOptions({ withDrawer: isDrawer, withShare: true })
     },
     {
+      initialParams,
       routeName: ScreenName.BookmarkCategory,
       screenComponent: BookmarkCategoryScreen
     },
     {
+      initialParams,
       routeName: ScreenName.Bookmarks,
       screenComponent: BookmarkScreen,
       screenOptions: getScreenOptions({ withDrawer: isDrawer })
     },
     {
+      initialParams,
       routeName: ScreenName.Category,
       screenComponent: IndexScreen
     },
     {
+      initialParams,
       routeName: ScreenName.ConsulHomeScreen,
       screenComponent: ConsulHomeScreen,
       screenOptions: { title: texts.screenTitles.consul.home }
     },
     {
+      initialParams,
       routeName: ScreenName.ConsulRegisterScreen,
       screenComponent: ConsulRegisterScreen,
       screenOptions: { title: texts.screenTitles.consul.register }
     },
     {
+      initialParams,
       routeName: ScreenName.ConsulRegisteredScreen,
       screenComponent: ConsulRegisteredScreen,
       screenOptions: { title: texts.screenTitles.consul.register }
     },
     {
+      initialParams,
       routeName: ScreenName.ConsulLoginScreen,
       screenComponent: ConsulLoginScreen,
       screenOptions: { title: texts.screenTitles.consul.login }
     },
     {
+      initialParams,
       routeName: ScreenName.ConsulIndexScreen,
       screenComponent: ConsulIndexScreen
     },
     {
+      initialParams,
       routeName: ScreenName.ConsulDetailScreen,
       screenComponent: ConsulDetailScreen
     },
     {
+      initialParams,
       routeName: ScreenName.ConsulStartNewScreen,
       screenComponent: ConsulStartNewScreen
     },
     {
+      initialParams,
       routeName: ScreenName.Company,
       screenComponent: getTilesScreen({
         matomoString: MATOMO_TRACKING.SCREEN_VIEW.COMPANY,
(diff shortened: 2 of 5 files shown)


src/config/navigation/defaultStackConfig.tsx
```diff
@@ -444,68 +501,82 @@ export const defaultStackConfig = ({
...
       screenComponent: WasteReminderScreen,
       screenOptions: { title: texts.screenTitles.wasteCollection }
     },
     {
+      initialParams,
       routeName: ScreenName.Weather,
       screenComponent: WeatherScreen
     },
     {
+      initialParams,
```
donni106, line 579: why is every route needing the initial params?

src/config/navigation/tabConfig.tsx
```diff
@@ -1,4 +1,5 @@
 /* eslint-disable react-native/no-inline-styles */
+import { upperFirst } from 'lodash';
```
donni106, line 2: please use `import _upperFirst from 'lodash/upperFirst';` as we did already in other places

src/config/navigation/tabConfig.tsx
```diff
@@ -102,3 +106,31 @@ export const tabNavigatorConfig: TabNavigatorConfig = {
...
   ]
 };
+
+export const createDynamicTabConfig = (
+  accessibilityLabel: string,
+  iconName: string,
+  iconSize: number,
+  index: number,
+  label: string,
+  length: number,
```
donni106, line 116: what about `total` or `totalCount` instead of `length`?

src/config/navigation/tabConfig.tsx
```diff
@@ -102,3 +106,31 @@ export const tabNavigatorConfig: TabNavigatorConfig = {
...
+  }),
+  tabOptions: {
+    tabBarAccessibilityLabel: `${accessibilityLabel || label} (Tab ${index + 1} von ${length})`,
+    tabBarLabel: label,
+    tabBarIcon: ({ color }: TabBarIconProps) => (
+      <OrientationAwareIcon
+        color={color}
+        Icon={Icon[upperFirst(iconName)]}
+        size={normalize(iconSize)}
+      />
```
donni106, line 133: what if an icon would need more specific props like

```
<OrientationAwareIcon
  color={color}
  Icon={Icon.About}
  landscapeStyle={{ marginRight: -normalize(6) }}
  size={normalize(28)}
  style={{ marginTop: normalize(3) }}
/>
```

src/config/navigation/tabConfig.tsx
```diff
@@ -102,3 +106,31 @@ export const tabNavigatorConfig: TabNavigatorConfig = {
     aboutTabConfig
   ]
 };
+
+export const createDynamicTabConfig = (
+  accessibilityLabel: string,
+  iconName: string,
+  iconSize: number,
```
donni106, line 113: `iconSize` should have a default value for cases, where nothing is configured like

```
{
  "label": "Custom",
  "screen": "VolunteerHome",
  "iconName": "Volunteer"
}
```

src/config/navigation/tabConfig.tsx
```diff
@@ -102,3 +106,31 @@ export const tabNavigatorConfig: TabNavigatorConfig = {
     aboutTabConfig
   ]
 };
+
+export const createDynamicTabConfig = (
+  accessibilityLabel: string,
+  iconName: string,
```
donni106, line 112: should we use a question mark icon as default icon, if nothing is set?

src/navigation/MainTabNavigator.tsx
```diff
@@ -1,37 +1,154 @@
...
 
 import { getStackNavigator } from './AppStackNavigator';
 
+const defaultTabRoutes: TabRoutes = {
+  tabConfig: {
+    activeBackgroundColor: colors.surface,
+    activeTintColor: colors.accent,
+    inactiveBackgroundColor: colors.surface,
+    inactiveTintColor: colors.primary,
+    tabs: ['Home', 'Service', 'Events', 'Index', 'About']
```
donni106, line 21: why is `Index` a default? i suggest a default of the ones we have in `tabNavigatorConfig`, so:

```
homeTabConfig,
serviceTabConfig,
pointOfInterestTabConfig,
eventsTabConfig,
aboutTabConfig
```

src/navigation/MainTabNavigator.tsx
```diff
@@ -1,37 +1,154 @@
...
+
+type TabRoutes = {
+  tabConfig: {
+    activeTintColor: string;
+    inactiveTintColor: string;
+    activeBackgroundColor: string;
+    inactiveBackgroundColor: string;
+    tabs: (Tab | string)[];
+  };
+};
```
donni106, line 42: can the types be moved in the types folder?

src/navigation/MainTabNavigator.tsx
```diff
@@ -1,37 +1,154 @@
...
+const useTabRoutes = () => {
+  const { data, error, loading } = useStaticContent<TabRoutes>({
+    name: 'tabNavigation',
+    type: 'json'
+  });
+
+  const [tabRoutes, setTabRoutes] = useState<TabRoutes>(defaultTabRoutes);
+
+  useEffect(() => {
+    const tabs: (Tab | string)[] = data?.tabConfig.tabs || defaultTabRoutes.tabConfig.tabs;
```
donni106, line 53: i would add a second ? for accessing `tabs` of `tabConfigs` in `data`, because it is not 100% sure that there is `tabConfig` if data exists.

and what about `data: tabRoutes` to have a better read here with `tabRoutes?.tabConfig?.tabs || defaultTabRoutes.tabConfig.tabs`?

src/navigation/MainTabNavigator.tsx
```diff
@@ -1,37 +1,154 @@
...
+    const { activeBackgroundColor, activeTintColor, inactiveBackgroundColor, inactiveTintColor } =
+      data?.tabConfig || defaultTabRoutes.tabConfig;
```
donni106, line 55: why not grabbing `tabs` here as well but separately a line before?

src/navigation/MainTabNavigator.tsx
```diff
@@ -1,37 +1,154 @@
...
+
+    setTabRoutes((prev) => {
+      const filteredTabs = defaultTabNavigatorConfig.tabConfigs.filter((tabConfig) => {
+        const screen = tabConfig.stackConfig.initialRouteName;
+
+        return tabs.some((tab) =>
+          typeof tab === 'string' ? tab === screen : tab.screen === screen
+        );
+      });
```
donni106, line 64: what is happening here in human words? filtering tabs with a `some` inside for what exact purpose?

src/navigation/MainTabNavigator.tsx
```diff
@@ -1,37 +1,154 @@
...
+
+      const dynamicTabs = tabs
+        .map((tab, index) => {
+          if (typeof tab === 'string') {
+            const screen = tab;
+            return filteredTabs.find(
+              (tabConfig) => tabConfig.stackConfig.initialRouteName === screen
+            );
```
donni106, line 72: `const screen = tab;` can be removed and result in `tabConfig.stackConfig.initialRouteName === tab`, right?

src/navigation/MainTabNavigator.tsx
```diff
@@ -1,37 +1,154 @@
...
+              tab.iconSize,
+              index,
+              tab.label,
+              tabs.length,
+              tab.screen,
+              tab.params
+            );
+          }
+        })
+        .filter(Boolean) as TabConfig[];
```
donni106, line 86: why is this filter needed? can it happen, that there are undefined entries in the array? if yes, how?

src/config/navigation/defaultStackConfig.tsx
ardasnturk: Because every screen can be tabbed and screens may need param. InitialParams value may not be valid for all screens, but I felt the need to add it in case you need it.

src/config/navigation/tabConfig.tsx
ardasnturk: I didn't add a customization option because we don't always need these props, but it can be added. I will develop this feature.

src/config/navigation/tabConfig.tsx
ardasnturk: yes you are right. I will give a value of 24 as standard. 👍

src/config/navigation/tabConfig.tsx
ardasnturk: Instead of specifying a standard icon, I find it better to integrate `Icon.NamedIcon` into the `OrientationAwareIcon` component. This way we can use Tabler icons as we do in `HomeService`. Currently we can only use the icons that are attached to the `Icon` component.

src/navigation/MainTabNavigator.tsx
ardasnturk: Yes, that's right. But I don't need to do that right now because I've changed the structure of the code. Currently we can access the desired data with `data?.tabConfigs`.

We have a state called `tabRoutes`. Maybe `tabRoutesData`?

src/navigation/MainTabNavigator.tsx
ardasnturk: Yes, that's right. I'll do it.

src/navigation/MainTabNavigator.tsx
ardasnturk: Actually I don't think we need this `filteredTabs` anymore.

To explain the code here:

default scrolls through the tabs and filters by `initialRouteName`. Then it uses this filtering in `dynamicTabs`. But it also works if we apply filtering directly in `dynamicTabs`. This is just creating confusion.

src/navigation/MainTabNavigator.tsx
ardasnturk: Yes, I saw that too and changed it.

src/navigation/MainTabNavigator.tsx
ardasnturk: Actually, this is a filter that is added because objects can be 'null' or 'undefined' during development. But we don't need it at the moment.

src/navigation/MainTabNavigator.tsx
ardasnturk: I updated it as `tabRoutesData`.

src/config/navigation/tabConfig.tsx
donni106: True, it is not needed always, but sometimes :D

src/config/navigation/tabConfig.tsx
donni106: Yes, being able to use icons the way we can use it for service tiles is a good idea.

src/navigation/MainTabNavigator.tsx
```diff
@@ -1,37 +1,114 @@
...
+      inactiveBackgroundColor,
+      inactiveTintColor,
+      tabConfigs
+    } = tabRoutesData || tabNavigatorConfig;
+
+    setTabRoutes((prev) => {
+      const dynamicTabs = tabConfigs?.map((tab, index) => {
+        if (typeof tab === 'string') {
+          return tabNavigatorConfig.tabConfigs.find(
+            (tabConfig) => tabConfig.stackConfig.initialRouteName === tab
```
donni106, line 37: should this always use `tabNavigatorConfig.tabConfigs` and not `tabConfigs` from `tabRoutesData || tabNavigatorConfig`?

ardasnturk: Yes. Here we are comparing defaultTabs with the array on main-server. If the string in the main-server array matches the `initialRouteName` in the `tabNavigatorConfig` (default), that tab is automatically selected.

src/components/OrientationAwareIcon.tsx
```diff
@@ -17,26 +18,45 @@ export const OrientationAwareIcon = (props: Props) => {
...
 
   // we need adjustments only on iOS, so otherwise just return the item with the usual props
   // TODO: is this still needed with Expo 48?
   if (!needLandscapeStyle || !(device.platform === 'ios')) {
-    return <Icon {...props} />;
+    if (SelectedIcon === IconComponent.NamedIcon) {
+      return <SelectedIcon {...props} name={iconName} />;
+    }
+
+    return <SelectedIcon {...props} />;
```
donni106, line 41: can you think of some refactoring to have this not that duplicated and with an own if?

src/navigation/MainTabNavigator.tsx
donni106: why did you only add `// if tab is a string, find the corresponding tabConfig from default config` and not your whole two sentences. please update that.

src/components/OrientationAwareIcon.tsx
ardasnturk: I greatly reduced code repetition.

[fa2b69f](https://github.com/smart-village-solutions/smart-village-app-app/pull/1226/commits/fa2b69f36d20317638b6c5f16cdd1891288aaa54)

src/components/OrientationAwareIcon.tsx
```diff
@@ -17,32 +18,43 @@ export const OrientationAwareIcon = (props: Props) => {
...
+
   const needLandscapeStyle =
     orientation === 'landscape' || dimensions.width > consts.DIMENSIONS.FULL_SCREEN_MAX_WIDTH;
 
-  // we need adjustments only on iOS, so otherwise just return the item with the usual props
-  // TODO: is this still needed with Expo 48?
-  if (!needLandscapeStyle || !(device.platform === 'ios')) {
-    return <Icon {...props} />;
-  }
+  const shouldAdjust = !(device.platform === 'ios') || !needLandscapeStyle;
```
donni106, line 35: is it still needed only for ios to make adjustments with our latest version?

ardasnturk: Yes, this check is still necessary. It is also clearly visible in the screenshots.

|with|without|
|--|--|
image|image

donni106: hmm...so android always need some adjustments (portrait and landscape) and ios just portrait?

a small request: `!(device.platform === 'ios')` should become `!device.platform !== 'ios'`

ardasnturk: Why do you want this? I tried to keep this control the same as before. What do you mean by `!device.platform !== 'ios'`?

src/components/OrientationAwareIcon.tsx
```diff
@@ -17,32 +18,43 @@ export const OrientationAwareIcon = (props: Props) => {
...
+
+  const getIconProps = () => {
+    const baseProps = {
+      ...props,
+      ...(isNamedIcon && { name: iconName })
+    };
+
+    if (shouldAdjust) {
+      return baseProps;
+    }
```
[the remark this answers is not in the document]
ardasnturk: In fact, we don't need to define an extra constant. It is already used 1 time and I suggest to update the code in this way.

```
...
    if (!(device.platform === 'ios') || !needLandscapeStyle) {
      return baseProps;
    }
...
```

src/components/OrientationAwareIcon.tsx
donni106: i mean `device.platform !== 'ios'`, sorry. i find it strange to have the `!` before a `===`

src/components/OrientationAwareIcon.tsx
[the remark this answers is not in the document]
donni106: that is a cool trick to not have confusing wordings!

src/components/OrientationAwareIcon.tsx
ardasnturk: Wouldn't it be more understandable if `device.platform === 'android'`? If it's not iOS, it's Android. 🤷‍♂️

src/components/OrientationAwareIcon.tsx
```diff
@@ -17,32 +18,41 @@ export const OrientationAwareIcon = (props: Props) => {
...
+    const baseProps = {
+      ...props,
+      ...(isNamedIcon && { name: iconName })
+    };
+
+    if (!(device.platform === 'ios') || !needLandscapeStyle) {
+      return baseProps;
+    }
+
+    const adjustedSize = size * LANDSCAPE_ADJUSTMENT_FACTOR;
```
donni106, line 45: this is also just used once, right? could be directly set for `size` without variable as well

src/components/OrientationAwareIcon.tsx
```diff
@@ -17,32 +18,41 @@ export const OrientationAwareIcon = (props: Props) => {
...
 
-  // we need adjustments only on iOS, so otherwise just return the item with the usual props
-  // TODO: is this still needed with Expo 48?
-  if (!needLandscapeStyle || !(device.platform === 'ios')) {
-    return <Icon {...props} />;
-  }
+  const getIconProps = () => {
+    const baseProps = {
+      ...props,
+      ...(isNamedIcon && { name: iconName })
```
donni106, line 38: if `isNamedIcon` is false, does `...` work or break?

src/components/OrientationAwareIcon.tsx
donni106: haha, yes! i mean, it could be web but we do not use it. so yes, letzt go for the check on android

src/components/OrientationAwareIcon.tsx
ardasnturk: It works without a problem.

src/components/OrientationAwareIcon.tsx
ardasnturk: Yes, that's right. But I still added it as a comment line so that later on we can see what that number means.

donni106: hä? the `LANDSCAPE_ADJUSTMENT_FACTOR` was great. i meant t oset `size: size * LANDSCAPE_ADJUSTMENT_FACTOR` directly

ardasnturk: But we use `adjustedSize` in 2 different places. So I thought we should keep it, so I return the code.

donni106: which second place do you mean? i have not seen it

ardasnturk: In line 49 and 50 we use `iconStyle` and `size` props.

```
    const adjustedSize = size * LANDSCAPE_ADJUSTMENT_FACTOR;

    return {
      ...baseProps,
      iconStyle: [iconStyle, styles.marginIcon, { width: adjustedSize }, landscapeIconStyle],
      size: adjustedSize,
      style: [style, landscapeStyle]
    };
```

donni106: ok thanks!

src/components/OrientationAwareIcon.tsx
ardasnturk: Hmm. I haven't tested it this way. Then I will update the code this way.

```
const baseProps = {
  ...props,
  ...(isNamedIcon ? { name: iconName } : {})
};
```

donni106: all right, yes, i think this is more secure, thanks

donni106 (approved):

PR #1227, opened 2025-01-13
fix(oparl): organisation dropdown select

donni106:
conditional rendering of `dropdownData?.length <= 2` did the trick

SVASD-65

### DIFF
--- a/src/screens/OParl/OParlOrganisationsScreen.tsx
+++ b/src/screens/OParl/OParlOrganisationsScreen.tsx
@@ -29,6 +29,8 @@ const useListData = (organizationData?: { [queryName]: OrganizationPreviewData[]
   return listData;
 };
 
+const keyExtractor = (item, index) => `index${index}-id${item.id}`;
+
 export const OParlOrganizationsScreen = ({ navigation }: Props) => {
   const [fetchingMore, setFetchingMore] = useState(false);
   const [finished, setFinished] = useState(false);
@@ -75,9 +77,8 @@ export const OParlOrganizationsScreen = ({ navigation }: Props) => {
     <SafeAreaViewFlex>
       <FlatList
         data={listData}
-        renderItem={({ item }) => (
-          <OParlPreviewComponent data={item} key={item.id} navigation={navigation} />
-        )}
+        keyExtractor={keyExtractor}
+        renderItem={({ item }) => <OParlPreviewComponent data={item} navigation={navigation} />}
         onEndReachedThreshold={1.5}
         onEndReached={onEndReached}
         // this does currently not work as intended, until we upgrade our apollo client dependency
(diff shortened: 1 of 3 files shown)


src/screens/OParl/OParlPersonsScreen.tsx
```diff
@@ -158,19 +171,20 @@ export const OParlPersonsScreen = ({ navigation }: Props) => {
...
-              searchPlaceholder="Suche"
-              showSearch
-              searchInputStyle={styles.searchInput}
-              setData={setDropdownData}
-            />
-          </Wrapper>
+          dropdownData?.length <= 2 ? null : (
+            <Wrapper style={styles.noPaddingTop}>
+              <DropdownSelect
+                data={dropdownData}
```
ardasnturk, line 180: Can we use the `&&` operator instead of returning `null`?

```
dropdownData?.length <= 2 && (...)
```

src/screens/OParl/OParlOrganisationsScreen.tsx
```diff
@@ -75,9 +75,8 @@ export const OParlOrganizationsScreen = ({ navigation }: Props) => {
     <SafeAreaViewFlex>
       <FlatList
         data={listData}
-        renderItem={({ item }) => (
-          <OParlPreviewComponent data={item} key={item.id} navigation={navigation} />
-        )}
+        keyExtractor={(item, index) => `index${index}-id${item.id}`}
```
ardasnturk, line 78: This is how we use `keyExtractor` in many parts of the code. How about creating a helper function?

For example:
```
const keyExtractor = (item, index) => `item${item?.id || item}-index${index}`;
```
```
<FlatList
  keyExtractor={keyExtractor}
  ...
/>
```

src/screens/OParl/OParlPersonsScreen.tsx
donni106: this could result in boolean `false` as value for `ListHeaderComponent`, which would be problematic i guess.

ardasnturk (approved):

src/screens/OParl/OParlOrganisationsScreen.tsx
ardasnturk: That's not what I meant by the latest update. Currently the repetition problem in the code is not solved. I mean creating a single `keyExtrator` helper function and using that function everywhere.

For Example: /src/helper/keyExtrator.ts
```
export const keyExtractor = (item, index) => `item${item?.id || item}-index${index}`;
```

I suggested to use all keyExtrators used in the code through this function.

donni106: that will not work for every list, as the items can be different from component to component.

ardasnturk (approved):

PR #1228, opened 2025-01-14
feat(eas): build and submit configuration

ardasnturk:
- updated `runtimeVersion.policy` value in `app.json` to `appVersion` so that eas can get the version in automation with eas
- added `autoIncrement` value to `eas.json` to automatically increment `buildNumber` and `versionCode`
- updated scripts in `package.json` for when you need to build locally or run submit

SVA-1350


eas.json
```diff
@@ -1,14 +1,19 @@
...
   },
   "build": {
     "production": {
       "node": "18.19.1",
       "yarn": "1.22.19",
       "channel": "production",
       "ios": {
-        "image": "latest"
+        "image": "latest",
+        "autoIncrement": true
```
donni106, line 13: does this auto increment reset to 1, when changing the main version for the app, as we always do manually?

eas.json
```diff
@@ -33,6 +38,14 @@
...
-    "production": {}
+    "production": {
+      "android": {
+        "track": "internal"
+      },
+      "ios": {
+        "ascAppId": "",
+        "appleTeamId": ""
+      }
+    }
```
donni106, line 49: can you also add some links to docs for this section in the commit message please

package.json
```diff
@@ -8,14 +8,16 @@
...
     "build:ios": "eas build --platform ios",
     "build:dev-android": "eas build --profile development --platform android",
     "build:dev-ios": "eas build --profile development --platform ios",
     "build:dev-ios-simulator": "eas build --profile development-simulator --platform ios",
     "build:local-ios-prod": "npx expo run:ios -d --configuration Release",
-    "submit:android": "eas submit --platform android",
-    "submit:ios": "eas submit --platform ios",
+    "submit:all": "eas submit --platform all --latest",
+    "submit:android": "eas submit --platform android --latest",
+    "submit:ios": "eas submit --platform ios --latest",
```
donni106, line 20: what if it would be not `latest`? what are other values or use cases?

eas.json
ardasnturk: Very good question. I couldn't find anything about it in the document. I imagine it could increase continuously, but I can't answer that without testing it.

app.json
```diff
@@ -8,7 +8,7 @@
     "githubUrl": "https://github.com/ikuseiGmbH/smart-village-app-app",
     "platforms": ["ios", "android"],
     "version": "4.0.3",
-    "runtimeVersion": { "policy": "nativeVersion" },
+    "runtimeVersion": { "policy": "appVersion" },
```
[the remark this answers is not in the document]
ardasnturk: Updated with Commit document link

eas.json
ardasnturk: Updated with Commit document link

package.json
ardasnturk: `latest` is a flag added to send the latest build created in eas to the stores. I removed this flag because I don't think it is needed in local. Because in local we have the possibility to choose the desired build.

eas.json
donni106: ok, we will see. and i think it would be no problem for us to change to a continuously increasing version code, right?

ardasnturk: That's what I think. We're always going to increase it for Android. For iOS, there won't be any problem.

ardasnturk: I did some thinking yesterday. To be honest, I don't think the `autoIncrement` option would be right for us, because we need an extra commit to trigger the pipeline. If we have `autoIncrement`, we can't send an extra commit. I think we should continue as we have been doing so far and remove the `autoIncrement` option from the code. This way we can control and increment `versionCode` and `buildNumber` ourselves. This way we can also send an extra commit to trigger the pipeline.

donni106: ok, we can have that in mind yes, when manually triggering per commit.

donni106 (approved):

PR #1230, opened 2025-01-15
chore(expo): update expo version to 52

ardasnturk:
This PR includes the Expo SDK 52 update

[SDK 52 Blog Post](https://expo.dev/changelog/2024/11-12-sdk-52)
[SDK 52 Changelog](https://github.com/expo/expo/blob/main/CHANGELOG.md#5200--2024-11-08)

SVA-1536


src/screens/EncounterScannerScreen.tsx
```diff
@@ -79,19 +79,12 @@ export const EncounterScannerScreen = ({ navigation }: { navigation: any }) => {
...
     }
-  }, [hasPermission, ref.current]);
-
-  useEffect(() => {
-    (async () => {
-      const { status } = await Camera.requestCameraPermissionsAsync();
-      setHasPermission(status === 'granted');
-    })();
-  }, []);
+  }, [permission, ref.current]);
```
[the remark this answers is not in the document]
ardasnturk: I don't know why, I just added `permission` here. `ref.current` was there before.

src/components/Chat.js
```diff
@@ -1,7 +1,6 @@
-import { Video } from 'expo-av';
 import * as FileSystem from 'expo-file-system';
-import { MediaTypeOptions } from 'expo-image-picker';
 import * as ScreenOrientation from 'expo-screen-orientation';
+import { useVideoPlayer, VideoView } from 'expo-video';
```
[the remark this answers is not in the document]
ardasnturk: commit error. I meant to say that the `expo-video` package is more performant than the `expo-av` package.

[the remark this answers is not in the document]
ardasnturk: I updated the commit.

src/screens/EncounterRegistrationScreen.tsx
```diff
@@ -64,7 +64,9 @@ export const EncounterRegistrationScreen = ({ navigation }: StackScreenProps<any
   // globally disable the button, when loading after pressing register
   const [registrationLoading, setRegistrationLoading] = useState(false);
 
-  const { imageUri, selectImage } = useSelectImage();
+  const { imageUri, selectImage } = useSelectImage({
+    allowsEditing: false
```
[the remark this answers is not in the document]
ardasnturk: no. I updated this.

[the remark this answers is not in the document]
donni106: we should not pass an empty object just to make the type check happy. it should be possible to use the hook without passing any params if not needed to use all the default values. this is our hook, can you have a check to update this please?

package.json
```diff
@@ -42,50 +42,49 @@
...
+    "expo-image-picker": "~16.0.4",
+    "expo-linear-gradient": "~14.0.2",
+    "expo-linking": "~7.0.4",
+    "expo-localization": "~16.0.1",
+    "expo-location": "~18.0.5",
+    "expo-media-library": "~17.0.5",
+    "expo-notifications": "~0.29.12",
+    "expo-screen-orientation": "~8.0.4",
+    "expo-secure-store": "~14.0.1",
+    "expo-splash-screen": "~0.29.20",
```
donni106, line 84: > Splash screens for Android cannot be fullscreen with this API (and this did not work particularly well before either), so you may need to update your splash screen. 

can this be something we need to check deeper? https://expo.dev/changelog/2024/11-12-sdk-52#big-improvements-to-expo-splash-screen

donni106: > The splash key will be deprecated in favor of the config plugin in a future release. For now, expo-splash-screen will attempt to map your splash config to the equivalent configuration for the config plugin automatically. For now, you won't need to make any changes to your existing splash configuration on iOS. However, on Android you will need to migrate away from using a full screen splash image if you are currently using one, and instead use a smaller image such as your app icon.

package.json
```diff
@@ -42,50 +42,49 @@
...
+    "expo-device": "~7.0.2",
+    "expo-document-picker": "~13.0.2",
+    "expo-font": "~13.0.3",
+    "expo-image-picker": "~16.0.4",
+    "expo-linear-gradient": "~14.0.2",
+    "expo-linking": "~7.0.4",
+    "expo-localization": "~16.0.1",
+    "expo-location": "~18.0.5",
+    "expo-media-library": "~17.0.5",
+    "expo-notifications": "~0.29.12",
```
donni106, line 81: important for us?

> expo-notifications trigger types changed: we simplified calendar trigger input types in response to feedback that the old approach was difficult to use and error prone.

src/screens/EncounterRegistrationScreen.tsx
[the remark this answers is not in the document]
ardasnturk: Okay. I solved this problem. 
I updated Hook as follows.

```
    setter: React.Dispatch<React.SetStateAction<T>>
  ) => React.Dispatch<React.SetStateAction<T>>;
  quality?: number;
  saveImage?: boolean;
} = {}) => {  <--
  const [imageUri, setImageUri] = useState<string>();
```

package.json
ardasnturk: Yes. That's going to be a problem for us. I've been trying to figure out how to solve it. If you tried the new development build you can see it. SplashScreen is too small. It looks like a logo. Maybe we can solve this problem by adding a standard width. But it needs to be tested.

package.json
ardasnturk: I think this is only for local notifications.

donni106: ok i found this PR...it is something with calendars we do not use https://github.com/expo/expo/pull/31598

package.json
donni106: ok, so this will be the last thing to figure put, right?

donni106: I do not understand to develop it in App.js. iOS legacy is not an option as it will be removed soon in the future.
Please make research how to do splash screens in the latest version and  make some proposals.

ardasnturk: It was updated in this way because we decided to add the app icon and primary color code as a background instead of a full screen splash screen.

[465e56d](https://github.com/smart-village-solutions/smart-village-app-app/pull/1230/commits/465e56d7db4d5a822c3151b10a85cc0632c249a1)

ardasnturk: |ios|android|
|--|--|
Simulator Screenshot - iPhone 16 Pro Max - 2025-01-17 at 14 41 31|Screenshot_20250117-144602

donni106: please use white for our dev app as background color. primary makes no sense here.

and for android the size of the logo needs to be corrected, so that it does not get cut.

ardasnturk: Yes, it is. That's better.

Screenshot_20250117-150926

donni106 (approved):

PR #1234, opened 2025-01-27
Fix/resolve critical errors shown in codeclimate

ardasnturk:
resolved some of the critical bugs shown in codeclimate


src/components/CardListItem.js
```diff
@@ -129,17 +129,34 @@ const renderCardContent = (item, index, horizontal, noOvertitle, bigTitle, sue)
...
+        if (address) {
+          cardContent.push(sequenceMap.sue.address());
+        }
+
+        if (status) {
+          cardContent.push(sequenceMap.sue.status());
+        }
+      };
+
+      addSueContent();
```
donni106, line 159: why do you add an function here that checks conditions instead of directly calling the conditions? the function is also called every time.

what about

```
if (sue) {
  if (!picture?.url) {
    cardContent.push(sequenceMap.sue.pictureFallback());
  }

  if (serviceName && requestedDatetime) {
    cardContent.push(sequenceMap.sue.category());
    cardContent.push(sequenceMap.sue.divider());
  }

  if (title) {
    cardContent.push(
      <WrapperHorizontal key={keyExtractor(title, index)}>
        {sequenceMap.title()}
      </WrapperHorizontal>
    );
  }

  if (address) {
    cardContent.push(sequenceMap.sue.address());
  }

  if (status) {
    cardContent.push(sequenceMap.sue.status());
  }
};
```

src/components/DataListSection.tsx
```diff
@@ -64,16 +64,23 @@ export const DataListSection = ({
...
+    <View>
+      {renderSectionHeader()}
+      <LoadingContainer>
+        <ActivityIndicator color={colors.refreshControl} />
+      </LoadingContainer>
+    </View>
+  );
+
+  if (loading) {
+    return renderLoadingState();
```
donni106, line 83: i can understand the refactoring of `renderSectionHeader` to avoid `&&` in rendering, but how is this better than before?

```
if (loading) {
  return (
    <View>
      {renderSectionHeader()}
      <LoadingContainer>
        <ActivityIndicator color={colors.refreshControl} />
      </LoadingContainer>
    </View>
  );
}
```

src/components/CardListItem.js
ardasnturk: I added a log inside the if(sue) control. There are 2 items in the list. In the old version of the code I can see this log 4 times. But in the new version of the code I can see the log 2 times. So it is called 1 time for each list item. I think this is acceptable. 

Also, looking at the previous version of the code, it is more difficult for me to understand.

src/components/DataListSection.tsx
ardasnturk: With `renderSectionHeader`, code repetition is avoided. As I noticed, the same code was used in 3 different sections, now only `renderSectionHeader` is used. Adding the if check instead of the `&&` operator solves the complexity of the code and I think it is more understandable.

src/queries/SUE/requests.tsx
```diff
@@ -32,16 +32,24 @@ export const postRequests = async (data: any) => {
...
+      firstName: 'first_name',
+      lastName: 'last_name',
+      lat: 'lat',
+      long: 'long',
+      phone: 'phone',
+      serviceCode: 'service_code',
+      title: 'title'
+    };
+
+    Object.entries(fieldMappings).forEach(([key, formKey]) => {
```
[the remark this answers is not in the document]
ardasnturk: Yes, this code works the same way. But I don't know if it solves code complexity according to codeclimate.

src/components/CardListItem.js
donni106: yes, the previous version is nothing we want. i was speaking about the additional method you want to introduce for some reason.

src/components/DataListSection.tsx
donni106: i think this does not answer my question. as is said, the refactoring of `renderSectionHeader` is clear to me.
but i see no reason for a method `renderLoadingState` that has benefits over my postet code before.

src/components/CardListItem.js
```diff
@@ -129,17 +129,35 @@ const renderCardContent = (item, index, horizontal, noOvertitle, bigTitle, sue)
...
-      title &&
-        cardContent.push(
-          <WrapperHorizontal key={keyExtractor(title, index)}>
-            {sequenceMap.title()}
-          </WrapperHorizontal>
-        );
-      address && cardContent.push(sequenceMap.sue.address());
-      status && cardContent.push(sequenceMap.sue.status());
+      const addSueContent = () => {
+        console.log('test');
```
donni106, line 133: whoop

src/components/DataListSection.tsx
ardasnturk: Yes, that's right. I removed `renderLoadingState` from the code because it doesn't have any extra benefit.

src/queries/SUE/requests.tsx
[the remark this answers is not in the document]
ardasnturk: Yes, that's right. So I updated the code with `&&` operator.

src/components/CardListItem.js
ardasnturk: I thought about it again. I decided that we don't need an extra method. I just changed the && operator control into an if control. Logically it works in the same way as the && operator. But I think it is more readable this way.

donni106 (approved):

PR #1236, opened 2025-01-27
feat(maestro): introduce maestro for e2e testing

donni106:
i recommend installing latest maestro on macOS with brew. delete earlier installations before if not installed per brew. you can find out with `which maestro` if it is installed and where.

to install per brew.

```
brew tap mobile-dev-inc/tap
brew install maestro
```

after that you should be able to run `yarn maestro`.

SVA-1086


package.json
```diff
@@ -17,6 +17,9 @@
     "submit:android": "eas submit --platform android",
     "submit:ios": "eas submit --platform ios",
     "lint": "./node_modules/.bin/eslint __tests__ src --ext .js,.jsx,.ts,.tsx",
+    "clear-state": "maestro test __maestro__/clear-state.yml -e APP_ID=de.dev.smartvillage.app",
+    "maestro": "maestro test __maestro__ -e APP_ID=de.dev.smartvillage.app",
+    "maestro:clear": "maestro test __maestro__ -e APP_ID=de.dev.smartvillage.app -e CLEAR_STATE=true",
```
donni106, line 22: for the future i plan to call these with passing the `APP_ID` per script as well to be able to be dynamic on different branches/apps

__maestro__/setup/onboarding.yml
```diff
@@ -0,0 +1,10 @@
+appId: ${APP_ID}
+---
+- tapOn: "WEITER"
+- tapOn: "WEITER"
+- tapOn: "WEITER"
+- tapOn: "WEITER"
+- tapOn: "WEITER"
+- tapOn: "WEITER"
+- tapOn: "WEITER"
+- tapOn: "Nein"
```
donni106, line 10: in the future this needs to be dynamic per script, calling the onboarding settings from the server and creating the number of taps on "WEITER" and possible interactions with alerts

package.json
```diff
@@ -17,6 +17,9 @@
     "submit:android": "eas submit --platform android",
     "submit:ios": "eas submit --platform ios",
     "lint": "./node_modules/.bin/eslint __tests__ src --ext .js,.jsx,.ts,.tsx",
+    "clear-state": "maestro test __maestro__/clear-state.yml -e APP_ID=de.dev.smartvillage.app",
+    "maestro": "maestro test __maestro__ -e APP_ID=de.dev.smartvillage.app",
```
ardasnturk, line 21: Flow fails when I run the `yarn maestro` command. Which script in the `__maestro__` folder does this command run by default?

image

`yarn maestro:clear` worked fine and I was able to open the bookmark screen.

__maestro__/setup/onboarding.yml
ardasnturk: Yes, that's right. But if you are going to use scripts for all apps, we will need much more customization options. Or do we only want to use it for the development version?

package.json
ardasnturk: Ok, I succeeded in running the `yarn maestro` command. In order for this command to work, the state must be cleared. So the onboarding screen must be passed.

ardasnturk (approved):

__maestro__/setup/onboarding.yml
ardasnturk: Yes, that's right. But if you are going to use scripts for all apps, we will need much more customization options. Or do we only want to use it for the development version?

donni106: we will see. but with the plans to get the settings from the server, it should work for others as well.

package.json
donni106: it depends. if your very first app state is cleared, the `yarn maestro` can not work, correct. so there is the `yarn maestro:clear` that also runs the onboarding. after the onboarding is passed once, `yarn maestro` can be used.
i will add a condition to run the onboarding also with `yarn maestro` if the first onboarding screen is visible.

__maestro__/config.yml
```diff
@@ -0,0 +1,2 @@
+flows:
+  - "flows/**"
```
donni106, line 2: > Which script in the __maestro__ folder does this command run by default?

It runs all scripts in `flows`, which is defined here.

package.json
donni106: there is a condition now to handle this:

```
# pass onboarding if it's visible
- runFlow:
    when:
      visible:
        text: "WEITER"
        below:
          text: "WILLKOMMEN"
    file: "onboarding.yml"
```

ardasnturk (approved):

PR #1237, opened 2025-01-27
chore: update packages without resolutions

donni106:
- removed resolutions and yarn.lock to obtain latest available versions of all packages
- called `rm -rf yarn.lock; rm -rf node_modules; yarn` for cleanup an reinstall
- called `npx expo-doctor@latest` for checkup
- ran `yarn start` and tested in iOS simulator -> app is running

`npx expo-doctor@latest` says:

image

but i had no problems running the app. i think it has benefits to see the invalid versions of `@expo/config-plugins` to keep us reminded about it, as long as we don't have any other problems.

please run and navigate around to check for some problems or bad behaviors, thanks.

### DIFF
--- a/package.json
+++ b/package.json
@@ -167,17 +167,5 @@
     "react-test-renderer": "^18.2.0",
     "set-value": "^4.1.0",
     "typescript": "~5.3.3"
-  },
-  "resolutions": {
-    "@expo/config-plugins": "^9.0.0",
-    "@babel/traverse": "~7.25.0",
-    "crypto-js": "~4.2.0",
-    "markdown-it": "~13.0.2",
-    "metro": "^0.81.0",
-    "metro-resolver": "^0.81.0",
-    "metro-config": "^0.81.0",
-    "micromatch": "~4.0.4",
-    "node-fetch": "~2.6.7",
-    "ws": "~8.17.1"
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -167,17 +167,5 @@
...
-  "resolutions": {
-    "@expo/config-plugins": "^9.0.0",
-    "@babel/traverse": "~7.25.0",
-    "crypto-js": "~4.2.0",
-    "markdown-it": "~13.0.2",
-    "metro": "^0.81.0",
-    "metro-resolver": "^0.81.0",
-    "metro-config": "^0.81.0",
-    "micromatch": "~4.0.4",
-    "node-fetch": "~2.6.7",
```
ardasnturk, line 180: I can't try right now. I have an internet problem I don't know why. When I do a speedtest, it shows 260+ mbit.

image

---

You didn't build a new developer, right? Or local build. You ran the app in a previously created build, right?

donni106: Yes

ardasnturk: I turned my computer off and on and the problem was solved. I was able to install the packages without any problem. I also created a build locally and it works. 👍

ardasnturk (approved):

PR #1239, opened 2025-01-29
fix(sue): reduce returns on sue report screen

ardasnturk:
- added `alertText` to reduce returns in `alertTextGeneratorForMissingData` function

SVA-1557

### DIFF
--- a/src/helpers/textHelper.js
+++ b/src/helpers/textHelper.js
@@ -1,4 +1,3 @@
-/* eslint-disable complexity */
 /**
  * Formatting a subtitle with a separating | if necessary
  *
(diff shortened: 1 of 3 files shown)


src/helpers/textHelper.js
```diff
@@ -9,19 +9,23 @@
...
+  } else if (time) {
+    formattedText = `${time} Uhr`;
+  } else if (last) {
+    formattedText = last;
+  }
 
-  return last;
+  return formattedText;
 };
 /* eslint-enable complexity */
```
donni106, line 30: ´/* eslint-disable complexity */` still needed?

src/screens/SettingsScreen.js
donni106, line 1: what is the reason that having a variable and `else if`s is better than `if`s with returns?

ardasnturk: Since we don't currently have return in every if, if we use if instead of else if, it will check all if checks one by one. But with the else if control it will go inside the matching control and skip all other unnecessary if checks.

src/helpers/textHelper.js
ardasnturk: No, we don't need it anymore.

src/screens/SettingsScreen.js
donni106: so another option would be to add `return;`s in the cases where we have no returns currently?

ardasnturk: I don't quite understand. Where do you want to add returns? We are already trying to reduce returns.

donni106 (approved):

src/screens/SettingsScreen.js
donni106: in `if`s where missing. but it's just an idea. we can go with the refactoring.

PR #1247, opened 2025-02-05
fix(map): the pin problem on the map for android

ardasnturk:
- patch added to fix the problem of half of the pins on the map being visible on android platform
  - https://github.com/react-native-maps/react-native-maps/issues/5247#issuecomment-2521958144

SVA-1569

|before|after|
|--|--|
image|image

### DIFF
--- a/patches/react-native-maps+1.18.0.patch
+++ b/patches/react-native-maps+1.18.0.patch
@@ -1,0 +1,16 @@
+diff --git a/node_modules/react-native-maps/android/src/main/java/com/rnmaps/maps/MapMarker.java b/node_modules/react-native-maps/android/src/main/java/com/rnmaps/maps/MapMarker.java
+index b755f9b..dd560f2 100644
+--- a/node_modules/react-native-maps/android/src/main/java/com/rnmaps/maps/MapMarker.java
++++ b/node_modules/react-native-maps/android/src/main/java/com/rnmaps/maps/MapMarker.java
+@@ -622,4 +622,11 @@ public class MapMarker extends MapFeature {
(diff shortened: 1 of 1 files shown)


app.json
```diff
@@ -47,6 +47,7 @@
       }
     },
     "android": {
+      "newArchEnabled": false,
```
donni106, line 50: oha ;) https://x.com/tomekzaw_/status/1886741028960424101
do you think this is a proper solution? this means everything in android is on old architecture, not only the map.
did you do some more research and can point to some links?

donni106: i found this https://github.com/react-native-maps/react-native-maps/issues/5247, which seems identical. there is a comment pointing t some PR with a fix in an android java file

ardasnturk: Yes. That wouldn't be my first choice either. But I don't see any other solution at this point.
I know this is not a fix. 🤷‍♂️
https://github.com/tomekvenits/react-native-map-clustering/issues/279

ardasnturk: Yes. It might be a clue for me. I'll look into it closely. Thank you.

patches/react-native-maps+1.18.0.patch
donni106, line 1: how did you create this file?

donni106: what do you mean with `require file`? i really do not know how patches work. with running `yarn patch-package react-native-maps` the file will not be created but applied, right? so where does the patch file come from?

donni106: perfect, now i know for the future!

donni106 (approved):

PR #1248, opened 2025-02-05
fix(sue): input's header design problem

ardasnturk:
- updated the title of the `Input` in the feedback section with `RegularText` to make it look the same as rating in design

SUE-78

|before|after|
|--|--|
Simulator Screenshot - iPhone 16 Pro Max - 2025-02-05 at 15 08 45|Simulator Screenshot - iPhone 16 Pro Max - 2025-02-05 at 15 08 27

### DIFF
--- a/src/components/SUE/report/SueReportSend.tsx
+++ b/src/components/SUE/report/SueReportSend.tsx
@@ -131,7 +131,7 @@ export const SueReportSend = ({
                   </View>
 
                   <View style={styles.ratingContainer}>
-                    <RegularText small style={styles.ratingContainer}>
+                    <RegularText small style={styles.labelText}>
                       {texts.sue.report.sendReportDone.ratingTitle}
                     </RegularText>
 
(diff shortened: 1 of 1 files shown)


src/components/SUE/report/SueReportSend.tsx
```diff
@@ -153,11 +153,14 @@ export const SueReportSend = ({
                   </View>
 
                   <View style={styles.headerContainer}>
+                    <RegularText small style={styles.ratingContainer}>
```
donni106, line 156: `ratingContainer`? can you find a name, that fits both? something like `labelContainer` for example?

ardasnturk: I named it `labelText`. because this is actually the style of the text component.

donni106: true! well done!

donni106 (approved):

PR #1249, opened 2025-02-06
fix(waste): the unmanageability of notifications

ardasnturk:
This PR adds the ability to turn on notification settings if the device's notification settings are turned off by notifying the user via alert

GOS-118
SVAK-58


src/components/waste/WasteCollectionSettings.tsx
```diff
@@ -348,6 +354,16 @@ export const WasteCollectionSettings = ({
...
 
+  const getPermission = async () => {
+    const inAppPermission = await getInAppPermission();
+
+    setIsPushPermissionGranted(inAppPermission);
+  };
+
+  useEffect(() => {
+    getPermission();
+  }, [isPushPermissionGranted, showNotificationSettings]);
```
donni106, line 365: `getPermission` is only called in this effect, so it could also be defined only in here, right?
we do this in other places as well to be able to call async in effect

donni106 (approved):

PR #1252, opened 2025-02-11
fix(sue): the problem of not showing data in list view

ardasnturk:
This PR allows the data to be displayed again in the list view in the SUE module and allows the data count display to be updated according to the filters made.

|without filter|with filter|
|--|--|
Simulator Screenshot - iPhone 16 Pro Max - 2025-02-11 at 10 48 38|Simulator Screenshot - iPhone 16 Pro Max - 2025-02-11 at 10 48 50

SUE-144

### DIFF
--- a/src/components/filter/Filter.tsx
+++ b/src/components/filter/Filter.tsx
@@ -8,6 +8,7 @@ import { Divider, Header } from 'react-native-elements';
 
 import { Icon, colors, consts, normalize, texts } from '../../config';
 import { momentFormat } from '../../helpers';
+import { initialStartDate } from '../../screens';
 import { FilterProps, FilterTypesProps } from '../../types';
 
 import { Button } from './../Button';
@@ -20,7 +21,6 @@ const { a11yLabel } = consts;
 type Props = {
   filterTypes?: FilterTypesProps[];
   initialQueryVariables?: FilterProps;
-  initialStartDate?: { start_date: string };
   isOverlay?: boolean;
   queryVariables: FilterProps;
   setQueryVariables: React.Dispatch<FilterProps>;
@@ -44,7 +44,6 @@ const deleteInitialStartDateFromQueryVariables = (
 export const Filter = ({
   filterTypes,
   initialQueryVariables,
-  initialStartDate,
   isOverlay = false,
   queryVariables,
   setQueryVariables,
@@ -52,7 +51,7 @@ export const Filter = ({
 }: Props) => {
   const updatedQueryVariables = deleteInitialStartDateFromQueryVariables(
     queryVariables,
-    initialStartDate?.start_date
+    initialStartDate.start_date
   );
   const [filters, setFilters] = useState<FilterProps>(updatedQueryVariables);
   const [isCollapsed, setIsCollapsed] = useState(true);
@@ -63,7 +62,7 @@ export const Filter = ({
       setQueryVariables((prev) => {
         const newFilters = { ...filters };
 
-        if (newFilters.start_date === initialStartDate) {
+        if (newFilters.start_date === initialStartDate.start_date) {
           delete newFilters.start_date;
 
           return {
@@ -77,7 +76,7 @@ export const Filter = ({
           return {
             ...prev,
             search: prev.search || '',
-            start_date: initialStartDate,
+            ...initialStartDate,
             ...newFilters
           };
         }
@@ -96,7 +95,7 @@ export const Filter = ({
 
         setQueryVariables({
           ...queryVariables,
-          start_date: initialStartDate
+          ...initialStartDate
         });
       }, 500);
     } else {
--- a/src/components/filter/Filter.tsx
+++ b/src/components/filter/Filter.tsx
@@ -27,6 +27,16 @@ type Props = {
   withSearch?: boolean;
 };
 
+const cleanObject = (obj: FilterProps, initialStartDate?: string): FilterProps => {
+  const newObj = { ...obj };
+
+  if (initialStartDate && newObj.start_date === initialStartDate) {
+    delete newObj.start_date;
+  }
+
+  return newObj;
+};
+
 export const Filter = ({
   filterTypes,
   initialFilters,
@@ -36,15 +46,8 @@ export const Filter = ({
   setQueryVariables,
   withSearch = false
 }: Props) => {
-  const [filters, setFilters] = useState<FilterProps>(
-    initialStartDate
-      ? Object.fromEntries(
-          Object.entries(queryVariables).filter(
-            ([key, value]) => !(key === 'start_date' && value === initialStartDate)
-          )
-        )
-      : queryVariables
-  );
+  const initialQueryVariables = cleanObject(queryVariables, initialStartDate);
+  const [filters, setFilters] = useState<FilterProps>(initialQueryVariables);
   const [isCollapsed, setIsCollapsed] = useState(true);
   const [filterCount, setFilterCount] = useState(0);
 
@@ -55,10 +58,10 @@ export const Filter = ({
 
         if (newFilters.start_date === initialStartDate) {
           delete newFilters.start_date;
+
           return {
             ...prev,
             search: prev.search || '',
-            start_date: initialStartDate,
             ...newFilters
           };
         }
@@ -82,15 +85,7 @@ export const Filter = ({
       setIsCollapsed(!isCollapsed);
 
       setTimeout(() => {
-        setFilters(
-          initialStartDate
-            ? Object.fromEntries(
-                Object.entries(queryVariables).filter(
-                  ([key, value]) => !(key === 'start_date' && value === initialStartDate)
-                )
-              )
-            : queryVariables
-        );
+        setFilters(initialQueryVariables);
 
         setQueryVariables({
           ...queryVariables,
--- a/src/components/filter/Filter.tsx
+++ b/src/components/filter/Filter.tsx
@@ -110,7 +110,7 @@ export const Filter = ({
     if (isOverlay) {
       const activeFilters = _omit(filters, Object.keys(initialQueryVariables || {}));
       const filteredActiveFilters = Object.keys(activeFilters).reduce((acc, key) => {
-        if (key !== 'saveable' || activeFilters[key] !== false) {
+        if ((key !== 'saveable' || key !== 'onlyCurrentlyOpen') && activeFilters[key] !== false) {
           acc[key] = activeFilters[key];
         }
 
--- a/src/components/filter/Filter.tsx
+++ b/src/components/filter/Filter.tsx
@@ -19,39 +19,39 @@ const { a11yLabel } = consts;
 
 type Props = {
   filterTypes?: FilterTypesProps[];
-  initialFilters?: FilterProps;
+  initialQueryVariables?: FilterProps;
   initialStartDate?: { start_date: string };
   isOverlay?: boolean;
-  initialQueryVariables: FilterProps;
+  queryVariables: FilterProps;
   setQueryVariables: React.Dispatch<FilterProps>;
   withSearch?: boolean;
 };
 
 const deleteInitialStartDateFromQueryVariables = (
-  initialQueryVariables: FilterProps,
+  queryVariables: FilterProps,
   initialStartDate?: string
 ): FilterProps => {
-  if (initialQueryVariables?.start_date === initialStartDate) {
-    const newQueryVariables = { ...initialQueryVariables };
+  if (queryVariables?.start_date === initialStartDate) {
+    const newQueryVariables = { ...queryVariables };
     delete newQueryVariables.start_date;
 
     return newQueryVariables;
   }
 
-  return initialQueryVariables;
+  return queryVariables;
 };
 
 export const Filter = ({
   filterTypes,
-  initialFilters,
+  initialQueryVariables,
   initialStartDate,
   isOverlay = false,
-  initialQueryVariables,
+  queryVariables,
   setQueryVariables,
   withSearch = false
 }: Props) => {
   const updatedQueryVariables = deleteInitialStartDateFromQueryVariables(
-    initialQueryVariables,
+    queryVariables,
     initialStartDate?.start_date
   );
   const [filters, setFilters] = useState<FilterProps>(updatedQueryVariables);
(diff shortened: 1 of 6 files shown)


src/screens/SUE/SueListScreen.tsx
```diff
@@ -174,6 +166,15 @@ export const SueListScreen = ({ navigation, route }: Props) => {
     }, 50);
   }, []);
 
+  useEffect(() => {
+    const { limit, offset, ...rest } = queryVariables;
+
+    setDataCountQueryVariables({
+      ...rest,
+      start_date: queryVariables.start_date || queryVariables.initial_start_date
```
donni106, line 174: hmmm again...i don't know. now there are three places with setting the start date with fallback. did you ask chatpt for refactoring or optimizations?

src/screens/SUE/SueListScreen.tsx
```diff
@@ -125,12 +115,14 @@ export const SueListScreen = ({ navigation, route }: Props) => {
       query,
       {
         ...queryVariables,
+        start_date: queryVariables.start_date || queryVariables.initial_start_date,
```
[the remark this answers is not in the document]
ardasnturk: That's right, you're right. It can be used that way too. 👍

src/screens/SUE/SueListScreen.tsx
ardasnturk: No, I didn't ask ChatGPT anything. But I found a method. Can you please check again?

src/screens/SUE/SueListScreen.tsx
```diff
@@ -106,12 +96,11 @@ export const SueListScreen = ({ navigation, route }: Props) => {
...
 
   const limit = 20;
-  const initial_start_date = '1900-01-01T00:00:00+01:00';
-  const dataCountQueryVariables = {
-    start_date: initial_start_date
-  };
+  const initial_start_date = { start_date: '1900-01-01T00:00:00+01:00' };
+  const [dataCountQueryVariables, setDataCountQueryVariables] = useState({
+    initial_start_date
+  });
```
donni106, line 102: there will be an key `initial_start_date` this way, but i guess you want `start_date` in here.
so it needs to be `useState(initial_start_date)`

src/screens/SUE/SueListScreen.tsx
```diff
@@ -125,12 +114,14 @@ export const SueListScreen = ({ navigation, route }: Props) => {
       query,
       {
         ...queryVariables,
+        ...dataCountQueryVariables,
```
donni106, line 117: why `dataCountQueryVariables` after `queryVariables`?

src/screens/SUE/SueListScreen.tsx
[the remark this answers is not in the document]
donni106: why did you put `dataCountQueryVariables` after `queryVariables`? if query variables would be set with a start date, this will be overwritten with the initial?

src/screens/SUE/SueListScreen.tsx
ardasnturk: yes you are right. I will update.

src/screens/SUE/SueListScreen.tsx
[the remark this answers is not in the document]
donni106: > `queryVariables` can be added before `dataCountQueryVariables`

huh? it is already before.
please check whether we do not have unnecessary querying and re-renderings with setting `dataCountQueryVariables` in use effect. all the data should be placed once right, not multiple times for multiple queries and re-renderings.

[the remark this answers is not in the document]
ardasnturk: I meant later. I'm sorry I misspelled it. 
I checked the query and we only send 1 request in each filtering.

src/screens/SUE/SueListScreen.tsx
```diff
@@ -106,12 +96,9 @@ export const SueListScreen = ({ navigation, route }: Props) => {
   const query = route.params?.query ?? '';
 
   const limit = 20;
-  const initial_start_date = '1900-01-01T00:00:00+01:00';
-  const dataCountQueryVariables = {
-    start_date: initial_start_date
-  };
+  const initial_start_date = { start_date: '1900-01-01T00:00:00+01:00' };
```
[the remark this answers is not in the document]
donni106: @ardasnturk please have a look again, you forgot to camel case the const, right?

src/screens/SUE/SueListScreen.tsx
```diff
@@ -105,17 +98,17 @@ export const SueListScreen = ({ navigation, route }: Props) => {
...
     limit,
-    offset: 0
+    offset: 0,
+    ...initial_start_date
   };
   const [queryVariables, setQueryVariables] = useState(initialQueryVariables);
+  const [dataCountQueryVariables, setDataCountQueryVariables] = useState(() => {
+    const { limit, offset, ...rest } = queryVariables;
+
+    return { ...rest, ...initial_start_date };
```
donni106, line 110: if initial start date is already part of the initial query variables (line 104), which is the initial value for the query variables (line 106), than the start date is already in `rest`, or am i wrong?

donni106: what about 

```
useState({
  ...queryVariables,
  limit: undefined,
  offset: undefined
})
```

src/screens/SUE/SueListScreen.tsx
```diff
@@ -320,6 +319,7 @@ export const SueListScreen = ({ navigation, route }: Props) => {
                       placeholder: texts.sue.filter.alleSortingTypes
                     }
                   ]}
+                  initialStartDate={initial_start_date.start_date}
```
donni106, line 322: why do we pass that as prop. it is always the same value and just for the case to identify the initial date right? we can have it fix in the other file or export the variable from the top here. this is nothing we need to make the component aware of per props.

src/components/filter/Filter.tsx
```diff
@@ -29,21 +30,80 @@ type Props = {
...
+          };
+        }
+
+        return { ...prev, search: prev.search || '', ...newFilters };
+      });
     }
   }, [filters]);
 
+  const resetFilters = () => {
+    if (!isOverlay) {
```
donni106, line 81: how come the difference between overlay and not overlay?
why do we have separate logics here?

src/components/filter/Filter.tsx
```diff
@@ -29,21 +30,80 @@ type Props = {
...
+            ? Object.fromEntries(
+                Object.entries(queryVariables).filter(
+                  ([key, value]) => !(key === 'start_date' && value === initialStartDate)
+                )
+              )
+            : queryVariables
+        );
+
+        setQueryVariables({
+          ...queryVariables,
```
donni106, line 96: if this is a rest, shouldn't it also reset the query variables and not use the latest ones?

src/components/filter/Filter.tsx
```diff
@@ -29,21 +30,80 @@ type Props = {
...
+
+      setTimeout(() => {
+        setFilters(
+          initialStartDate
+            ? Object.fromEntries(
+                Object.entries(queryVariables).filter(
+                  ([key, value]) => !(key === 'start_date' && value === initialStartDate)
+                )
+              )
+            : queryVariables
```
donni106, line 92: duplicate content here, as we have this already above.
could this be simplified? why do we need to do object from entries and object entries.
what is the goal here? to filter out a start date if it is the initial one?

src/screens/SUE/SueListScreen.tsx
ardasnturk: We use `initialQueryVariables` in two parts. We need `initialQueryVariables` to reset the filter. This way, we can reset the filter.

src/screens/SUE/SueListScreen.tsx
ardasnturk: Yes, that’s correct. This is just a fixed date. But we use the Filter component in many places, and we only need to consider `start_date` in this specific case. If we add `start_date` as a fixed value to the entire Filter component, I think it might break other instances of the Filter component.

src/components/filter/Filter.tsx
ardasnturk: Because the Filter component is used in two different ways. One is the Overlay-Filter and the other is the SUE-Filter. We have the same logic in the return section.

src/screens/SUE/SueListScreen.tsx
donni106: did you comment right? this conversation is not about resetting

src/screens/SUE/SueListScreen.tsx
donni106: moving `export const initialStartDate = { start_date: '1900-01-01T00:00:00+01:00' };` in the other file should not brake anything

src/components/filter/Filter.tsx
ardasnturk: I don't really understand the question.

donni106: you are setting query variables with the current values of `queryVariables`. this can be edited already, so this has not initial values anymore and seems not to be a reset?

src/components/filter/Filter.tsx
ardasnturk: Yes, that's right, there is a repetition here. I will create a helper function for this.

If the `start_date` value is `initialStartDate`, then remove it from the filter object. This way we prevent the `initialStartDate` date from appearing in the `start_date` input.

donni106: ok. i understand but this is not what you do. you are not removing something from the filter object. you are iteration multiple times to filter everything else except the initial start date. can you find a way that matches your sentence and translates it into code?

donni106: still `Object.fromEntries(` and `Object.entries...filer...`
and everytime `queryVariables` changes that line will be run again? you only want it for setting the filters initial state, right?

src/components/filter/Filter.tsx
ardasnturk: I think there is a naming error here. This is actually an `initialQueryVariable`. So it's right to reset.

```
queryVariables={initialQueryVariables}
```

donni106: where is this line from?

ardasnturk: SueListScreen -> Filter component (Line: 323)

src/components/filter/Filter.tsx
ardasnturk: That's right. We just want the filter to be that way to start with. So what do you suggest?

src/screens/SUE/SueListScreen.tsx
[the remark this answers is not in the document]
ardasnturk: [171b753](https://github.com/smart-village-solutions/smart-village-app-app/pull/1252/commits/171b75321c0c919670bb73f35fc32a6465171a21)

src/screens/SUE/SueListScreen.tsx
ardasnturk: [171b753](https://github.com/smart-village-solutions/smart-village-app-app/pull/1252/commits/171b75321c0c919670bb73f35fc32a6465171a21)

src/components/filter/Filter.tsx
ardasnturk: I updated the code in this way. Yes `Object...` is still in the code. I couldn't think of any other method.

[171b753](https://github.com/smart-village-solutions/smart-village-app-app/pull/1252/commits/171b75321c0c919670bb73f35fc32a6465171a21)

src/components/filter/Filter.tsx
donni106: shouldn't it be `initialFilters` in that case?

src/screens/SUE/SueListScreen.tsx
donni106: in which cases you do not want an initial start date?

src/components/filter/Filter.tsx
ardasnturk: You are absolutely right. But I think there's a reason why we don't have this nomenclature.

src/components/filter/Filter.tsx
ardasnturk: I think it's better now.

[5bebc99](https://github.com/smart-village-solutions/smart-village-app-app/pull/1252/commits/5bebc9965c1c5c3c4ba0f98dc243729ca5439848)

src/screens/SUE/SueListScreen.tsx
ardasnturk: Only for SUE Module

src/screens/SUE/SueListScreen.tsx
```diff
@@ -320,7 +319,8 @@ export const SueListScreen = ({ navigation, route }: Props) => {
                       placeholder: texts.sue.filter.alleSortingTypes
                     }
                   ]}
-                  queryVariables={initialQueryVariables}
+                  initialQueryVariables={initialQueryVariables}
+                  initialStartDate={initialStartDate.start_date}
```
donni106, line 323: here `initialStatDate` is an object but in the other component it is a string? i suggest using the same object when naming it the same

src/components/filter/Filter.tsx
```diff
@@ -20,33 +20,92 @@ const { a11yLabel } = consts;
 type Props = {
   filterTypes?: FilterTypesProps[];
   initialFilters?: FilterProps;
+  initialStartDate?: string;
   isOverlay?: boolean;
-  queryVariables: FilterProps;
+  initialQueryVariables: FilterProps;
```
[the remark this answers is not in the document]
ardasnturk: Ach yes. That's exactly what I was talking about. We use different nomenclature for the SUE but we use different nomenclature for the Overlay filter. And it doesn't fit together. 

https://github.com/smart-village-solutions/smart-village-app-app/pull/1252#discussion_r1951165630

src/screens/SUE/SueListScreen.tsx
```diff
@@ -105,17 +98,17 @@ export const SueListScreen = ({ navigation, route }: Props) => {
...
   const initialQueryVariables = route.params?.queryVariables ?? {
-    initial_start_date,
     limit,
-    offset: 0
+    offset: 0,
+    ...initialStartDate
   };
   const [queryVariables, setQueryVariables] = useState(initialQueryVariables);
+  const [dataCountQueryVariables, setDataCountQueryVariables] = useState({
+    ...queryVariables,
```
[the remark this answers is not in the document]
ardasnturk: Yes, that's absolutely right.

src/components/filter/Filter.tsx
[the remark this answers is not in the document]
ardasnturk: [5792dc1](https://github.com/smart-village-solutions/smart-village-app-app/pull/1252/commits/5792dc187f753735615ff3b0f3d4abd5d3df49a4)

src/screens/SUE/SueListScreen.tsx
ardasnturk: I tested it again. using export/import instead of prop does not cause any problem. 

[8b97b8c](https://github.com/smart-village-solutions/smart-village-app-app/pull/1252/commits/8b97b8c8c9eef7c0d4bcb33854c0c682b5fbe646)

src/components/filter/Filter.tsx
donni106: shouldn't it be the `initialQueryVariables` here to reset to, instead of the `queryVariables`

ardasnturk: That's the only confusion left. No problem for the overlay filter. But for the SUE filter the `queryVariables` are actually `initialQueryVariables`.

donni106 (approved):

PR #1261, opened 2025-02-18
feat: update design

ardasnturk:
Updated the header with white background and the color of the items inside the header to black as a design standard.
Also the icons on the tabBar have been updated
- Service
- About
- Calendar

SVA-1572

|before|after|
|--|--|
Simulator Screenshot - iPhone 16 Pro Max - 2025-02-18 at 16 45 21|Simulator Screenshot - iPhone 16 Plus - 2025-06-24 at 15 03 55

### DIFF
--- a/src/components/DiagonalGradient.js
+++ b/src/components/DiagonalGradient.js
@@ -1,6 +1,7 @@
 import { LinearGradient } from 'expo-linear-gradient';
 import PropTypes from 'prop-types';
 import React from 'react';
+import { View } from 'react-native';
 
 import { colors } from '../config';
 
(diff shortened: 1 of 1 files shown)


src/navigation/screenOptions.tsx
```diff
@@ -45,7 +45,7 @@ export const getScreenOptions =
     return {
       // header gradient:
       // https://stackoverflow.com/questions/44924323/react-navigation-gradient-color-for-header
-      headerBackground: () => <DiagonalGradient />,
+      headerBackground: () => <DiagonalGradient colors={[colors.surface, colors.surface]} />,
```
donni106, line 48: please remove the diagonal gradient entirely here, as we habe one color to show now

src/components/SectionHeader.tsx
```diff
@@ -32,7 +32,7 @@ export const SectionHeader = ({
   if (!title) return null;
 
   const innerComponent = (
-    <WrapperRow>
+    <WrapperRow spaceBetween>
```
donni106, line 35: how come this `spaceBetween` necessary now?

src/components/DiagonalGradient.js
```diff
@@ -6,7 +6,7 @@ import { colors } from '../config';
 
 export const DiagonalGradient = ({
   children,
-  colors: gradientColors = [colors.primary, colors.secondary],
+  colors: gradientColors = [colors.primary, colors.primary],
```
donni106, line 9: shouldn't this stay `secondary` to keep the buttons like in the screenshot image

ardasnturk: The screenshots were taken before the code was updated. I've updated it now. We use gradient buttons less than before, but that doesn't mean we won't use them at all.

src/navigation/screenOptions.tsx
donni106: ok, right. can happen. what if we make a condition on the array? if the two color values are different, we use gradient, otherwise just a simple background color? i would like to avoid rendering a gradient component if we do not want to have gradient.

src/components/DiagonalGradient.js
donni106: so everywhere we use the `DiagonalGradient` component without passing colors, we will have the primary color without gradient, right? this is what we want?

donni106:
@codex review

PR #1265, opened 2025-02-19
fix: the divider problem in lists

ardasnturk:
- wrapped `Divider` in `WasteCollectionListItem` with `WrapperHorizontal` for consistency
- introduced `bottomDivider` option in `useRenderItem` for `publicJsonFile` query

SVA-1364

|before|after|before|after|
|--|--|--|--|
|Simulator Screenshot - iPhone 16 Pro Max - 2025-02-19 at 12 29 21|Simulator Screenshot - iPhone 16 Pro Max - 2025-02-19 at 12 28 54|Simulator Screenshot - iPhone 16 Pro Max - 2025-02-19 at 12 29 25|Simulator Screenshot - iPhone 16 Pro Max - 2025-02-19 at 12 28 57|

### DIFF
--- a/src/hooks/listHooks.js
+++ b/src/hooks/listHooks.js
@@ -99,6 +99,13 @@ export const useRenderItem = (query, navigation, options = {}) => {
   const listType = options.listType || getListType(query, listTypesSettings);
   let renderItem;
 
+  const calculateBottomDivider = (item, index, section) => {
+    return (
+      item.bottomDivider ??
+      (_isArray(section?.data) ? section.data.length - 1 !== index : undefined)
+    );
+  };
+
   switch (listType) {
     case LIST_TYPES.CARD_LIST: {
       renderItem = ({ item, index }) => {
@@ -159,9 +166,7 @@ export const useRenderItem = (query, navigation, options = {}) => {
             <TextListItem
               item={{
                 ...item,
-                bottomDivider:
-                  item.bottomDivider ??
-                  (_isArray(section?.data) ? section.data.length - 1 !== index : undefined)
+                bottomDivider: calculateBottomDivider(item, index, section)
               }}
               navigation={navigation}
               noSubtitle={options.noSubtitle}
@@ -200,9 +205,7 @@ export const useRenderItem = (query, navigation, options = {}) => {
           <TextListItem
             item={{
               ...item,
-              bottomDivider:
-                item.bottomDivider ??
-                (_isArray(section?.data) ? section.data.length - 1 !== index : undefined)
+              bottomDivider: calculateBottomDivider(item, index, section)
             }}
             {...{
               navigation,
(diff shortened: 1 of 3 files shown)


src/hooks/listHooks.js
```diff
@@ -288,6 +288,20 @@ export const useRenderItem = (query, navigation, options = {}) => {
...
+              item={{
+                ...item,
+                bottomDivider: true
+              }}
+              navigation={navigation}
+              noSubtitle={options.noSubtitle}
+              noOvertitle={options.noOvertitle}
+            />
+          );
+        }
```
donni106, line 303: i guess this should be handled in line 309 and not with an additional condition just for the divider

donni106 (approved):

src/hooks/listHooks.js
```diff
@@ -288,11 +288,18 @@ export const useRenderItem = (query, navigation, options = {}) => {
           return item.component;
         }
 
+        const bottomDivider =
+          query === QUERY_TYPES.PUBLIC_JSON_FILE
```
donni106, line 292: there shouldn't be an exception for public json files. we want all lists behave the same, right?!

src/hooks/listHooks.js
```diff
@@ -218,6 +218,13 @@ export const useRenderItem = (query, navigation, options = {}) => {
     default: {
       /* eslint-disable complexity */
       renderItem = ({ item, index, section, target }) => {
+        const bottomDivider =
+          query === QUERY_TYPES.PUBLIC_JSON_FILE
```
donni106, line 222: there shouldn't be an exception for public json files. we want all lists behave the same, right?!

ardasnturk: I don't think there will be a problem. When I look at other textListItems I can see the Divider at the end of the list. For example:

image

donni106: it is not a problem of working or not working. the code is bad with this specific condition on some query, if we want a divider in all cases

ardasnturk: I updated the code, now instead of the QUERY control, there is a `bottomDivider` control. And I added `calculateBottomDivider` function to prevent code repetition. This way I think I prevent repetitions in the code. Now I think we have a more understandable code.

src/hooks/listHooks.js
```diff
@@ -218,6 +218,8 @@ export const useRenderItem = (query, navigation, options = {}) => {
     default: {
       /* eslint-disable complexity */
       renderItem = ({ item, index, section, target }) => {
+        const bottomDivider = item.bottomDivider ? true : calculateBottomDivider(index, section);
```
donni106, line 221: we can do `const bottomDivider = item.bottomDivider ?? calculateBottomDivider(index, section)` like in the other places, right?

ardasnturk: Yes. To do this, I updated the `skipLastDivider` value in `DataListSection` to `false`.

src/hooks/listHooks.js
```diff
@@ -99,6 +99,10 @@ export const useRenderItem = (query, navigation, options = {}) => {
   const listType = options.listType || getListType(query, listTypesSettings);
   let renderItem;
 
+  const calculateBottomDivider = (index, section) => {
+    return _isArray(section?.data) ? section.data.length - 1 !== index : undefined;
+  };
```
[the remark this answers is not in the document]
ardasnturk: You are right. I added the `item` to the function. Now we do this check in `calculateBottomDivider`.

donni106 (approved):

PR #1267, opened 2025-02-20
fix: the problem of not showing daily events

ardasnturk:
With this PR, the problem that the daily event screen redirected via widget still shows all events has been fix and `dateRange` dates are shown in the filter component

SVA-1384

|before|after|before|after|before|after|
|--|--|--|--|--|--|
Simulator Screenshot - iPhone 16 Pro Max - 2025-02-20 at 10 57 04|Simulator Screenshot - iPhone 16 Pro Max - 2025-02-20 at 10 56 43|Simulator Screenshot - iPhone 16 Pro Max - 2025-02-20 at 10 57 12|Simulator Screenshot - iPhone 16 Pro Max - 2025-02-20 at 10 56 45|Simulator Screenshot - iPhone 16 Pro Max - 2025-02-20 at 10 57 18|Simulator Screenshot - iPhone 16 Pro Max - 2025-02-20 at 10 56 48

### DIFF
--- a/src/components/widgets/EventWidget.tsx
+++ b/src/components/widgets/EventWidget.tsx
@@ -53,7 +53,7 @@ export const EventWidget = ({ text, additionalProps }: WidgetProps) => {
       title: text ?? texts.homeTitles.events,
       query: QUERY_TYPES.EVENT_RECORDS,
       queryVariables: {
-        order: 'listDate_ASC',
+        ...queryVariables,
         limit: additionalProps?.limit || 15
       },
       rootRouteName: ROOT_ROUTE_NAMES.EVENT_RECORDS,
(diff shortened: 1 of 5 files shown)


src/components/widgets/EventWidget.tsx
```diff
@@ -53,7 +53,7 @@ export const EventWidget = ({ text, additionalProps }: WidgetProps) => {
       title: text ?? texts.homeTitles.events,
       query: QUERY_TYPES.EVENT_RECORDS,
       queryVariables: {
-        order: 'listDate_ASC',
+        ...queryVariables,
```
donni106, line 56: is the `order` still applied here?

PR #1270, opened 2025-02-21
refactor: improve tab bar layout and styles

donni106:
- removed tab bar height calculations to use default across the systems
- adjusted tab bar item styles for better spacing in portrait mode for different devices


src/navigation/MainTabNavigator.tsx
```diff
@@ -108,18 +106,18 @@ export const MainTabNavigator = () => {
...
           backgroundColor: colors.surface,
           borderTopColor: colors.gray20,
-          borderTopWidth: 1,
-          height: tabBarHeight
+          borderTopWidth: 1
         },
+        tabBarItemStyle:
+          isPortrait && !hasSafeArea
+            ? { marginBottom: normalize(7), marginTop: normalize(4) }
+            : undefined,
```
ardasnturk, line 117: How about using the && operator instead of `undefined` here?

```
 tabBarItemStyle: isPortrait && !hasSafeArea && { marginBottom: normalize(7), marginTop: normalize(4) },
```

src/navigation/MainTabNavigator.tsx
```diff
@@ -108,18 +106,18 @@ export const MainTabNavigator = () => {
...
         tabBarLabelStyle: {
-          fontSize: normalize(12),
-          lineHeight: normalize(14),
-          marginBottom: orientation === 'portrait' ? normalize(10) : undefined,
-          marginTop: orientation === 'portrait' ? normalize(4) : undefined
+          fontSize: normalize(10),
+          lineHeight: normalize(12)
```
ardasnturk, line 120: Do these values match our design forms?

src/navigation/MainTabNavigator.tsx
```diff
@@ -108,18 +106,18 @@ export const MainTabNavigator = () => {
...
         tabBarStyle: {
           backgroundColor: colors.surface,
           borderTopColor: colors.gray20,
-          borderTopWidth: 1,
-          height: tabBarHeight
+          borderTopWidth: 1
         },
+        tabBarItemStyle:
+          isPortrait && !hasSafeArea
+            ? { marginBottom: normalize(7), marginTop: normalize(4) }
```
ardasnturk, line 116: How did you calculate these values?

src/navigation/MainTabNavigator.tsx
```diff
@@ -108,18 +106,18 @@ export const MainTabNavigator = () => {
           tabRoutes?.inactiveBackgroundColor || tabNavigatorConfig.inactiveBackgroundColor,
         tabBarInactiveTintColor:
           tabRoutes?.inactiveTintColor || tabNavigatorConfig.inactiveTintColor,
-        tabBarItemStyle: { marginTop: normalize(0) },
         tabBarStyle: {
           backgroundColor: colors.surface,
           borderTopColor: colors.gray20,
-          borderTopWidth: 1,
-          height: tabBarHeight
+          borderTopWidth: 1
```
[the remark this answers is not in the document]
donni106: yes, after looks more accurate and aligns more with default system values

src/navigation/MainTabNavigator.tsx
donni106: because using only && would return false instead of undefined if it does not resolve to true

src/navigation/MainTabNavigator.tsx
donni106: ah this is goslar, i will reset these

PR #1271, opened 2025-02-24
refactor(waste): remove old street logics

donni106:
- removed unused state and action types related to old street settings
- updated reducer and screen logic to reflect changes in state management
- streamlined the update process for active waste types

GOS-118

### DIFF
--- a/src/components/HeaderLeft.tsx
+++ b/src/components/HeaderLeft.tsx
@@ -12,12 +12,22 @@ export const HeaderLeft = ({
   backImage,
   text
 }: HeaderBackButtonProps & { text?: string }) => {
-  if (!onPress && !backImage) {
+  if (!onPress && !backImage && !text) {
     return null;
   }
 
-  if (!onPress && backImage) {
-    return <View>{backImage({ tintColor: colors.darkText })}</View>;
+  if (!onPress) {
+    return (
+      <View>
+        {backImage ? (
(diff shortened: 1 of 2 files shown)


src/reducers/wasteSettingsReducer.ts
```diff
@@ -42,16 +35,7 @@ export enum WasteSettingsActions {
...
-      payload: {
-        [key: string]: {
-          active: boolean;
-          storeId?: number | string;
-        };
-      };
-    }
-  | {
-      type: WasteSettingsActions.resetActiveTypesWithOldStreetCleanup;
+      type: WasteSettingsActions.resetActiveTypes;
```
ardasnturk, line 38: If I have already saved a street and I want to update it, as a user it would be nice to see a loading icon on the screen after pressing the “save” button. Because depending on the internet connection, requests can be slow and I don't see any changes on the screen, so I thought that the “save” button wasn't working and pressed it several times in a row. Then I found myself in HomeScreen. Because many times the `goBack()` function worked.

donni106: Not only a visual feedback for the user. You cannot press the save multiple times, which was your issue.

ardasnturk: Yes. That's how we can solve both problems. 👍

ardasnturk (approved):

PR #1273, opened 2025-02-24
feat(navigation): update react-navigation from 6 to 7

donni106:
- needed to upgrade in terms of new arch
- refactored `unmountOnBlur`: https://reactnavigation.org/docs/upgrading-from-6.x#the-unmountonblur-option-is-removed-in-favor-of-poptotoponblur-in-bottom-tab-navigator-and-drawer-navigator


src/navigation/screenOptions.tsx
```diff
@@ -75,8 +74,6 @@ export const getScreenOptions =
     };
   };
 
-/* eslint-enable complexity */
```
ardasnturk, line 78: If we now have standard tabbar sizes. How about doing the same for the header? Currently we are also using a customized sizing for the header.
If we want a standard header, the following codes can be deleted.

`src/navigation/screenOptions`

```
 const headerHeight = !isPortrait ? normalize(35) + insets.top : normalize(56) + insets.top;
...
headerStyle: {
  height: headerHeight
},
```

ardasnturk: Portrait
|before|after|
|--|--|
image|image

Landscape
|before|after|
|--|--|
image|image

donni106: yeah I am completely convinced by the idea

PR #1274, opened 2025-02-25
release: add script for EAS Update

donni106:
New command `yarn update:all "EAS UPDATE MESSAGE` to make EAS Update more efficient with automatically incrementing a new `otaVersion` that will be shown in version number string to know about the latest version on devices.

image

GOS-105

### DIFF
--- a/app.json
+++ b/app.json
@@ -27,6 +27,7 @@
     "assetBundlePatterns": ["**/*"],
     "ios": {
       "bundleIdentifier": "de.dev.smartvillage.app",
+      "buildNumber": "1",
       "supportsTablet": true,
       "requireFullScreen": true,
       "infoPlist": {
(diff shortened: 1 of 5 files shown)


app.json
```diff
@@ -7,9 +7,18 @@
     "githubUrl": "https://github.com/smart-village-solutions/smart-village-app-app",
     "platforms": ["ios", "android"],
     "version": "4.1.0",
-    "runtimeVersion": { "policy": "appVersion" },
-    "extra": { "eas": { "projectId": "20540f3a-6e82-4038-8987-adb6d97f0e75" } },
-    "updates": { "url": "https://u.expo.dev/20540f3a-6e82-4038-8987-adb6d97f0e75" },
+    "otaVersion": 1,
```
ardasnturk, line 10: It looks very good. If we run `yarn update:all`, the otaVersion will be increased automatically, right? We don't need to increase this value extra. Well I think we will commit the updated otaVersion later.

ardasnturk (approved):

app.json
donni106: I do not understand? The script increments the `otaVersion` as i wrote in the PR description. And Git actions are performed as well, see the last 7 lines of the new script.

PR #1276, opened 2025-02-25
feat(tab): add `tabBarLabelStyle` to tab configuration

ardasnturk:
- introduced `tabBarLabelStyle` parameter in `createDynamicTabConfig`
- updated `MainTabNavigator` to utilize new `tabBarLabelStyle`
- modified `CustomTab` type to include `tabBarLabelStyle`

SVA-1531


src/config/navigation/tabConfig.tsx
```diff
@@ -147,6 +148,7 @@ export const createDynamicTabConfig = (
         size={normalize(iconSize)}
         style={iconStyle}
       />
-    )
+    ),
+    tabBarLabelStyle
```
donni106, line 152: please move it under `tabBarLabel` to have it close to the `tabBarLabel` logically

PR #1279, opened 2025-02-26
feat(report): add configuration prop to `SueReportLocation`

ardasnturk:
- introduced configuration prop to pass `clusterDistance` for map clustering
- updated `SueReportScreen` to include the new configuration prop

SUE-154

### DIFF
--- a/src/components/SUE/report/SueReportLocation.tsx
+++ b/src/components/SUE/report/SueReportLocation.tsx
@@ -77,6 +77,7 @@ enum SueStatus {
 /* eslint-disable complexity */
 export const SueReportLocation = ({
   areaServiceData,
+  configuration,
   control,
   errorMessage,
   requiredInputs,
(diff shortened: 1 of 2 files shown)


src/components/SUE/report/SueReportLocation.tsx
```diff
@@ -296,6 +302,7 @@ export const SueReportLocation = ({
       <WrapperHorizontal>
         <Map
           calloutTextEnabled
+          clusterDistance={configuration.geoMap?.clusterDistance}
```
donni106, line 305: thats all? wow. and the given value per configuration is the correct one that the map understands?

donni106 (approved):

src/components/SUE/report/SueReportLocation.tsx
ardasnturk: Yes. The value comes as the map wants. 👍

PR #1288, opened 2025-03-10
fix(navigation): update tab bar style for better appearance

ardasnturk:
- set tab bar inactive background color to a variable for consistency
- clean up unused imports in `screenOptions`

SVA-1578

|without colored|with colored|
|--|--|
Simulator Screenshot - iPhone 16 Pro - 2025-03-10 at 08 01 36|Simulator Screenshot - iPhone 16 Pro - 2025-03-10 at 08 02 05

### DIFF
--- a/src/navigation/MainTabNavigator.tsx
+++ b/src/navigation/MainTabNavigator.tsx
@@ -90,17 +90,19 @@ export const MainTabNavigator = () => {
 
   if (!tabConfigs || loading) return <LoadingSpinner loading />;
 
+  const { inactiveBackgroundColor: backgroundColor } = tabRoutes || tabNavigatorConfig;
+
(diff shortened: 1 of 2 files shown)


src/navigation/MainTabNavigator.tsx
```diff
@@ -90,17 +90,20 @@ export const MainTabNavigator = () => {
 
   if (!tabConfigs || loading) return <LoadingSpinner loading />;
 
+  const backgroundColor =
```
donni106, line 93: what about `const { inactiveBackgroundColor: backgroundColor } = tabRoutes || tabNavigatorConfig;`? does this work?

donni106 (approved):

PR #1293, opened 2025-03-21
feat(onboarding): add 'skip' option to app intro

ardasnturk:
- added `backgroundColor` prop to update the background color of onboarding
- updated the `renderSlider` function to show the incoming order of content
- added `handleSkip` to redirect to that slide if there is `termsAndConditions` checkbox when skip button is pressed
- added `handleSlideChange` to show alert with next button if there is `termsAndConditions` checkbox
- added `bottomButton` prop to style and `AppIntroSlider` component to show next and skip buttons above the dots
- updated text file to show on skip button

SVA-1577

|before|after|before|after|before|after|
|--|--|--|--|--|--|
Simulator Screenshot - iPhone 16 Pro Max - 2025-03-21 at 13 02 42|Simulator Screenshot - iPhone 16 Pro Max - 2025-03-21 at 13 01 33|Simulator Screenshot - iPhone 16 Pro Max - 2025-03-21 at 13 02 44|image|Simulator Screenshot - iPhone 16 Pro Max - 2025-03-21 at 13 02 47|Simulator Screenshot - iPhone 16 Pro Max - 2025-03-21 at 13 02 16

### DIFF
--- a/src/components/Checkbox.js
+++ b/src/components/Checkbox.js
@@ -8,9 +8,40 @@ import { useOpenWebScreen } from '../hooks';
 import { OrientationContext } from '../OrientationProvider';
 
 import { BoldText, RegularText } from './Text';
-import { WrapperHorizontal, WrapperRow } from './Wrapper';
+import { WrapperHorizontal } from './Wrapper';
 
 const { a11yLabel } = consts;
+
+const renderTitleContent = ({
+  boldTitle,
+  lightest,
+  link,
+  linkDescription,
+  navigate,
+  openWebScreen,
+  title
+}) => {
+  const TextComponent = boldTitle ? BoldText : RegularText;
+
+  return (
+    <WrapperHorizontal>
+      <TextComponent small lightest={lightest}>
+        {title}
+        {(!!link || !!navigate) && !!linkDescription && (
+          <RegularText
+            small
+            primary
+            underline
+            onPress={link ? openWebScreen : navigate}
+            style={{ flexShrink: 1 }}
+          >
+            {linkDescription}
+          </RegularText>
+        )}
+      </TextComponent>
+    </WrapperHorizontal>
+  );
+};
 
 // eslint-disable-next-line complexity
 export const Checkbox = ({
@@ -50,26 +81,15 @@ export const Checkbox = ({
       }
       size={normalize(21)}
       center={center}
-      title={
-        <WrapperHorizontal>
-          <WrapperRow>
-            {boldTitle ? (
-              <BoldText small lightest={lightest}>
-                {title}
-              </BoldText>
-            ) : (
-              <RegularText small lightest={lightest}>
-                {title}
-              </RegularText>
-            )}
-            {(!!link || !!navigate) && !!linkDescription && (
-              <RegularText small primary underline onPress={link ? openWebScreen : navigate}>
-                {linkDescription}
-              </RegularText>
-            )}
-          </WrapperRow>
-        </WrapperHorizontal>
-      }
+      title={renderTitleContent({
+        boldTitle,
+        lightest,
+        link,
+        linkDescription,
+        navigate,
+        openWebScreen,
+        title
+      })}
       onPress={onPress}
       checkedIcon={checkedIcon}
       checked={checked}
--- a/src/screens/AppIntroScreen.tsx
+++ b/src/screens/AppIntroScreen.tsx
@@ -5,9 +5,11 @@ import {
   ListRenderItem,
   Modal,
   ScrollView,
+  StyleProp,
   StyleSheet,
   TouchableOpacity,
-  View
+  View,
+  ViewStyle
 } from 'react-native';
 import AppIntroSlider from 'react-native-app-intro-slider';
 import { getStatusBarHeight } from 'react-native-status-bar-height';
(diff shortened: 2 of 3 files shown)


src/screens/AppIntroScreen.tsx
```diff
@@ -95,41 +105,82 @@ const TermsAndConditionsSection = ({ setShowButtonTermsAndConditions }) => {
...
+    <BoldText big style={styles.leftAligned}>
+      {item.title.toUpperCase()}
+    </BoldText>
+  );
+
+  const TextComponent = () => (
+    <WrapperVertical>
+      <RegularText style={styles.leftAligned}>{item.text}</RegularText>
+    </WrapperVertical>
+  );
```
donni106, line 144: i can't say what happened where, but the style got worse. content is misaligned and over the screen area...see before and after:

425444399-08d2c6e0-9e85-4d08-b06b-21e0681d2cfc-tile

src/screens/AppIntroScreen.tsx
```diff
@@ -158,69 +246,60 @@ export const AppIntroScreen = ({ setOnboardingComplete, onlyTermsAndConditions }
...
-          })
-        }
-        renderNextButton={() => <SliderButton label={texts.appIntro.continue} />}
+        renderSkipButton={() => (
+          <SliderButton label={texts.appIntro.skip} style={styles.skipButton} />
+        )}
         scrollEnabled={false}
         showDoneButton={showDoneButtonTermsAndConditions}
-        showNextButton={showNextButtonTermsAndConditions}
+        showNextButton
```
donni106, line 299: why is this set to always `true` now?

src/screens/AppIntroScreen.tsx
```diff
@@ -158,69 +246,60 @@ export const AppIntroScreen = ({ setOnboardingComplete, onlyTermsAndConditions }
...
+                slider.current?.goToSlide(currentSlideIndex + 1, true);
+              }}
+            >
+              <View style={buttonStyle}>
+                <BoldText lightest={!(isTermsSlide && !hasAcceptedTerms)}>
+                  {texts.appIntro.continue.toUpperCase()}
+                </BoldText>
+              </View>
+            </TouchableOpacity>
+          );
```
donni106, line 292: please make this a component like `SliderButton` to not have so much code here in the render part. it is very hard to read.

src/screens/AppIntroScreen.tsx
```diff
@@ -158,69 +246,60 @@ export const AppIntroScreen = ({ setOnboardingComplete, onlyTermsAndConditions }
...
-            >
-              <SliderButton
-                isPlaceholder
-                label={texts.appIntro.continue}
-                style={styles.sliderButtonAdditional}
-              />
-            </TouchableOpacity>
-          </View>
-        </View>
-      )}
```
donni106, line 223: how come all of this could be removed?

src/screens/AppIntroScreen.tsx
```diff
@@ -158,69 +246,60 @@ export const AppIntroScreen = ({ setOnboardingComplete, onlyTermsAndConditions }
...
-          if (
-            index + 1 ==
-            slides.findIndex((slide) => slide.onLeaveSlideName === Initializer.TermsAndConditions)
-          ) {
-            setShowNextButtonTermsAndConditions(false);
-          }
-
-          slides[index]?.onLeaveSlide?.(true);
+        onSlideChange={handleSlideChange}
+        renderDoneButton={() => <SliderButton label={texts.appIntro.continue} isLightest />}
```
donni106, line 260: why is this always `lightest` now?

src/screens/AppIntroScreen.tsx
ardasnturk: It may look bad in terms of design. But I tried to make it the same as the design in figma. Also, it should not be forgotten that the order of the content can be updated. If you want, I can fix the content in the slider at the top again like before.

image

src/screens/AppIntroScreen.tsx
ardasnturk: Because now we create the buttons through the props in `AppIntreSlider`.

src/screens/AppIntroScreen.tsx
ardasnturk: This actually changes the color of the text inside the button. I updated it to `isLightestText` to increase understandability

src/screens/AppIntroScreen.tsx
ardasnturk: Yeah, we don't actually need this prop.

src/screens/AppIntroScreen.tsx
donni106: yes, much better now. but the padding between image and checkbox needs to be larger.

src/components/Checkbox.js
```diff
@@ -8,10 +8,41 @@ import { useOpenWebScreen } from '../hooks';
...
+            small
+            primary
+            underline
+            onPress={link ? openWebScreen : navigate}
+            style={{ flexShrink: 1 }}
+          >
+            {linkDescription}
+          </RegularText>
+        )}
+      </TextComponent>
```
donni106, line 41: this will nest a text component inside a text component. i think there can be issues on some operating systems? this was not the case before, how come this is changed?

src/screens/AppIntroScreen.tsx
```diff
@@ -26,31 +35,96 @@ import { HtmlScreen } from './HtmlScreen';
...
 const SliderButton = ({
+  isLightestText = false,
+  isPlaceholder = false,
   label,
-  style,
-  isPlaceholder = false
+  style
 }: {
+  isLightestText?: boolean;
+  isPlaceholder?: boolean;
```
donni106, line 44: why does one have `...Text` and the other not? please name it similar. `isLightest` and `isPlaceholder` for example.

src/screens/AppIntroScreen.tsx
```diff
@@ -86,7 +160,7 @@ const TermsAndConditionsSection = ({ setShowButtonTermsAndConditions }) => {
             params: {
               title: texts.profile.privacyCheckLink,
               query: QUERY_TYPES.PUBLIC_HTML_FILE,
-              queryVariables: { name: 'datenschutzProfile' }
+              queryVariables: { name: contentName }
```
donni106, line 163: where does the `contentName` comes from? is this something we need to create for existing apps?

src/screens/AppIntroScreen.tsx
```diff
@@ -158,69 +316,39 @@ export const AppIntroScreen = ({ setOnboardingComplete, onlyTermsAndConditions }
...
-          // hide the next button on the terms and conditions slide
-          if (
-            index + 1 ==
-            slides.findIndex((slide) => slide.onLeaveSlideName === Initializer.TermsAndConditions)
-          ) {
-            setShowNextButtonTermsAndConditions(false);
-          }
-
-          slides[index]?.onLeaveSlide?.(true);
-        }}
```
donni106, line 180: this is entirely gone? what about the `showNextButtonTermsAndConditions` logic? did you try the case, where it was needed before with the new code?

src/screens/AppIntroScreen.tsx
ardasnturk: Is this space enough or should I increase it?
|before|after|
|--|--|
Simulator Screenshot - iPhone 16 Plus - 2025-05-09 at 11 49 22|Simulator Screenshot - iPhone 16 Plus - 2025-05-09 at 11 48 57

src/components/Checkbox.js
ardasnturk: Previously, two text components were wrapped with `WrapperRow`. This was causing the text in the `Checkbox` to jump off the screen. In this way, it is ensured that it descends to the bottom line without going off the screen.
Maybe we should use wrapping with `WrapperWrap`?

|before|after|after (with WrapperWrap)|
|--|--|--|
Simulator Screenshot - iPhone 16 Plus - 2025-05-09 at 11 52 38|Simulator Screenshot - iPhone 16 Plus - 2025-05-09 at 11 53 05|Simulator Screenshot - iPhone 16 Plus - 2025-05-09 at 11 52 54

src/screens/AppIntroScreen.tsx
ardasnturk: It was already `isLightest` before, but I updated it to `isLightestText` to make it more understandable. The `isLightest` prop seems to be related to the button. Should I make it `isPlaceholderText` then?

src/screens/AppIntroScreen.tsx
ardasnturk: If we have the `termsAndConditions` option in the slider, then the name of the content we want to show must be added to the item of the slider with the `contentName` key. Otherwise the `datenschutzProfile` content always needs to be created.

src/screens/AppIntroScreen.tsx
ardasnturk: I don't think we need this logic anymore, because we already create the buttons with the props. This logic was used to show buttons outside the `AppIntroSlider` component.

src/screens/AppIntroScreen.tsx
ardasnturk: I updated it like this.

Simulator Screenshot - iPhone 16 Plus - 2025-05-09 at 12 11 44

src/screens/AppIntroScreen.tsx
donni106: seems legit, all right

src/screens/AppIntroScreen.tsx
donni106: no, the props names are without text, so i suggest the booleans should as well.

src/components/Checkbox.js
donni106: if it works well for all platforms, we can keep the text nesting.

ardasnturk: Yes, I tested it for both platforms and did not encounter any problems.

src/screens/AppIntroScreen.tsx
ardasnturk: We don't need the `isPlaceholder` prop because we don't use it. We don't use it. I update `isLightestText` to `isLightest`.

donni106 (approved):

PR #1294, opened 2025-03-21
feat(service): improve tile layout and centering for last row

ardasnturk:
- added orientation context to determine layout based on device orientation
- adjusted tile rendering to center the last row when necessary
- updated styles for tiles container and centering behavior

SVA-1577

|before (portrait)|after (portrait)|before (landscape)|after (landscape)|
|--|--|--|--|
Simulator Screenshot - iPhone 16 Pro Max - 2025-03-21 at 13 05 53|Simulator Screenshot - iPhone 16 Pro Max - 2025-03-21 at 13 05 14|Simulator Screenshot - iPhone 16 Pro Max - 2025-03-21 at 13 05 59|Simulator Screenshot - iPhone 16 Pro Max - 2025-03-21 at 13 05 18


src/components/screens/Service.tsx
```diff
@@ -109,7 +117,12 @@ export const Service = ({
     </DiagonalGradient>
   ) : (
     <>
-      <WrapperWrap spaceBetween>{tiles?.map(renderItem)}</WrapperWrap>
+      <WrapperWrap
+        style={[styles.tilesContainer, shouldCenter && styles.centerLastRow]}
+        spaceBetween
```
donni106, line 122: that wrapper has props already:

image

so we should decide wether to have `spaceBetween` or `spaceAround` applied based on `shouldCenter` instead of adding a new individual component style, right!?

ardasnturk: Yes, you're absolutely right. I will update the code.

donni106 (approved):

PR #1296, opened 2025-03-26
feat(event): remove dates from sub list in calendar view

ardasnturk:
This PR prevents multi-day events from repeating in the calendar view and updates the style of the sticky header.

SVASD-130

|before list view|after list view|before calendar view|after calendar view|
|--|--|--|--|
Simulator Screenshot - iPhone 16 Pro Max - 2025-03-26 at 08 09 07|Simulator Screenshot - iPhone 16 Pro Max - 2025-03-26 at 08 08 44|Simulator Screenshot - iPhone 16 Pro Max - 2025-03-26 at 08 09 14|Simulator Screenshot - iPhone 16 Pro Max - 2025-03-26 at 08 08 52

### DIFF
--- a/src/hooks/listHooks.js
+++ b/src/hooks/listHooks.js
@@ -17,7 +17,7 @@ import { TextListItem } from '../components/TextListItem';
 import { VolunteerApplicantListItem } from '../components/volunteer/VolunteerApplicantListItem';
 import { VolunteerConversationListItem } from '../components/volunteer/VolunteerConversationListItem';
 import { VolunteerPostListItem } from '../components/volunteer/VolunteerPostListItem';
-import { consts, normalize, texts } from '../config';
+import { colors, consts, normalize, texts } from '../config';
 import { momentFormat } from '../helpers';
 import { QUERY_TYPES } from '../queries';
 import { ScreenName } from '../types';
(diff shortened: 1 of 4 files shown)


src/hooks/listHooks.js
```diff
@@ -35,23 +35,25 @@ const getListType = (query, listTypesSettings) => {
...
-                ...options.queryVariables,
-                dateRange: [momentFormat(item, 'YYYY-MM-DD'), momentFormat(item, 'YYYY-MM-DD')]
-              },
-              rootRouteName: ROOT_ROUTE_NAMES.EVENT_RECORDS
-            })
-        : undefined
-    }
-  />
+  <View style={styles.sectionHeader}>
+    <SectionHeader
```
donni106, line 39: couldn't this be the `containerStyle` of the `SectionHeader` instead of a new `View` wrapping here?

ardasnturk: I updated it to `sectionHeaderContainer`, I think this name is better.

donni106: setting the `containerStyle` with `paddingTop` should overwrite the default values of the styled component.

donni106 (approved):

PR #1297, opened 2025-03-26
feat(waste): improve waste calendar performance

ardasnturk:
This PR improves the performance of the waste calendar and fixes minor UI bugs

|before|after|before|after|
|--|--|--|--|
Simulator Screenshot - iPhone 16 Pro Max - 2025-03-26 at 23 08 44|Simulator Screenshot - iPhone 16 Pro Max - 2025-03-26 at 23 09 59|Simulator Screenshot - iPhone 16 Pro Max - 2025-03-26 at 23 08 57|Simulator Screenshot - iPhone 16 Pro Max - 2025-03-26 at 23 10 03

SVA-1539
SVA-1510

### DIFF
--- a/src/components/waste/WasteInputForm.tsx
+++ b/src/components/waste/WasteInputForm.tsx
@@ -80,16 +80,16 @@ export const WasteInputForm = ({
     ((!isInputAutoFocus && inputValue) || isInputAutoFocus)
   );
 
-  const listContainerHeight = useMemo(
-    () =>
-      (inputValueCitySelected ? filteredStreets : filteredCities)?.length < 6
-        ? device.platform === 'ios'
-          ? 'auto'
-          : (inputValueCitySelected ? filteredStreets : filteredCities)?.length *
-            (normalize(22) + 2 * normalize(16))
-        : dimensions.height / 2.5,
-    [inputValueCitySelected, filteredStreets, filteredCities, dimensions.height]
-  );
+  const listContainerHeight = useMemo(() => {
+    const activeList = inputValueCitySelected ? filteredStreets : filteredCities;
+    const listLength = activeList?.length ?? 0;
+
+    if (listLength < 6) {
+      return device.platform === 'ios' ? 'auto' : listLength * (normalize(22) + 2 * normalize(16));
+    } else {
+      return dimensions.height / 2.5;
+    }
+  }, [inputValueCitySelected, filteredStreets, filteredCities, dimensions.height]);
 
   return (
     <>
@@ -214,7 +214,8 @@ const styles = StyleSheet.create({
       android: {
         borderColor: colors.gray20,
         borderRadius: 0,
-        borderWidth: normalize(1)
+        borderWidth: normalize(1),
+        maxHeight: normalize(300)
       }
     })
   },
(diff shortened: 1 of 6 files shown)


src/screens/WasteCollectionScreen.js
```diff
@@ -313,7 +313,7 @@ export const WasteCollectionScreen = ({ navigation }) => {
     setIsDayOverlayVisible(true);
   }, []);
 
-  if (isRehydrating || loading || typesLoading) {
+  if (isRehydrating || typesLoading) {
```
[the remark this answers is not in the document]
ardasnturk: No. This suggestion is not correct. If we add `addressesData` control, `LoadingSpinner` always appears on the screen and this if check cannot be passed.

src/config/texts.js
```diff
@@ -1531,7 +1531,7 @@ export const texts = {
     hintCityAndStreet: 'Bitte geben Sie Ihre Ortschaft und anschließend Ihre Straße an.',
     hintStreet: 'Bitte geben Sie Ihre Straße an.',
     location: 'Ortschaft',
-    myLocation: 'Meine Straße:',
+    myLocation: 'Meine Straße',
```
donni106, line 1534: whats the difference here on the screenshots to look at?

image

src/components/waste/WasteInputForm.tsx
```diff
@@ -84,7 +85,7 @@ export const WasteInputForm = ({
       (inputValueCitySelected ? filteredStreets : filteredCities)?.length < 6
         ? device.platform === 'ios'
           ? 'auto'
-          : (inputValueCitySelected ? filteredStreets : filteredCities)?.length *
+          : (inputValueCitySelected ? filteredCities : filteredStreets)?.length *
```
donni106, line 88: i do not understand this change for height calculation. can you please explain the swap

src/config/texts.js
ardasnturk: Meine Straße has the value `:` instead of `::`

src/components/waste/WasteInputForm.tsx
ardasnturk: If `inputValueCitySelected` is true then the `filteredStreet.length` is 0 because the city is selected but the `filteredCities.length` is greater than 0. Therefore the list of suggestions for Android was not visible. So I changed `filteredScreets` to `filteredCities`.

donni106: hm, I do not get it. if the city is selected, we want to show streets next, so we need to calculate with street length.
if a city is not selected, we want to show cities and calculate with cities. this is how can think about the situation.

otherwise the line before `(inputValueCitySelected ? filteredStreets : filteredCities)?.length < 6` needs an update as well?

ardasnturk: Yeah, you're right. That doesn't make sense right now. How about updating the code in this way to make it more understandable? We're currently doing the same check in 2 different places. I haven't tested this code yet.

```
 const listContainerHeight = useMemo(() => {
    const activeList = inputValueCitySelected ? filteredStreets : filteredCities;
    const listLength = activeList?.length ?? 0;

    if (listLength < 6) {
      return device.platform === 'ios' ? 'auto' : listLength * (normalize(22) + 2 * normalize(16));
    } else {
      return dimensions.height / 2.5;
    }
  }, [inputValueCitySelected, filteredStreets, filteredCities, dimensions.height]);
```

donni106 (approved):

PR #1299, opened 2025-03-28
feat(tabbar): add the ability to use different icons when tabbar is active

ardasnturk:
- added `activeIconName` to add the option to use a different icon when the tabbar is active

SVA-1555

In the screenshots, the active and deactive icon of the 3rd icon from the right is activated. It can be seen that the icon changes when Tab is selected.

|||
|--|--|
Simulator Screenshot - iPhone 16 Pro Max - 2025-03-28 at 10 58 22|Simulator Screenshot - iPhone 16 Pro Max - 2025-03-28 at 11 00 11

### DIFF
--- a/src/config/navigation/tabConfig.tsx
+++ b/src/config/navigation/tabConfig.tsx
@@ -124,6 +124,7 @@ export const createDynamicTabConfig = (
   label: string,
   totalCount: number,
   screen: ScreenName,
+  activeIconName?: keyof typeof Icon,
   iconLandscapeStyle?: ViewStyle,
   iconStyle?: ViewStyle,
   initialParams?: Record<string, any>,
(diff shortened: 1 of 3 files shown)


src/config/navigation/tabConfig.tsx
```diff
@@ -142,17 +143,29 @@ export const createDynamicTabConfig = (
...
+        <OrientationAwareIcon
+          color={color}
+          Icon={Icon[iconName as keyof typeof Icon]}
+          iconName={iconName}
+          landscapeStyle={iconLandscapeStyle}
+          size={normalize(iconSize)}
+          strokeColor={strokeColor}
+          strokeWidth={strokeWidth}
+          style={iconStyle}
+        />
```
donni106, line 168: too much code duplicated here sadly. isn't it possible to set the icon name conditionally and then return the `OrientationAwareIcon` once?

ardasnturk: Yes, you are right. I prevented code repetition with the commit I just sent. I added !`!activeIconName && focused` control to `iconName` and `Icon` props.

https://github.com/smart-village-solutions/smart-village-app-app/commit/f04288440c71c50bbfec9e3945ba714d4baf4e0f

donni106 (approved):

src/config/navigation/tabConfig.tsx
donni106: wow, such great small changes now!

PR #1307, opened 2025-04-04
feat(service): integrate app design system for service tiles

ardasnturk:
- added `serviceTiles` from `appDesignSystem` object to `Service.tsx` to implement service tiles design that can be updated via main server
- added styles to components in `ServiceTile.tsx` to update service tiles according to styles added to `globalSettings.appDesignSystem.serviceTiles` object
- added style controls to prevent the app from crashing if there are no styles in `globalSettings`
- added `IconUrl` component to add the ability to show svg icon via link on service tiles
- added `svg` type to show svg format icons with `IconUrl` component
- reduced the size of the `Image` component used for the icon from 40 to 30 to be the same size as the icons

SVA-1614

upgrade app to version 5.1.3 for testing
the following styles should be added to `globalSettings` by default

```
"appDesignSystem": {
  "serviceTiles": {
    "numberOfLines": 2,
    "tileStyle": {
      "backgroundColor": "#000",
      "borderColor": "#000",
      "marginBottom": 7,
      "borderRadius": 8,
      "height": 120,
      "borderWidth": 1,
      "alignSelf": "center"
    },
    "fontStyle": {
      "fontFamily": "regular",
      "fontWeight": "100",
      "fontSize": 19,
      "color": "#fff"
    },
    "iconStyle": {
      "strokeWidth": 2,
      "iconSize": 20,
      "iconStroke": 1,
      "color": "#fff"
    }
  },
}
```

Simulator Screenshot - iPhone 16 Pro Max - 2025-04-04 at 09 28 24

### DIFF
--- a/src/components/screens/Service.tsx
+++ b/src/components/screens/Service.tsx
@@ -85,7 +85,7 @@ export const Service = ({
         item={item}
         key={`item${item.title || item.accessibilityLabel}-index${index}`}
         onToggleVisibility={onToggleVisibility}
-        serviceTileStyles={serviceTiles}
+        serviceTiles={serviceTiles}
         tileSizeFactor={tileSizeFactor}
       />
     ),
--- a/src/components/screens/ServiceTile.tsx
+++ b/src/components/screens/ServiceTile.tsx
@@ -11,6 +11,26 @@ import { Badge } from '../profile';
 import { ServiceBox } from '../ServiceBox';
 import { BoldText } from '../Text';
 
+const normalizeStyleValues = (styleObj: any) => {
+  if (!Object.keys(styleObj).length) return styleObj;
+
+  const normalizedStyle = {};
+
+  for (const key in styleObj) {
+    const value = styleObj[key];
+
+    if (typeof value === 'number') {
+      normalizedStyle[key] = normalize(value);
+    } else if (typeof value === 'object' && value !== null) {
+      normalizedStyle[key] = normalizeStyleValues(value);
+    } else {
+      normalizedStyle[key] = value;
+    }
+  }
+
+  return normalizedStyle;
+};
+
 export type TServiceTile = {
   accessibilityLabel: string;
   icon: string;
@@ -35,7 +55,7 @@ export const ServiceTile = ({
   item,
   onToggleVisibility,
   tileSizeFactor = 1,
-  serviceTileStyles
+  serviceTiles
 }: {
   draggableId: string;
   hasDiagonalGradientBackground?: boolean;
@@ -47,7 +67,7 @@ export const ServiceTile = ({
     setIsVisible: (isVisible: boolean) => void
   ) => void;
   tileSizeFactor?: number;
-  serviceTileStyles?: any;
+  serviceTiles?: any;
 }) => {
   const navigation = useNavigation<StackNavigationProp<any>>();
   const { orientation, dimensions } = useContext(OrientationContext);
@@ -61,9 +81,13 @@ export const ServiceTile = ({
     [isEditMode, isVisible]
   );
   const ToggleVisibilityIcon = isVisible ? Icon.Visible : Icon.Unvisible;
-  const { fontStyle = {}, iconStyle = {}, numberOfLines, tileStyle = {} } = serviceTileStyles;
+  const { fontStyle = {}, iconStyle = {}, numberOfLines, tileStyle = {} } = serviceTiles;
 
   const hasTileStyle = !!Object.keys(tileStyle).length;
+
+  const normalizedFontStyle = normalizeStyleValues(fontStyle);
+  const normalizedIconStyle = normalizeStyleValues(iconStyle);
+  const normalizedTileStyle = normalizeStyleValues(tileStyle);
 
   return (
     <ServiceBox
@@ -71,7 +95,7 @@ export const ServiceTile = ({
       dimensions={dimensions}
       numberOfTiles={item?.numberOfTiles}
       orientation={orientation}
-      style={tileStyle}
+      style={normalizedTileStyle}
     >
       <TouchableOpacity
         style={[hasTileStyle && styles.button]}
@@ -93,24 +117,32 @@ export const ServiceTile = ({
           {item.iconName ? (
             <Icon.NamedIcon
               color={
-                iconStyle.color || (hasDiagonalGradientBackground ? colors.lightestText : undefined)
+                normalizedIconStyle.color
+                  ? normalizedIconStyle.color
+                  : hasDiagonalGradientBackground
+                  ? colors.lightestText
+                  : undefined
               }
               name={item.iconName}
-              size={iconStyle.size || normalize(30)}
-              strokeColor={iconStyle.strokeColor}
-              strokeWidth={iconStyle.strokeWidth}
-              style={[styles.serviceIcon, iconStyle]}
+              size={normalizedIconStyle.size || normalize(30)}
+              strokeColor={normalizedIconStyle.strokeColor}
+              strokeWidth={normalizedIconStyle.strokeWidth}
+              style={[styles.serviceIcon, normalizedIconStyle]}
             />
           ) : item.svg ? (
             <IconUrl
               color={
-                iconStyle.color || (hasDiagonalGradientBackground ? colors.lightestText : undefined)
+                normalizedIconStyle.color
+                  ? normalizedIconStyle.color
+                  : hasDiagonalGradientBackground
+                  ? colors.lightestText
+                  : undefined
               }
               iconName={item.svg}
-              size={iconStyle.size || normalize(30)}
-              strokeColor={iconStyle.strokeColor}
-              strokeWidth={iconStyle.strokeWidth}
-              style={[styles.serviceIcon, iconStyle]}
+              size={normalizedIconStyle.size || normalize(30)}
+              strokeColor={normalizedIconStyle.strokeColor}
+              strokeWidth={normalizedIconStyle.strokeWidth}
+              style={[styles.serviceIcon, normalizedIconStyle]}
             />
           ) : (
             <Image
@@ -118,7 +150,7 @@ export const ServiceTile = ({
               childrenContainerStyle={[
                 styles.serviceImage,
                 !!item.icon && {
-                  height: iconStyle.size || normalize(30)
+                  height: normalizedIconStyle.size || normalize(30)
                 },
                 !!item.tile &&
                   stylesWithProps({
@@ -142,7 +174,7 @@ export const ServiceTile = ({
               center
               accessibilityLabel={`(${item.title}) ${consts.a11yLabel.button}`}
               numberOfLines={numberOfLines}
-              style={fontStyle}
+              style={normalizedFontStyle}
             >
               {item.title}
             </BoldText>


src/components/screens/Service.tsx
```diff
@@ -51,6 +52,8 @@ export const Service = ({
   const { globalSettings } = useContext(SettingsContext);
   const { settings = {} } = globalSettings;
   const { personalizedTiles = false, tileSizeFactor = 1 } = settings;
+  const { appDesignSystem } = useContext(ConfigurationsContext);
+  const { serviceTiles = {} } = appDesignSystem;
```
donni106, line 56: why does this happen here, just to pass it to `ServiceTile`? shouldn't it be placed directly in `ServiceTile` to use the styles there?

src/components/screens/ServiceTile.tsx
```diff
@@ -21,6 +21,7 @@ export type TServiceTile = {
   params?: any;
   query?: string;
   routeName: string;
+  svg?: string;
```
donni106, line 24: where does this svg came from and what is its purpose? i could not find any hint in the commit message or PR description.

src/components/screens/ServiceTile.tsx
```diff
@@ -84,16 +94,35 @@ export const ServiceTile = ({
         <View style={!isVisible && styles.invisible}>
           {item.iconName ? (
             <Icon.NamedIcon
-              color={hasDiagonalGradientBackground ? colors.lightestText : undefined}
+              color={
+                iconStyle.color || hasDiagonalGradientBackground ? colors.lightestText : undefined
```
donni106, line 98: as copilot already mentioned, this should be `iconStyle.color || (hasDiagonalGradientBackground ? colors.lightestText : undefined)` or `(iconStyle.color || hasDiagonalGradientBackground) ? colors.lightestText : undefined`. which one is the correct one?

src/components/screens/ServiceTile.tsx
```diff
@@ -115,6 +144,8 @@ export const ServiceTile = ({
               primary={!hasDiagonalGradientBackground}
               center
               accessibilityLabel={`(${item.title}) ${consts.a11yLabel.button}`}
+              numberOfLines={numberOfLines}
+              style={hasFontStyle && fontStyle}
```
donni106, line 148: why do you check if the object `fontStyle` has values? `style={}` for `fontStyle = {}` should work with applying no styles.

src/components/screens/ServiceTile.tsx
```diff
@@ -58,15 +61,22 @@ export const ServiceTile = ({
...
+  const { fontStyle = {}, iconStyle = {}, numberOfLines, tileStyle = {} } = serviceTileStyles;
+
+  const hasFontStyle = !!Object.keys(fontStyle).length;
+  const hasIconStyle = !!Object.keys(iconStyle).length;
+  const hasTileStyle = !!Object.keys(tileStyle).length;
 
   return (
     <ServiceBox
-      orientation={orientation}
+      bigTile={!!item.tile || hasTileStyle}
```
donni106, line 72: this reads weird, if there is a tile or it has tile style then it is a big tile? having tile style always means big tile?
what about having a variable `isBigStyle` with some comments, to clarify the logic?

src/components/screens/ServiceTile.tsx
```diff
@@ -84,16 +94,35 @@ export const ServiceTile = ({
...
+              style={[styles.serviceIcon, hasIconStyle && iconStyle]}
             />
           ) : (
             <Image
               source={{ uri: item.icon || item.tile }}
               childrenContainerStyle={[
                 styles.serviceImage,
+                hasIconStyle &&
+                  !!item.icon && {
+                    height: iconStyle.size || normalize(30)
```
donni106, line 124: the height was 40 before, why 30 now? info in commit message missing.

src/components/screens/ServiceTile.tsx
```diff
@@ -58,15 +61,22 @@ export const ServiceTile = ({
     [isEditMode, isVisible]
   );
   const ToggleVisibilityIcon = isVisible ? Icon.Visible : Icon.Unvisible;
+  const { fontStyle = {}, iconStyle = {}, numberOfLines, tileStyle = {} } = serviceTileStyles;
```
donni106, line 64: every values are missing normalize, right? shouldn't there be a logic to apply all the number values from the app design system with normalize?

src/config/appDesignSystem/default.ts
```diff
@@ -1,4 +1,10 @@
 export const defaultAppDesignSystemConfig = {
+  serviceTiles: {
+    fontStyle: {},
+    iconStyle: {},
+    numberOfLines: 3,
```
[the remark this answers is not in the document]
donni106: I think you understood wrong. 3 is the default, 2 is the example that overrides the default in that case.

src/components/screens/Service.tsx
ardasnturk: I find it better to place it here for optimization. If I place it in the ServiceTile component and we have for example 10 service tiles, a request is sent to the context api for all of them. But here only one request is sent and it is applied for all of them.

src/components/screens/ServiceTile.tsx
```diff
@@ -58,15 +61,22 @@ export const ServiceTile = ({
...
-      orientation={orientation}
+      bigTile={!!item.tile || hasTileStyle}
       dimensions={dimensions}
-      bigTile={!!item.tile}
       numberOfTiles={item?.numberOfTiles}
+      orientation={orientation}
+      style={hasTileStyle && tileStyle}
     >
       <TouchableOpacity
+        style={hasTileStyle && styles.button}
```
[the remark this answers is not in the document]
ardasnturk: For this part in the rest of the code, then I should update it to array, right?
```
<View style={!isVisible && styles.invisible}>
```

src/components/screens/ServiceTile.tsx
ardasnturk: `iconStyle.color || (hasDiagonalGradientBackground ? colors.lightestText : undefined)`

I think this is correct. First we take the `iconStyle.color` value. If `iconStyle.color` is undefined then the second part is applied.

src/components/screens/ServiceTile.tsx
ardasnturk: Yes, it makes sense. You're right. We already create an empty object by default.

src/components/screens/ServiceTile.tsx
```diff
@@ -127,13 +158,18 @@ export const ServiceTile = ({
...
+    justifyContent: 'center',
+    width: '100%'
+  },
   serviceIcon: {
     alignSelf: 'center',
     paddingVertical: normalize(7.5)
   },
   serviceImage: {
     alignSelf: 'center',
-    height: normalize(40),
```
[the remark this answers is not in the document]
ardasnturk: Because the `normalize(30)` value has been added as standard to the `childrenContainerStyle` prop in the `Image` component. If you want to know why it went from 40 to 30, it's because I think it should be the same size as the icons.

src/components/screens/ServiceTile.tsx
ardasnturk: We already had the `normalize(30)` value for `Icon.NamedIcon` before. So I also added `IconUrl` as `normalize(30)`. I made this change in the `Image` component because I thought it should be the same size as the icons.

src/components/screens/ServiceTile.tsx
ardasnturk: Yes, that's right. I asked ChatGPT for this and he suggested a script like this. Do you think we should use it?

```
const normalizeStyleValues = (styleObj: any) => {
  const normalizedStyle = {};

  if (!Object.keys(styleObj).length) return styleObj;

  for (const key in styleObj) {
    const value = styleObj[key];

    if (typeof value === 'number') {
      normalizedStyle[key] = normalize(value);
    } else if (typeof value === 'object' && value !== null) {
      // Eğer nested object varsa (örneğin shadowOffset: { width: 2, height: 3 })
      normalizedStyle[key] = normalizeStyleValues(value);
    } else {
      normalizedStyle[key] = value;
    }
  }

  return normalizedStyle;
};
```

src/components/screens/ServiceTile.tsx
ardasnturk: Yes. If `tileStyle` is available, it always means bigTile. Until now we had no other option to create a bigTile than adding an image. We are trying to add the option to create the bigTile from the main-server. If there is no `tileStyle` then this is a normal service tile.

src/components/screens/ServiceTile.tsx
ardasnturk: [9fa123f](https://github.com/smart-village-solutions/smart-village-app-app/commit/9fa123f31738e43ad010657b72c7bd4219f63428)

src/components/screens/ServiceTile.tsx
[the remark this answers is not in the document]
ardasnturk: [9fa123f](https://github.com/smart-village-solutions/smart-village-app-app/commit/9fa123f31738e43ad010657b72c7bd4219f63428)

src/components/screens/ServiceTile.tsx
ardasnturk: [9fa123f](https://github.com/smart-village-solutions/smart-village-app-app/commit/9fa123f31738e43ad010657b72c7bd4219f63428)

src/components/screens/ServiceTile.tsx
ardasnturk: [9fa123f](https://github.com/smart-village-solutions/smart-village-app-app/commit/9fa123f31738e43ad010657b72c7bd4219f63428)

src/components/screens/Service.tsx
donni106: ok, but then please find a better prop name for `serviceTileStyles` as it is not a style object as we use it for styling elements but a self defined object of configs

```
const { fontStyle = {}, iconStyle = {}, numberOfLines, tileStyle = {} } = serviceTileStyles;
```

so as it is named `const { serviceTiles = {} } = appDesignSystem;` from the global configs, why not using that name further as the prop?

src/components/screens/ServiceTile.tsx
donni106: why not, seems appropriate.
only change i would suggest is to move the `const normalizedStyle = {};` under the early return, to not init a variable, that will not be used.

src/components/screens/ServiceTile.tsx
donni106: ok thanks for clearing out.
and what is with the second part?
> what about having a variable isBigStyle with some comments, to clarify the logic?

src/components/screens/Service.tsx
ardasnturk: I honestly thought `serviceTileStyles` would be more descriptive. But of course it can also be used as `serviceTiles`.

ardasnturk: https://github.com/smart-village-solutions/smart-village-app-app/pull/1307/commits/c34455a461b7746ffd807194727c8dd2162f62b0

src/config/appDesignSystem/default.ts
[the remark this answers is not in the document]
ardasnturk: Yes. 3 is the default value. Unless otherwise specified, `numberOfLines` is 3.

PR #1311, opened 2025-04-16
feat(profile): login with Keycloak

ardasnturk:
This PR allows the user to log in with Keycloak and save the resulting accessToken to the device. In order to test it, a new build needs to be created due to the native requirements of expo-web-browser.

To test:

To activate the profile property, firstly, the `withProfile` value in the `ScreenName.Home` object in the `defaultStackConfig` must be updated to `true`

Then the keys required for the login process must be added to the `globalSettings.settings` object as follows.

```
  "profile": {
    "clientId": "",
    "clientSecret": "",
    "scopes": [],
    "serverUrl": ""
  },
```

|not logged user|login alert|login with web browser|logged in user|
|--|--|--|--|
Simulator Screenshot - iPhone 16 Pro Max - 2025-04-16 at 13 08 01|Simulator Screenshot - iPhone 16 Pro Max - 2025-04-16 at 13 08 04|Simulator Screenshot - iPhone 16 Pro Max - 2025-04-16 at 13 09 22|Simulator Screenshot - iPhone 16 Pro Max - 2025-04-16 at 13 08 10

SVA-1511


src/config/consts.js
```diff
@@ -43,6 +43,10 @@ export const consts = {
     infoProvider: 'Anbieterinformationen (Taste)',
     lastName: 'Nachname',
     link: '(Link)',
+    login: 'Anmelden (Taste)',
+    loginHint: 'Zu den Anmeldedaten wechseln',
+    logout: 'Abmelden (Taste)',
+    logoutHint: 'Zu den Anmeldedaten wechseln',
```
donni106, line 49: is it intended to habe the same texts for `loginHint` and `logoutHint`? and where does `Zu den Anmeldedaten wechseln` come from? i don't get the meaning correctly, what should it tell someone?

src/components/LoginHeader.tsx
```diff
@@ -0,0 +1,49 @@
...
+export const LoginHeader = ({ route, style }: Props) => {
+  const { globalSettings } = useContext(SettingsContext);
+  const { settings } = globalSettings;
+  const { hasProfile = false } = settings;
+
+  const { isLoggedIn, loading, login, logout } = useLoginProfile();
+
+  if (!hasProfile || loading) {
+    return null;
+  }
```
donni106, line 28: checking `hasProfile` should not happen here but already considered in the stack config boolean flag `withLogin`

src/hooks/useLoginProfile.ts
```diff
@@ -0,0 +1,183 @@
...
+const serverUrl = profile?.serverUrl;
+
+const keycloakDiscovery = {
+  authorizationEndpoint: serverUrl + '/auth',
+  tokenEndpoint: serverUrl + '/token',
+  revocationEndpoint: serverUrl + '/revoke'
+};
+
+const CLIENT_ID = profile?.clientId;
+const CLIENT_SECRET = profile?.clientSecret;
```
donni106, line 28: what about using these values from global settings to be able to change them on the run instead of having to make new releases as they are shipped in the secrets file?

src/hooks/useLoginProfile.ts
```diff
@@ -0,0 +1,183 @@
...
+    scheme: appJson.expo.scheme,
+    path: 'redirect'
+  });
+
+  const [request, response, promptAsync] = useAuthRequest(
+    {
+      clientId: CLIENT_ID,
+      codeChallengeMethod: CodeChallengeMethod.S256,
+      redirectUri,
+      responseType: ResponseType.Code,
```
donni106, line 45: why do we add it here, if it is the default? 

image

https://docs.expo.dev/versions/latest/sdk/auth-session/#authrequestconfig

src/hooks/useLoginProfile.ts
```diff
@@ -0,0 +1,183 @@
...
+      scopes: SCOPES,
+      usePKCE: false
```
donni106, line 47: why do we set this to `false`?

src/hooks/useLoginProfile.ts
```diff
@@ -0,0 +1,183 @@
...
+    try {
+      setLoading(true);
+      const tokenResult = await exchangeCodeAsync(
+        {
+          code,
+          clientId: CLIENT_ID,
+          clientSecret: CLIENT_SECRET,
+          redirectUri,
+          extraParams: {
+            grant_type: 'authorization_code'
```
donni106, line 73: why is this hardcoded if everything else is set per variables?

src/hooks/useLoginProfile.ts
```diff
@@ -0,0 +1,183 @@
...
+    } finally {
+      setLoading(false);
+    }
+  };
+
+  useEffect(() => {
+    (async () => {
+      await checkToken();
+    })();
+  }, []);
```
donni106, line 92: wouldn't `useEffect(checkToken, []);` work here? why would we need to await something if we are not doing anything with its returned value?

src/config/consts.js
ardasnturk: Yes, you are right. loginHint and logoutHint should not be the same. These texts were added here for accessibility. If the user is logged in, the device reads logoutHint to the user.

src/hooks/useLoginProfile.ts
ardasnturk: Yeah, you're right. It can be done. We've done it before. But I have some fears about it. Like security. After all, I need clientId and secretKey here and if we get them from `globalSettings`, there may be a security problem, right? After all, this request comes to the app as a result of a request and these requests can be easily intercepted nowadays.

src/hooks/useLoginProfile.ts
```diff
@@ -0,0 +1,183 @@
...
+
+  const redirectUri = makeRedirectUri({
+    scheme: appJson.expo.scheme,
+    path: 'redirect'
+  });
+
+  const [request, response, promptAsync] = useAuthRequest(
+    {
+      clientId: CLIENT_ID,
+      codeChallengeMethod: CodeChallengeMethod.S256,
```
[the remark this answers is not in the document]
ardasnturk: It can be removed. I overlooked it.

src/hooks/useLoginProfile.ts
ardasnturk: Because it is used 1 time in the code, I did not feel the need to create an extra variable.

src/hooks/useLoginProfile.ts
ardasnturk: This only enables the checkToken function to work. checkToken checks whether there is a token stored in the device and whether this token is expired. If the token has expired, a new token is automatically generated and stored in the device. 

But yes, you are right, if we don't need the value returned from the function, then the async/await structure can be removed.

src/hooks/useLoginProfile.ts
```diff
@@ -0,0 +1,183 @@
...
+
+    try {
+      const parsedToken = JSON.parse(storedToken);
+
+      const now = Math.floor(Date.now() / 1000);
+      const expiresAt = parsedToken.issuedAt + parsedToken.expiresIn;
+
+      if (now >= expiresAt) {
+        // Token expired, try refresh
+        const refreshed = await refreshToken(parsedToken.refreshToken);
```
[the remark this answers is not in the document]
ardasnturk: I'm replacing it with `refreshedToken`.

src/hooks/useLoginProfile.ts
```diff
@@ -0,0 +1,183 @@
...
+  };
+
+  useEffect(() => {
+    (async () => {
+      await checkToken();
+    })();
+  }, []);
+
+  const checkToken = async () => {
+    const storedToken = await readFromStore(PROFILE_ACCESS_TOKEN);
```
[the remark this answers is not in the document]
ardasnturk: I'm replacing it with `storedTokenData`.

src/hooks/useLoginProfile.ts
```diff
@@ -0,0 +1,177 @@
...
+import * as appJson from '../../app.json';
+import { addToStore, readFromStore } from '../helpers';
+
+const PROFILE_ACCESS_TOKEN = 'profileAccessToken';
+
+type TProfile = {
+  serverUrl: string;
+  clientId: string;
+  clientSecret: string;
+  scopes: string[];
```
donni106, line 18: do we want them sorted alphabetically?

src/hooks/useLoginProfile.ts
```diff
@@ -0,0 +1,177 @@
...
+    tokenEndpoint: serverUrl + '/token',
+    revocationEndpoint: serverUrl + '/revoke'
+  };
+
+  const [isLoggedIn, setIsLoggedIn] = useState(false);
+  const [loading, setLoading] = useState(false);
+
+  const redirectUri = makeRedirectUri({
+    scheme: appJson.expo.scheme,
+    path: 'redirect'
```
donni106, line 35: do we want them sorted alphabetically?

src/hooks/useLoginProfile.ts
donni106: where do you know that from?

src/hooks/useLoginProfile.ts
```diff
@@ -0,0 +1,177 @@
...
+          redirectUri,
+          extraParams: {
+            grant_type: 'authorization_code'
+          }
+        },
+        keycloakDiscovery
+      );
+
+      setIsLoggedIn(true);
+      await addToStore(PROFILE_ACCESS_TOKEN, JSON.stringify(tokenResult));
```
donni106, line 76: shouln't we use the `SecureStore` with `SecureStore.setItemAsync(...)` for the token?

src/hooks/useLoginProfile.ts
```diff
@@ -0,0 +1,178 @@
...
+      if (!token?.idToken) return;
+
+      await fetch(`${serverUrl}/logout?id_token_hint=${token.idToken}`, {
+        method: 'GET',
+        headers: {
+          Authorization: `Bearer ${token.accessToken}`
+        }
+      });
+
+      await addToStore(PROFILE_ACCESS_TOKEN, '');
```
donni106, line 164: if not using secure store, this should be `removeFromStore`, right?

src/hooks/useLoginProfile.ts
ardasnturk: Yeah, you're right. I'll update it with `SecureStore`. 👍

src/hooks/useLoginProfile.ts
ardasnturk: Yes 🙌 But I'm already updating with `SecureStore` now, so I'll use `SecureStore.deleteItemAsync(...)`.

src/hooks/useLoginProfile.ts
ardasnturk: In my opinion, I see no reason why this code should not work. But if I update the code as you say, I get this error. Therefore, I think it should remain as it is now.

image

donni106: huh, those are the same things. weird. and nothing has to do with some `create`, right? can you share the git diff of that change?

ardasnturk: I only have this modification. I have no idea why the error message mentions `create`.

image


The issue was caused by passing an async function directly to useEffect like this: 
```
  useEffect(checkToken, []);
```
Since React doesn’t allow async functions as effect callbacks (because they return a Promise), it led to a runtime error.That's why we have to leave the code as it is.

src/hooks/useLoginProfile.ts
ardasnturk: Because the pkce value is off in keycloak. But I don't know why it's off.

image

You can see this setting by following the route below.
keycloak -> Identity providers -> ITEBO

donni106: https://oauth.net/2/pkce, `true` per default: https://docs.expo.dev/versions/latest/sdk/auth-session/#authrequestconfig

so if someone changes this in keycloak, we need a new app release with also an adjusted value. can we consider this to be part of the global settings?

ardasnturk: https://github.com/smart-village-solutions/smart-village-app-app/pull/1311/commits/ddabdcba94f62db4dbb1ad567f24403cc2e39ef8

src/hooks/useLoginProfile.ts
donni106: getting this now

> Value being stored in SecureStore is larger than 2048 bytes and it may not be stored successfully. In a future SDK version, this call may throw an error.

ardasnturk: We are currently using the `issueAt`, `expiresIn` and `refreshToken` values from the `tokenResult` object we saved in `SecureStore`. Different keys can be created for these 3 values and saved differently. We can also adapt the `token` value, which I think we will need in the future.

donni106: Only `refreshToken` needs to be put in `SecureStore`, right? The others can be in `AsyncStorage`. What do you think is the effort to refactor this?

ardasnturk: I developed this code a long time ago. Therefore, I must first carefully analyse the changes that need to be made. I estimate it will take about two hours.

src/hooks/useLoginProfile.ts
```diff
@@ -0,0 +1,204 @@
...
+    const storedRefreshToken = await SecureStore.getItemAsync(PROFILE_REFRESH_TOKEN);
+    const storedAccessMeta = await readFromStore(PROFILE_ACCESS_META);
+
+    if (!storedRefreshToken || !storedAccessMeta) {
+      setIsLoggedIn(false);
+      return null;
+    }
+
+    try {
+      const parsedTokenMeta = JSON.parse(storedAccessMeta);
```
donni106: now "access meta" is called "token meta"? please keep the namings in line.

src/hooks/useLoginProfile.ts
```diff
@@ -0,0 +1,204 @@
...
+          code,
+          extraParams: { grant_type: 'authorization_code' },
+          redirectUri
+        },
+        keycloakDiscovery
+      );
+
+      setIsLoggedIn(true);
+
+      const { refreshToken: refresh, issueAt, expiresIn } = tokenResult;
```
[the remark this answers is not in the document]
ardasnturk: I chose this because of the name conflict. However, I will fix it.

src/hooks/useLoginProfile.ts
```diff
@@ -0,0 +1,204 @@
...
+import { dismissAuthSession } from 'expo-web-browser';
+import { useEffect, useState } from 'react';
+import { Alert } from 'react-native';
+
+import * as appJson from '../../app.json';
+import { device, texts } from '../config';
+import { addToStore, readFromStore, removeFromStore } from '../helpers';
+
+const PROFILE_REFRESH_TOKEN = 'profileRefreshToken';
+const PROFILE_ACCESS_META = 'profileAccessMeta';
```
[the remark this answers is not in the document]
ardasnturk: I named it this way because the metadata returned from the server during login does not contain a token. I will update the naming. I don't like it this way.

I will change it to `PROFILE_TOKEN` and `PROFILE_META`.

src/hooks/useLoginProfile.ts
```diff
@@ -0,0 +1,204 @@
...
+    if (!storedRefreshToken || !storedAccessMeta) {
+      setIsLoggedIn(false);
+      return null;
+    }
+
+    try {
+      const parsedTokenMeta = JSON.parse(storedAccessMeta);
+
+      const now = Math.floor(Date.now() / 1000);
+      const expiresAt = parsedTokenMeta.issuedAt + parsedTokenMeta.expiresIn;
```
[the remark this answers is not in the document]
ardasnturk: You've made a very good point. I've made a naming error here. The name of the data stored in Asyncstorage should also be `issuedAt`.

donni106:
I think this should be merged with https://github.com/smart-village-solutions/smart-village-app-app/pull/1519 first.

PR #1314, opened 2025-04-23
feat(map): add pin to map

ardasnturk:
This PR adds the ability to add new pins to the map view in the Maplibre component

|without new pin|with new pin|with new pin without cluster|
|--|--|--|
Simulator Screenshot - iPhone 16 Plus - 2025-04-23 at 15 12 16|Simulator Screenshot - iPhone 16 Plus - 2025-04-23 at 15 12 22|Simulator Screenshot - iPhone 16 Plus - 2025-04-23 at 15 12 31

SVA-1365

### DIFF
--- a/src/components/map/MapLibre.tsx
+++ b/src/components/map/MapLibre.tsx
@@ -174,6 +174,19 @@ export const MapLibre = ({
 
   const [centerCoordinate] = useState([initialRegion.longitude, initialRegion.latitude]);
 
+  const handleMapPressToSetNewPin = (event: any) => {
+    const { geometry } = event;
+    if (!geometry) return;
+
+    if (geometry.coordinates) {
+      const newPin = point(geometry.coordinates, {
+        iconName: MAP.DEFAULT_PIN, // TODO: find a proper default icon for setting a pin
+        id: `new-pin-${Date.now()}`
+      });
+      setNewPins([newPin]);
+    }
+  };
+
   const handleOnPress = async (event: any) => {
     const feature = event.features[0];
     if (!feature) return;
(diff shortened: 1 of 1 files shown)


src/components/map/MapLibre.tsx
```diff
@@ -0,0 +1,392 @@
...
+  };
+
+  const handleMapPress = (event: any) => {
+    const { geometry } = event;
+    if (geometry?.coordinates) {
+      const newPin = point(geometry.coordinates, {
+        iconName: MAP.DEFAULT_PIN,
+        id: `${Date.now()}`
+      });
+      setNewPins([newPin]);
```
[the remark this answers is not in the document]
donni106: no, it should be always only the one pin from the last press

src/components/map/MapLibre.tsx
```diff
@@ -209,6 +222,7 @@ export const MapLibre = ({
         rotateEnabled={false}
         style={[styles.map, mapStyle]}
         zoomEnabled
+        onPress={!!onMapPress && handleMapPress}
```
donni106, line 225: this should not depend on `onMapPress`. we need some other flag to indicate, that setting a pin is allowed.
and if this is activated, other map press actions should not be applied i guess. imagine a map press would navigate somewhere, you would not want this to happen after setting a new pin.

ardasnturk: I added this control here because the `handleMapPress` function is useless without the `onMapPress` function. We need the changes made inside `onMapPress`. Also, if you look at the `handleMapPress` function, you can see that the `onMapPress` function is inside it. But I think that with the controls made in `handleMapPress`, we prevent the `onMapPress` control.

donni106: ok, in this case it should be handled inside `handleMapPress` and not when passing the prop.

donni106: i will add the flag `setPinEnabled` to determine the case

PR #1323, opened 2025-05-05
feat(sue): update pin on report screen

ardasnturk:
Added border to increase the accessibility of this PR and to make the pin added to the map visible

|before|after|before|after|
|--|--|--|--|
Simulator Screenshot - iPhone 16 Plus - 2025-05-05 at 11 12 18|Simulator Screenshot - iPhone 16 Plus - 2025-05-05 at 11 11 50|Simulator Screenshot - iPhone 16 Plus - 2025-05-05 at 11 12 23|Simulator Screenshot - iPhone 16 Plus - 2025-05-05 at 11 11 55

SUE-165

### DIFF
--- a/src/components/SUE/report/SueReportLocation.tsx
+++ b/src/components/SUE/report/SueReportLocation.tsx
@@ -28,6 +28,7 @@ import { Map } from '../../map';
 import { getLocationMarker } from '../../settings';
 
 const { a11yLabel, INPUT_KEYS } = consts;
+export const SELECTED_MARKER_ID = 'selectedMarkerId';
 
 export const locationServiceEnabledAlert = ({
   currentPosition,
@@ -218,7 +219,10 @@ export const SueReportLocation = ({
   let mapCenterPosition = {} as { latitude: number; longitude: number };
 
   if (selectedPosition) {
-    locations = [...mapMarkers, { ...baseLocationMarker, position: selectedPosition }];
+    locations = [
+      ...mapMarkers,
+      { ...baseLocationMarker, position: selectedPosition, id: SELECTED_MARKER_ID }
+    ];
   }
 
   if (alternativePosition) {
@@ -315,6 +319,7 @@ export const SueReportLocation = ({
             navigation.navigate(ScreenName.SueReportMapView, {
               calloutTextEnabled: true,
               clusteringEnabled: true,
+              selectedMarker: selectedPosition ? SELECTED_MARKER_ID : undefined,
               currentPosition,
               isMyLocationButtonVisible: !!locationService,
               locations,
@@ -325,6 +330,7 @@ export const SueReportLocation = ({
               showsUserLocation: true
             })
           }
+          selectedMarker={selectedPosition ? SELECTED_MARKER_ID : undefined}
           updatedRegion={
             !!selectedPosition && (updatedRegion || updateRegionFromImage)
               ? { ...selectedPosition, latitudeDelta: 0.01, longitudeDelta: 0.01 }
--- a/src/components/map/Map.tsx
+++ b/src/components/map/Map.tsx
@@ -331,7 +331,7 @@ export const Map = ({
                 />
               )}
 
-              {calloutTextEnabled && (
+              {calloutTextEnabled && (!!serviceName || !!title) && (
                 <Callout style={styles.callout}>
                   {!!serviceName && (
                     <BoldText smallest center>
--- a/src/screens/SUE/SueMapViewScreen.tsx
+++ b/src/screens/SUE/SueMapViewScreen.tsx
@@ -1,11 +1,16 @@
 import { StackNavigationProp } from '@react-navigation/stack';
 import * as Location from 'expo-location';
-import React, { useEffect, useState } from 'react';
+import React, { useMemo, useState } from 'react';
 import { Alert, StyleSheet } from 'react-native';
 
-import { Map, locationServiceEnabledAlert } from '../../components';
+import { Map } from '../../components';
+import {
+  SELECTED_MARKER_ID,
+  locationServiceEnabledAlert
+} from '../../components/SUE/report/SueReportLocation';
 import { colors, normalize, texts } from '../../config';
 import { useLocationSettings } from '../../hooks';
+import { MapMarker } from '../../types';
 
 export const SueMapViewScreen = ({
   navigation,
@@ -25,6 +30,7 @@ export const SueMapViewScreen = ({
     isMaximizeButtonVisible,
     isMyLocationButtonVisible,
     locations,
+    selectedMarker,
     mapCenterPosition,
     onMapPress,
     onMarkerPress,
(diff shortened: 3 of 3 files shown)


src/components/SUE/report/SueReportLocation.tsx
```diff
@@ -218,7 +218,10 @@ export const SueReportLocation = ({
   let mapCenterPosition = {} as { latitude: number; longitude: number };
 
   if (selectedPosition) {
-    locations = [...mapMarkers, { ...baseLocationMarker, position: selectedPosition }];
+    locations = [
+      ...mapMarkers,
+      { ...baseLocationMarker, position: selectedPosition, id: 'selected' }
+    ];
   }
```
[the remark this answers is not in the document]
donni106: my question would be, why the `id` is hardcoded to the string `'selected'` here? why is that needed to be a string? aren't we expecting ids to be something different?

src/components/map/Map.tsx
```diff
@@ -331,18 +331,16 @@ export const Map = ({
                 />
               )}
 
-              {calloutTextEnabled && (
+              {calloutTextEnabled && title && (
```
donni106, line 334: i cannot understand this. please read out loud in real language: "if callout text is enabled and there is a title, render a callout, that probably has a service name but always a title"

shouldn't it be more like: "if callout text is enabled and there is a title or a service name, render a callout with a service name a title or both"?

src/screens/SUE/SueMapViewScreen.tsx
```diff
@@ -36,18 +38,17 @@ export const SueMapViewScreen = ({
...
-        item?.iconName === iconName ? { iconName, position: selectedPosition } : item
-      )
-    );
-  }, [selectedPosition]);
+  if (selectedPosition) {
+    locationsWithPin = [
+      ...locations.filter((item) => item.id !== 'selected'), // Remove old pin
+      { iconName, position: selectedPosition, id: 'selected' } // Add new pin
+    ];
+  }
```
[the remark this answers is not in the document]
donni106: could we use `useMemo` with dendency to `selectedPosition` to not always re-loop and recalculate the locations object?

src/components/map/Map.tsx
ardasnturk: I only added title control because we have title value in other parts of the app but we have serviceName value only in Sue module. Therefore, I thought of adding only title control to make it work smoothly in other parts of the app. But what you're saying makes more sense. I will update it this way:
```
{calloutTextEnabled && (!!title || !!serviceName)...
```

Yes, you are right, I need to convert it to a boolean expression.

src/components/SUE/report/SueReportLocation.tsx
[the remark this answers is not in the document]
ardasnturk: Here we add an extra pin to the map. The id value of this pin is undefined. But if we want to show it as an active pin, the ids in `marker.id` and `selectedMarker` values must match. I think it is more logical to create a constant and update all id values with this constant.

src/components/SUE/report/SueReportLocation.tsx
```diff
@@ -28,6 +28,7 @@ import { Map } from '../../map';
 import { getLocationMarker } from '../../settings';
 
 const { a11yLabel, INPUT_KEYS } = consts;
+export const SELECTED_MARKER_ID = 'selected';
```
donni106, line 31: what about `selectedMarkerId` as value?

src/components/map/Map.tsx
```diff
@@ -331,18 +331,16 @@ export const Map = ({
...
                     </BoldText>
                   )}
-                  {!!title && (
-                    <RegularText smallest center>
-                      {title}
-                    </RegularText>
-                  )}
+                  <RegularText smallest center>
+                    {title}
+                  </RegularText>
```
donni106, line 343: check for title presence is needed again

src/components/map/Map.tsx
```diff
@@ -331,18 +331,16 @@ export const Map = ({
                 />
               )}
 
-              {calloutTextEnabled && (
+              {calloutTextEnabled && (!!title || !!serviceName) && (
```
donni106, line 334: as the service name is rendered before the title, the check should be the same for better reaability and context

src/screens/SUE/SueMapViewScreen.tsx
```diff
@@ -1,11 +1,12 @@
 import { StackNavigationProp } from '@react-navigation/stack';
 import * as Location from 'expo-location';
-import React, { useEffect, useState } from 'react';
+import React, { useState } from 'react';
 import { Alert, StyleSheet } from 'react-native';
 
-import { Map, locationServiceEnabledAlert } from '../../components';
+import { Map, SELECTED_MARKER_ID, locationServiceEnabledAlert } from '../../components';
```
donni106, line 6: i need to think about importing the constant from components...hm..it is not a component...same as the alert. we have a bad structure here.
maybe you can import it explicitly from the the file where it is exported from? the same for `locationServiceEnabledAlert`

src/screens/SUE/SueMapViewScreen.tsx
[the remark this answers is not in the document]
ardasnturk: Yeah, I'll do that. 👌

src/screens/SUE/SueMapViewScreen.tsx
ardasnturk: https://github.com/smart-village-solutions/smart-village-app-app/commit/d81789c4260af837113bb002f8dbd1334b778ffd

src/screens/SUE/SueMapViewScreen.tsx
```diff
@@ -36,17 +42,19 @@ export const SueMapViewScreen = ({
...
-    setLocationsWithPin((prevData) =>
-      prevData.map((item) =>
-        item?.iconName === iconName ? { iconName, position: selectedPosition } : item
-      )
-    );
+  const locationsWithPin = useMemo(() => {
+    if (selectedPosition) {
+      return [
+        ...locations.filter((item) => item.id !== SELECTED_MARKER_ID), // Remove old pin
+        { iconName, position: selectedPosition, id: SELECTED_MARKER_ID } // Add new pin
```
donni106, line 53: with seeing this again, why are we filtering all but the one and adding the one again? couldn't we update the position of the selected marker inside the locations?

ardasnturk: Directly mutating the position like `locations.find(...).position = selectedPosition` won’t trigger a re-render because React won’t detect the change if the reference to the locations array stays the same. Using `useMemo` with a new array ensures immutability and proper reactivity, so the component updates correctly.

donni106: ok, too bad, but I thought so

donni106 (approved):

PR #1326, opened 2025-05-12
fix(map): the problem of not being able to add pins to the map

ardasnturk:
This PR solves the problem of not being able to add pins to the map and not being able to remove the added pin

SVA-1365

### DIFF
--- a/src/components/map/MapLibre.tsx
+++ b/src/components/map/MapLibre.tsx
@@ -64,19 +64,25 @@ type Props = {
     scrollEnabled: boolean;
     zoomEnabled: boolean;
   };
-  isLocationSelectable?: boolean;
   isMultipleMarkersMap?: boolean;
   isMyLocationButtonVisible?: boolean;
   locations: MapMarker[];
   mapCenterPosition?: LocationObjectCoords;
   mapStyle?: StyleProp<ViewStyle>;
   minZoom?: number;
-  onMapPress?: ({ nativeEvent }: { nativeEvent?: any }) => void;
+  onMapPress?: ({
+    geometry: { coordinates }
+  }: {
+    geometry: {
+      coordinates: number[];
+    };
+  }) => Promise<{ isLocationSelectable?: boolean }>;
   onMarkerPress?: (arg0?: string) => void;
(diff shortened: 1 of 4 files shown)


src/components/map/MapLibre.tsx
```diff
@@ -58,14 +58,17 @@ type Props = {
...
   locations: MapMarker[];
   mapCenterPosition?: LocationObjectCoords;
   mapStyle?: StyleProp<ViewStyle>;
   minZoom?: number;
-  onMapPress?: ({ nativeEvent }: { nativeEvent?: any }) => void;
+  onMapPress?: (arg0: {
+    geometry: {
+      coordinates: number[];
+    };
+  }) => Promise<{ isLocationSelectable?: boolean }>;
```
donni106, line 71: `arg0` is weird, shouldn't this work?

```
onMapPress?: ({
    geometry: {
      coordinates: number[];
    };
  }) => Promise<{ isLocationSelectable?: boolean }>;
```

src/components/map/MapLibre.tsx
```diff
@@ -193,14 +195,16 @@ export const MapLibre = ({
...
-  const handleMapPressToSetNewPin = (event: any) => {
+  const handleMapPressToSetNewPin = async (event: any) => {
     const { geometry } = event;
     if (!geometry) return;
 
     const coordinates = geometry.coordinates as number[];
     if (!coordinates?.length) return;
 
-    if (isLocationSelectable === undefined || isLocationSelectable) {
+    const { isLocationSelectable = false } = (await onMapPress?.(event)) ?? {};
```
donni106, line 205: please call `await onMapPress({ geometry })` like in SueMapViewScreen.tsx

src/components/map/MapLibre.tsx
ardasnturk: I've updated what you sent a bit. It's more accurate this way. 

```
  onMapPress?: ({
    geometry: { coordinates }
  }: {
    geometry: {
      coordinates: number[];
    };
  }) => Promise<{ isLocationSelectable?: boolean }>;
```

donni106 (approved):

PR #1327, opened 2025-05-12
fix(map): the problem of not updating the added pin when switching between pages

ardasnturk:
This PR ensures that the pin added/updated in the large map view is updated in the small map view.


donni106 (approved):

src/components/SUE/report/SueReportLocation.tsx
```diff
@@ -160,6 +169,13 @@ export const SueReportLocation = ({
     [data]
   );
 
+  // TODO: do we need this?
+  // useFocusEffect(() => {
+  //   if (selectedPosition) {
+  //     setSelectedPosition(selectedPosition);
+  //   }
+  // });
```
donni106, line 177: don't miss this, so I add a comment here

ardasnturk: No. We don't need it anymore because we don't switch to another screen for the big map view. We do everything from one map.

src/components/SUE/report/SueReportLocation.tsx
```diff
@@ -159,6 +159,12 @@ export const SueReportLocation = ({
     [data]
   );
 
+  useFocusEffect(() => {
+    if (selectedPosition) {
+      setSelectedPosition(selectedPosition);
```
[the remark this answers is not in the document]
ardasnturk: We don't `useFocusEffect` here anymore. It is not in our code. So I ignore this comment.

PR #1328, opened 2025-05-12
feat(map): return enhancements and align several code spots

donni106:
SVA-1365

### DIFF
--- a/src/components/map/MapLibre.tsx
+++ b/src/components/map/MapLibre.tsx
@@ -64,19 +64,25 @@ type Props = {
     scrollEnabled: boolean;
     zoomEnabled: boolean;
   };
-  isLocationSelectable?: boolean;
   isMultipleMarkersMap?: boolean;
   isMyLocationButtonVisible?: boolean;
   locations: MapMarker[];
   mapCenterPosition?: LocationObjectCoords;
   mapStyle?: StyleProp<ViewStyle>;
   minZoom?: number;
-  onMapPress?: ({ nativeEvent }: { nativeEvent?: any }) => void;
+  onMapPress?: ({
+    geometry: { coordinates }
+  }: {
+    geometry: {
+      coordinates: number[];
+    };
+  }) => Promise<{ isLocationSelectable?: boolean }>;
   onMarkerPress?: (arg0?: string) => void;
   onMaximizeButtonPress?: () => void;
-  onMyLocationButtonPress?: () => void;
+  onMyLocationButtonPress?: ({ isFullScreenMap }: { isFullScreenMap?: boolean }) => void;
   selectedMarker?: string;
   selectedPosition?: LocationObjectCoords;
+  setPinEnabled?: boolean;
   showsUserLocation?: boolean;
   style?: StyleProp<ViewStyle>;
   updatedRegion?: Region;
@@ -87,7 +93,6 @@ export const MapLibre = ({
   calloutTextEnabled = false,
   clusterDistance,
   geometryTourData,
-  isLocationSelectable,
   isMultipleMarkersMap = true,
   isMyLocationButtonVisible = true,
   locations,
@@ -100,6 +105,7 @@ export const MapLibre = ({
   onMyLocationButtonPress,
   selectedMarker = '',
   selectedPosition,
+  setPinEnabled,
   style,
   ...otherProps
 }: Props) => {
@@ -183,6 +189,8 @@ export const MapLibre = ({
   }, [mapReady, loading, isMultipleMarkersMap, locations, selectedMarker]);
 
   useEffect(() => {
+    if (!setPinEnabled) return;
+
     if (!selectedPosition) {
       setNewPins([]);
       return;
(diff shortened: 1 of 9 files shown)


src/components/map/MapLibre.tsx
```diff
@@ -244,11 +267,10 @@ export const MapLibre = ({
         compassEnabled={false}
         mapStyle="https://tileserver-gl.smart-village.app/styles/osm-liberty/style.json"
         ref={mapRef}
-        rotateEnabled={false}
         style={[styles.map, mapStyle]}
-        zoomEnabled
-        onPress={handleMapPressToSetNewPin}
         onDidFinishLoadingMap={() => setMapReady(true)}
+        onPress={handleMapPress}
+        {...otherProps.interactivity}
```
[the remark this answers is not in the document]
donni106: no, we ensure that this will not happen.

src/components/map/MapLibre.tsx
```diff
@@ -368,7 +390,7 @@ export const MapLibre = ({
         )}
       </MapView>
 
-      {isMyLocationButtonVisible && showsUserLocation && (
+      {isMyLocationButtonVisible && showsUserLocation && !onMapPress && (
```
ardasnturk, line 393: Why did you add the onMapPress control here? I think we should give the user the ability to choose their own location as well as the `onMapPress` feature.

src/components/screens/PointOfInterest.js
```diff
@@ -210,11 +210,20 @@ export const PointOfInterest = ({ data, hideMap, navigation, route }) => {
                 [category.iconName || MAP.DEFAULT_PIN]: 1,
                 iconName: category.iconName || MAP.DEFAULT_PIN,
                 id,
-                position: { latitude, longitude }
+                position: { latitude, longitude },
+                serviceName: title,
+                title
```
ardasnturk, line 215: Do we really need these two properties? After all, `serviceName` and `title` properties are valid for CalloutText. But we do not use CalloutText in the POI map.

src/components/screens/PointOfInterest.js
```diff
@@ -210,11 +210,20 @@ export const PointOfInterest = ({ data, hideMap, navigation, route }) => {
...
               }
             ]}
-            selectedMarker={id}
             mapStyle={styles.mapStyle}
+            interactivity={{
+              pitchEnabled: false,
+              rotateEnabled: false,
+              scrollEnabled: false,
+              zoomEnabled: false
+            }}
```
ardasnturk, line 224: We can't scroll the map, we can't zoom in, but we can show our own position. Is that right? If I am too far away and once I press the show my location button, I can no longer go back to the location that shows the location of the POI.

image

src/components/screens/PointOfInterest.js
```diff
@@ -210,11 +210,20 @@ export const PointOfInterest = ({ data, hideMap, navigation, route }) => {
...
+            isMultipleMarkersMap={false}
+            selectedMarker={id}
```
ardasnturk, line 226: I don't know the exact cause of the problem, but in some POI categories (Gastronomie) the app crashes and I get this error.

Simulator Screenshot - iPhone 16 Plus - 2025-05-13 at 09 29 26

src/components/screens/PointOfInterest.js
```diff
@@ -210,11 +210,20 @@ export const PointOfInterest = ({ data, hideMap, navigation, route }) => {
...
                 id,
-                position: { latitude, longitude }
+                position: { latitude, longitude },
+                serviceName: title,
+                title
               }
             ]}
-            selectedMarker={id}
             mapStyle={styles.mapStyle}
+            interactivity={{
```
ardasnturk, line 219: I can add new pins to the POI map. This shouldn't happen. I added the red marked pin in the screenshot myself.

Bildschirmfoto 2025-05-13 um 09 32 17

ardasnturk (changes requested):

src/components/screens/PointOfInterest.js
ardasnturk: I added `onMapPress` control to `handleMapPressToSetNewPin` function and the problem is solved.

src/components/screens/PointOfInterest.js
ardasnturk: I added the `isMyLocationButtonVisible` prop to the component as `false` and now the button is not visible.

src/components/screens/PointOfInterest.js
ardasnturk: This section has been removed as it was unnecessary.

src/components/screens/PointOfInterest.js
```diff
@@ -210,9 +210,7 @@ export const PointOfInterest = ({ data, hideMap, navigation, route }) => {
                 [category.iconName || MAP.DEFAULT_PIN]: 1,
                 iconName: category.iconName || MAP.DEFAULT_PIN,
                 id,
-                position: { latitude, longitude },
-                serviceName: title,
-                title
```
donni106: we need them only for callouts and we have no callouts in poi detail screen, that is why we can remove them here?

ardasnturk: I don't quite understand the question. I already removed them because we don't need CalloutText, so why are you asking if we can remove them again?

donni106: the question was, why we can remove them and i tried to answer it myself. do you agree with "because it is just for callouts and we have no callouts at POI details"?

ardasnturk: Yes, absolutely. https://github.com/smart-village-solutions/smart-village-app-app/pull/1328#discussion_r2086113956 This is exactly what I was trying to say in the comment here.

PR #1330, opened 2025-05-16
feat(waste): export selected garbage types

ardasnturk:
This PR allows the export of selected waste calendar types.

SVA-1539


src/hooks/waste.tsx
```diff
@@ -287,44 +287,54 @@ export const useRenderSuggestions = (selectionCallback?: () => void) => {
...
 
-    if (city) {
-      params += `&city=${encodeURIComponent(city)}`;
-    }
+      if (!!city) {
+        params += `&city=${encodeURIComponent(city)}`;
+      }
 
-    const combinedUrl = baseUrl + params;
+      if (!!selectedTypes) {
```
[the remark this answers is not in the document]
donni106: same for `if (!!zip) {` and `if (!!city) {`

src/screens/WasteCollectionScreen.js
```diff
@@ -441,7 +445,11 @@ export const WasteCollectionScreen = ({ navigation, route }) => {
           ]}
         >
           <Wrapper noPaddingBottom>
-            <Button title={wasteTexts.exportButton} notFullWidth onPress={triggerExport} />
+            <Button
+              title={wasteTexts.exportButton}
+              notFullWidth
+              onPress={() => triggerExport({ selectedTypes })}
```
donni106, line 451: i guess that we can use `waste.selectedTypeKeys` here to avoid `Object.keys(selectedTypes).filter((key) => selectedTypes[key])` in `useTriggerExport` hook.

src/hooks/waste.tsx
```diff
@@ -287,44 +287,54 @@ export const useRenderSuggestions = (selectionCallback?: () => void) => {
...
+        params += `&waste_types=${Object.keys(selectedTypes)
+          .filter((key) => selectedTypes[key])
+          .map((key) => key.replace(/_/g, ''))
```
donni106, line 309: what is `key.replace(/_/g, ''))` needed for?

src/screens/WasteCollectionScreen.js
ardasnturk: Yes, that's right. I didn't see that. I think it's better now.

src/hooks/waste.tsx
ardasnturk: This was an unnecessary code. It is now removed and we don't use it anymore.

PR #1334, opened 2025-06-04
fix(service-tiles): prevent last row tiles from appearing too close t…

ardasnturk:
With this PR, it solves the problem of service tiles that are not fully placed on the screen.

|before|after|before|after|
|--|--|--|--|
Simulator Screenshot - iPhone 16 Plus - 2025-06-04 at 15 16 38|Simulator Screenshot - iPhone 16 Plus - 2025-06-04 at 15 15 47|Simulator Screenshot - iPhone 16 Plus - 2025-06-04 at 15 16 41|Simulator Screenshot - iPhone 16 Plus - 2025-06-04 at 15 15 50

### DIFF
--- a/src/components/screens/Service.tsx
+++ b/src/components/screens/Service.tsx
@@ -4,6 +4,7 @@ import React, { useCallback, useContext } from 'react';
 import { StyleSheet, TouchableOpacity, View } from 'react-native';
 
 import { colors, consts, normalize, texts } from '../../config';
+import { ConfigurationsContext } from '../../ConfigurationsProvider';
 import { usePersonalizedTiles } from '../../hooks';
 import { OrientationContext } from '../../OrientationProvider';
 import { SettingsContext } from '../../SettingsProvider';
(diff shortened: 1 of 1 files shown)


src/components/screens/Service.tsx
```diff
@@ -123,9 +128,20 @@ export const Service = ({
...
-      <WrapperWrap spaceAround={shouldCenter} spaceBetween={!shouldCenter}>
-        {tiles?.map(renderItem)}
-      </WrapperWrap>
+      {rows.map((row, index) => {
+        const isLastRow = index === rows.length - 1;
+        const isIncompleteRow = row.length < itemsPerRow;
+
+        return (
+          <WrapperWrap
+            key={index}
```
donni106, line 137: we learned somwhen, that the index ist not the best choice for a key. please think of something more unique here.

ardasnturk: I updated the key with `tile.title || accessibilityLabel`.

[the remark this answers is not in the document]
ardasnturk: I updated the key with `tile.title || accessibilityLabel`

PR #1342, opened 2025-06-25
feat: add button to carousel

ardasnturk:
This PR gives the button the invert button property with white border and adds the ability to add multiple buttons to the home carousel

If we want to show a button, the button object can be continued. But to add more than one button, buttons array should be added.

before: 
```
[
  {
   ...
    "button": {
      "style": {
        "big": false,
        "disabled": false,
        "iconColor": "white",
        "iconPosition": "right",
        "invert": false,
        "noFullWidth": false,
        "small": false,
        "smallest": false
      },
      "routeName": "SueReport",
      "iconName": "Plus",
      "title": "Etwas melden",
      "params": {}
    }
  }
]
```

after:
```
[
  {
    ...
    "buttons": [
      {
        "style": {
          "big": false,
          "disabled": false,
          "iconColor": "white",
          "iconPosition": "right",
          "invert": false,
          "noFullWidth": false,
          "small": false,
          "smallest": false
        },
        "routeName": "SueReport",
        "iconName": "Plus",
        "title": "Etwas melden",
        "params": {}
      },
      {
        "style": {
          "big": false,
          "disabled": false,
          "iconColor": "white",
          "iconPosition": "right",
          "invert": true,
          "lightest": true,
          "noFullWidth": false,
          "small": false,
          "smallest": false
        },
        "routeName": "SueList",
        "iconName": "ArrowRight",
        "title": "Meldungen ansehen",
        "params": {
          "viewType": "map",
          "title": "Meldungskarte"
        }
      }
    ]
  }
]
```

|before|after|
|--|--|
Simulator Screenshot - iPhone 16 Plus - 2025-06-25 at 12 39 44|Simulator Screenshot - iPhone 16 Plus - 2025-06-25 at 12 38 19

SUE-167

### DIFF
--- a/src/components/Button.js
+++ b/src/components/Button.js
@@ -15,12 +15,12 @@ export const Button = ({
   icon,
   iconPosition = 'right',
   invert = false,
+  lightest = false,
   notFullWidth = false,
   onPress,
   small,
   smallest,
-  title,
-  whiteInvert = false
+  title
 }) => {
   const { orientation, dimensions } = useContext(OrientationContext);
   const needLandscapeStyle =
@@ -62,7 +62,7 @@ export const Button = ({
         big && styles.bigTitle,
         small && styles.smallTitle,
         smallest && styles.smallestTitle,
-        whiteInvert && styles.titleWhiteInvert
+        lightest && styles.titleLightest
       ]}
       disabledStyle={styles.buttonDisabled}
       disabledTitleStyle={styles.title}
@@ -74,10 +74,10 @@ export const Button = ({
         big && [styles.bigButton, styles.bigButtonRadius],
         small && [styles.smallButton, styles.smallButtonRadius],
         smallest && [styles.smallestButton, styles.smallestButtonRadius],
-        whiteInvert && [styles.buttonInvert, styles.whiteInvert]
+        lightest && styles.lightestBorderColor
       ]}
       containerStyle={[styles.container, needLandscapeStyle && styles.containerLandscape]}
-      ViewComponent={invert || whiteInvert || isDelete || disabled ? undefined : DiagonalGradient}
+      ViewComponent={invert || isDelete || disabled ? undefined : DiagonalGradient}
       useForeground={!invert}
       accessibilityLabel={`${title} ${consts.a11yLabel.button}`}
       disabled={disabled}
@@ -150,6 +150,9 @@ const styles = StyleSheet.create({
     marginLeft: normalize(-14),
     paddingRight: normalize(14)
   },
+  lightestBorderColor: {
+    borderColor: colors.surface
+  },
   rejectButton: {
     backgroundColor: colors.error
   },
@@ -183,11 +186,8 @@ const styles = StyleSheet.create({
   titleLandscape: {
     paddingHorizontal: normalize(14)
   },
-  titleWhiteInvert: {
-    color: colors.surface
-  },
-  whiteInvert: {
-    borderColor: colors.surface
+  titleLightest: {
+    color: colors.lightestText
   }
 });
 
@@ -197,10 +197,10 @@ Button.propTypes = {
   icon: PropTypes.node,
   iconPosition: PropTypes.string,
   invert: PropTypes.bool,
+  lightest: PropTypes.bool,
   notFullWidth: PropTypes.bool,
   onPress: PropTypes.func.isRequired,
   small: PropTypes.bool,
   smallest: PropTypes.bool,
-  title: PropTypes.string.isRequired,
-  whiteInvert: PropTypes.bool
+  title: PropTypes.string.isRequired
 };
--- a/src/components/Button.js
+++ b/src/components/Button.js
@@ -19,7 +19,8 @@ export const Button = ({
   onPress,
   small,
   smallest,
-  title
+  title,
+  whiteInvert = false
 }) => {
   const { orientation, dimensions } = useContext(OrientationContext);
   const needLandscapeStyle =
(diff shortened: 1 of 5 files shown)


src/components/Image.js
```diff
@@ -133,6 +134,8 @@ export const Image = ({
           <View style={styles.contentContainerStyle}>
             {!!message && <ImageMessage message={message} />}
             {!!button && <ImageButton button={button} />}
+            {!!buttons?.length &&
+              buttons.map((button, index) => <ImageButton key={index} button={button} />)}
```
donni106, line 138: we do not want to use just index for react component keys. can you find a better solution please

src/components/Button.js
```diff
@@ -179,6 +182,12 @@ const styles = StyleSheet.create({
   },
   titleLandscape: {
     paddingHorizontal: normalize(14)
+  },
+  titleWhiteInvert: {
+    color: colors.surface
```
donni106, line 187: for texts, we use `lightestText` and not `surface`

src/components/Button.js
```diff
@@ -19,7 +19,8 @@ export const Button = ({
   onPress,
   small,
   smallest,
-  title
+  title,
+  whiteInvert = false
```
donni106, line 23: i have difficulties to use this prop. this seems not very sustainable to me. hat if the surface is a different color than white? than this prop is wrong. `white` should not be part of the prop name.
couldn't we use the existing `invert` prop with an additional `lightest` or something like that?

src/components/Image.js
ardasnturk: Yes, I will definitely find a better solution.

src/components/Button.js
```diff
@@ -71,10 +73,13 @@ export const Button = ({
...
         big && [styles.bigButton, styles.bigButtonRadius],
         small && [styles.smallButton, styles.smallButtonRadius],
-        smallest && [styles.smallestButton, styles.smallestButtonRadius]
+        smallest && [styles.smallestButton, styles.smallestButtonRadius],
+        invert && lightest && [styles.buttonInvert, styles.lightestInvert]
       ]}
       containerStyle={[styles.container, needLandscapeStyle && styles.containerLandscape]}
-      ViewComponent={invert || isDelete || disabled ? undefined : DiagonalGradient}
+      ViewComponent={
+        invert || (invert && lightest) || isDelete || disabled ? undefined : DiagonalGradient
```
donni106, line 81: i think this is not needed. if `invert` should be enough, right?

src/components/Button.js
```diff
@@ -60,7 +61,8 @@ export const Button = ({
         needLandscapeStyle && styles.titleLandscape,
         big && styles.bigTitle,
         small && styles.smallTitle,
-        smallest && styles.smallestTitle
+        smallest && styles.smallestTitle,
+        invert && lightest && styles.titleLightestInvert
```
donni106, line 65: why checking `invert` again here? `invert` and `lightest` should be seen independent of each other. if we use both props, both will be applied. so `lightest` should work like all the others with `lightestTitle` here.

src/components/Button.js
ardasnturk: yes

src/components/Button.js
```diff
@@ -71,7 +73,8 @@ export const Button = ({
         !invert && isDelete && styles.rejectButton,
         big && [styles.bigButton, styles.bigButtonRadius],
         small && [styles.smallButton, styles.smallButtonRadius],
-        smallest && [styles.smallestButton, styles.smallestButtonRadius]
+        smallest && [styles.smallestButton, styles.smallestButtonRadius],
+        invert && lightest && [styles.buttonInvert, styles.lightestInvert]
```
donni106, line 77: why checking `invert` again here? `invert` and `lightest` should be seen independent of each other. if we use both props, both will be applied. `styles.buttonInvert` is already there and `lightest` should work like all the others with explicit `lightestBorderColor` here.

src/components/Button.js
ardasnturk: I made a new force-commit and applied the changes you mentioned with this commit.

src/components/Button.js
ardasnturk: I made a new force-commit and applied the changes you mentioned with this commit.

src/components/Button.js
```diff
@@ -147,6 +150,10 @@ const styles = StyleSheet.create({
     marginLeft: normalize(-14),
     paddingRight: normalize(14)
   },
+
```
donni106, line 153: empty line accidently?

src/components/ImagesCarouselItem.js
```diff
@@ -13,7 +13,16 @@ import { Image } from './Image';
...
+    buttons,
+    containerStyle,
+    message,
+    navigation,
+    refreshInterval,
+    source
+  }) => {
     const { routeName: name, params } = source;
 
     if (name && params) {
```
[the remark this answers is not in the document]
ardasnturk: I don't think we need it, but it's a good suggestion. If the image in the Carousel is clickable, do we want to show an extra button? @donni106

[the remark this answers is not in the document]
ardasnturk: This is the situation here. We have two different returns in ImagesCarouselItem. Copilot suggests that we need to add the buttons prop to the Image component inside the clickable carousel. But I don't think we need that.

```
 if (name && params) {
      return (
        <TouchableOpacity
          accessibilityLabel={`${
            source.captionText ? source.captionText : consts.a11yLabel.imageCarousel
          } ${consts.a11yLabel.button}`}
          onPress={() => navigation.navigate({ name, params })}
          activeOpacity={0.8}
        >
          <Image {...{ button, source, message, containerStyle, aspectRatio }} /> <--- Here!
        </TouchableOpacity>
      ...
```

src/components/Button.js
ardasnturk: yes

PR #1346, opened 2025-07-01
feat(settings): make list titles on the settings screen remotely upda…

ardasnturk:
This PR adds the ability to update the headers of the list elements in the settings page and the header with `globalSettings`.

How does it work?

This feature can be activated by adding the following object to the `settings` object in `globalSettings`. Keys are fixed.

```
  "settings": {
    "settingsScreenListItemTitles": {
      "augmentedRealitySettings": "AR-Einstellungen2",
      "listSettings": "App-Aussehen2",
      "locationSettings": "Standort2",
      "matomo": "Matomo Analytics2",
      "mowasRegionSettings": "MoWaS-Regionen2",
      "onboarding": "App-Einführung2",
      "permanentFilterSettings": "Datenquellen2",
      "pushNotifications": "Push-Benachrichtigungen2",
      "termsAndConditions": "Einwilligung zu Datenschutzhinweisen2"
    },
    ...
  },
```

|before|after|after with header title|
|--|--|--|
image|Simulator Screenshot - iPhone 16 Plus - 2025-07-01 at 09 00 09|Simulator Screenshot - iPhone 16 Plus - 2025-07-01 at 09 00 11

SVA-1626

### DIFF
--- a/src/screens/SettingsScreen.js
+++ b/src/screens/SettingsScreen.js
@@ -53,17 +53,19 @@ export const SETTINGS_SCREENS = {
   PERMANENT_FILTER: 'permanentFilterSettings'
 };
 
-const renderItem = ({ item, navigation, listsWithoutArrows }) => {
+/* eslint-disable complexity */
+const renderItem = ({ item, navigation, listsWithoutArrows, settingsScreenListItemTitles }) => {
   let component;
+  const title = settingsScreenListItemTitles[item];
 
   if (item === SETTINGS_SCREENS.LOCATION) {
     component = (
       <TextListItem
         item={{
           isHeadlineTitle: false,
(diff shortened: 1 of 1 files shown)


src/screens/SettingsScreen.js
```diff
@@ -207,7 +212,7 @@ export const SettingsScreen = ({ navigation, route }) => {
         settingsList.push({
           data: [
             {
-              title: texts.settingsTitles.analytics,
+              title: settingsScreenListItemTitles.matomo || texts.settingsTitles.analytics,
```
donni106, line 215: why a different naming here with `matomo` instead of `analytics`? would be nice to keep it consistent with the others

donni106: i see, we already have inconsistencies there :(

PR #1354, opened 2025-07-18
feat: add `CitySelectionScreen` for city selection functionality

ardasnturk:
This PR is integrated with the city selection module.

|dropdown screen|dropdown screen|city tiles screen|
|--|--|--|
Simulator Screenshot - iPhone 16 Plus - 2025-07-18 at 12 25 55|Simulator Screenshot - iPhone 16 Plus - 2025-07-18 at 12 25 56|Simulator Screenshot - iPhone 16 Plus - 2025-07-18 at 12 26 00

SVA-1588


src/screens/CitySelectionScreen.tsx
```diff
@@ -0,0 +1,186 @@
...
+    if (city) {
+      await addToStore(SELECTED_CITY, city);
+      setStoredCity(city);
+    }
+  }, []);
+
+  const resetCity = useCallback(async () => {
+    await removeFromStore(SELECTED_CITY);
+    setStoredCity(null);
+  }, []);
```
donni106, line 47: it is the first time that the store is used per hook instead of helper, right? what is the benefit of this?

src/screens/CitySelectionScreen.tsx
```diff
@@ -0,0 +1,186 @@
...
+
+  return { storeCity, storedCity, loadStoredCity, loading, resetCity };
+};
+
+export const CitySelectionScreen = () => {
+  const { data: citiesData, loading: citiesLoading } = useStaticContent<Array<string>>({
+    refreshTimeKey: 'cities',
+    name: 'cities',
```
donni106, line 55: you do not need to set a `refreshTimeKey` if it the same as `name`

src/screens/CitySelectionScreen.tsx
```diff
@@ -0,0 +1,186 @@
...
+    type: 'json'
+  });
+
+  const { data: htmlContent, loading: htmlLoading } = useStaticContent({
+    refreshTimeKey: 'cityHtmlContent',
+    name: 'cityHtmlContent',
```
donni106, line 61: you do not need to set a `refreshTimeKey` if it the same as `name`

src/screens/CitySelectionScreen.tsx
```diff
@@ -0,0 +1,186 @@
...
+    type: 'html'
+  });
+
+  const initiallySelectedItem = {
+    id: 0,
+    index: 0,
+    value: 'z.B. Stuttgart',
```
donni106, line 68: the placeholder needs to be part of some server config

src/screens/CitySelectionScreen.tsx
```diff
@@ -0,0 +1,186 @@
...
+
+  const [selectedCity, setSelectedCity] = useState<string | null>(null);
+  const [contentName, setContentName] = useState<string | null>(null);
+  const { storeCity, loadStoredCity, storedCity, loading, resetCity } = useCitySelection();
+
+  useEffect(() => {
+    const setContentNameWithCity = async () => {
+      const city = await loadStoredCity();
+      if (city) {
+        setContentName(`city${city}`);
```
donni106, line 81: why do we need the prefix `city`? and wouldn't it be more correct to separate the value from the prefix with a `-` for example like `city-Augsburg`? and could it be clever to lowercase the city value to get `city-augsburg`? and hat about city names with space, shouldn't they get kebab-cased like `city-linz-am-rhein`?

src/screens/CitySelectionScreen.tsx
```diff
@@ -0,0 +1,186 @@
...
+  }, [citiesData]);
+
+  const onResetPress = useCallback(() => {
+    Alert.alert('Ort zurücksetzen', 'Möchten Sie den ausgewählten Ort wirklich zurücksetzen?', [
+      {
+        text: 'Abbrechen',
+        style: 'cancel'
+      },
+      {
+        text: 'OK',
```
donni106, line 121: texts to texts please

src/screens/CitySelectionScreen.tsx
```diff
@@ -0,0 +1,186 @@
...
+      }
+    ]);
+  }, [updateDropdownData]);
+
+  if (loading || htmlLoading || citiesLoading) {
+    return <LoadingSpinner loading />;
+  }
+
+  if (dropdownData?.length < 2) {
+    return <EmptyMessage title="Orte nicht verfügbar" />;
```
donni106, line 137: text to texts please

src/screens/CitySelectionScreen.tsx
```diff
@@ -0,0 +1,186 @@
...
+
+  if (!storedCity && !contentName) {
+    return (
+      <SafeAreaViewFlex>
+        <Wrapper>
+          <HtmlView html={htmlContent} />
+
+          <DropdownSelect
+            data={dropdownData}
+            placeholder="z.B. Stuttgart"
```
donni106, line 148: the placeholder needs to be part of some server config

src/screens/CitySelectionScreen.tsx
```diff
@@ -0,0 +1,186 @@
...
+  };
+  const [dropdownData, setDropdownData] = useState<DropdownProps[]>([initiallySelectedItem]);
+
+  const [selectedCity, setSelectedCity] = useState<string | null>(null);
+  const [contentName, setContentName] = useState<string | null>(null);
+  const { storeCity, loadStoredCity, storedCity, loading, resetCity } = useCitySelection();
+
+  useEffect(() => {
+    const setContentNameWithCity = async () => {
+      const city = await loadStoredCity();
```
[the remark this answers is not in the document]
ardasnturk: Because I created this hook and it not only fetches the city name from storage but also updates the `storeCity` state.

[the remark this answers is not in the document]
donni106: and why does this need a hook? the store can be fetched directly and the state handled as well?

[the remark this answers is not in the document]
ardasnturk: [84a55af](https://github.com/smart-village-solutions/smart-village-app-app/pull/1354/commits/84a55affdeb41989ae14b955e53a0c6b0e904c44)

src/screens/CitySelectionScreen.tsx
```diff
@@ -0,0 +1,146 @@
...
+        <WrapperRow itemsCenter>
+          <Title>{storedCity}</Title>
+
+          <Touchable onPress={onResetPress}>
+            <Icon.Pen color={colors.darkText} size={normalize(18)} style={styles.paddingLeft} />
+          </Touchable>
+        </WrapperRow>
+      </Wrapper>
+
+      {contentName && <ServiceTiles staticJsonName={contentName} />}
```
donni106, line 137: `!!string`!!!

src/screens/CitySelectionScreen.tsx
ardasnturk: I fixed this mess with the last commit. Now we have helper functions instead of hooks.

src/screens/CitySelectionScreen.tsx
```diff
@@ -0,0 +1,146 @@
...
+          text: texts.citySelection.alerts.ok,
+          onPress: async () => {
+            setSelectedCity(null);
+            await resetCity();
+            updateDropdownData();
+          }
+        }
+      ]
+    );
+  }, [updateDropdownData, resetCity]);
```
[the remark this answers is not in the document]
ardasnturk: I added it here because we use the resetCity function in the hook, but we don't need it. Because updating updatedDropdownData is enough for us.

src/helpers/citySelectionHelper.ts
```diff
@@ -0,0 +1,21 @@
...
+    await addToStore(SELECTED_CITY, city);
+    return city;
+  }
+  return null;
+};
+
+export const resetCity = async () => {
+  await removeFromStore(SELECTED_CITY);
+  return null;
+};
```
donni106, line 21: did you check our existing helpers?
have am look at `voucherHelper` for example:

```
export const storeVoucherMemberId = (memberId?: string) => {
  if (memberId) {
    addToStore(VOUCHER_MEMBER_ID, memberId);
  } else {
    removeFromStore(VOUCHER_MEMBER_ID);
  }
};
```

this is the way we used to implement adding and removing

src/helpers/citySelectionHelper.ts
```diff
@@ -0,0 +1,21 @@
...
+
+export const storeCity = async (city: string | null) => {
+  if (city) {
+    await addToStore(SELECTED_CITY, city);
+    return city;
+  }
+  return null;
+};
+
+export const resetCity = async () => {
```
donni106, line 18: i suggest `resetSelectedCity` to be more clear

src/helpers/citySelectionHelper.ts
```diff
@@ -0,0 +1,21 @@
+import { addToStore, readFromStore, removeFromStore } from '../helpers';
+
+export const SELECTED_CITY = 'selectedCity';
+
+export const loadStoredCity = async () => {
```
donni106, line 5: i suggest `storedSelectedCity` to be more clear. in other cases we have a check and fallback removal if something is not working with the given key, see `voucherHelper`:

```
export const voucherMemberId = async () => {
  let memberId;

  try {
    memberId = await readFromStore(VOUCHER_MEMBER_ID);
  } catch {
    // Token deleted here so that it can be recreated
    removeFromStore(VOUCHER_MEMBER_ID);
  }

  return memberId;
};
```

or as `SELECTED_CITY` is exported here, you could use `await readFromStore(SELECTED_CITY)` directly where needed?

ardasnturk: Yes. Previously we needed the loadStoredCity function because it updated the state in the helper. But now that we have moved it to `CitySelectionScreen`, we can use it directly with `readFromStore`.

donni106 (approved):

PR #1358, opened 2025-07-29
fix(waste): update suggestion rendering and loading states

donni106:
The changes correct some behaviors and reduce processes, that could be refactored in a more efficient wy.

SVASD-269

### DIFF
--- a/src/screens/WasteCollectionScreen.js
+++ b/src/screens/WasteCollectionScreen.js
@@ -271,83 +271,85 @@ export const WasteCollectionScreen = ({ navigation }) => {
               setSelectedStreetId={setSelectedStreetId}
             />
           </DefaultKeyboardAvoidingView>
-        ) : showCalendar ? (
-          <>
-            {wasteHeader()}
-            <CalendarListToggle showCalendar={showCalendar} setShowCalendar={setShowCalendar} />
-            <Wrapper>
-              <RegularText small>{wasteTexts.calendarIntro}</RegularText>
-            </Wrapper>
-            <RNCalendar
-              dayComponent={DayComponent}
-              firstDay={1}
-              markedDates={markedDates}
-              markingType="multi-dot"
-              onDayPress={onDayPress}
-              renderArrow={renderArrow}
-              theme={{
-                todayTextColor: colors.primary,
-                todayBackgroundColor: colors.lighterPrimaryRgba,
-                indicatorColor: colors.refreshControl,
-                dotStyle: {
-                  borderRadius: DOT_SIZE / 2,
-                  height: DOT_SIZE,
-                  marginBottom: normalize(8),
-                  marginTop: normalize(8),
-                  width: DOT_SIZE
-                }
-              }}
-            />
-
-            {!!hasExport && (
-              <View style={styles.paddingTop}>
-                <Wrapper style={styles.noPaddingBottom}>
-                  <Button
-                    title={wasteTexts.exportButton}
-                    notFullWidth
-                    onPress={() => triggerExport({ selectedTypes: waste.selectedTypeKeys })}
-                  />
-                </Wrapper>
-              </View>
-            )}
-            <Overlay
-              animationType="fade"
-              isVisible={isDayOverlayVisible}
-              onBackdropPress={() => setIsDayOverlayVisible(false)}
-              windowBackgroundColor={colors.overlayRgba}
-              overlayStyle={[styles.overlay, styles.overlayWidth]}
-              supportedOrientations={['portrait', 'landscape']}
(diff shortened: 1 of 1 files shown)


src/screens/WasteCollectionSettingsScreen.tsx
```diff
@@ -266,20 +273,6 @@ export const WasteCollectionSettingsScreen = () => {
...
-  useEffect(() => {
-    if (
-      !hasStoredStreetSettings.current &&
-      loadedStoredSettingsInitially &&
-      waste?.streetId !== selectedStreetId
-    ) {
-      setSelectedStreetId(selectedStreetId);
-      hasStoredStreetSettings.current = true;
-    }
-  }, [loadedStoredSettingsInitially, waste.streetId, selectedStreetId]);
```
donni106, line 281: `setSelectedStreetId(selectedStreetId);` made no sense. i think this condition never evaluated to true. but if it would, the state would be set with the current value of the state? strange. i did not encounter any issues after removal.

src/screens/WasteCollectionSettingsScreen.tsx
```diff
@@ -292,7 +285,7 @@ export const WasteCollectionSettingsScreen = () => {
     if (!!item && selectedStreetId !== item.id) {
       setSelectedStreetId(item.id);
     }
-  }, [addressesData, getStreetString, inputValue, selectedStreetId]);
```
donni106, line 295: set order of usage in effect

src/screens/WasteCollectionScreen.js
```diff
@@ -85,33 +83,29 @@ export const WasteCollectionScreen = ({ navigation }) => {
...
-    inputValueCity,
-    setInputValueCity,
-    setInputValueCitySelected
-  } = renderSuggestions;
+  const renderSuggestions = useRenderSuggestions((item) => {
+    if (item?.id) {
+      setSelectedStreetId(item.id);
+      setIsReset(false);
+      navigation.navigate(ScreenName.WasteCollectionSettings, { currentSelectedStreetId: item.id });
+    }
```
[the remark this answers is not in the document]
donni106: please have another look with my latest changes

src/screens/WasteCollectionSettingsScreen.tsx
```diff
@@ -244,6 +256,10 @@ export const WasteCollectionSettingsScreen = () => {
   useEffect(() => {
     if (usedTypes) {
       dispatch({ type: WasteSettingsActions.setInitialWasteSettings, payload: usedTypeKeys });
+      // Activate notifications if the user has allowed system permissions
+      handleSystemPermissions().then((permission) => {
+        if (permission) dispatch({ type: WasteSettingsActions.toggleNotifications });
+      });
```
ardasnturk, line 262: Is the `permission` value here a boolean expression?

src/screens/WasteCollectionScreen.js
```diff
@@ -185,7 +155,7 @@ export const WasteCollectionScreen = ({ navigation }) => {
 
     setSelectedTypes(
       Object.fromEntries(
-        (waste.selectedTypeKeys?.length ? waste.selectedTypeKeys : Object.keys(usedTypes)).map(
+        (waste.selectedTypeKeys ? waste.selectedTypeKeys : Object.keys(usedTypes)).map(
```
ardasnturk, line 158: I think `selectedTypeKeys` is not a boolean expression, so I suggest using the `!!` control.

src/screens/WasteCollectionSettingsScreen.tsx
donni106: yes

image

src/screens/WasteCollectionScreen.js
ardasnturk: As a result it is not a boolean expression. Here we are checking true/false. I think the !! check is necessary here, as we always do.

donni106: we are not checking true/false. we are checking if defined or undefined. where do you think is the boolean condition placed or needed?

src/screens/WasteCollectionScreen.js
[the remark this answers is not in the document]
ardasnturk: It's working.

ardasnturk (approved):

src/screens/WasteCollectionScreen.js
donni106: ok i guess, if `selectedTypeKeys` is an array, we can use it. otherwise it falls back to `Object.keys(usedTypes)`. seems legit.

PR #1367, opened 2025-08-13
feat(push): add server non response control

ardasnturk:
With this PR, it adds a warning to the user if there is no response from the server when turning the notification settings on or off

IMG_0312

SVA-1644

### DIFF
--- a/src/pushNotifications/TokenHandling.ts
+++ b/src/pushNotifications/TokenHandling.ts
@@ -1,7 +1,8 @@
 import * as SecureStore from 'expo-secure-store';
+import { Alert } from 'react-native';
 
 import * as appJson from '../../app.json';
-import { device, secrets } from '../config';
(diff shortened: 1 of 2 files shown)


src/pushNotifications/TokenHandling.ts
```diff
@@ -56,7 +57,14 @@ const removeTokenFromServer = async (token: string) => {
 
     // 204 means that it was a success on the server
     // 404 means that the token was already not on the server and can be treated as a success
-    return response.status === 204 || response.status === 404;
+    const isSuccess = response.status === 204 || response.status === 404;
+
+    if (!isSuccess) {
+      Alert.alert(texts.errors.errorTitle, texts.errors.noData);
+      return false;
```
donni106, line 64: this extra return is not needed. we will return  `isSuccess` anyways, which will be false in this case. i will remove it.

PR #1374, opened 2025-08-18
feat(switch): enhance disabled state handling for thumb and track colors

ardasnturk:
Updates the disable style of the switcher for the android platform with this PR.

|before ios with false value|after ios with false value|before ios with true value|after ios with true value|
|--|--|--|--|
Bildschirmfoto 2025-08-19 um 11 16 42|IMG_0326|Bildschirmfoto 2025-08-19 um 11 16 47-1|IMG_0325

|before android  with false value|after android  with false value|before android  with true value|after android  with true value|
|--|--|--|--|
Screenshot_20250819-111634|Screenshot_20250819-112055|Screenshot_20250819-111624|Screenshot_20250819-111111

SVA-1644

### DIFF
--- a/src/components/Switch.js
+++ b/src/components/Switch.js
@@ -19,6 +19,16 @@ const thumbColorEnabled = Platform.select({
   android: colors.gray20,
   ios: colors.lightestText
 });
+const disabledTrackColor = {
+  ...Platform.select({
+    android: { false: colors.shadow, true: colors.primary },
+    ios: { false: colors.shadow, true: colors.primary }
+  })
+};
+const disabledThumbColor = Platform.select({
+  android: colors.gray40,
+  ios: colors.lightestText
+});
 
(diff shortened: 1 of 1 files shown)


src/components/Switch.js
```diff
@@ -19,6 +19,20 @@ const thumbColorEnabled = Platform.select({
   android: colors.gray20,
   ios: colors.lightestText
 });
+const disabledTrackColor = {
+  ...Platform.select({
+    android: { false: colors.shadow, true: colors.shadow },
+    ios: { false: colors.shadow, true: colors.shadow }
```
donni106, line 25: you also changed iOS but are not showing an iOS screenshot. please add a before and after for both platforms.
iOS was quite good before if i remember correctly, why did you also change that?

ardasnturk: I have updated the screenshots. 👍

donni106 (approved):

PR #1376, opened 2025-08-20
feat(push): show alert in case of no internet connection

ardasnturk:
With this PR, during the change of switchers that require internet connection, the internet connection is checked and if there is no connection, an alert is shown to the user.

Bildschirmfoto 2025-08-20 um 11 10 19

SVA-1644

### DIFF
--- a/src/components/SettingsToggle.js
+++ b/src/components/SettingsToggle.js
@@ -1,9 +1,11 @@
 import PropTypes from 'prop-types';
-import React, { useEffect, useState } from 'react';
+import React, { useContext, useEffect, useState } from 'react';
 import { ActivityIndicator, StyleSheet } from 'react-native';
 import { ListItem } from 'react-native-elements';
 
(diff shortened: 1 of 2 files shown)


src/components/SettingsToggle.js
```diff
@@ -1,17 +1,20 @@
...
+import { serverConnectionAlert } from '../pushNotifications';
 
 import { Switch } from './Switch';
 import { BoldText, RegularText } from './Text';
 import { Touchable } from './Touchable';
 import { WrapperRow } from './Wrapper';
 
 // TODO: snack bar / toast als nutzerinfo
-export const SettingsToggle = ({ item }) => {
+export const SettingsToggle = ({ isNeedConnection = true, item }) => {
```
donni106, line 16: is need connection? what about needs connection? i think this would grammatically correct.

donni106 (approved):

PR #1502, opened 2026-03-24
refactor(bus): switched to live API via react-query

donni106:
Replaced BUS GraphQL/Apollo data loading with direct REST queries against the new BUS proxy, because the BUS module should no longer depend on the importer-based GraphQL path. 

Renamed BB-BUS structures to BUS, because the old module naming was wrong prefixed with BB for Brandenburg, the system is global -> BREAKING CHANGE! -> need a new versioned global settings.

Removed the area dropdown content, because this will be updated with SBB-194.

SBB-193


src/hooks/bus.js
```diff
@@ -0,0 +1,97 @@
...
+  const { bus = {} } = settings;
+  const hasBusConfig = !!bus?.uri && !!areaId;
+
+  const { data, isLoading, refetch } = useQuery(
+    [BUS_QUERY_KEYS.ROOT, BUS_QUERY_KEYS.SERVICES, areaId],
+    () => findPublicServices({ areaId, bus }),
+    {
+      enabled: hasBusConfig
+    }
+  );
```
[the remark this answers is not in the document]
donni106: There are no plans for these to change in production

src/types/Navigation.ts
```diff
@@ -10,8 +10,8 @@ export enum ScreenName {
   ARShow = 'ARShow',
   ARInfo = 'ARInfo',
   ArtworkDetail = 'ArtworkDetail',
-  BBBUSDetail = 'BBBUSDetail',
-  BBBUSIndex = 'BBBUSIndex',
+  BusDetail = 'BusDetail',
+  BusIndex = 'BusIndex',
```
ardasnturk: Why were the page names changed? Because of this change, we will have to update the staticContent of all apps that use BbBus.

src/screens/BUS/IndexScreen.js
```diff
@@ -0,0 +1,148 @@
...
+  const results = getListItems(areaId, services);
+
+  return (
+    <SafeAreaViewFlex>
+      <DefaultKeyboardAvoidingView>
+        <IndexFilterWrapperAndList filter={filter} setFilter={setFilter} />
+        {shouldShowLoadingOverlay ? (
+          <LoadingContainer>
+            <ActivityIndicator color={colors.refreshControl} />
+          </LoadingContainer>
```
ardasnturk: Why wasn't the `LoadingSpinner` component used here?

src/screens/BUS/IndexScreen.js
```diff
@@ -0,0 +1,148 @@
...
+    (selectedFilter?.id === FILTER_IDS.TOP10 && isLoadingTop10);
+  const shouldShowInitialLoading = loading && !services.length && !top10.length;
+  const shouldShowLoadingOverlay = loading && selectedFilter?.id !== FILTER_IDS.TOP10;
+
+  if (shouldShowInitialLoading) {
+    return (
+      <LoadingContainer>
+        <ActivityIndicator color={colors.refreshControl} />
+      </LoadingContainer>
+    );
```
ardasnturk: Why wasn't the LoadingSpinner component used here?

src/hooks/bus.js
ardasnturk, line 1: I think the form should be in TypeScript format.

donni106:
@codex[agent] please do a Code Review and leave your comments here.

src/hooks/bus.js
donni106: The file is now converted to TypeScript. I kept the hook API the same, added local types for bus settings, service items and hook args.

src/hooks/bus.js
```diff
@@ -0,0 +1,97 @@
...
+  const { bus = {} } = settings;
+  const hasBusConfig = !!bus?.uri && !!areaId;
+
+  const { data, isLoading, refetch } = useQuery(
+    [BUS_QUERY_KEYS.ROOT, BUS_QUERY_KEYS.SERVICE, areaId, id],
+    () => getPublicService({ areaId, bus, id }),
+    {
+      enabled: hasBusConfig && !!id
+    }
+  );
```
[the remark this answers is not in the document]
donni106: Fixed by making the BUS React Query keys depend on the active BUS config as well, while only storing a non-raw API key fingerprint in the key.

PR #1510, opened 2026-04-02
feat(bus): migrate area lookup to political-area endpoints

donni106:
The BUS area lookup now uses the dedicated `political-area` proxy endpoints for both search and initial area resolution by ID. The autocomplete contract was stabilized so `searchPoliticalAreas` always returns an array, with tests updated to cover repeated `searchWords` params, detail lookups, and safe fallbacks for unexpected responses. In addition, the area autocomplete UI and reset behavior were refined so the field now behaves consistently with the service search input across selection, reset, and follow-up search flows.

SBB-194


src/components/BUS/AreaAutocomplete.js
```diff
@@ -0,0 +1,331 @@
...
+        previousSelection.current.areaId !== areaId ||
+        previousSelection.current.areaName !== areaName;
+
+      if (!hasSelectionChanged) return;
+
+      previousSelection.current = { areaId, areaName };
+
+      const timeoutId = setTimeout(() => {
+        setInputValue(areaName || '');
+      }, 0);
```
ardasnturk: Why did you set a 0 ms timeout here? What is the point of that?

src/components/BUS/AreaAutocomplete.js
```diff
@@ -0,0 +1,331 @@
...
+    backgroundColor: colors.surface,
+    minHeight: normalize(48),
+    justifyContent: 'center',
+    paddingHorizontal: normalize(12),
+    paddingVertical: normalize(10)
+  }
+});
+
+AreaAutocomplete.displayName = 'AreaAutocomplete';
+AreaAutocomplete.propTypes = {
```
ardasnturk: Why did you choose to create a new JavaScript file? Actually, we wanted to create the new files in TypeScript format.

src/screens/BUS/IndexScreen.js
```diff
@@ -83,66 +90,58 @@ export const IndexScreen = ({ navigation }) => {
     isLoading: isLoadingTop10,
     refetch: refetchTop10
   } = useBusTop10(services);
+  const resolvedAreaName = useMemo(() => {
+    if (`${areaId}` === `${initialAreaId}`) {
+      return areaName || initialAreaName;
+    }
+
+    return areaName;
+  }, [areaId, areaName, initialAreaId, initialAreaName]);
```
ardasnturk: If the two values are already equal, why are we using the || operator in the return statement? Isn’t it enough to choose just one value?
Is a check required to ensure that `areaId` or `initialAreaId` is not undefined?

src/components/BUS/AreaAutocomplete.js
```diff
@@ -0,0 +1,331 @@
...
+};
+
+export const AreaAutocomplete = memo(
+  ({ areaId, areaName, initialAreaId, initialAreaName, onSelectArea }) => {
+    const [inputValue, setInputValue] = useState(areaName || '');
+    const [isFocused, setIsFocused] = useState(false);
+    const [searchBarRenderKey, setSearchBarRenderKey] = useState(0);
+    const previousSelection = useRef({ areaId, areaName });
+    const { data = [], isEnabled, isError, isFetching, isLoading } = useBusAreas(inputValue);
+    const areas = useMemo(
```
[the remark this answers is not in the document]
donni106: Ok, done

src/screens/BUS/IndexScreen.js
```diff
@@ -83,66 +90,58 @@ export const IndexScreen = ({ navigation }) => {
...
+          navigation={navigation}
+          selectedFilter={selectedFilter}
+          results={results}
+          areaId={areaId}
+          areaName={resolvedAreaName}
+          initialAreaId={initialAreaId}
+          initialAreaName={initialAreaName}
+          setArea={(area) => {
+            setAreaId(area.id);
+            setAreaName(area.label);
```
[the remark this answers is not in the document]
donni106: No need now

src/hooks/bus.js
```diff
@@ -12,21 +17,33 @@ const BUS_QUERY_KEYS = {
...
+  const { bus = {} } = settings;
+  const hasBusConfig = !!bus?.uri && searchTerm.length >= 3;
+
+  const { data, error, isError, isFetching, isLoading, refetch } = useQuery(
+    [BUS_QUERY_KEYS.ROOT, BUS_QUERY_KEYS.AREAS, searchTerm],
+    () => searchPoliticalAreas({ searchTerm, bus }),
+    {
+      enabled: hasBusConfig,
+      keepPreviousData: true
+    }
```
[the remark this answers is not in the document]
donni106: Ok, done

src/screens/BUS/IndexScreen.js
donni106: @codex[agent] please explain it to us or suggest a refactoring.

donni106: @copilot please explain it to us or suggest a refactoring.

src/components/BUS/AreaAutocomplete.js
donni106: It wasn't on purpose, changed now

src/components/BUS/AreaAutocomplete.js
donni106: updating the controlled value in the same effect tick can be overwritten by the search bar's stale native text after an external selection change, so the previous query briefly reappears and the selected area label gets lost until the next render.

PR #1514, opened 2026-04-21
fix(HtmlView): resolve YouTube iframe error 153 and 152-4 via Referer header

ardasnturk:
With this PR, the issue where YouTube content embedded via iframes could not be displayed due to an error has been resolved.

|before|after|
|--|--|
Simulator Screenshot - iPhone 13 mini - 2026-04-21 at 14 26 31|Simulator Screenshot - iPhone 13 mini - 2026-04-21 at 14 27 11

SVASD-548

### DIFF
--- a/src/components/HtmlView.js
+++ b/src/components/HtmlView.js
@@ -72,6 +72,40 @@ const htmlConfig = {
   }
 };
 
+const isYouTubeUri = (uri) => {
+  if (!uri) {
+    return false;
+  }
+  try {
+    const { hostname } = new URL(uri);
+    const normalizedHostname = hostname.toLowerCase();
+    return (
+      normalizedHostname === 'youtu.be' ||
+      normalizedHostname === 'youtube.com' ||
+      normalizedHostname.endsWith('.youtube.com')
+    );
+  } catch {
+    // Invalid URLs must not break HTML rendering.
+    return false;
+  }
+};
+
+// YouTube embeds require a Referer header, otherwise YouTube returns error 153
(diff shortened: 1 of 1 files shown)


src/components/HtmlView.js
```diff
@@ -72,6 +72,22 @@ const htmlConfig = {
...
+// ("The request contains no HTTP Referer or equivalent client identity").
+// provideEmbeddedHeaders is called by @native-html/iframe-plugin for each iframe src
+// and the returned headers are passed to the underlying WebView source.
+// Note: using 'https://www.youtube.com' as Referer causes error 152-4 because YouTube
+// applies stricter embed rules when the Referer matches its own domain. Using a neutral
+// origin like 'https://localhost' satisfies the presence check without triggering those
+// domain-specific restrictions.
+const provideEmbeddedHeaders = (uri) => {
+  if (uri && (uri.includes('youtube.com') || uri.includes('youtu.be'))) {
+    return { Referer: 'https://localhost' };
```
donni106: localhost? how come?

ardasnturk: I tried it with youtube.com earlier and got error 152-4 instead of error 153. Actually, it’s enough to just specify a reference for YouTube here. You could even use example.com. I will replace it with example.com.

donni106: but localhost does not make any sense at all. why not smart-village.app?

donni106: @copilot what do you think about this discussion?

donni106 (approved):

PR #1522, opened 2026-05-05
fix: resolve in-app redirect issue via notification

ardasnturk:
This PR resolves the issue where redirects within the app failed to work when the app launched from a cold start after a push notification was tapped.

Cold start: the state where the app is completely closed and then restarted.


donni106 (approved):

donni106:
🤞

PR #1524, opened 2026-05-06
fix: resolve gallery access permission issue

ardasnturk:
This PR updates the system so that the permission request to access the gallery is only displayed when permission has not been granted while on the report screen

SUE-207

### DIFF
--- a/src/hooks/imagePicker.ts
+++ b/src/hooks/imagePicker.ts
@@ -33,12 +33,16 @@ const saveImageToGallery = async (uri: string) => {
   const { status: existingStatus, canAskAgain } = await getPermissionsAsync();
   const appName = appJson.expo.name;
 
-  let status = existingStatus;
-
-  if (status !== PermissionStatus.GRANTED) {
-    if (canAskAgain) {
-      ({ status } = await requestPermissionsAsync(undefined, ['photo']));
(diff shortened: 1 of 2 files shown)


src/hooks/imagePicker.ts
```diff
@@ -27,11 +28,25 @@ export const MediaTypeOptions: Record<'Images' | 'Videos' | 'All', TMediaTypeOpt
...
+  // picker overlay when the user previously granted only limited access.
+  const { status: existingStatus, canAskAgain } = await getPermissionsAsync();
   const appName = appJson.expo.name;
 
+  let status = existingStatus;
+
   if (status !== PermissionStatus.GRANTED) {
-    return;
+    if (canAskAgain) {
+      ({ status } = await requestPermissionsAsync(undefined, ['photo']));
```
donni106: @copilot how come that the brackets are enclosing the whole line? shouldn't it be `{ status } = await requestPermissionsAsync(undefined, ['photo'])`?

donni106: @ardasnturk what do you think about the changes?

ardasnturk: Yes. That makes more sense. 👍

donni106 (approved):

PR #1538, opened 2026-05-22
feat: enhance `AvailableVehicles` component for park house functionality

ardasnturk:
With this PR, the occupancy rate of the city's parking garages is displayed in DetailScreen.

|detail view|map view|
|--|--|
image|image

SVASD-523

### DIFF
--- a/src/components/screens/AvailableVehicles.tsx
+++ b/src/components/screens/AvailableVehicles.tsx
@@ -10,6 +10,11 @@ import { LoadingSpinner } from '../LoadingSpinner';
 
 export const vehiclePropertyKey = 'Datastreams/0/Observations/0/result';
 
+// Status strings that are guaranteed to map to a named map-marker icon asset.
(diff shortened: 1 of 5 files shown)


src/components/screens/AvailableVehicles.tsx
```diff
@@ -45,18 +50,45 @@ export const fetchAvailableVehicles = async (
...
+            <RegularText>{status}</RegularText>
+          </ListItem.Content>
+        </ListItem>
+
+        <WrapperHorizontal>
+          <Divider />
+        </WrapperHorizontal>
+      </>
+    );
+  }
```
donni106, line 117: can we get rid of the `isSpecialForParkHaus` flag entirely, if we go without the "Status" headline?

ardasnturk: but then it will look like this.

image

